PR #2, opened 2015-03-26
These are a bunch of fixes and changes we made here at Blizzard. 

tvandijck:
Same as on BitBucket, with the os.delete, and the parse for known options removed..
the parse for known options I'll put in a separate request to discuss further..


TurkeyMan:
This is awesome. Thanks heaps guys!
I wonder if we should split these into many PR's for individual consideration?
I can't see a lot in here that's controversial though.

tvandijck:
that editorconfig plugin is the shizzle ;) thank you very much for sharing that little gold gem...

TurkeyMan:
Yeah, really. Couldn't live without it! Especially at work ;)

TurkeyMan:
Yeah, really. Couldn't live without it! Especially at work ;)

tvandijck:
Fixed the tab, and I found a bug in the oven.bubbleFields yesterday which affects usage of tokens in "vpaths", which we enabled.

tvandijck:
Closing this... I split most of these out into individual pull requests.

PR #3, opened 2015-03-31
pic api

TurkeyMan:
Added PIC (position independent code) flag, used all over the place.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -645,6 +645,16 @@
 	}
 
 	api.register {
+		name = "pic",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Off",
+			"On",
+		}
+	}
+
+	api.register {
 		name = "platforms",
 		scope = "project",
 		kind = "list:string",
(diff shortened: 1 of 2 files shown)


TurkeyMan:
TODO: support the PIC flag for GCC/Clang builds.
This is here to sample for objections first...

starkos:
IMO this would be better as a pic() (or similar) API call because a) flags aren't the best for usability (overlapping or removing values) and I'd like to trend away from them and b) I think there are actually a few different kinds of position independent code ("pic" vs. "PIC", relaxed, simple) that we may want to support in the future.

I'd suggest a simple pic ( "On" | "Off" ) for starters?

TurkeyMan:
Yeah, I wasn't feeling convinced either way.
I'll change it and implement for clang/gcc.

TurkeyMan:
Okay, I changed it to a pic api, updated the code, tested it works, unittests pass.
I'm not sure why travis didn't rebuild my PR when I amended... but it does build and passes the unittests.

starkos:
Cool—can you add a page to the wiki? Thanks!

PR #4, opened 2015-03-31
'fpu' option

TurkeyMan:
Added an 'fpu' option, to select software/hardware FPU (used by arm, mips, etc)

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -448,6 +448,17 @@
 	}
 
 	api.register {
+		name = "fpu",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Software",
+			"Hardware",
+		}
+	}
+
+
+	api.register {
 		name = "framework",
 		scope = "config",
 		kind = "string",


starkos:
Could this overlap with the existing floatingpoint() call? Should we just add the values there?

TurkeyMan:
Yeah, I wasn't sure about that.

There are 3 distinct concepts:
1) hardware availibility ('fpu' as i propose here)
2) abi ('floatabi')
3) float codegen options ('floatingpoint', fast/strict/etc)

It doesn't really feel right to conflate any of those. The only one that might not create a mutual exclusion is 'floatingpoint', but i'm not sure about the same mechanism that says to choose fast over strict float support may say there is no FPU hardware? ...maybe it doesn't sound so bad when I say it this way... but it still seems a bit crooked to me.

starkos:
My opinion would be to only use two fields if it makes sense to set values from both in parallel, e.g. if something like this would be valid:

    fpu "software"
    floatingpoint "strict"

Otherwise I would add "software" and "hardware" to floatingpoint(). If the architecture doesn't have FPU hardware then I would expect those values to simply be ignored as we do elsewhere.

floatabi(), on the other hand, does feel like it should be a separate call.

TurkeyMan:
Well there would never be `floatingpoint "hardware"`, because as soon as you have hardware FPU, there will always be various fast/strict/etc tradeoffs.
So `floatingpoint "software"/"fast"/"strict"/...` seems weird to me, and it seems conceivable that fast/strict tradeoff may be made even in the software context; FPU emulation is slow, and I imagine all sorts of approximations that could be taken to speed it up.

I think I have this correct as presented in this PR.

TurkeyMan:
So how are we feeling about this? I think this PR is right...

PR #5, opened 2015-03-31
'endian' option

TurkeyMan:
Added 'endian' option to select target endian (arm, ppc, mips)

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -324,6 +324,16 @@
 		tokens = true,
 	}
 
+	api.register {
+		name = "endian",
+		scope = "config",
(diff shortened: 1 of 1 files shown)


starkos:
This one seems fine. Do you want to add a little implementation here, or merge this one and submit the implementations separately?

TurkeyMan:
x86 is practically the only architecture with no support for this :)
ARM/MIPS/PPC all want this option, but there's nothing in core that refer to those.

TurkeyMan:
Just wanted to check up on this. There's nothing for me to implement in core (it's all x86).
Android will make use of it initially.

PR #6, opened 2015-03-31
These are ALL our changes to premake-core to get our build up and running.

tvandijck:
This Pull request is not intended to be merged, it's just here to create visibility into the changes and fixes we made. I can pull out changes and make individual pull requests for those where needed.
At this time however some of the changes depend on earlier changes, so it be nice if we could at least work somewhat in order...


starkos:
This one can be closed?

tvandijck:
I just kept it around as kind of a lookup for what other changes we still have... some of those are not merged into master at all, since I didn't make pull request for them... but I would still like to receive some sort of review on these changes so that we may be able to submit them as individual pull requests in the future.

tvandijck:
Most notable are:
- Update to lua 5.2.3
- Buffered I/O, and only write generated file if it changed.
- Use most descriptive vpath for visual studio virtual folders.
- fix project dependencies.

starkos:
Fine with these except for the Lua upgrade, which I know will break some clients that are integrating with other Lua 5.1 systems (debuggers, etc.). What's the reason for the upgrade? If we are going to switch, would it make more sense to go to 5.3 instead of breaking it again in the future?

Not entirely opposed to the upgrade, just want to make sure we're doing it for a reason other than "because it's there".

tvandijck:
Yeah, mostly because it's there, but I also wanted to try out LuaJit to see what the performance benefits would be. For us with the big projects we have, premake takes up to 20 seconds to generate everything, so anything to speed up the process would help. So I did the upgrade to 5.2.3 to assure compatibility with LuaJit... then tried LuaJit... and well, we kinda just stuck with it.

As for 5.3, I tried it recently... doesn't work... so we stuck with 5.2.3..

The other commits, I'll try to make pull requests for them, after that I'm sure we can close this one.

PR #7, opened 2015-03-31
add string.sha1 function

tvandijck:


tvandijck:
Our xcode.newid now looks like this:

function premake.xcode.newid(...)
	local name = table.concat({...}, ';');
	return string.sub(name:sha1(), 1, 24)
end

PR #10, opened 2015-03-31
deprecate ReleaseRuntime flag, use 'runtime "Release"' instead.:

tvandijck:


TurkeyMan:
I want this fix. It's actually plaguing me right now... but I'm not sure if this is quite the right approach.
Trouble is, that `config.isDebugBuild()` business has always irk-ed me. It's making a lot of assumptions, and in my experience, they're not reliable.
I tend to think `config.isDebugBuild()` should be removed, `runtime` should be used exclusively, and sensible defaults (that match `isDebugBuild`?) should be applied...?

tvandijck:
Yeah, we went for most backward compatibility... so when runtime is set, it'll be used, if not then it uses the 'old' system... I'm all with you though, but it would require that we set runtime to some sort of default.

TurkeyMan:
Well, it's true that the default needs to be semi-intelligent... but I don't think the current logic is quite correct. I've fallen for edge cases on numerous occasions.

starkos:
I'm going to merge this one as I think it is a good starting point. I'm happy to continuing iterating on it before the full release though.

PR #11, opened 2015-03-31
Fix two bugs in embed action.

tvandijck:


premake5.lua
```diff
@@ -27,6 +27,8 @@
 		trigger = "embed",
 		description = "Embed scripts in scripts.c; required before release builds",
 		execute = function ()
+			_MAIN_SCRIPT_DIR = os.getcwd()
+			_SCRIPT_DIR = path.join(_MAIN_SCRIPT_DIR, "scripts")
```
TurkeyMan: How did this help?

tvandijck: Before this change, none of my lua scripts would get added to the embed... these two lines are in the premake4.lua script, but not in the premake5.lua script, so I copied them... made it work.

starkos: I had to add those lines to premake4.lua because the 4.x code doesn't set those variables. Premake 5.x does, so these shouldn't be needed. You might want to try again without them, or provide more information about what's going wrong so we can figure out the right fix.

starkos:
I'm rejecting this one, because those variables should be already set, and I believe there is some other problem at work. If you can provide a little more information about what's going on we can treat it as a bug and try to find the right fix.

PR #13, opened 2015-03-31
Add C++11, C++14 flags for GCC & Clang

tvandijck:


starkos:
As discussed, I'm merging this one but these flags will (hopefully) be deprecated soon in favor of additional language() values.

jmansion:
They don't seem to be mentioned on the flags page - will they be terained?  Or can we expect new languages?  Should also be useful with MSVC.

tvandijck:
MSVC doesn't let you specify this as an option... not sure what you mean.

we will eventually deprecate this for proper API, but at this time, they are still there.. we'll fix the documentation.

PR #15, opened 2015-03-31
add table.tostring

tvandijck:


TurkeyMan:
I've wanted this for ages. It seems like a big implementation though. Is this particularly different/better than the seemingly standard 'tprint' function that's plastered all over the net?
I'm sure you Bliz guys know lua better than anyone ;)

tvandijck:
I don't think we know lua all that well ;) at least not me... so some of these things just come from the need to display something, and it seems useful to add... the tprint looks simpler, and I'm not sure if it prints things the same way, but either one could work I guess... we can always modify it to get the thing out of it that we need.

tvandijck:
Oh, and actually looking at tprint... that really just prints things... table.tostring returns a string... we save that to a log sometimes... I'm sure we can modify tprint to return a string, although I have no idea what send does... also the 'tostring' method doesn't always return something useful, primarily for <userdata> values.

TurkeyMan:
I'm sure it's fine. It was just at a cursory glance I found it harder to digest, and didn't immediately see any significant difference in functionality.

PR #18, opened 2015-03-31
64-bit compiler doesn't like SSE/SSE2 flags, they are implied.

tvandijck:


TurkeyMan:
LGTM, except do you need to update gmake too?

This creates a disparity between vs and gmake builds?
Perhaps it would be better to filter these out in the bake, then all generators will get the fix?

tvandijck:
The gcc/clang compilers are perfectly fine with the option... it's only the msvc compiler that complains... even though I agree that the option on gcc/clang may be entirely ignored.

TurkeyMan:
Fair enough.

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1053,10 +1053,12 @@
 			v = "AdvancedVectorExtensions"
 		elseif x == "AVX2" and _ACTION > "vs2012" then
 			v = "AdvancedVectorExtensions2"
-		elseif x == "SSE2" then
-			v = "StreamingSIMDExtensions2"
-		elseif x == "SSE" then
-			v = "StreamingSIMDExtensions"
+		elseif cfg.architecture ~= "x64" then
```
TurkeyMan: I was going to pull this, but a recent change has renamed x32->x86 and x64->x86_64
This needs to be updated to x86_64.

I also wonder if it should be `== "x86"` rather than `~= "x86_64"`, since it's precisely one architecture that supports this, not 'anything other than x86_64'.
Is there a possible 'unspecified' case for architecture that the ~= also catches?

tvandijck: Well, I can certainly rename the if to x86_64.... I think x86 would be wrong however, since we still want this enabled on 32-bit platforms, where this feature is not implied... old x86 processors don't always have SSE2, although you have to go back in history quite a bit for that you still have to explicitly enable this in Visual Studio.

TurkeyMan:
@starkos You did see that this uses the compatibility alias rather than the modern "x86_64" yeah?

PR #23, opened 2015-04-03
Cleanup Hg

TurkeyMan:
Clean up the Hg litter.


TurkeyMan:
Haha, good thing we have CI! I never would have imagined this change could fail the build!

PR #25, opened 2015-04-05
Add modules to .gitignore

starkos:
Prevents private or in-progress modules from appearing in pending changes, makes it a little easier to work against core.


TurkeyMan:
Good call!

PR #27, opened 2015-04-06
add bindirs

tvandijck:
not sure if this needs to be added to gcc back-end, but we use this for pre/post build steps in MSVC.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -894,25 +894,14 @@
 	end
 
 
-	function m.filterEmpty(dirs)
-		return table.translate(dirs, function(val)
-			if val and #val > 0 then
-				return val
-			else
-				return nil
-			end
-		end)
-	end
-
-
 	function m.additionalIncludeDirectories(cfg, includedirs)
 		if #includedirs > 0 then
 			local dirs = project.getrelative(cfg.project, includedirs)
-			dirs = m.filterEmpty(dirs)
+			dirs = table.filterempty(dirs)
 
 			if #dirs > 0 then
 				table.sort(dirs)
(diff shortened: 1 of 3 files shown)


TurkeyMan:
It's kinda strange how this is failing the unittests. The slashes all got turned around somehow, and I can't see what in this patch would have done that...

tvandijck:
I think it's a missing path.translate somewhere... I'll look into it...

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -892,11 +893,26 @@
 	end
 
 
+	function m.filterEmpty(dirs)
```
starkos: This would actually make a useful addition to the table.* namespace; I can think of other places I'd use it.

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -892,11 +893,26 @@
...
+				return nil
+			end
+		end)
+	end
+
+
 	function m.additionalIncludeDirectories(cfg, includedirs)
 		if #includedirs > 0 then
 			local dirs = project.getrelative(cfg.project, includedirs)
-			dirs = path.translate(table.concat(dirs, ";"))
```
starkos: I think you need to put this path.translate() back.

tvandijck: Yes, I just ran into this locally as well ;) I accidentally mistook the "table.translate" in the m.filterEmpty method with a path.translate, so thought the m.filterEmpty already did the translation.... doh... I'll fix this one up, and move the filterEmpty method into table.lua

tvandijck:
this one is reworked and tested... should work now.

tvandijck:
seeing as you made a vstudio.path() method, we could put the filter empty call in there?

PR #28, opened 2015-04-06
add time information.

tvandijck:
We like this for stats...

### DIFF
--- a/src/host/premake_main.c
+++ b/src/host/premake_main.c
@@ -5,11 +5,16 @@
  */
 
 #include "premake.h"
+#include <time.h>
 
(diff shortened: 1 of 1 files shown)


starkos:
Sorry for the late reply on this one. While it makes the overall timing a little less accurate, I think this should be moved to the Lua side, in _premake_main.lua, to allow people to override it.

You could put the start/top in installModuleLoader() and postAction(), or perhaps it would be better to introduce a beginMain() and endMain()?

tvandijck:
sure, that works for me... will rework when I have the time for this one then.

PR #29, opened 2015-04-07
cygwin build

zpgaal:
- uuid for guid generation
 - use proc/exe for binary folder detection
 - buildscript modifications
 - some other smaller compile fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ before_install:
 script:
  - premake4 embed
  - premake4 gmake
- - make -j8
+ - pkg-config uuid --libs
+ - pkg-config uuid --cflags
+ - make verbose=1 -j8
  - bin/release/premake5 test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,5 @@ before_install:
 script:
  - premake4 embed
  - premake4 gmake
- - pkg-config uuid --libs
- - pkg-config uuid --cflags
- - make verbose=1 -j8
  - bin/release/premake5 test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,6 @@ before_install:
 script:
  - premake4 embed
  - premake4 gmake
+ - make -j8
  - bin/release/premake5 test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 before_install:
  - sudo add-apt-repository -y ppa:codegear/release
  - sudo apt-get -qq update
- - sudo apt-get install premake4
+ - sudo apt-get install premake4 uuid-dev
  
 script:
  - premake4 embed
--- a/premake5.lua
+++ b/premake5.lua
@@ -168,9 +168,11 @@
 			links       { "CoreServices.framework" }
 
 		configuration { "solaris" }
+			includedirs ( "/usr/include/uuid" )
 			linkoptions { "-Wl,--export-dynamic" }
 
 		configuration "aix"
+			includedirs ( "/usr/include/uuid" )
 			links       { "m" }
 
 		configuration {}
--- a/premake5.lua
+++ b/premake5.lua
@@ -84,6 +84,40 @@
 	solution "Premake5"
 		configurations { "Release", "Debug" }
 		location ( _OPTIONS["to"] )
+		
+		configuration "Debug"
+			defines     "_DEBUG"
+			flags       { "Symbols" }
+
+		configuration "Release"
+			defines     "NDEBUG"
+			flags       { "OptimizeSize" }
+
+		configuration "vs*"
+			defines     { "_CRT_SECURE_NO_WARNINGS" }
+
+		configuration "vs2005"
+			defines   {"_CRT_SECURE_NO_DEPRECATE" }
+			
+		configuration "cygwin"
+			--gccprefix "i686-pc-cygwin-"
+			defines   { "LUA_USE_POSIX" }
+
+		configuration "linux or bsd or hurd"
+			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
+
+		configuration "macosx"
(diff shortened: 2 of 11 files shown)


src/base/os.lua
```diff
@@ -80,6 +80,9 @@
 			if os.is("macosx") then
 				formats = { "lib%s.dylib", "%s.dylib" }
 				path = os.getenv("DYLD_LIBRARY_PATH") or ""
+			elseif os.is("cygwin") then
+				formats = { "lib%s.a", "%s.a", "lib%s.dll.a", "%s.dll.a" }
+				path = os.getenv("LD_LIBRARY_PATH") or ""
```
TurkeyMan: I'm not sure I'm sold on this. Why do you think os.is() shouldn't return 'windows', when it's running on windows?
Cygwin isn't an OS, it's a toolchain, but I guess it's a bit awkward that it also implies a few tools that make it feel more linux-like.

premake5.lua
```diff
@@ -84,6 +84,40 @@
...
+
+		configuration { "macosx", "gmake" }
+			toolset "clang"
+			buildoptions { "-mmacosx-version-min=10.4" }
+			linkoptions  { "-mmacosx-version-min=10.4" }
+
+		configuration "aix"
+			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
+
+		configuration {}
```
TurkeyMan: What happened here? Why rearrange this file so much?

TurkeyMan:
Other than the fact that it doesn't pass the unittests, I'm not entirely sold here.
It's nice to have a patch that allows premake to build under cygwin, that's cool, but the platform bits seem a little awkward. Particularly, is cygwin really an OS? Surely the OS is Windows?

premake5.lua
zpgaal: There's another patch ongoing that adds a binary module and requires the same LUA settings as premake and links to pramke5.
I've moved all non-exe related things out of premake project as other project should also get the same (LUA) config parameters.
see: https://groups.google.com/d/msg/premake-development/QGnDJ_m1elI/SB5xcGPo4doJ

src/base/os.lua
zpgaal: Writing code for cygwin is much closer to linux than to windows. If you check the patch the "linux"  implementation almost always wins. So I think a cygwin should be treated as a different os.
And actually AIX, Solaris, Bsd, linux, etc, are all a unix based something with different os names.

premake5.lua
TurkeyMan: If that's a separate patch, this change should be bundled with that PR.

src/base/os.lua
TurkeyMan: Hmmm. Yeah, I'm semi-convinced.
It raises the question; what about mingw?
And what about support for cygwin cross-compiling under linux? I think that's equally important. cygwin/mingw is often the target when cross-compiling for windows.
I've had some similar thoughts wrt mingw recently. @starkos?

tvandijck: I would much rather introduce a os.is("posix") that way it would cover any posix compliant OS... cygwin, mingw, are not the only "unix" based OS's out there... the "os.is" function could easily have this list internally, and return true on "posix" instead..

starkos:
FWIW, Premake used to build just fine under Cygwin already (I haven't tried it in a while though).

premake5.lua
starkos: Agreed—this is a reasonable change and could have been accepted easily, now it is going to get all tangled up in these other changes.

src/base/os.lua
starkos: I agree with Tom on os.is("posix"), but do we still need a Cygwin identifier too? Those crazy ".dll.a" extensions could be justification.

src/host/os_chmod.c
```diff
@@ -21,7 +21,7 @@ int os_chmod(lua_State* L)
 
 	int mode = (int)strtol(modeStr, &endPtr, 8);
 
-#if PLATFORM_WINDOWS
+#if PLATFORM_WINDOWS && !defined(PLATFORM_CYGWIN)
```
starkos: If PLATFORM_WINDOWS is defined then PLATFORM_CYGWIN is not; they are mutually exclusive (as they should be) in premake.h.

src/host/os_realpath.c
```diff
@@ -17,7 +17,7 @@ int os_realpath(lua_State* L)
 
 	const char* path = luaL_checkstring(L, 1);
 
-#if PLATFORM_POSIX
+#if PLATFORM_POSIX || defined(PLATFORM_CYGWIN)
```
starkos: Cygwin is a Posix platform; PLATFORM_POSIX should be all that's needed here.

.travis.yml
```diff
@@ -2,7 +2,7 @@
 before_install:
  - sudo add-apt-repository -y ppa:codegear/release
  - sudo apt-get -qq update
- - sudo apt-get install premake4
+ - sudo apt-get install premake4 uuid-dev
```
starkos: What do you guys think of adding this dependency? Personally I'd rather avoid it unless we are reasonably confident that it will be available out of the box (without needing an extra install step) on a significant majority of systems.

TurkeyMan: Are there also lib packages present for mingw (cross-compiling for windows)? And how to source it for OSX? BSD?
I'm not really good with this unless it addresses the build for Windows, OSX, BSD, etc aswell.

TurkeyMan: I think this might lead back into that 'package'/'dependency' discussion raised on the forum...

starkos:
zpgaal: can you resubmit these changes as separate pull requests: cygwin support, Premake project script reorganization, and this UUID library change? I think the first can be approved with minor changes (see comments here), and the second can be approved as is. The last one needs a little testing before I think we'll feel comfortable using it. Thanks!

PR #30, opened 2015-04-08
update xcode submdoule revision

fzwoch:
updated the revision of the xcode submodule. I don't know if there is a way to auto update it to point it to the latest or if that is something what we want?


TurkeyMan:
I'm gonna pull this since starkos was happy to accept the xcode patch.

starkos:
> I don't know if there is a way to auto update it to point it to the latest or if that is something what we want?

I haven't decided yet if this is a bug or a feature. Still too new at working with the sub-repos. 

Thanks for doing the pull Manu.

TurkeyMan:
Yeah, I've been changing my mind back and fourth on this for years actually.
Thing is, git submodules behave exactly 'correctly', ie, every historical revision in the outer repo will always build, since the precise submodule revision at each point in time is versioned too; going back in time works as expected.
That's good, except it's a pain in the arse, since 99% of the time, you have to update the outer repo too along with the submodule...
I read somewhere that it is a recently added feature for git submodules to track a branch, rather than an absolute revision number, but obviously, you lose historic coherence.
I wonder if a best-of-both-worlds solution may exist; master tracks submodules master branches, while release branches submodules reference precise revision numbers? I've never tried this, but maybe it's what we want, since I think we're gonna and up with a lot of submodules...

gorlak:
So I know that pain well, having using submodules extensively.  In my experience I have most often found that I want some submodules to actively tracking branches (and get the latest commit from that tracking branch committed to the parent whenever someone notices it has changed), and some submodules I want to stay at the specified revision until I elect to change it (best for third-party submodules like libraries).  The best solution I have found to get both desired behaviors when you want it is to script it via a wrapper :(  My solution is here: https://github.com/gorlak/gitr.  The gist of it is:

* `git submodule` will do a headless checkout by default (to preserve history)
* gitr.py will always treat headless checkouts as "should be updated to the specified commit in the parent during a pull to the parent", basically: `git submodule update --init`, which is the 'classic' behavior to preserve history
* if you checkout a submodule to a branch head (opt it into the tracking behavior), gitr.py will instead pull that branch head (a workalike to the new git 1.8 submodule feature @TurkeyMan is talking about)

The end result is that the user gets to elect per-clone, per-submodule which submodules should track branches, and which to treat as historical.  You can transition a submodule between these two state relatively easily.  Gitr.py also includes a reporting mode to tell you which submodules are tracking branch heads, and which are headless checkouts.  It also has a utility to checkout every submodule to headless.

PR #31, opened 2015-04-08
Fix handling of "../../.." sequences in path.normalize()

starkos:
The previous implementation of path.normalize() would begin stripping "../" after the second sequence: ../../../test would become ../../test. Fix and a unit test to cover.

### DIFF
--- a/src/host/path_normalize.c
+++ b/src/host/path_normalize.c
@@ -35,13 +35,11 @@ int path_normalize(lua_State* L)
 		if (ch == '.' && last == '.') {
 			ptr = dst - 3;
 			while (ptr >= buffer) {
-				if (*ptr != '/') {
-					--ptr;
-				}
-				else {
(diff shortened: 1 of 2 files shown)


TurkeyMan:
lgtm

TurkeyMan:
out of interest, is this why you're posting your own changes as pr's? to get reviews, rather than just pushing your own changes straight up as usual?
do you want to be responsible for merging these yourself? (i've merged a couple i thought looked both good and benign now... hope you're cool with that)

TurkeyMan:
or do you just want the build machine to have a go at it? ;)

starkos:
Both! Feel free to merge if you think it looks good, or you can leave it for me. If the pull requests are going to keep coming so quickly I'll need some help to process them with too long a delay.

TurkeyMan:
It's the power of github!
I've had numerous projects where experience has shown simply moving to github increases participation by an order of magnitude at least ;)

PR #35, opened 2015-04-10
More debugging api's

TurkeyMan:
Specify a debugger to use.
Specify the debugger binaries, args to run with.
Specify path remaps: It's typical to remap paths embedded in dwarf (as the exe was built) to the location of the source on your own pc.

I have another module which adds VisualGDB as a debugger for use with visual studio, and the D module supports an alternative debugger called Mago, which has D language expression parsing and better visualisations. There's another called WinGDB which can be used for Android.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -241,6 +241,24 @@
 	}
 
 	api.register {
+		name = "debugger",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"GDB",
+			"LLDB",
+		}
+	}
+
+	api.register {
+		name = "debugpathmap",
+		scope = "config",
+		kind = "list:keyed:path",
+		tokens = true,
+	}
+
+	api.register {
 		name = "debugport",
 		scope = "config",
 		kind = "integer",
@@ -265,6 +283,22 @@
 		scope = "config",
 		kind = "list:string",
 		tokens = true,
+	}
+
+	api.register {
+		name = "debugtoolargs",
+		scope = "config",
+		kind = "list:string",
+		tokens = true,
+		pathVars = true,
+	}
+
+	api.register {
+		name = "debugtoolcommand",
+		scope = "config",
+		kind = "path",
+		tokens = true,
+		pathVars = true,
 	}
 
 	api.register {


src/_premake_init.lua
```diff
@@ -241,6 +241,24 @@
...
 	api.register {
+		name = "debugger",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"VisualStudio",
+			"GDB",
+			"LLDB",
+		}
+	}
```
TurkeyMan: This is a high level debugger selection. It is intended to generate project bindings for the chosen debugger.

src/_premake_init.lua
```diff
@@ -241,6 +241,24 @@
...
+
+	api.register {
+		name = "debugpathmap",
+		scope = "config",
+		kind = "list:keyed:path",
+		tokens = true,
+	}
```
TurkeyMan: This allows to register path maps, which are common when using GDB, since DWARF often embeds absolute paths relative to the filesystem a binary was built on, and a debuggee will need to remap that directory structure to their own local source tree.

src/_premake_init.lua
```diff
@@ -268,6 +286,22 @@
...
+		pathVars = true,
+	}
+
+	api.register {
+		name = "debugtoolcommand",
+		scope = "config",
+		kind = "path",
+		tokens = true,
+		pathVars = true,
+	}
```
TurkeyMan: Ie, gdb.exe
I wanted to call this 'debugcommand', but that's already taken.
Is this a reasonable name?

src/_premake_init.lua
```diff
@@ -268,6 +286,22 @@
 	}
 
 	api.register {
+		name = "debugtoolargs",
+		scope = "config",
+		kind = "list:string",
+		tokens = true,
+		pathVars = true,
+	}
```
TurkeyMan: gdb at least often needs to be invoked with some arguments.
These are args passed directly to gdb, not to the client application we are launching to debug.

src/_premake_init.lua
TurkeyMan: I wondered if this should be called 'debugtool' to match the other 2 new options below, but it didn't feel right.
I don't think "VisualStudio" is a 'debugtool'. That terminology just felt a bit awkward to me.

starkos: VisualStudio feels wrong to me. Should it just be "Default"?

src/_premake_init.lua
starkos: Wouldn't this be inferred from debugger() above?

src/_premake_init.lua
TurkeyMan: Hmmm, yeah okay. I accept.

src/_premake_init.lua
TurkeyMan: Not really. 'debugger' says high level information about which debugger is in use, so everything can be configured accordingly. Some toolset's have debugger's in awkward places; NaCl for instance, it's gdb binary is separate from the various arch toolchains, and its name doesn't follow typical GCC convention. So in that case, it's necessary to specify the debugger binary explicitly.

That said, I'm still not super jazzed about the api name.

src/_premake_init.lua
```diff
@@ -241,6 +241,24 @@
 	}
 
 	api.register {
+		name = "debugger",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
```
TurkeyMan: It is Default. I'm not sure I love it, but I think it's better.

TurkeyMan:
Okay, changed to 'Default'.

src/_premake_init.lua
starkos: Ideally, the NaCL module would figure that out for you, right? But I get the idea, you sometimes need to set the path manually. What about debuggerpath()?

TurkeyMan: The NaCl module does set that for you. It just calls the APIs to set a default: https://github.com/TurkeyMan/premake-nacl/blob/master/nacl.lua#L62

The NaCl sets up a lot of defaults via API calls. Mainly to configure the debugger (which is a comprehensive process for NaCl).

TurkeyMan: Oh, and 'debuggerpath'... not sure.
We use 'debugcommand' to specify the binary to run, and other instances of 'path' in the api list refer to directories, not commands to execute. I don't think there's precedent for that.

starkos: Could be just debugtool() then? Or make it debuggerpath() and use it in conjunction with debugger()? So we get the tool's executable name from debugger() and combine it with debuggerpath()?

TurkeyMan: >Could be just debugtool() then?

I guess. I just wrote 'command' to extend what appeared to be precedent from elsewhere. I just feel whatever it is needs to be precedented, and follow (or extend) any existing patterns.

>Or make it debuggerpath() and use it in conjunction with debugger()? So we get the tool's executable name from debugger() and combine it with debuggerpath()?

Nar, 'debugger' needs to be something high-level and abstract. 'Default' isn't an executable, and 'VisualGDB', 'WinGDB', 'Mago' are VS plugins. They don't have any such executable name, so I don't think that pattern works.

It also leads to debugger, debuggerpath... debuggerargs, and i'd say debuggerargs is too similar to debugargs. People would look at that and never perceive the distinction.
That's mainly why I went with 'debugtool', I think I felt something about the word 'tool' really made the distinction clear.

starkos: I'm fine with letting this go through so you can move forward, but my gut feeling is that there is too much overlap going on and these are going to need to be consolidated at some point.

TurkeyMan: I have the same feeling, but I can't think of a simplification that both 1) doesn't significantly change the long-term api's, and 2) expresses everything I need.
I think part of it is just a misconception that debug configuration should be relatively simple, but it turns out it's actually really not. There's a ton of stuff that needs to be configured properly for it to work in any situation. A lot of these options are niche, and wouldn't usually be touched. Most of them can have automated defaults. Perhaps I could just automate the outputs directly, rather than have api's at all... but I think it's better to create an API and assign it a default, that way it can be overridden in the inevitable case where it's necessary.

If you're happy with this for now, that'd be good, and I'm more than happy to rearrange it in the future if we can pin down a simpler set of api's that are able to express everything.

starkos: Yes, I approve, make it so.

PR #36, opened 2015-04-10
Added aliases for explicit architectures

TurkeyMan:
The 'architecture' api has always bothered me; none of the entries are actually architectures... and it's often necessary to select architectures explicitly.

@starkos I see you were going for something a little bit abstract, associating with the x86 toolchains '-m32'/'-m64' compile args. That's kind reasonable for x86, but doesn't make sense for anything else.

Obviously, we can't remove those entries, but I think we need to add explicit architectures alongside those abstract entries. Many of my modules introduce various explicit architectures, but I think explicit x86 variants should be defined in core.

I've always felt that x32 and x64 create confusion since they look like they may specify an explicit architecture ('x' implies x86 line of processors), but they're only meant to specify -m32/-m64. I think for premake5, we should rename those 2 entries to m32 and m64, which is what they are. I have created aliases for x32->m32 and x64->m64, which I think should be deprecated.

I then added x86 and x86_64; the proper names for those explicit architectures.

I'm not actually sure m32/m64 should exist, but they are certainly long-standing legacy, so I think it's better to keep them, and the rename makes their conceptual meaning more clear and avoid confusion with absolute architecture identifiers.

Then finally, I've painstakingly made everything work...

Please tell me this seems correct to you, I spent a lot of time on this ;)

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -23,8 +23,14 @@
 		kind = "string",
 		allowed = {
 			"universal",
-			"x32",
-			"x64",
+			"x86",
+			"x86_64",
+		},
+		aliases = {
+			i386 = "x86",
+			amd64 = "x86_64",
+			x32 = "x86",	-- these should be DEPRECATED
+			x64 = "x86_64",
 		},
 	}
 
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -21,8 +21,8 @@
 
 	vstudio.vs200x_architectures =
 	{
-		x32     = "x86",
-		x64     = "x64",
+		x86     = "x86",
+		x86_64  = "x64",
 		xbox360 = "Xbox 360",
 	}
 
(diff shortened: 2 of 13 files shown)


starkos:
Yes, I'm good with this. The x32/x64 were added when PPC was still a thing, and were a handy, portable way to target a specific bit length at the time. I think we could drop m32/m64 and alias them directly to x86 and x86_64 if you like. Unless there is 32-bit ARM architecture to consider?

TurkeyMan:
Well, PPC still exists, I don't think we should write it off. There is AArch64, but I haven't seen a toolchain with -m32/-m64 to switch between them; they're totally separate toolchains (same with ppc32/64).

What I never understood, is how exactly x32/x64 was intended to map across platforms? If I build x64 windows, that has no bearing on whether I want 32 or 64bit ppc builds (wii? x360?).

If you're happy to remove x32/x64 completely and roll with explicit architectures, I'm behind that, and I can amend this PR for that.

There's another question though. I have x86 and x86_64, but if we want to be precise, it should probably be i386 and x86_64... I have also seem i686 builds of linux, which we may also want to be able to describe. So, what do you reckon; we move to the proper arch names, or stick with 'x86'?

TurkeyMan:
I've removed m32/m64. This now just has the deprecated arch's aliasing the new explicit architectures directly.

starkos:
Looks good. Merging…

PR #42, opened 2015-04-14
libCurl Http / Https Download Support

grbd:
Hi,
I've recently imported some of the changes from this original patch
https://bitbucket.org/premake/premake-dev/pull-request/122/initial-http-support/diff
to support downloading of files via http / https using libcurl

I'm not sure if the modules directory is the correct place to put the code, also I know there are other comments in the original link that are outstanding, but it seems to work when I tested it
http.download("https://linktofile","somelink.txt")

By default this feature is disabled to avoid compile conflicts with macosx, but can be enabled by setting
local ENABLE_CURL = true
within premake5.lua
so far I've only tested it on the rpi2 / arm / gentoo


TurkeyMan:
I imagine it would live in a directory under src/host/? but that said, while I love the feature, I'm not convinced including the full source is the best things to do. On *nix platforms, it should (dynamically) link a lib from the package manager, since that will receive regular security updates and what not... for windows (always the pain in the arse!), not so sure...
Surely OSX has libcurl.dylib available by default?

Perhaps embedding the source is the best thing to do for the windows build?
I don't like it. But then, I don't like the windows dev environment either, and there's not too many options. If it's only for windows (assuming it's available by packaging, or by default on other platforms), perhaps it better to include a .lib?

TurkeyMan:
Surely we don't need the docs in there at least?

grbd:
I think the original idea was to statically compile everything so that premake would be standalone instead of relying on dynamic libs
that being said, it looks as if libcurl pulls in other libs dynamically anyway even when you tell it to do a static compile (looking at ldd)

including binary .lib blobs is definitley not the way to go though, the only reason I know of the issues surrounding macosx is because of the previous linked thread above, I've actually been testing this on raspeberry pi with gentoo linux installed. I've not yet tested this under windows

I think I can remove the docs directory, but I'll have to modify the configure script, I'll make a couple more tweaks then submitt a new pull

TurkeyMan:
>I think the original idea was to statically compile everything so that premake would be standalone instead of relying on dynamic libs
>that being said, it looks as if libcurl pulls in other libs dynamically anyway even when you tell it to do a static compile (looking at ldd)

Well, like you say, curl invokes a whole heap of other libs. So it might be a bit misguided in this case. And on all systems I know except windows, curl is a stock-standard library that everyone has present.

>including binary .lib blobs is definitley not the way to go though, the only reason I know of the issues surrounding macosx is because of the previous linked thread above, I've actually been testing this on raspeberry pi with gentoo linux installed. I've not yet tested this under windows

OSX would want to link against the one provided by default with the os no? Linuxes have the package manager for this. Windows is the only problem.
So, assuming we would only need to provide a binary lib for windows... what's the problem? That's super common practise for windows because of the lack of package manager.
It would eliminate a significant complication to the build process which would only affect one platform. I don't like non-uniform complexity.
Is the curl build self-contained on windows, or does it link any other libs? I wonder if the windows problem will cascade?

Also, considering that we're cross-compiling the windows builds, we can probably get it from the linux package manager in that case too. Leaving the one case of user-built binaries, in which case, I think it would be better to supply a binary than to require an extra step in the build process for that one case.

It just feels like pretty poor form for linux/osx to re-build curl, since curl is such a terribly standard lib, and especially for internet libs, it is desirable for the os to be able to roll-out security updates and whatnot.

Anyway, that's just my 2c. I don't know what @starkos had in mind.

starkos:
Before this goes too far: as I mention in the original discussions, I don't want to require autoconf to build Premake, Premake should be able to configure itself. In my investigation it looked very possible, but required some work (easy, but a lot of it) to generate config.h, or hand assemble the necessary headers for each platform, or a mix of both.

If you're planning to package this up as a third-party module that's cool, you can do what you want there (though they we need to dynamically link Lua to enable native code plugins, another whole issue).

TurkeyMan:
@starkos What's your feeling on just sourcing curl from the system? It shouldn't need anything special in the build system in that case. Just `links "curl"`, and for windows, provide a libpath which points to a prebuilt lib.

grbd:
hmm this seems to leave 3 options
1. wait for or try to implement a 3rd party module loader
2. re-implement the libcurl build system with premake scripts
3. load libcurl from a pre-existing linux install as TurkeyMan has suggested, although windows would need special consideration in terms of compilation / pre-existing binaries

any opinions on option 3? or would libcurl need to be statically compiled in?

tvandijck:
We have it in a 'contrib' folder, and with a premade config.h that requires no configure stage... if anyone is interested...

we also added zlib and libzip, so we can download .zip's and unpack them...

starkos:
@tvandijck: That would probably be good to see if possible, thanks!

The Windows version of curl uses a hand-built config.h already, so we potentially build from source for that version and use the system libraries everywhere else, without having to jump through too many hoops ourselves.

And yes, curl will need to be static linked into the Windows version so we can continue to ship a single executable.

tvandijck:
I submitted it as a separate pull request... I did not add any lua API in that pull though... it just adds the libs, and for us it builds on linux/mac and windows so far...

grbd:
That's cool, I'll have a go at getting it to build and adding in the lua support

tvandijck:
yeah, our API is really very limited.. all we have as API is:

download(url, destinationFolder, [requestType])
and
unzip(sourceFilename, destinationFolder)

we didn't need more, but obviously, if you're going to pull in curl, you might as well provide a more comprehensive 'curl' API.. we'd still want a simple 'download' api, but I can imagine others would want more.

grbd:
To be honest I was only looking towards providing enough functionality for the download of dependencies as part of CppSharp / QtSharp, since that needs to download copies of LLVM and Clang from github via https.
And that simple API would probably be enough for that, the code I submitted here is pretty much the same thing, just a rehash of the code by João Matos for simple http / https downloads. Although it could always be expanded later on.

grbd:
Closing this pull request in favour of tvandijck's which uses premake for the build process under contrib

PR #43, opened 2015-04-14
adding contrib libs (curl, zlib and libzip)

tvandijck:
as requested...


grbd:
I've created a patch here to add in basic http download support
https://raw.githubusercontent.com/grbd/grbd.github.io/master/patches/premake-http-1.patch
also a couple of tweaks to enable / disable the including of the contrib libraries

example:
http.download("http://www.lua.org/pil/1.1.html","test1.txt")

for info:

1. with this version http works, but https doesn't so it looks like openssl or an additional library might be needed in contrib to get https downloads working with libcurl

2. the http support I've added in the patch is very basic, but good enough for downloading files at least, this is just a copy / paste of the code from https://bitbucket.org/premake/premake-dev/pull-request/122/initial-http-support/diff

3. Only tried this under linux so far, need to also try this under windows

tvandijck:
something seems to be amiss:

tdijck@TVANDIJCK-W7 /c/dev/premake (contrib-libs)
$ git apply premake-http-1.patch
error: src/host/http.c: No such file or directory

do you have these changes in a branch somewhere? then I can just add the remote and cherry-pick the changes, and update this pull request..

grbd:
try this one
https://github.com/grbd/premake-core/commit/5129c10e33602274d7dac543aa76f671536c5ede

tvandijck:
awesome, that worked ;)

starkos:
Would be great to get this stuff in—how to we go about it? @grbd can you submit a pull request with everything? Or should we merge this one first?

tvandijck:
I think 'grbd' already send me a pull request with the http api changes, which I then added to this pull request... if you look on the commits tab, there is two commits in this pull... one from me, and one from 'grdb'. so I think we're good...

grbd:
I think this one contains everything needed for http downloads, I closed the other one down
we're still lacking ssl support but it's a start

tritao:
Glad to see the HTTP support being suggested again. I can help getting this in.

@grbd IIRC my pull had working SSL support on all platforms. What are we lacking in this pull that makes it not work?

tvandijck:
This is a "stripped" down version of curl with just http support... because for SSL support we'd have to include OpenSSL, or some other compatible SSL library.... making OpenSSL build from source as a static lib using premake is going to be a pain..

Basically, we don't want any of this to rely on ".so/.dll" files that may or may not be present on the system... basically, there is a desire for zero external dependencies, because it makes deployement of premake harder, although I think on linux/mac this isn't really a big deal, because it has pretty good package management and dependency solutions.. but on windows having .dll's is just a big P.I.T.A.

Maybe curl isn't the answer here though, and maybe on windows there is easier solutions using just the win32 sdk...

tritao:
That's not correct.

My pull supported SSL with no external dependencies. On Windows by using the Windows-native SChannel backend on curl. On OSX it used DarwinSSL backend and on Linux it used OpenSSL backend.

See https://bitbucket.org/triton/premake-cs-fixes/commits/460a6580b6ac2ca58e301026037be7082751b63f#Lpremake5.luaT53.

tvandijck:
sweet,  would you be able & willing to fix this one up? Just send me a pull request on my branch, and we'll get this one up and running...

https://github.com/Blizzard/premake-core/tree/contrib-libs

starkos:
What should we do with this one?

leeonix:
@grbd why not use `os.execute`?
use `curl` and `wget` to download file?
```Lua
os.execute('http://www.lua.org/pil/1.1.html -O test1.txt")
```
Or you can use `luasocket` or `lsocket` or other any lua socket module, to download.
and check `test1.txt` download success.

because premake is a lua console. you can write lua file as yourself submodule.

I think premake must be focus generate build files. not CMake.
my opinion is premake will keep simple and clear.

leeonix:
I love Doug McIlroy's words: 
`
Write programs that do one thing and do it well. Write programs to work together. 
`
This is why I use premake not other tools.

tvandijck:
I'm kind of in the same boat... but unfortunately Premake doesn't work very well with "Mixed Mode" modules... pure lua is OK, but anything that needs C/C++ or any other native code is basically out, and so luasocket of lsocket are not really an options right now...

This is mostly due to the fact that premake statically links against lua... even for our own modules this is a hindrance.

leeonix:
@tvandijck
I can't imagine you said "Mixed Mode" didn't work.
why pure lua is OK? premake is not OK?
maybe my working environment is not complex.
using native code, I prefer to use `require` module to do someting.

tvandijck:
Yes, but because lua is statically linked inside of premake, and modules like "luasocket" depend on "lua.dll", luasocket just fails to load... this is that way so that both the module and the lua scripts that 'require' it actually use the same runtime... but in premake's case, the lua runtime is embedded, and so any module that you require that uses any C/C++ and calls into lua runtime methods, is going to require that runtime, except it is inaccessible.

starkos:
FWIW I'm in favor of adding these libraries, I think they align with Premake's purpose, and bundling them in fits in with Premake's "just one file" philosophy.

That said, it *would* be great if we had the ability to link against a system Lua library and mix-and-match Lua modules. My gut feeling is that it will be possible to get the `luaL_loadfile()` changes into the script side, but I just don't have the bandwidth to tackle it right now.

So…when I asked "what should we do with this one?" I really meant "is it ready to merge or should I hold off a bit?"

tvandijck:
It builds, but it does not expose any of the curl and zlib methods to lua with lua bindings... 

I was going to do that after it got merged as a separate commit, and also a little based on feedback from other users to see what kind of functionality we want exposed... right now all we (Blizzard) needs is a download file and unzip API... but obviously we can do a lot more with curl and libzip.

So is it ready to merge... in my opinion yes... Some other have asked for SSL support, which I disabled because I didn't want to pull in openssl as well... but I think someone made suggestions on how to fix that, but then never made a pull request for that.. So that too is something we can add later if it comes up again...

So I think this is just a start, and more pull requests will follow based on this change...

leeonix:
@tvandijck OH, you said is pool sad.
As your said, merge these module into premake, `this is just a start`. ;-)
that is why I worry about. those will make `premake` become complex.

tvandijck:
yes, absolutely more complex.... with this merge you get the ability to download files and unzip them... which in itself isn't very impressive.... 

however, here we made the following API, which is something I'd love to contribute, although some things will have to be properly abstracted away a little..

```lua
packageman.server = "http://someurl";

import {
        ['zlib']                    = '1.2.8-src',
        ['libpng']                  = '1.6.16-src',
}

project 'test'

     includedependencies { 'zlib', 'libpng' }
     linkdependendencies { 'zlib', 'libpng' }

     files { '**.cpp' }

```

just as a simple example, this will automatically download zlib and libpng from the package server, invoke the premake5-package.lua scripts that are available in those packages, and then generate the vcxproj/makefiles for those packages, as well as the test project and setup the dependencies correctly.... now obviously this is just a simple example... and you can imagine that for Starcraft2 and Heroes of the Storm that dependency graph is quite a bit more complex. But updating to a new version of zlib, is literally a one line (5 character) change.

In our case it supports both prebuilt binary packages, and build from source packages. 
apart from include and link dependencies we have:

```lua
bindirdependencies   -- to setup the bin folder for tools.
copybindependencies  -- copy dlls to the target output..
```

as well as filtering for libs, dlls, etc, so you can do things like:

```lua
linkdependencies {
        ['3DSMaxSDK'] = { 
            '^bmm', '^core', '^geom', '^gfx', 
            '^gup', '^maxutil', '^mesh','^MAXScrpt',
            '^Paramblk2', '^rtmax'
         }
    }
```

leeonix:
yeah, thanks your explaination. I understand your idea.
add a module, using it add other module from remote site, like `Go language`.
that's good. it should be design a package manager.
have to say, Lua haven't these, it is weak.

starkos:
Great, thanks!

tvandijck:
I never expected this to actually become part of premake ;) this is awesome, and we can really build forward on this... So let me ask you the following question... the above API that I very briefly explained, I have full documentation for it here, which I can package up for you to read if you want.. But even though the above works for us, I can imagine the community has it's opinions about it, so maybe we should post it as a proposal API first, get some feedback and then work towards that? what do you think?

tvandijck:
oh, and we made it as a module... because we didn't want to make it as modifications... it's how we started, but with our move to premake 5, we moved everything over to a module...

starkos:
I'd say release it as a module initially, let people test drive it and we'll see where it goes (I'm super curious, to be sure).

leeonix:
@tvandijck hi, I saw you define this
```Lua
	local ENABLE_CURL = false
	local ENABLE_COMPRESSION = true
```
why are you define a newoption for option?
like `--to=path` option
```Lua
	newoption {
		trigger = "to",
		value   = "path",
		description = "Set the output location for the generated files"
	}
```

tvandijck:
That actually came from a change @grbd made on top of my pull request... 
I would remove those options entirely, but that is up to @starkos really...

starkos:
Yeah, I'm not a big fan of those to be honest, and had planned to rework them the next time I needed to touch that file. IMHO they should just away, giving a single canonical build of Premake, but I'd be okay with switchable options too.

PR #44, opened 2015-04-16
modules are now correctly loaded in all situations

dcourtois:
Hi,
I did 2 fixes related to issues with embedded modules loading, but they were not bullet proof. This PR is : now the module loader will try the usual locate with the various paths, and if none worked, it will try finding the same paths from the embedded scripts.
This should cover all the possible uses of the modules.

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -47,17 +47,32 @@
 		end
 
 		local full = dir .. "/" .. base .. ".lua"
-		local p = os.locate("modules/" .. full) or
-		          os.locate(full) or
-		          os.locate(name .. ".lua")
 
-		if not p then
-			-- try to load from the embedded script
-			p = "$/" .. full
+		-- list of paths where to look for the module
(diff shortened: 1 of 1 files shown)


TurkeyMan:
I'll see if this solves the problem I raised on the forum.
Did you test it against one of my scenarios; many modules with same named files?

dcourtois:
If you are refering to the "Most of my modules aren't actions... and they don't work" part when you spoke about embedding modules, it might, but I'm not so sure ...
The problem this pull request fixes is a require failing with the usual error "module not found". So unless the problem you're experiencing is this one, I don't think this PR will solve it.
Going back on the forum to ask more details about that, I missed this part when I first read your post.

dcourtois:
Hi, any news about this pull request ?

starkos:
Sorry, trying to get to these put other things keep coming up!

dcourtois:
Yep, no problem, thanks for accepting it :)

PR #46, opened 2015-04-27
Visual Studio utility project improvements

starkos:
Two new APIs to help with some utility projects that we need locally.

* buildlog() sets the path to generated build log file
* vsplatformtools() sets the VC platform toolset (e.g. "v100", "v110")

Both are currently implemented for VC 2010+ only.


TurkeyMan:
Hmmm, I'm not really convinced on vsplatformtools.
I've been wrestling over this idea in my head too, but in a different context; gcc and clang both have selectable compiler versions, and they can be installed in parallel. Android for instance ships gcc 4.6, 4.8, 4.9, and clang 3.2, 3.4, 3.6, etc, and you can choose which one to use (and in vs-android, you even choose with the same selectbox!)
I think this maps very well to the same concept that may allow you to choose the MSC compiler version.
So I'm thinking this should be something like toolsetversion rather than what you propose. It is related to toolset.
An alternative would be to do what you've been arguing with me to do with 'language', which is to have the fully justified name in 'toolset', ie "gcc4.8", "msc100", and make sure 'toolset.allowed' matches the start of the string, not the whole string, and assume the suffix is a version specifier...?

starkos:
Good points, I'll revisit it.

PR #48, opened 2015-04-28
Modify toolset() to accept an option version number

starkos:
Examples:
GCC with no specific version: toolset("gcc")
GCC version 4.8: toolset("gcc-4.8")
MSC with platform toolset v100: toolset("msc-100") 
Shortcut for above: toolset("v100")

The MSC version gets placed in the <PlatformToolset> element of VC 2010+ project files.


TurkeyMan:
I know I suggested this, but I'm not actually sold on this idea ;)
(I'm still not sold on doing this for language!)
Let me think on it overnight?

Incidentally, why the '-' character? That kinda seems like superfluous syntax. Does it really offer anything other than the opportunity for a user to forget or mistype it?

Also, I'm not entirely sure about this 'v100' alias mechanism. It seems a bit complex. At least, it's not really precedented among the rest of remake code; we're more likely to use a table of allowed values elsewhere. What would such an alias look like for gcc/clang?

starkos:
> why the '-' character

I work with a (proprietary) tool with a number in the name. The dash indicates where the version starts and the tool name ends. It should be possible to make it optional but I think I'd rather have just one format instead of two. If folks complain I'll make it optional.

> Also, I'm not entirely sure about this 'v100' alias mechanism. It seems a bit complex. At least, it's not really precedented among the rest of remake code; we're more likely to use a table of allowed values elsewhere. What would such an alias look like for gcc/clang?

There is still a table of values and the "v" simply aliases to "msc-" before the lookup. The "v100" form is how people will be used to seeing it appear in VS. There would not be an alias for other toolsets (unless it is the main toolset for some widely used UI that presents one?).

TurkeyMan:
Okay. Well I have no reasonable objections, so I think the change is good. I guess the dash is fine.
I'll be happy to have this, I can remove my spurious "gccversion" and "clangversion" api's that I never offered in PR's because I didn't like them.

src/base/tools.lua
```diff
@@ -0,0 +1,40 @@
...
+--    second return value. If no corresponding tool adapter exists,
+--    returns nil.
+---
+
+	function p.tools.canonical(identifier)
+		local parts
+		if identifier:startswith("v") then
+			parts = { "msc", identifier:sub(2) }
+		else
+			parts = identifier:explode("-")
```
TurkeyMan: This line will break for any tools with dash in the name... but since I don't know of any of those yet, we can ignore this for now.

starkos:
Thanks for the feedback, I agree this is better than my first approach!

PR #56, opened 2015-05-07
Fix external casing

starkos:
Standardized the names of the generated external…() functions (externalproject, externalrule) to be all lowercase, matching the rest of the project API.


TurkeyMan:
looks good

PR #59, opened 2015-05-08
Forcing debugcommand for vs201x to be an absolute path

amc522:
Visual Studio will not let you use a relative path for the debug command. cfg.debugcommand is already translating to an absolute path, so I just added a check.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj_user.lua
+++ b/src/actions/vstudio/vs2010_vcxproj_user.lua
@@ -87,7 +87,12 @@
 
 	function m.localDebuggerCommand(cfg)
 		if cfg.debugcommand then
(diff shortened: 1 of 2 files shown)


src/actions/vstudio/vs2010_vcxproj_user.lua
```diff
@@ -87,7 +87,12 @@
 
 	function m.localDebuggerCommand(cfg)
 		if cfg.debugcommand then
-			local dir = p.project.getrelative(cfg.project, cfg.debugcommand)
+			if not path.isabsolute(cfg.debugcommand) then
+				print("Error: debugcommand must be an absolute path")
```
starkos: Premake makes all paths absolute automatically, so this should never happen. It should be enough to remove the call to getrelative().

amc522:
Updated

PR #60, opened 2015-05-08
Fix rule var paths

starkos:
Correct rule variables containing paths for Visual Studio by translating and making relative to the project.

### DIFF
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -383,6 +383,24 @@
 	end
 
 
+---
+-- Prepare a path value for output in a Visual Studio project or solution.
+-- Converts path separators to backslashes, and makes relative to the project.
+--
+-- @param cfg
+--    The project or configuration which contains the path.
+-- @param value
+--    The path to be prepared.
+-- @return
+--    The prepared path.
+---
+
+	function vstudio.path(cfg, value)
+		cfg = cfg.project or cfg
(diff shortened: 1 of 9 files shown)


TurkeyMan:
I kinda wonder if this function could also do substitutions for the builtin variables, ie $(SolutionDir), $(ProjectDir), $(TargetDir), etc?
Any paths that are recognised to be relative to those locations for which there are variables, we could substitute with the proper variables. That would make VS project generation far more natural.

TurkeyMan:
Anyway, this looks good, that was just something to think on.

tvandijck:
> I kinda wonder if this function could also do substitutions for the builtin variables

That would really only be useful if you generate your solution and projects in place though.... in our case we have the build output going into a different folder, for some people even on an entirely different drive (RAMDISK orso)... $(ProjectDir) at that point becomes rather pointless to use.

TurkeyMan:
But in that case, the substitutions wouldn't happen, in that they wouldn't be relative to those variables...
I suggest only if the absolute paths ARE relative to those variables that they should substitute as I think most VS users would expect/prefer.

starkos:
> I kinda wonder if this function could also do substitutions for the builtin variables, ie $(SolutionDir), $(ProjectDir), $(TargetDir), etc?

In theory yes, but it probably makes more sense to do it during or after expanding the tokens instead, to be generic across exporters. The detoken code can already map specific tokens to toolset specific variables (search for "pathVar").

PR #64, opened 2015-05-11
Fixed Copy command

akaStiX:
This contribution fixed copy command


src/base/os.lua
```diff
@@ -11,7 +11,6 @@
 --
 
 	premake.override(os, "execute", function(base, cmd)
-		cmd = path.normalize(cmd)
```
akaStiX: This line caused some weird issues. I don't understand, why it is here.
I have tryed to run simple command to copy files:
```lua
os.execute("{COPY} ../../Code/Dependencies/FBX/lib/".._ACTION.."/x64/debug/libfbxsdk.dll ../../Bin64")
```
But this command is put through path.normalize and it's output is:
**"{COPY} ../../Code/Dependencies/FBX/lib/vs2013/x64/Bin64"**

src/base/os.lua
```diff
@@ -504,7 +503,7 @@
 				return "chdir " .. path.translate(v)
 			end,
 			copy = function(v)
-				return "xcopy /Q /E /Y /I " .. path.translate(v)
+				return "xcopy /Q /E /Y /I " .. path.translate(v) .. " > nul"
```
akaStiX: This is just to suppress confirmation message - i.e. **Files Copied: 1**
If there is going to be a error in a command arguments, xcopy will show the in console. I've tested that :)

akaStiX:
I've fixed the broken tests in another contribution - https://github.com/premake/premake-xcode/pull/5
Looks like Travis-CI doesn't run tests on xcode module during contribution :(

src/base/os.lua
starkos: Yeah no, you can't just remove stuff until the problem goes away. There is probably an issue in path.normalize() in the say it is removing the "../" sequences.

src/base/os.lua
starkos: This really should have been submitted as a separate pull request; I could have merged it while we sorted out the path.normalize() issue. If you want to remove the change above we can let this one go through.

src/base/os.lua
akaStiX: i did i quick look on path.normalize implementation and i couldn't understand, what it's intention is. So if I'd fix it - I could potentially break other places.
I'll take closer look on it path.normalize.

moomalade: Calling path.normalize on os.execute() does seem dodgy. It will treat all the arguments of the command as one long path and munge the arguments.

For example: `os.execute("echo a/b ../c")` will output `a/c`

I don't think we want path.normalize doing special work to figure out if you passed in one path, or multiple separate paths, that would almost always be surprising for the caller, so the two neat solutions I can image are:

1. os.execute() takes multiple arguments. `os.execute("echo","a/b","../c")`. Not quite sure how that could map to os.executef().
2. Teach os.execute() to parse command lines.

Or, 3, stop os.execute() being clever and normalizing paths (what does the normalization functionally add?).

Thoughts?

modules/xcode
```diff
@@ -1 +1 @@
-Subproject commit b0f33f4c9fd5ad54adceb23e0caa7362cd6d5ac3
+Subproject commit de23410910c0afcb1264e1e5f150ec6982d02a5b
```
TurkeyMan: Should you really be updating a submodule reference in this PR?

starkos:
Main fix (" > nul" on xcopy) was merged in #66.

PR #67, opened 2015-05-14
Add shell scripts to generate embedded lua scripts file and Makefile to build premake5

noresources:
Add shell scripts to generate embedded lua scripts file and Makefile to build premake5
* sh/embed.sh: Generate scripts.c like the embed action
* sh/gmake.sh: Generate GNU makefile to build premake5

Changed in "embed" action
* fix inline comments stripping when line starts with whitespace
* strip blank lines
* sort manifest files for better consistency accross platforms

Rationale
One issue I found while building premake is that premake itself is required to build premake. Precompiled binaries are available for Windows and Mac but Linux ones does not work most of the time, especially on ARM.
The two shell scripts of this pull request mimics the "embed" and "gmake" actions. Only bash, perl and sed are required.

Typical usage
```
#  Generate scripts.c in build directory
./sh/embed.sh -o build
#  Generate Premake5.make in build directory
./sh/gmake.sh -o build -e build/scripts.c -t install
# Build premake5 in install directory
make -f build/Premake5.make
```

Tested on Ubuntu 14.04, OS X 10.6 and a Linux BusyBox (ARM)

The changes made to "embed" action are mostly to get a consistent and very similar scripts.c on all platforms

### DIFF
--- a/scripts/embed.lua
+++ b/scripts/embed.lua
@@ -23,13 +23,13 @@
 		s = s:gsub("[^\"']%-%-%[==%[.-%]==%]", "")
 
 		-- strip out inline comments
-		s = s:gsub("\n%-%-[^\n]*", "\n")
+		s = s:gsub("\n%s*%-%-[^\n]*", "\n")
 
 		-- escape backslashes
 		s = s:gsub("\\", "\\\\")
 
-		-- strip duplicate line feeds
-		s = s:gsub("\n+", "\n")
+		-- strip duplicate line feeds and blank lines
+		s = s:gsub("\n%s*\n+", "\n")
 
 		-- strip out leading comments
 		s = s:gsub("^%-%-[^\n]*\n", "")
@@ -89,6 +89,7 @@
 
 	local mask = path.join(_MAIN_SCRIPT_DIR, "**/_manifest.lua")
 	local manifests = os.matchfiles(mask)
+	table.sort(manifests)
 
 
 -- Generate an index of the script file names. Script names are stored
--- a/sh/embed.sh
+++ b/sh/embed.sh
@@ -1,0 +1,327 @@
+#!/usr/bin/env bash
+###################################################
+# Generate embedded lua scripts file 
+# like the premake 'embed' action
+
+usage()
+{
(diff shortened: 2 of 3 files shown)


tvandijck:
looks good to me.. bummer it's a .sh though... windows is out, although I guess maybe it would work in msys/cygwin...

it would be cool if it was a python script for the embed part, that would cover all platforms and would require only python, which is easy to install on all those platforms.

TurkeyMan:
"Require only python"... I think it's far more likely to find bash on a windows machine than Python ;)

TurkeyMan:
That said, my feeling is that it would be better to assure that a bootstrap build is reliably available on all platforms...
If linux is the only existing problem in that department, then a (temporary?) bash script solution might be reasonable. I dunno.

starkos:
> my feeling is that it would be better to assure that a bootstrap build is reliably available on all platforms

I guess this is my feeling as well. The source packages contain generated Makefiles and a scripts.c and should be able to build-and-go everywhere. If they don't, I would love to see that addressed.

That said, I don't see any real harm in having these scripts. Minor nitpick: do we need to introduce an `sh` folder? Can't we just drop these into `scripts`?

noresources:
As you said, source packages contains the pre-generated build scripts and they work perfectly on Linux. 

One of my goals with these scripts was to be able to use premake-core repository as a submodule of other projects which use premake as a build tool. This way ensure I will always have the correct version of premake for the given project (premake API is still a moving target). It is also easier to tweak or merge new commits.

Another goal was to generate all things out-of-source (scripts.c, Makefiles and binaries).

I introduce a "sh" folder to avoid to mix lua and sh files but your are free to move it to "scripts"

starkos:
See [new pull request #96](https://github.com/premake/premake-core/pull/96) — that one has an approach that works on Windows too, and can target VS.

starkos:
Is there anything in this request we want to keep, or did #96 take care of it?

noresources:
Since you merged the other PR, not really.
I think I will move my changes in a dedicated git submodule.

PR #70, opened 2015-05-19
Performance improvement by handling edge .. directories during path join

tvandijck:
On our repository this results in about a 30% performance improvements, because the path.normalize now doesn't have to do as much, and the join can achieve the same result with much less code and less copy-ing of strings.

### DIFF
--- a/src/host/path_join.c
+++ b/src/host/path_join.c
@@ -10,7 +10,8 @@
 
 int path_join(lua_State* L)
 {
-	int i, len;
+	int i;
+	size_t len;
 	const char* part;
 	char buffer[0x4000];
 	char* ptr = buffer;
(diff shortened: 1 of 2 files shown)


TurkeyMan:
Do you think it's the right thing to do? Imply a semi-path.normalize when joining?
I like it, but I wonder if it's empirically 'correct'?
Reckon there's cases where people expect path.join to produce a straight concatenation? (obviously with the appropriate '/' being added)
I wonder if this could cause unexpected breakage...

tvandijck:
I don't really know, but I don't expect that this could really break anything. we added extra tests to test for some odd corner cases, but in the end 'a/b/../c' and 'a/c' are entirely the same as far as the filesystem in concerned.. I'm not sure what use-case there really is for having the 'un-normalized' path. But I agree, if someone has a use for that, then this is breaking that.

TurkeyMan:
Yeah, it's just that it changes behaviour, and I wonder if people expect the new behaviour. They may have logic that relies on what people would typically assume would happen (unlikely?).
I suspect the typical assumed behaviour of this function is not what this change does.

Dunno. @starkos? :)

tvandijck:
Yup, I understand... quite honestly... I wish Path.Combine in C# and PathConcatenate in the Windows SDK would give me a cleaned up normalized path as well actually ;) It's so much easier to read and debug... but yes, it is definitely a change in behavior.

starkos:
I like this idea, actually, even if there wasn't an associated performance improvement (though 30% is a nice bonus!).

TurkeyMan:
He has spoken! :)

TurkeyMan:
I wonder if the next step for this though is to make sure all paths are normalised always somehow?

starkos:
At this point, they probably will be. All paths coming through the project APIs are run through path.getabsolute(), and most others are assembled via path.join().

PR #72, opened 2015-05-19
support framework folders and bundles in gcc for mac.

tvandijck:
This allows you to link against frameworks on mac, and to create .bundle files.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -446,6 +446,8 @@
 			"WPF",
 			"C++11",
 			"C++14",
+
+			"MacOSXBundle"
 		},
 		aliases = {
 			FatalWarnings = { "FatalWarnings", "FatalCompileWarnings", "FatalLinkWarnings" },
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -139,6 +139,20 @@
 
 	end
 
+--
+-- get the right output flag.
+--
+	function clang.getsharedlibarg(cfg)
+		if cfg.system == premake.MACOSX then
+			if cfg.flags.MacOSXBundle then
+				return "-bundle"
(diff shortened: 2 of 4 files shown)


starkos:
I think this might be better as a new kind ("Framework", maybe?) than a flag. The tight coupling with the SharedLib flag feels wrong. Thoughts?

tvandijck:
that's an idea...

tvandijck:
Actually to get back on that, according to the xcode documentation there is several types of bundles...

application bundles, framework bundles and loadable bundles....
so the fact that we want a bundle, is not really a "kind".

- an application bundle is an "application" kind + bundle flag.
- a framework bundle is a "static lib" kind + bundle flag.
- a loadable bundle is a "dynamic lib" kind + bundle flag.

starkos:
A WindowedApp on Mac OS X creates an application bundle, so that case is already covered. Frameworks are very common and I think it would be worth having a specific kind for them. Bundles aren't used nearly as often, and are basically an application with (optionally) a different entry point, so perhaps we treat them like that?

Anyway, getting back to this specific change, I still think the introduction of a "Framework" kind is the way to go.

tvandijck:
OK, what about "dynamic libs" then? which is actually the case we're using it for..

"Battle.net.bundle"

starkos:
Oh, yes, true. I guess I still prefer a kind of "Bundle" over "SharedLib" + a flag. For incompatible configurations (not OS X) we should raise an error during validation, for now at least. What do you think?

tvandijck:
OK, that works for me... I'll have to refactor this on our end then, and see where that takes us... Let me talk it over with our mac team, see what they have to say, since they are the guys writing the new xcode 6 action.

PR #73, opened 2015-05-19
added NoLibSysDir option for gcc.

tvandijck:
for some libs we wish to override default libraries.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -446,6 +446,7 @@
 			"WPF",
 			"C++11",
 			"C++14",
+			"NoLibSysDir",         -- disable adding /usr/lib on linux
 		},
 		aliases = {
 			FatalWarnings = { "FatalWarnings", "FatalCompileWarnings", "FatalLinkWarnings" },
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -229,7 +229,11 @@
 	}
 
 	function gcc.getLibraryDirectories(cfg)
-		local flags = config.mapFlags(cfg, gcc.libraryDirectories)
+		local flags = {}
+
+		if not cfg.flags.NoLibSysDir then
+			flags = config.mapFlags(cfg, gcc.libraryDirectories)
+		end
 
 		-- Scan the list of linked libraries. If any are referenced with
 		-- paths, add those to the list of library search paths


TurkeyMan:
Can you describe this problem some more?

tvandijck:
basically on linux the gcc backend automatically add /sys/lib and /sys/include etc to the makefile so that if you include some kind of system header it will find it... very useful, unless you want a 100% deterministic build against a particular version of these header files.... in our case for example our "svn repository" contains these include files and libs, and we want to explicitly include those, and want to be absolutely sure that no files from the OS are included automatically... so we set this flag.

we do the same for the visual studio sdk and windows sdk...

TurkeyMan:
Do does this mean we're FINALLY getting a bloody Linux SC2 client?! ;)
I'll port it myself if bliz will contract! :P

Back on topic; I have a bikeshed complaint... should it be `NoSysLibDir`, or `NoSystemLibDir`? I find the name grammatically hard to process.

tvandijck:
Our servers are mostly linux, so unfortunately... don't get your hopes up :( 

As for the name: "NoSystemDirectories", or something like that... I can work with that for sure.

TurkeyMan:
The word 'dirs' is precedented throughout. Should this apply to include paths too? `NoSystemDirs` is more general...

starkos:
Possibly related: [I'm adding support for system include/lib directories](https://groups.google.com/forum/#!topic/premake-development/pYGdydZ0Oog).

Kind of leaning toward "NoSysDirs" as the flag names, since you would probably want to turn off both include and library directories at the same time?

tvandijck:
Right now it only applies to the sys/lib directory, so I guess hence the name nosyslibdir... but I think this definitely should apply to the sys/include dir too... we're also only using this flag in the gcc backend, in the visual studio backend we have an entirely different method to achieve this.. so maybe we should come up with something that is more 'common'..

TurkeyMan:
Incidentally, I've never seen an include or lib dir under `/sys`... that's not a place you really ever touch.
Why would there be things in sys? When does that happen?

starkos:
I think I'm going to push back on this, because I think it should be more generic/common. MS C has a "ignore default libraries" flag that does a similar thing; the naming and functionality of this feature should probably encompass that as well. And it should likely also disable the system include directories, as mentioned. The idea of disabling the system libraries is useful though.

tvandijck:
Agreed, if we can come up with a common API, I'd be happy to implement this the way we all agree on.

Issue #33, opened 2015-04-09
buildcommands() don't support multiple commands


starkos:
The current method of specifying custom build commands (the buildcommands, buildinputs, and buildoutputs APIs) don't allow multiple rule sets, with their own inputs and outputs to be specified. The values all get lumped together during project baking and the intended associations get lost.

Probably need to bring back the previous buildrule { ... } syntax.

PR #74, opened 2015-05-19
buildoutputsasinputs

tvandijck:
This flag (true or false) can be used in custom build steps to indicate that the files in the buildoutput table are to be added to the generated solution, without the need to create 'dummy' files for these first, or explicitly mentioning these files in the 'files' table..

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -126,6 +126,12 @@
 	}
 
 	api.register {
+		name = "buildoutputsasinputs",
+		scope = "config",
+		kind = "string"
+	}
+
+	api.register {
 		name = "buildrule",     -- DEPRECATED
 		scope = "config",
 		kind = "table",
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -31,6 +31,7 @@
 		"phonyRules",
 		"cppConfigs",
 		"cppObjects",
+		"customBuildDepends",
 		"shellType",
 		"cppTargetRules",
 		"targetDirRules",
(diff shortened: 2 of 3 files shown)


TurkeyMan:
Should `kind` be boolean instead of string? Or shouldn't `allowed = { "true", "false" }`?
Additionally, I'm not sure I'm personally convinced by the API name...

tvandijck:
yeah, I was kind of wondering the same to be honest... this is an API added almost 8 months ago, so it was a little lost on me... but it's a very useful flag, and we could definitely fix things as needed.

TurkeyMan:
The functionality seems useful, but I wonder if there are other ways?
Is there any other way to imply this relationship without making it explicit?

tvandijck:
I'm not sure if I understand the question ;)

currently if you do:
files { '*.cpp', '*.script' }
filter { 'files:*.script' }
buildcommand { "p.exe <input> <name>.cpp" }
buildoutput { "<name>.cpp" }

<name>.cpp is not added to the vcxproj, because premake has no idea about it...
specifying "buildoutputasinput 'true'" will then add whatever is in buildoutput...

maybe buildoutput should just always do that? although I can see cases where I definitely don't want that so not sure if that would be a good default. but I also don't want to create dummy files so that I get my not yet existing .cpp files in the vcxproj.

TurkeyMan:
So, the outputs are meant to be added as `files` right? There's nothing else that's special going on here?

tvandijck:
I think so yes ;) we should probably really add tests for this before we merge this back... since there is changes to the GCC backend as well... like I said... and old change, but we're still using it.

TurkeyMan:
Yeah, I can see the value in it.
I just find the name awkward, and wonder if it can be described in some other way.

tvandijck:
I'm open to suggestions ;)

starkos:
> Is there any other way to imply this relationship without making it explicit?

My intention, though I never was able to get to it, was to automatically add generated source files back to the build. So if path.iscpp(filename) then add-to-build(). But I can still see the value in a call like this, for non-C/C++ files that maybe have a custom rule associated with them or something.

Though I'm wondering if it would hurt to just automatically drop all build outputs into files()? And then you can removefiles() anything you don't want?

Also, just FYI, these APIs are [likely to be changed at some point](https://github.com/premake/premake-core/issues/33).

starkos:
Going to push back on this one as well: we do need a way to feed files back into the build, but I don't think it should be all-or-nothing like this patch will implement. It isn't uncommon to generate a source file that should be built, and a support file (metadata for editor bindings comes to mind) that should not. Ideally, anything that could be identified as a source file (.c, .cpp, etc.) should get fed back in automatically.

PR #76, opened 2015-05-19
Add system include and library search paths

starkos:
Adds sysincludedir() and syslibdir() APIs, which are added to the "VC++ Directories" in Visual Studio, and decorated with -isystem and -L for GCC and Clang.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -829,6 +829,13 @@
 	}
 
 	api.register {
+		name = "syslibdirs",
+		scope = "config",
+		kind = "list:directory",
+		tokens = true,
+	}
+
+	api.register {
 		name = "system",
 		scope = "config",
 		kind = "string",
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -822,6 +822,13 @@
 	}
 
 	api.register {
+		name = "sysincludedirs",
+		scope = "config",
+		kind = "list:directory",
+		tokens = true,
+	}
+
+	api.register {
 		name = "system",
 		scope = "config",
 		kind = "string",
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -394,7 +394,7 @@
 
 
 	function make.includes(cfg, toolset)
-		local includes = premake.esc(toolset.getincludedirs(cfg, cfg.includedirs))
+		local includes = premake.esc(toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs))
 		_p('  INCLUDES +=%s', make.list(includes))
 	end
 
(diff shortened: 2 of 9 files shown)


TurkeyMan:
That's a whole lot of commits for a few lines O_O

Can you tell what distinguishes a 'sys' dir from a normal one? I think I must have missed something important related to these sys path discussions.

starkos:
The biggest difference is that system includes (as opposed to user includes) are treated as always correct, so no warnings are shown. System directories are also processed after the user directories. In Visual Studio, they show up in a different location in the property sheet UI ("VC++ Directories").

We have an issue supporting DirectX on Windows 7 that requires some directories to be set there.

tvandijck:
We have this in our "Visual Studio/10" package..
basically in our premake, if a program has  "import { ['VisualStudio'] = '10.0' }"
the package manager will download the "base" for that package, which contains this scripts and runs it.. and basically sets to VC++ directories to point to itself, instead of the standard SDK included with the installation... Once these paths are set, everything else is ignored, and we can safely use the normal "includes" for things like the Windows SDK, and Direct X...

What problems are you having supporting DirectX on Windows 7?? we're using it as well, and have no issues with it...

   return function(operation)
        if _ACTION ~= 'vs2010' then
            error('Wrong Visual Studio package version selected for action \'' .. _ACTION .. '\'.', -1)
        end

        p.override(vc2010.elements, "outputProperties", function(base, cfg)
            local calls = base(cfg)
            table.insertafter(calls, vc2010.executablePath, vc2010.overrideSDK)
            return calls
        end)

        function vc2010.overrideSDK(cfg)
            local dir = project.getrelative(cfg.project, path.join(cwd, 'include'))
            _x(2,'<IncludePath>%s</IncludePath>', path.translate(dir))

            if (cfg.architecture == "x86_64") then
                dir = path.join(cwd, '../win32-x86_64')
                package.declare('VisualStudio/10.0', 'win32-x86_64')
            else
                dir = path.join(cwd, '../win32-i386')
                package.declare('VisualStudio/10.0', 'win32-i386')
            end

            local libdir = project.getrelative(cfg.project, path.join(dir, 'lib'))
            local bindir = project.getrelative(cfg.project, path.join(dir, 'bin'))

            _x(2,'<LibraryPath>%s</LibraryPath>', path.translate(libdir))
            _x(2,'<ReferencePath>%s</ReferencePath>', path.translate(libdir))

            _x(2,'<SourcePath />')
            _x(2,'<ExcludePath />')
        end
    end

tvandijck:
FYI, I think this addition to premake would be really nice, it would remove a bunch of code in the package script I pasted above, since I can just use "sysincludepaths { }" and "syslibpaths {}"

Unfortunately it doesn't cover enough though, because I will still have to set SourcePath, ExcludePath and ReferencePath as well...

starkos:
> What problems are you having supporting DirectX on Windows 7

The bug report is from a build machine with the Windows 8 Framework SDK and DirectX, and building for Win7, and running into header file conflicts. The workaround is described in "Using DirectX SDK projects with Visual Studio" section of [this KB article](https://msdn.microsoft.com/en-us/library/windows/desktop/ee663275(v=vs.85).aspx).

PR #79, opened 2015-05-20
Restore ordering of include file search paths

starkos:
Sorry I didn't catch this sooner, but we need to maintain control over the ordering of the include file search paths. I removed the table.sort() call, and since I was in there swapped out the path.getrelative() and path.translate() with the new vstudio.path() call.


TurkeyMan:
lgtm

PR #81, opened 2015-05-21
add validation of valid_kinds and valid_languages

tvandijck:
it just prints the "errors" right now, but it should really just abort.

### DIFF
--- a/src/actions/vstudio/vs2010.lua
+++ b/src/actions/vstudio/vs2010.lua
@@ -112,7 +112,7 @@
 
 		-- The capabilities of this action
 
-		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None" },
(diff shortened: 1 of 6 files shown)


starkos:
This should probably use premake.warn(), but should we just raise an error instead? (Also, looks like tests are failing).

tvandijck:
Ah yes, needs an additional nil check, I'll fix it... 
and I think for now .warn is probably a little safer, because even the Visual Studio backends are failing to correctly identify what they support... (missing "Utility" for example), and similar in a lot of the unit-tests...

so maybe we should have .warn for now, and once that is all fixed, change it to .error

tvandijck:
I cleaned it up a little further too, makes the API a little easier to use. passes tests locally..

starkos:
FYI, I'd like to fix up the exporters so they aren't warning for kinds and languages that are actually supported. Will merge if I can find the time to do it…

tvandijck:
I reworked this one to fix those warnings for the visual studio actions, also use premake.warn

PR #84, opened 2015-05-26
add "system" to the filter for solutions, so we can do system specific solution configurations.

tvandijck:
this allows you to write:

```
solution "foo"
     filter { "system:win32" }
         platforms { "x86", "x86_64" }
     filter { "system:linux" }
         platforms { "x86_64" }
```


TurkeyMan:
Awesome, this is so useful!

Lgtm, but I'll let @starkos merge since I haven't messed with the filter stuff.

PR #96, opened 2015-06-06
Add bootstrap scripts to create premake initial executable

moomalade:
On a fresh clone, `bootstrap.sh` or `bootstrap.bat` can be used to prepare the initial premake executable on Windows, OSX and Linux.
The bootstrap scripts are stripped from the source releases.

### DIFF
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -26,7 +26,15 @@ BUILDING FROM A SOURCE PACKAGE
 
 BUILDING FROM THE REPOSITORY
 
- If you have pulled sources from the Premake source repository, you will need
+ If you have pulled sources from the Premake source repository, you can run
+ the `bootstrap.sh` or `bootstrap.bat` script to generate your first premake
+ executable.
+
+    $ git submodule init
+    $ git submodule update
+    $ ./bootstrap.sh
+
+ If your toolset is not supported by the bootstrap script, you will need
  to embed the scripts into a C source file so they may be built into the
  executable, and also generate the project files for your chosen toolset. In
  order do either of these things, you will need a working Premake executable.
--- a/bootstrap.bat
+++ b/bootstrap.bat
@@ -1,0 +1,58 @@
+@echo off
+setlocal enabledelayedexpansion
+
+rem Detect Visual Studio Version
+
+if not "%VS140COMNTOOLS%" == "" (
+  set _VSCOMMON="%VS140COMNTOOLS%"
+  set _VSVERSION=vs2015
+) else ( if not "%VS120COMNTOOLS%" == "" (
+  set _VSCOMMON="%VS120COMNTOOLS%"
+  set _VSVERSION=vs2013
+) else ( if not "%VS110COMNTOOLS%" == "" (
+  set _VSCOMMON="%VS110COMNTOOLS%"
+  set _VSVERSION=vs2012
+) else (
+  echo Visual Studio 2012, 2013 or 2014 not found!
+  exit /B 1
+) ) )
+
+echo Found %_VSVERSION%
+call %_VSCOMMON%vsvars32.bat
+
+rem Figure out which source files we want
+
+set LUA_SRC_DIR=%~dp0src\host\lua-5.1.4\src
+set SRC_DIR=%~dp0src\host
+set BOOTSTRAP_DIR=%~dp0build\bootstrap
+
+set SRC=
+for %%i in ("%SRC_DIR%\*.c") do (
+    if "%%i" == "%SRC_DIR%\scripts.c" ( rem
+    ) else ( set SRC=!SRC!"%%i" )
+)
+
+set LUA_SRC=
+for %%i in ("%LUA_SRC_DIR%\*.c") do (
+    if "%%i" == "%LUA_SRC_DIR%\lauxlib.c" ( rem
+	) else if "%%i" == "%LUA_SRC_DIR%\lua.c" ( rem
+	) else if "%%i" == "%LUA_SRC_DIR%\luac.c" ( rem
+	) else if "%%i" == "%LUA_SRC_DIR%\print.c" ( rem
+    ) else ( set LUA_SRC=!LUA_SRC!"%%i" )
+)
+
+rem Build a bootstrap executable
+
+IF NOT EXIST "%BOOTSTRAP_DIR%" ( mkdir "%BOOTSTRAP_DIR%" )
+cd "%BOOTSTRAP_DIR%"
+cl /Fepremake_bootstrap.exe /DPREMAKE_NO_BUILTIN_SCRIPTS /I"%LUA_SRC_DIR%" %LUA_SRC% %SRC% user32.lib ole32.lib
+cd %~dp0
+
+rem Embed scripts and build premake on the detected version of Visual Studio
+
+"%BOOTSTRAP_DIR%"\premake_bootstrap embed
+"%BOOTSTRAP_DIR%"\premake_bootstrap %_VSVERSION% --to="%BOOTSTRAP_DIR%"
+msbuild.exe "%BOOTSTRAP_DIR%"\Premake5.sln -nologo -consoleloggerparameters:NoSummary;Verbosity=minimal -m -p:Configuration=Release
+
+echo Bootstrap complete.
+echo %~dp0bin\release\premake5.exe
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+set -e
+TOP="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/. && pwd )"
+
+UNAME=`uname`
+case $UNAME in
+'Darwin')
+    EXTRA_PREMAKE_CFLAGS="-framework CoreServices"
+    CC=${CC-clang}
+    ;;
+'Linux')
+    EXTRA_PREMAKE_CFLAGS="-lm"
+    CC=${CC-gcc}
+    ;;
+MINGW32*)
+
+    # Use gcc on mingw if we have it,
+    # Otherwise chain to the Visual Studio bootstrap.
+
+    if [ "$CC" == "" ] ; then
+        if [ "$(which mingw32-gcc)" != "" ] ; then
+            CC="mingw32-gcc"
+        elif [ "$(which gcc)" != "" ] ; then
+            CC="gcc"
+        fi
+    fi
+
+    if [ "$MAKE" == "" ] ; then
+        if [ "$(which mingw32-make)" != "" ] ; then
+            MAKE="mingw32-make"
+        fi
+    fi
+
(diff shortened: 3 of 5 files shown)


moomalade:
Tested on Windows 7 with VS2015, Ubuntu 13.10 with GCC 4.8.1, and OSX with Apple LLVM version 6.1.0.

If the lua distro source files we don't want were moved to an 'unused' directory, these scripts would be simpler. Strictly speaking we don't need to exclude scripts.c either, it's benign to add in.

moomalade:
Updated with mingw gcc support on windows. If gcc isn't installed it will still use Visual Studio.

starkos:
Hey @TurkeyMan — thoughts on this? We now have two pull requests (see [#67](https://github.com/premake/premake-core/pull/67)) to improve bootstrapping. This is probably the better of the two.

I think it may still be better/easier to direct people to grab one of the recent source code packages from the download site and use that to bootstrap?

TurkeyMan:
Yeah... like, I'm not personally sold. In some ways it's kinda cool, but it also seems a bit impure.
It doesn't really hurt and some may find it useful, but do you want to find yourself maintaining these scripts?

I guess my key issue is that it doesn't seem simpler than just committing a makefile. Should we just consider committing a pre-rolled makefile instead if this is really a major problem? I've just never had substantial trouble with the bootstrap process, so maybe I can't really comment.

moomalade:
Understood, fair points. If it's not a pain point for anyone else I'll close this out. If it doesn't feel right to the primary maintainers, it's probably not.

The options I see being used to address automated builds in projects using premake:
1. Drop in prebuilt binaries into your repository for each platform.
2. Only use source drops of premake and vendor it into your repo.
3. Drop in a script to build premake.

Maybe you could argue the relative merits of each approach in which is more likely to generate upstream patches back to premake to address fixes and create improvements, but I honestly can't say I've put enough thought into it, just scratching my own itch.

moomalade:
That said, if you do have any thoughts or 'wouldn't it be nice if we did X', then I'm happy to scratch that itch too.

TurkeyMan:
So, personally, I've always used option 1. I think premake historically endorsed option one, and that was the principle reason for the key design decision that premake be a single (and small) self-contained exe.
Do you feel there's issues or limitations to option 1?

Your suggestion is that 2 and 3 may encourage greater participation on premake? It's an interesting thought.

I admit it can be a nuisance to build premake 3 times for committing to your repo, but I probably did this just once like 5 years ago, and then it was done. I've also done it again recently for projects that depend on premake5. It's probably more of a nuisance now that premake5 is yet to have a final release that people can rely on...

Perhaps we could address this by creating a distribution that contains a pre-built binary for each platform in one package? Then average users would never need to touch the source.
Or just commit a pre-built makefile to the repo...? Trouble with makefile is, in windows, it's hard to know what compiler to build for.

moomalade:
I think the maintenance burden is a completely valid reason to be wary, and I'm fairly convinced it's not a great approach now. If it's not under CI and perhaps not a common executed path (ie. maybe 3 people use it), it's just complex enough that it will rot.

All the options have merits. Maybe this approach would have general value if enough people were submoduling or checking out tags/heads rather than vendoring source/binary drops. Maybe the shape of an acceptable solution in that case, as you suggested, would be a Lua like super basic gmake and nmake compatible makefile. As it stands, especially pre-stable versions, it's likely better to reject the extra complexity until there's a clearer need.

tvandijck:
I recently needed to make a premake5 build on Centos 5... you have no idea how much this script would have helped.... the "download an existing premake and use that to bootstrap" simply didn't work very well... I have to go back to a version of premake4 that was at least 3 years old, which did not correctly generate the "scripts.c" file.... so I had to basically make a debug build, with effectively an empty/incorrect scripts.c, then use the /scripts= argument to run the debug build to regenerate the scripts.c, then do a release build... it was a big pain in the <beep>..

I've been pondering on how to do the embed better and more independent, and also have it generate lua bytecode instead of script, because it would contain better debug information, such as correct line information. right now a debugtrace on a release build makes absolutely no sense at all.

moomalade:
One approach I tried was making embed.lua plain-old-lua, but it didn't feel maintainable. If it broke, it would not be obvious why, adding to the confusion. Regardless, it's easy enough to build without scripts.c with an #ifdef, there just needs to be a method to getting a compatible host running once.

Where the complexity comes in these scripts is:
- Handles file excludes.
- Handles Visual Studio versions.
- Builds the final executable, rather than just bootstrapping the final executable build files.

If I cut all those out, it's likely it can be a simpler nmake/gmake file.

starkos:
> I recently needed to make a premake5 build on Centos 5.

Now that we have alpha builds available, the recommended process would be to grab the most recent source code package and build using the appropriate provide Makefile in the build/ folder. Would that have helped?

I'm not opposed to shipping a bootstrapping script for folks building out of Hg, just want to make sure it is really going to add value.

tvandijck:
I think if the makefiles are provided that would certainly help yes... do those packages have pre-generated embed files? that at least would get me a working executable that I can then use to do a second embed...

It would still be an extra 2 steps compared to having some other way to embed.
1. get source package
2. build
3. embed with that binary on my own repo
4. make

with a sh/bat file (or something else).
1. run embed thingie
2. make

I think this is one of the biggest barriers to entry for those who wish to contribute to premake btw.

starkos:
It does have a pre-generated embed file too, yes.

TurkeyMan:
@tvandijck Sorry, just to be clear, are you saying there's a significant difference in process between a bootstrap script and if a makefile and scripts.c were commit directly?
If the makefile were commit, you'd just clone, then make... no? What are the extra steps you describe?

tvandijck:
> If the makefile were commit, you'd just clone, then make... no? What are the extra steps you describe?

yes, for vanilla premake that be true.... but we have about 57 commits of changes, and an additional  "blizzard" module to take care of package management, etc... (zlib, curl, etc).

So whatever you give me, will only give me a premake that I can then use to run on my own project to run embed on.... so the steps are:

1. clone, then make (as you suggest)
2. copy executable to my own repository.
3. run embed and gmake.
4. then make... and I have an executable I can actually use...

Now... if there was an .sh/.bat (or something else) that does the embed completely independently, then this whole process becomes...

1. run embed and gmake
2. then make... and I have an executable I can actually use...

TurkeyMan:
Right. But I think yours is a very exceptional case. You also have comprehensive CI systems building your builds. You likely have the most robust premake development infrastructure there is (...shame it's proprietary!) ;)

Yeah I dunno, I can't really comment on this, since it doesn't affect me.
Perhaps it wouldn't hurt to include these scripts, at least until premake5 is gold, and there are build machines producing binaries for every conceivable OS, and versions available in the distro package managers...
@starkos: Perhaps defer to these gentlemen's good judgement on this? (I'm still not convinced, but hey)

tvandijck:
>  (...shame it's proprietary!) ;)

it's not really... but it does need the curl/zip stuff etc.... unfortunately the 'module' system as it currently stands doesn't really allow for C/C++ modules, since Lua is statically linked, so we either need to link curl/zip statically into the binary as well, or we need to make lua a dll... both have their merrits, but things are just hard to integrate...

That said, I've not even discussed internally whether this is a part I can just share, so we're going to need to have to discuss this, but personally I'd be happy to share what we got. It's not like I'm including the source code for Starcraft or something ;)

> Perhaps it wouldn't hurt to include these scripts, 
Just to be clear.... it would make our lives somewhat easier, but it is currently not preventing us from building... things have been working pretty good for that last couple months orso.. So if there is no consensus on scripts like this, then leave uhm out... we can always have our own little python script that does the embed for us if we really run into problems.

Personally, I would like to see the embed actually compile to lua bytecode...

TurkeyMan:
> It's not like I'm including the source code for Starcraft or something ;)

Incidentally, my mate just got a beta invite for LOTV, but I can't play AT with him (which is what we play)... can you fix me up by any chance? ;)

> Personally, I would like to see the embed actually compile to lua bytecode...

That would be nice. Retaining proper file and line references at least would be really nice.

starkos:
You guys are going too fast…I need to get my work done so I can get back to watching sense8 ;)

> Perhaps defer to these gentlemen's good judgement on this? (I'm still not convinced, but hey)

I'm fine with it. What form should it take? Should we accept this request as is?

> so we either need to link curl/zip statically into the binary as well, or we need to make lua a dll... both have their merrits, but things are just hard to integrate…

I have no objection to linking in curl and zip (though we're getting a bit off topic here…)

> Personally, I would like to see the embed actually compile to lua bytecode…

It used to do this. The byte code actually ended up being larger than the stripped sources, but maintaining file and line references would be nice. I'm open to it. Note that the byte code is endian dependent; there is a patch in the history somewhere to work around that.

TurkeyMan:
> The byte code actually ended up being larger than the stripped sources

Hey, if we linked zlib, then we could compress the bytecode :P

moomalade:
> I'm fine with it. What form should it take? Should we accept this request as is?

I'd like to have a stab simplifying this over the weekend to address some of the points raised, as well as make it less to understand/carry if we rework in the future. Byte code builds are worth taking into account here as well, so it's worth a couple of hours hacking to see the shape of it.

moomalade:
I think this is as simple as a plain Makefile could be. Updated branch confused the PR. Commit linked here: moomalade/premake-core@eeea1a453b8b64d46f863aa5e568947a512087d6

Only a couple of cute things. Uses devenv /Upgrade to avoid Visual Studio detection and gets the processor count for make using the posix method. Works on gmake and nmake.

Lua source list could just be wild cards if we rearranged the lua files, but listing them seems more straight forward and easy to understand. Also updated travis.yml to show how osx/linux builds could done in CI easily to avoid rot.

Another alternative is to add a 'bootstrap' configuration to the Premake files and commit those to the branch, suggesting a "make -C build/bootstrap/osx". Post build command can trigger an embed/build as Bootstrap.mak does.

TurkeyMan:
I'm liking this.
Incidentally, do you wanna work this to support mingw aswell? The plan is to do all builds with Travis, by cross-compiling for MinGW from linux.

moomalade:
Host platform mingw bootstrap would be nice, the mingw make env is a little weird on windows, I'll update and open a new PR with this approach for poking.

For a cross compile we only need the host platform bootstrap:
`make -f Bootstrap.mak linux`, `premake4 --os=windows gmake`, then `CC=i686-w64-mingw32-gcc make` as I saw from your branch the painful part is getting the cross compiler to play nice.

TurkeyMan:
Yeah, it's not particularly important. Just thought it might be a one-liner somewhere.
We'll mostly be cross-compiling MinGW, and we can bootstrap a linux build fine.

The reason I thought if it is cus I'm working on a Qt app recently, and the Qt distribution for windows comes bundled with a MinGW toolchain, so you can comfortably dev Qt apps out-of-the-box without VS installed at all. In that case, a MinGW bootstrap might be useful, but don't lose any sleep over it ;)

PR #97, opened 2015-06-06
Add Visual Studio calling convention support

moomalade:
Add callingconvention directive, possible values are: Cdecl, FastCall, StdCall or VectorCall.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -467,6 +467,18 @@
 	}
 
 	api.register {
+		name = "callingconvention",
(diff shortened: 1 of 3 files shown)


TurkeyMan:
LGTM.
Incidentally, gcc also has one called 'regcall' (`-regcall` gcc arg). It's also available in windows via `/Qregcall`, although I think it's only supported by the intel compiler(?) (which we don't support).

I would merge this, but I think any extensions to the core API should go through @starkos.

PR #99, opened 2015-06-06
Enable Makefile environment overrides for CC,CXX and AR.

moomalade:
Overriding toolchains from the command line is useful when swapping between compilers.
For example: `CC=gcc-4.8.1 make -C build`

CC can be set by make itself so we can use following idiom to allow overrides:

     CC ?= cc
       ifeq ($(origin CC), default)
       CC = cc
     endif

If the value is not set, or if it's set to the default make value, then use the premake defined value, otherwise leave it alone.

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -359,17 +359,23 @@
 	function make.cppTools(cfg, toolset)
 		local tool = toolset.gettoolname(cfg, "cc")
 		if tool then
-			_p('  CC = %s', tool)
+			_p('  ifeq ($(origin CC), default)')
+			_p('    CC = %s', tool)
+			_p('  endif' )
 		end
 
 		tool = toolset.gettoolname(cfg, "cxx")
 		if tool then
-			_p('  CXX = %s', tool)
+			_p('  ifeq ($(origin CXX), default)')
+			_p('    CXX = %s', tool)
+			_p('  endif' )
 		end
 
 		tool = toolset.gettoolname(cfg, "ar")
 		if tool then
-			_p('  AR = %s', tool)
+			_p('  ifeq ($(origin AR), default)')
+			_p('    AR = %s', tool)
+			_p('  endif' )
 		end
 
 		tool = toolset.gettoolname(cfg, "rc")
--- a/tests/actions/make/cpp/test_clang.lua
+++ b/tests/actions/make/cpp/test_clang.lua
@@ -31,9 +31,15 @@
 		make.cppConfigs(prj)
 		test.capture [[
 ifeq ($(config),debug)
-  CC = clang
-  CXX = clang++
-  AR = ar
+  ifeq ($(origin CC), default)
+    CC = clang
+  endif
+  ifeq ($(origin CXX), default)
+    CXX = clang++
+  endif
+  ifeq ($(origin AR), default)
+    AR = ar
+  endif
   		]]
 	end
 


starkos:
@TurkeyMan — thoughts on this one?

TurkeyMan:
Haha, yeah I'll admit I pondered this for a bit, and I wasn't really sure... then figured I'd leave it for you ;)

@moomalade So we're clear, is the intent such that when the makefile specifies these variables, the makefile still supports overrides being supplied by the command line?

I think it's just that the make code looks odd to me. Am I correct in guessing that `?=` says that it will assign the variable only if it is unassigned? If so, what does the if block achieve?
I think the premise is fine, I just don't follow the code... not sure if I should be embarrassed or not ;)

moomalade:
This one dates back to a very old premake4 patch we've been maintaining.

    CC = my_compiler

Kills any environment overrides.
You cannot say `CC=/path/to/my/gcc make -C build`, as you would normally expect with an autotools generated makefiles, as your CC override will be ignored.

    CC ?= my_compiler

Means, if CC isn't set, then set it to 'my_compiler'.
On OSX, the `?=` assignment does what you expect. If you override CC in the environment, it'll use the value you specified, otherwise it will default to 'my_compiler'.
On Linux and mingw it won't. If you override CC in the environment it will use it, otherwise it will retain the default value of 'cc'. 

    ifeq ($(origin CC), default)
      CC = my_compiler
    endif

Means, if CC is set to the initial value Make has assigned it, then set it to 'my_compiler', which handles the case correctly.

I've just re-tested and the `ifeq` is all that's actually needed to have expected behavior on all platforms. 
If in principle you're in agreement that environment overrides are a useful feature I can remove the ?= lines.

TurkeyMan:
I think this should be fine.
It just a shame that it feels kinda awkward, but I guess that's the whole problem with makefiles, and the main reason that premake exists ;)

moomalade:
Updated.

TurkeyMan:
I have to wonder, what is the likeliness that people will have CC defined in their environment by default, or unintentionally, and that will cause their `premake gmake` output to fail unexpectedly?

moomalade:
If you define CC I think the expectation is that it will be picked up.
If you set it unwittingly then you have problems everywhere you use Make or a configure script.
That is the intended behavior of Make, and why it defaults it to 'cc' if you don't set it.

It is also the behavior of premake5 now _unless_ you've called toolset().
You could take the position that toolset() act's as a forced override, and that if you want to play nicely with the host environment, don't ever call toolset. That's an option, though I think it's a little artificial unless that's really the intended purpose of toolset(), but could be solved with some documentation.

TurkeyMan:
Yeah, I think this is right. @starkos?
It's a shame it's such an ugly bit of make syntax!

starkos:
Yes that looks good. I suspect it is going to break some of the unit tests though. @moomalade — have you tried to run the tests (`bin/debug/premake5 test` in the top Premake folder)?

moomalade:
The only test that failed I updated in the PR (tests/actions/make/cpp/test_clang.lua).
I ran on windows, osx and linux when I was testing each platform's make variants.

Is there a situation where the tests might fail but travis might pass? Just want to check I've not missed something.

starkos:
> Is there a situation where the tests might fail but travis might pass? Just want to check I've not missed something.

Sorry, of course. I guess I'm still new to this programming thing. :p

TurkeyMan:
I'll take that as a rubber stamp. Merging... :)

PR #102, opened 2015-06-08
add new exceptionhandling and rtti API's, deprecate the corresponding flags.

tvandijck:
OK,

So with the "flags" system, you can set a flags, and later possibly remove it, which reverts the behavior to the "default" system behavior... which is not what I want... sometime I want to explicitly turn things ON or OFF, such as with these values..

I don't want to have an external project having to add a "removeflags { "NoExceptions" }", because it needs exception handling, just because someone might add the flag. But more importantly, removeflags is just not enough, it only allows me to get the default behavior, which may or may not be exceptionhandling enabled.

So internally, we're deprecating ALL flags that have to do with compiler settings, and make proper API out of it...

The question in this case is however, should we have "on", "off" and "default" as possible flags, or should it just be "on" and "off" like what I have in this commit... The downside is pretty obvious to just having "on" and "off", since it invalidated a bunch of tests which I fixed, but it basically now always outputs these flags into the visual studio project... with "default" we could get around that, where "default" would mean, don't output anything into the vcxproj. But, if we go that route, I would want the same for "editandcontinue"

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -24,14 +24,14 @@
 		kind = "string",
 		allowed = {
 			"universal",
-			"x86",
-			"x86_64",
+			p.X86,
+			p.X86_64,
 		},
 		aliases = {
-			i386 = "x86",
-			amd64 = "x86_64",
-			x32 = "x86",	-- these should be DEPRECATED
-			x64 = "x86_64",
+			i386  = p.X86,
+			amd64 = p.X86_64,
+			x32   = p.X86,	-- these should be DEPRECATED
+			x64   = p.X86_64,
 		},
 	}
 
@@ -343,19 +343,34 @@
 	api.register {
 		name = "editandcontinue",
 		scope = "config",
-		kind = "boolean",
+		kind = "string",
+		allowed = {
+			"Default",
+			"On",
+			"Off",
+		},
 	}
 
 	api.register {
 		name = "exceptionhandling",
 		scope = "config",
-		kind = "boolean",
+		kind = "string",
+		allowed = {
+			"Default",
+			"On",
+			"Off",
+		},
 	}
 
 	api.register {
 		name = "rtti",
 		scope = "config",
-		kind = "boolean",
+		kind = "string",
+		allowed = {
+			"Default",
+			"On",
+			"Off",
+		},
 	}
 
 	api.register {
@@ -1264,9 +1279,6 @@
 -----------------------------------------------------------------------------
 
 	clr "Off"
-	editandcontinue "On"
-	exceptionhandling "On"
-	rtti "On"
 
 	-- Setting a default language makes some validation easier later
 
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -347,6 +347,18 @@
 	}
 
 	api.register {
+		name = "exceptionhandling",
+		scope = "config",
+		kind = "boolean",
+	}
+
+	api.register {
+		name = "rtti",
+		scope = "config",
+		kind = "boolean",
+	}
+
+	api.register {
 		name = "enablewarnings",
 		scope = "config",
 		kind = "list:string",
@@ -416,7 +428,7 @@
 			"No64BitChecks",
 			"NoCopyLocal",
 			"NoEditAndContinue",   -- DEPRECATED
-			"NoExceptions",
+			"NoExceptions",        -- DEPRECATED
 			"NoFramePointer",
 			"NoImplicitLink",
 			"NoImportLib",
@@ -426,7 +438,7 @@
 			"NoNativeWChar",       -- DEPRECATED
 			"NoPCH",
 			"NoRuntimeChecks",
-			"NoRTTI",
+			"NoRTTI",              -- DEPRECATED
 			"NoBufferSecurityCheck",
 			"NoWarnings",          -- DEPRECATED
 			"OmitDefaultLibrary",
@@ -1118,6 +1130,24 @@
 	end)
 
 
+	api.deprecateValue("flags", "NoExceptions", nil,
+	function(value)
+		exceptionhandling "Off"
+	end,
+	function(value)
+		exceptionhandling "On"
+	end)
+
+
+	api.deprecateValue("flags", "NoRTTI", nil,
+	function(value)
+		rtti "Off"
+	end,
+	function(value)
+		rtti "On"
+	end)
+
+
 	api.deprecateValue("flags", "Unsafe", nil,
 	function(value)
 		clr "Unsafe"
@@ -1235,6 +1265,8 @@
 
 	clr "Off"
 	editandcontinue "On"
+	exceptionhandling "On"
+	rtti "On"
 
 	-- Setting a default language makes some validation easier later
 
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1097,10 +1097,14 @@
 
 
 	function m.exceptionHandling(cfg)
-		if cfg.flags.NoExceptions then
+		if cfg.exceptionhandling == false then
 			p.w('<ExceptionHandling>false</ExceptionHandling>')
-		elseif cfg.flags.SEH then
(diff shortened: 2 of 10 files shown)


starkos:
I'm in agreement, the flags need to go.

Premake currently defaults to exceptions and RTTI enabled unless you set the flags to turn them off. I think this is reasonable for these features because a) in my experience the toolsets always turns them on by default and b) you really need consistent behavior across toolsets for these features. I think that's a good behavior to keep in this case.

Since VS defaults them to on, I don't think you have to write those elements out if they are set to on. I would have to go generate a default project to see what it outputs.

tvandijck:
Yeah, in addition there appear to be several exceptionhandling modes in Visual Studio

- Sync
- Async
- SyncCThrow

Do we want to somehow support those?? I would have no idea what the equivalents for gcc/clang are to be honest, so I just stuck with "on" and "off" in premake, but I could see someone wanting to have this ability..

starkos:
We'll need to have "on" and "off" for sure. If you want to try to tackle the additional values great, otherwise someone else can add them if they need them later.

tvandijck:
Ok, I'll go with the "someone else can add them if they need them later." then :)

Just to get a clear picture, is there anything that needs to be changed in this? and what?

TurkeyMan:
I don't know if I like that this patch makes the state of these new API's explicit even if they're undefined. Surely it's reasonable that they could be in an undefined state?

Most API's support an undefined state, and in that state they either leave it to the toolset to imply defaults, or premake will produce defaults, but I don't think implying an explicit setting is the best way to implement a default, since it will usually lead to explicit fields in projects rather than absence of those fields (ie, defaults)... if that made sense.

src/_premake_init.lua
```diff
@@ -1118,6 +1130,24 @@
 	end)
 
 
+	api.deprecateValue("flags", "NoExceptions", nil,
+	function(value)
+		exceptionhandling "Off"
+	end,
+	function(value)
+		exceptionhandling "On"
+	end)
```
TurkeyMan: Here we can see that `exceptionhandling` is explicitly defined either way, even if the user never expressed an opinion.

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1607,8 +1611,10 @@
 	end
 
 	function m.runtimeTypeInfo(cfg)
-		if cfg.flags.NoRTTI and cfg.clr == p.OFF then
-			_p(3,'<RuntimeTypeInfo>false</RuntimeTypeInfo>')
+		if cfg.rtti == false and cfg.clr == p.OFF then
+			p.w('<RuntimeTypeInfo>false</RuntimeTypeInfo>')
+		elseif cfg.rtti == true then
+			p.w('<RuntimeTypeInfo>true</RuntimeTypeInfo>')
```
TurkeyMan: Here again, it's explicit either way.
Should we leave the default to VS, or does VS's default contradict premakes prescribed default value?

src/tools/snc.lua
```diff
@@ -41,20 +41,20 @@
...
 	function snc.getcxxflags(cfg)
-		local flags = table.translate(cfg.flags, snc.cxxflags)
+		local flags = {}
 
 		-- turn on exceptions and RTTI by default, to match other toolsets
-		if not cfg.flags.NoExceptions then
+		if cfg.exceptionhandling then
 			table.insert(flags, "-Xc+=exceptions")
+		else 
+			table.insert(flags, "-Xc-=exceptions")
```
TurkeyMan: Explicit either way. Is there room for an 'unspecified' state here?

tests/actions/vstudio/vc2010/test_compile_settings.lua
```diff
@@ -36,6 +36,8 @@
 	<PrecompiledHeader>NotUsing</PrecompiledHeader>
 	<WarningLevel>Level3</WarningLevel>
 	<Optimization>Disabled</Optimization>
+	<ExceptionHandling>Sync</ExceptionHandling>
+	<RuntimeTypeInfo>true</RuntimeTypeInfo>
```
TurkeyMan: And here is the crux of my issue, evidence that the number of explicitly defined things increases, regardless if the user actually defined them or not.
If the case is that premake has prescribed firm defaults for these API's, and VS's defaults contradict that prescription, then I may accept this as necessity.

TurkeyMan:
You know... I'm gonna be sad to lose flags.
I understand the reasons to move away from them, but that doesn't change the fact they're a nice succinct way to declare those binary settings. I feel like this patch might be a mistake.

Maybe a better approach would be a re-think of the flags api, rather than breaking out every boolean flag into their own api?
I wonder if we can have "+Flag" and "-Flag", or something to that effect, rather than this "NoXX" convention we seem to have?

TurkeyMan:
Umm, yeah sorry, I totally didn't read your last paragraph properly! You completely addressed my issues while I was already commenting away! ;)
I think I need to go to bed now.

tvandijck:
> I wonder if we can have "+Flag" and "-Flag", or something to that effect, rather than this "NoXX" convention we seem to have?

there is an idea I can get behind, it would be more inline with how flags work in clang/gcc as well... All I want to achieve however is a library saying, "I NEED exceptions", not "I don't want No Exception", because that says "I'm fine with whatever the default is", which is not explicit enough...

If we stick with flags, instead of proper API... I want "flags { 'Exception' }" for things that need it, and "flags { 'NoExceptions' }" for things that really don't want it.... and "flags {}" for things that don't care.

So the alternative to my change would be to get rid of the "No" in all flags, but auto-generate those. That way it is backwards compatible with what there currently is, and I have my "explicit" behavior.

starkos:
> If we stick with flags, instead of proper API... I want "flags { 'Exception' }" for things that need it, and "flags { 'NoExceptions' }" for things that really don't want it.... and "flags {}" for things that don't care.

What does this do?

```lua
flags { "NoExceptions", "Exceptions" }
```

The problem with flags is a) the issue above and b) there are very few values that are really truly binary. As mentioned above, VS has more options available than just "on" and "off". Even for currently binary values, having an call instead of a flag keeps it more easily extensible (modules, etc.)

tvandijck:
Hehe, yeah I see your point there ;) 

but kind of understand Manu's point too... I like flags for their simplicity, but I often run into problems with the fact that is "On" or "DontCare", instead of "On", "Off" or "DontCare", and so quite honestly even in this particular change, I don't think it is done yet... because I made a boolean out of it, but really it should be a "tristate" (except we don't have that type yet).

Unfortunately in lua the difference between "false" and "nil" is hard to test for... so I would really prefer to make a "string" out of this call, and test for "On", "Off" or "nil", instead of specifying a default state in the _premake_init.lua

TurkeyMan:
I think @starkos has convinced me. I was just throwing thoughts out there :)

The tri-state you are looking for can exist via the boolean flags; it can be either true, false, or undefined, except your implementation doesn't work that way. I don't really like how your implementation explicitly sets the boolean api one way or another. I think if a user doesn't specify the option, it should remain undefined, and the action may imply a default state if it feels it needs to, but from the api perspective, 'undefined' is a useful concept.

starkos:
> I don't really like how your implementation explicitly sets the boolean api one way or another.

That's a legacy thing. When all we had was the NoExceptions flag, not specifying the flag meant exceptions were enabled. I'm just keep that behavior so there are no surprises.

TurkeyMan:
It was a different pair of binary states though. Before it was 'undefined' or 'off'. Now it's 'on' or 'off'... I'm not actually sure it has any effect in terms of project generation, but there is a distinction regardless, and from that perspective, perhaps it would be an improvement to open it to a (preferable?) tri-state situation? I think supporting an 'undefined' state which simply implements the same behaviour that the undefined flag does now (I guess it implies 'on', but it's still not explicit), and adding an explicit on state is closer to the current implementation no?

tvandijck:
Yes, in my local repo I've experimented with this, and will update the pull request with those changes, since I actually like it a lot better... update coming soon...

tvandijck:
All fixed...

PR #103, opened 2015-06-08
add inlining API.

tvandijck:


TurkeyMan:
I'm not wild that this patch doesn't support gmake at least, but I suspect this api may be much more far reaching than that?

src/_premake_init.lua
```diff
@@ -467,6 +467,18 @@
...
 
 	api.register {
+		name = "inlining",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"Disabled",
+			"Explicit",
+			"Implicit"
```
TurkeyMan: I don't think it's super-obvious what 'Implicit' mean in this context.
The other names do make sense to me, but I'm not sure they're the best possible names either.

TurkeyMan: Maybe 'Enabled' would be better than 'Implicit'?

tvandijck: "Explicit" -> "WheneverISaySo"
"Implicit" -> "WhateverTheCompilerThinksIsRight"

But I'm open to suggestions...

tvandijck:
We can certainly add the gmake support for this.... or you know, feel free to send me a pull request for it, and I'll include it into this patch ;)

src/_premake_init.lua
TurkeyMan: Haha.
Yeah, see it's not really 'implicit' (is it implied by anything in particular?), perhaps it's more like 'automatic', or just 'enabled'?

starkos: I think I might prefer "Auto" to "Implicit", or just "On" (and "Off"?)

tvandijck:
Fixed to Auto... which is so far the best suggestion I've heard...

TurkeyMan:
@starkos I still don't feel good about these names. Default and Auto don't feel intuitively distinct to me. Is it just me?

tvandijck:
Not feeling good isn't really very constructive...if I may be so offensive... ;)

These are the visual studio names, we can directly map those as well, but I felt those were even less desirable.

- Default
- Disabled
- OnlyExplicitInline
- AnySuitable

We can debate about this until the end of time I guess, but it will never match both gcc/clang/xcode and visual studio all at once... so pick something and go with it...

TurkeyMan:
Sorry, I'm not trying to be unconstructive, I just didn't have a better suggestion (and deferred to starkos). That doesn't mean we should merge a bad name just because we haven't thought of a good one yet.
This is obviously subjective. If other people feel good about it, then it's just me and it's fine.

I like `OnlyExplicit`, and `AnySuitable`... but obviously they're closely related to the MSC nomenclature. That said, I think we're already fairly closely bound to MSC with this API. GCC/Clang isn't really well expressed here, it has way more options:
-finline-functions  (AnySuitable)
-fno-inline  (Disabled)
-finline-functions-called-once
-finline-limit=n
-finline-small-functions
... [and other minor/uninteresting permutations]

It's not entirely clear to me how to best map those additional concepts, and I don't know if GCC has a parallel for OnlyExplicit.

tvandijck:
Right... I think a lot of those GCC options simply fall under the "AnySuitable" flag to be honest... except for the inline limit maybe.... MSC also has that option, but its a pragma you have to put in code..

#pragma inline_recursion( [{on | off}] )
#pragma inline_depth( [n] )

kinda dumb...

Anyway, again, I'm not against changing the names to whatever it is we "like", but in all honesty I don't think we can ever come up with something that will satisfy ALL compilers, you gotta sometimes pick the least common denomitor and in many cases that turns out to be MSC...

Also at this point changing it has a cost to us, which is not an excuse for 'bad' code obviously so don't let that stop you from change, but I somehow have to add backwards compatibility with the "old" names, so it is mildly annoying ;)

TurkeyMan:
Well I'm happy with `OnlyExplicit` and `AnySuitable`, I think they're suitably explicit. I just didn't like how it's difficult to imagine what the difference between Default and Auto would be, I think AnySuitable is much better.

I think @starkos should have the final say on this. I'm in favour of the change, and I'm sorry that leads to inconvenience, but it's clearly marked 'alpha' ;)

starkos:
I'm actually okay with Auto here, and we'll just be sure to document properly. Another option is to use `AnySuitable` as the official value and support `Auto` as an alias?

starkos:
Thanks for this!

PR #108, opened 2015-06-10
stricter fix for /O2 not liking /RTC1, only when using the debug runtime.

tvandijck:
outputs a lot less of these fields by default...


TurkeyMan:
looks good

PR #110, opened 2015-06-11
Bootstrap Makefile

moomalade:
Following on from discussion in PR #96 .

mingw gcc distros on windows don't install a 'cc', so you need to say: `CC=mingw32-gcc mingw32-make -f Bootstrap.mak mingw`, which makes the instructions a little less neat. But that's also true of the gmake.windows build files. I can take or leave mingw32 bootstrap host platform support.

Travis changes to build for osx would need osx builds enabled on premake travis, which appears to be 'by request'. Right now I see it only building linux on my account, regardless of what you ask for. Maybe this should be teased out into a separate PR for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
+os:
+  - linux
 
-before_install:
- - sudo add-apt-repository -y ppa:codegear/release
- - sudo apt-get -qq update
- - sudo apt-get install premake4
- 
 script:
- - premake4 embed
- - premake4 gmake
- - make -j8
+ - make -f Bootstrap.mak $TRAVIS_OS_NAME
  - bin/release/premake5 test
 
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -33,9 +33,14 @@ BUILDING FROM THE REPOSITORY
     $ git submodule update
     $ make -f Bootstrap.mak PLATFORM
 
- Where PLATFORM can be osx or linux. On Windows use nmake:
+ Where PLATFORM can be osx or linux.
+ On Windows with Visual Studio use nmake:
 
     $ nmake -f Bootstrap.mak windows
+
+ Or on Windows with MinGW use mingw32-make:
+
+    $ CC=mingw32-gcc mingw32-make -f Bootstrap.mak mingw
 
  If your toolset is not supported by the bootstrap Makefile, you will need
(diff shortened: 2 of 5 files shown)


TurkeyMan:
@noresources How does this look to you, since you also have another PR with a stake in this pie?

TurkeyMan:
And yeah, might be a good idea to split the Travis update into a separate PR. I'll test it on the master account before merging.
At least split the line that adds osx to a different PR.

noresources:
Looks a step in the good direction to ease premake build

I'm not sure you need to define a new -DPREMAKE_NO_BUILTIN_SCRIPTS. AFAIK Premake does not use built-in scripts if it is build in debug

moomalade:
Removed the osx line from .travis.yml.

Debug builds still reference the builtin_scripts and builtin_scripts_index symbols.
Either they need to be ifdef'd out, or a stub file added to declare empty versions.
I considered committing a stub scripts.c, or adding a scripts_stub.c that is only included in the bootstrap but opted for the most obvious solution. Happy to swap to another approach.

TurkeyMan:
LGTM! :+1:

leeonix:
I only have vs2010. when use 	nmake -f Bootstrap.mak windows. it has error.
because generate vs2012's solution.

moomalade:
I forgot vs2010 was a supported target. I won't be near a Windows box for a
few days. Changing the bootstrap target to vs2010 should work, but would
need testing on a later VS to ensure it upgrades correctly.




--
Tim Wharton

leeonix:
@moomalade yeah, I saw your Bootstrap.bat, it's good! and I write my bat too, like yours. I don't know why don't use your bat file. if use `nmake` i have to run vsvars32.bat by myself. it isn't automatic.

moomalade:
Yes, with nmake you have to have run vsvars yourself already. If you're
building the from the CLI you'd have to do that regardless, and if you have
multiple versions of VS installed, you can be specific about which version
you want to use.

Anyhow, the problem I think you'll be seeing is on this line:
https://github.com/premake/premake-core/blob/master/Bootstrap.mak#L71

It build vs2012 projects, then upgrades them to whatever you're running on
your CLI.
If it built vs2010 instead, I think you'd be set. We just need to check
they upgrade to a later vs correctly (I assume they do, but who knows).




--
Tim Wharton

PR #111, opened 2015-06-12
Narrow the Visual Studio /O2+/RTC1 compatibility test

starkos:
As discussed in pull request #12, disable runtime checks only when optimizations AND the debug runtime are enabled.


tvandijck:
ah, I did the same in pull request 108...

starkos:
D'oh! Yours wins.

PR #112, opened 2015-06-12
Make application of flag mappings deterministic (and potentially prioritized)

starkos:
Currently, the ways flags are ordered in config.mapFlags() is arbitrary, and changes each time a new entry is added to the mapping table. Make the application deterministic by enumerating a sorted list of fields, and using the field names to look up the values in the map, instead of enumerating the map keys directly.

This also opens up the possibility of controlling the order in which Premake's fields are applied, if that becomes necessary (we may have a need for that, but aren't using it currently).

Trade-off is that the mapping runs a little slower, since it has to enumerate all of the fields instead of just the map keys.

### DIFF
--- a/src/base/config.lua
+++ b/src/base/config.lua
@@ -469,46 +469,52 @@
 			table.insertflat(flags, replacement)
 		end
 
-		-- For each configuration field provided in the mapping, pull the
-		-- corresponding list of values from the configuration
+		-- To ensure we get deterministic results that don't change as more keys
+		-- are added to the map, and to open the possibility to controlling the
+		-- application order of flags, use a prioritized list of fields to order
+		-- the mapping, even though it takes a little longer.
 
-		for field, map in pairs(mappings) do
-			local values = cfg[field]
-			if type(values) ~= "table" then
-				values = { values }
-			end
+		for field in p.field.eachOrdered() do
+			local map = mappings[field.name]
+			if map then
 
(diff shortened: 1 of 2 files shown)


TurkeyMan:
Is the trade-off significant? Are there applications other than Bliz where people have experiencing poor premake performance?
Wonder if it's worth considering Bliz's luajit solution?

TurkeyMan:
Oh yeah, and lgtm.

starkos:
> Is the trade-off significant?

On my machine, a full test suite run is about 0.2 secs slower with this change. The actual difference is going to depend on the number configurations that need to be mapped in your solution, but since the tests are mapping quite a few configurations themselves I wouldn't expect the difference in a typical project to be very noticeable.

tvandijck:
> Wonder if it's worth considering Bliz's luajit solution?

We tried it, but it's not making a significant impact, so we never moved to it...
it's not hard, but it requires a few updates to the premake code to move away from 5.0 deprecated API, which we're still using, such as loadstring, and package.loaders, and some stuff in the detoken API, which have all been replaced with something else. Most of those changes are in the lua 5.2.3 upgrade change...

Anyway, I'm not sure how the LuaJit engine works, but it looks to me like it is unable to optimize some stuff, and then essentially reverts back to interpreted mode. And since I didn't see any improvements I feel like about 90% of premake is in that situation... who knows, your mileage may vary..

PR #115, opened 2015-06-15
fixed a crash when a token contains non-escaped backslashes

dcourtois:
If you use a token containing a non-escaped backslash (e.g. "%{cfg.linktarget.directory:gsub('/', '\\\\')}" then the detoken function will crash with the following error "unfinished string near '<eof>'". This is due to the fact that even though the backslash is escaped, once sent to the loadstring function, there is only one backslash left, thus the crash.

This pull request simply adds a unit test for this case (which I use in almost all my scripts, and which worked fine until recently, oddly enough ..) and the fix simply consists on escaping again any backslashes.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -23,7 +23,7 @@ Since 5.0-alpha2:
 * Fix: Allow tokens in vpaths()
 * Fix: Silence xcopy output from {COPY} command token (StiX)
 * Fix: Remove ".." sequences in path.join()
-* Fix: Allow CC and CXX overrides from command line (Time Wharton)
+* Fix: Allow CC and CXX overrides from command line (Tim Wharton)
 * Fix: Enable solution level filtering on system values
 * Fix: Make configuration flag mapping order deterministic
 * Fix: Map "Win32" to x86 architecture for Visual Studio
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,8 +7,30 @@ for the complete list of changes from the Premake 4.x series.
 
 Since 5.0-alpha2:
 
+* New: CodeLite support
+* New: D language support
+* New: MonoDevelop support
 * New API: buildlog()
+* New API: callingconvention() (Tim Wharton)
 * New API: os.writefile_ifnotequal()
+* New API: sysincludedirs()
+* New API: syslibdirs()
+* New: toolset() can now specify a version number
+* New: Default values and categories for command line options
+* New: Add bootstrapping script for Git repository (Tim Wharton)
+* Fix: Modules are now loaded correctly in all situations (Damien Courtois)
+* Fix: Make Visual Studio debug commands absolute (amc522)
+* Fix: Allow tokens in vpaths()
+* Fix: Silence xcopy output from {COPY} command token (StiX)
+* Fix: Remove ".." sequences in path.join()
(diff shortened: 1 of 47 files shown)


dcourtois:
Someone can take a look at this ? There's a workaround that consist on double escaping backslashes, but it's a bit tedious :p

TurkeyMan:
Fixes relating to escaping always make me super-nervous! ;)
Your fix looks contained, I can't imagine how it can go wrong, so, LGTM.

starkos:
Yes, me too, which is why this hasn't landed yet. Hoping to run it through all my client projects this weekend to check for side effects.

dcourtois:
My description was a bit erroneous : github escaped my example ^^'' (basically the "e.g." thing in my first comment just crash when it shouldn't :p)

starkos:
Okay, sorry for the delay, finally got a chance to try it. The issue I'm seeing is that we have lots of commands that are already properly escaped:

```lua
buildcommands { '"$(MY_SDK_DIR)\\host_tools\\bin\\toolname.exe" "%(FullPath)" -o "$(IntDir)%(Filename).x"' }
```

This did generate (correctly):

```xml
"$(MY_SDK_DIR)\host_tools\bin\toolname.exe" "%(FullPath)" -o "$(IntDir)%(Filename).x"
```

With this change it generates (incorrectly):

```xml
"$(MY_SDK_DIR)\\host_tools\\bin\\toolname.exe" "%(FullPath)" -o "$(IntDir)%(Filename).x"
```

Which points out the *real* reason for your crash: '\\' is equivalent to one backslash, not two. `gsub("one/two/three", "/", "\\")` results in "one\two\three" and not "one\\two\\three" as I think you were expecting.

My thought is that Lua requires backslashes to be escaped in strings, and that this isn't an uncommon requirements. I think in this case the bug was in your own code. Perhaps Premake should output a nicer error message, but otherwise I don't think this is something that should be "fixed" (I may way to use an escaped character code here, after all, though I can't think of a good reason for it off the top of my head…outputting formatted text into a file perhaps?)

I vote to close. Any one else have thoughts on this?

dcourtois:
Hi,
Maybe I wasn't specific enough, but the bug is only for tokens ! It's not a problem with escaping characters in general, it's a real problem due to the way detoken works. What happens is this (with tokens) :

1. In Lua, you have a string like this: "%{cfg.linktarget.directory:gsub('/', '\\')}"
2. In premake, it becomes expanded to this: "return cfg.linktarget.directory:gsub('/', '\')"
3. And then the loadstring is executed, e.g. loadstring("return cfg.linktarget.directory:gsub('/', '\')") which is invalid.

The example you provided with build commands do not have the problem of the tokens since the string are just passed through to the end tool, so the usual escape works. But for tokens, the string is "interpreted" twice, leading to this problem.

starkos:
That second snippet I posted is an actual result after applying this pull request. The extra escaping is not currently being limited to token expansions.

dcourtois:
Ah yes you're right. I'll add your example as a unit test and modify my PR to ensure that only the inside of a token gets correctly escaped. Thanks for taking the time to check this :)

dcourtois:
I updated the pr: now it only escapes what's inside a token (e.g. what's passed to the "loadstring" function) and doesn't modify the rest. I added another unit test to ensure of that.

starkos:
Looks better, thanks!

PR #118, opened 2015-06-17
Embedded modules should now return "should load" test function from _preload.lua

starkos:
Each _preload.lua script should now end by returning a function of the form `function (cfg) … end`, which should itself return true if the module is required by the provided configuration

````
-- for the Xcode module
return function ()
   return _ACTION == "xcode4"
end

-- for the D language module
return function (cfg) 
   return cfg.language == "D" 
end
```

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -25,6 +25,11 @@
 	local m = p.main
 
 
+-- Keep a table of modules that have been preloaded, and their associated
+-- "should load" test functions.
+
+	m._preloaded = {}
+
 
 ---
 -- Add a new module loader that knows how to use the Premake paths like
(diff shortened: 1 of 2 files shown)


TurkeyMan:
Where did the `cfg` arg come from?

src/_premake_main.lua
```diff
@@ -256,6 +267,24 @@
 ---
 
 	function m.postBake()
+		local function shouldLoad(func)
+			for sln in p.global.eachSolution() do
+				for prj in p.solution.eachproject(sln) do
+					for cfg in p.project.eachconfig(prj) do
+						if func(cfg) then
```
starkos: cfg arg comes in right here

TurkeyMan:
Belated LGTM! :)

PR #127, opened 2015-06-24
Finish switching to exceptionhandling() and rtti()

starkos:
Finish switching away from the use of flags. 

* Deprecated SEH flag, added new value to exceptionhandling()
* Fix implementation in a few places where "Default" or unset values weren't considered
* Moved inline if-then implementation to flag maps to enable overrides
* Replaced use of flags everywhere

Once this hits I will also update the modules.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -359,17 +359,7 @@
 			"Default",
 			"On",
 			"Off",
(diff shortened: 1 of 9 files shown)


tvandijck:
LGTM.

Issue #125, opened 2015-06-24
Use environment value add to output dir on Windows, make by gmake.


leeonix:
I write a TotalCommander Plugins **[wlx_csv](https://github.com/leeonix/wlx_csv)**, use it's environment path `COMMANDER_PATH`, you can see my [Premake5.lua](https://github.com/leeonix/wlx_csv/blob/master/Premake5.lua).
```Lua
targetdir '$(COMMANDER_PATH)/Plugins/wlx/csv'
```

my COMMANDER_PATH value is `D:\tools\TC8\totalcmd`
But, I use `premake5 gmake` and make the project and show this
```
==== Building wlx_csv (debug) ====
Running prebuild commands
on
ragel -T1 ../csv/csv.rl
csv.c
Linking wlx_csv
d:/prog/tdm-gcc-32/bin/../lib/gcc/mingw32/4.8.1/../../../../mingw32/bin/ld.exe: cannot open output file D:toolsTC8totalcmd/Plugins/wlx/csv/wlx_csv.wlx: No such file or directory
collect2.exe: error: ld returned 1 exit status
make[1]: *** [D:\tools\TC8\totalcmd/Plugins/wlx/csv/wlx_csv.wlx] Error 1
make: *** [wlx_csv] Error 2
```
the backslash isn't double backslash, so make COMMANDER_PATH invalid.

I wrote a [Makefile](https://github.com/leeonix/wlx_csv/blob/master/Makefile) and use `$@` for `-o` param. this COMMANDER_PATH is available.

I see the Premake generated Makefile
```
TARGETDIR = $(COMMANDER_PATH)/Plugins/wlx/csv
TARGET = $(TARGETDIR)/wlx_csv.wlx
...
LINKCMD = $(CC) -o $(TARGET) $(OBJECTS) $(RESOURCES) $(ALL_LDFLAGS) $(LIBS)
```
Maybe use TARGETDIR directly make that error.
I modify  $(TARGET) to $@, it is pass.
```
LINKCMD = $(CC) -o "$@" $(OBJECTS) $(RESOURCES) $(ALL_LDFLAGS) $(LIBS)
```
or modify $(TARGET), it is pass again
```
LINKCMD = $(CC) -o $(subst \,/,$(TARGET)) $(OBJECTS) $(RESOURCES) $(ALL_LDFLAGS) $(LIBS)
```
But I think, is it `-o` problem?
```
[...]

PR #128, opened 2015-06-24
modify gmake Makefile LINKCMD $(TARGET) to "$@" issue #125

leeonix:
solve issue #125

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -426,9 +426,9 @@
 	function make.linkCmd(cfg, toolset)
 		if cfg.kind == premake.STATICLIB then
 			if cfg.architecture == premake.UNIVERSAL then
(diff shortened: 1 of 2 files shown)


starkos:
Perfect, thanks!

TurkeyMan:
I wonder if this issue also affects dlang: https://github.com/premake/premake-dlang/blob/master/actions/gmake.lua
I didn't entirely understand the context of the bug.

leeonix:
@TurkeyMan I haven't test that dlang module.
if you mean this code lines:
https://github.com/premake/premake-dlang/blob/master/actions/gmake.lua#L237-L242

I think, `$(TARGETDIR)` or `$(TARGET)` using environment value on windows.
If gmake read it, that bug probably come out. Because the backslash in path value is not a "real backslash char"

in issue #125, I test the some code. you can check it.

PR #129, opened 2015-06-25
fix linking of project references

tbasnoopy:
fix the issue, that it makes no different if you make dependson or link statement.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -843,7 +843,7 @@
 -- Generate the list of project dependencies.
 --
 
-	m.elements.projectReferences = function(prj, ref)
+	m.elements.projectReferences = function(prj, ref, linkit)
 		if prj.clr ~= p.OFF then
 			return {
 				m.referenceProject,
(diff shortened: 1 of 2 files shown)


tbasnoopy:
It took a closer look on that you say.

I expect following from premake:
If I make a `dependson "B"` I want a project reference for build order.
if I make a `links "B"` I want a project reference for build order and the linker should automatically add the `b.lib` as input for linking.

The diamond inhirence you talk about, is a problem with static libs. But the world know more than static linking. I found no indicator that, if you input a .lib from a shared lib (or executable), you do duplicate symbols.

Refering to your link (the link in the link ;-) ):
http://blogs.msdn.com/b/vcblog/archive/2010/05/03/flexible-project-to-project-references.aspx

Project A will create a project reference to Project B and Project C *(this is dependson)*
For the Project B set the “Link Library Dependencies” property to “False” *(this is dependson, too)*
Project D will create a project reference to Project B and Project C. In this case the global defaults for the properties (“Link Library Dependencies” set to “True” and “Use Library Dependency Inputs” set to “False”) will define the behavior. *(this is links)*

or take a look on what microsoft says:
https://msdn.microsoft.com/en-us/library/024awkd1.aspx
Gives you the choice of linking in the .lib files that are produced by dependent projects. **Typically, you will want to link in the .lib file.**

So it seems that it is not to use with *extremly care*.

tvandijck:
> I expect following from premake:
> If I make a dependson "B" I want a project reference for build order.
> if I make a links "B" I want a project reference for build order and the linker should automatically add > the b.lib as input for linking.

and that is where we differ in opinion I guess..
If I make a "dependson", I expect a project dependency in the solution for build order, and NOT a link reference.... since I don't actually want to link against it, since if I wanted that, I would have used a links...
If I make a "links" I expect a project references, since I actually want to link...

The whole "propagate" link dependencies down flag is entirely a different story, and while useful I really don't want that kind of implicit behavior.. certainly not because it does not in any way translate to make or any other compiler...

tbasnoopy:
just for clarification:
project dependency != link dependency

and the commit does exact one thing:
if you tell premake link something - then really link it and make only a build order dependency!

tvandijck:
> and the commit does exact one thing:

No... it doesn't... it introduces and extra flag, that I don't want...
now surely I can override this behavior, so kind of whatever I guess... Apart from that, it breaks the build right now, but I'm assuming that can be fixed easily...

but again, this fix doesn't fix the core issue, and whatever it does introduces a flag I don't want. So in my opinion this is just plain wrong... but alas, I concede, I don't think I can explain it much better that I have...

tbasnoopy:
which extra flag is added?

tvandijck:
<LinkLibraryDependencies>true</LinkLibraryDependencies>

starkos:
As @tvandijck has stated, you can't do it this way, it has side effects, and it will break existing projects (including mine). Can you open an issue ticket, or start a discussion on the forums, and we'll figure out the right way to fix it?

PR #131, opened 2015-06-25
allow changing of local debugger type

tbasnoopy:
allow changing of local debugger type

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -251,6 +251,17 @@
 	}
 
 	api.register {
+		name = "debuggertype",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Mixed",
+			"NativeOnly",
+			"ManagedOnly",
+		}
+	}
+
+	api.register {
 		name = "debugpathmap",
 		scope = "config",
 		kind = "list:keyed:path",
--- a/src/actions/vstudio/vs2010_vcxproj_user.lua
+++ b/src/actions/vstudio/vs2010_vcxproj_user.lua
@@ -66,6 +66,7 @@
 			m.localDebuggerWorkingDirectory,
 			m.debuggerFlavor,
 			m.localDebuggerCommandArguments,
+			m.localDebuggerDebuggerType,
 			m.localDebuggerEnvironment,
 			m.localDebuggerMergeEnvironment,
 		}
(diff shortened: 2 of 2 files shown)


TurkeyMan:
I feel like this can probably be worked into one of the (many!) existing debug related API's.
Perhaps `debugger`? These are effectively different debug engines, and I don't think it's too far of a stretch to place them there. D lang also has an alternative VS integrated debugger called Mago which I intend to add to that list.

starkos:
I agree, we're getting overloaded with debugger APIs, merging this with `debugger()` sounds reasonable.

tbasnoopy:
yeah... i hate all of these git clients.
So i have no fun to change this.

starkos:
Is there any other debugger that supports these options? If so, that would be a good reason to split it out into a separate API. Does Eclipse support this in its Java debugger, for instance?

TurkeyMan:
Support mixed/managed mode for Java? These are .NET concepts. As far as I know, they're MSC/.NET specific.
I'm pretty sure they're related to edit-and-continue, which causes VC to use a different debugger.

It does kinda feel awkward to express these all as a different debugger... but I don't know of any situation where the concept is applicable to any other debug engine.
Doesn't feel great either way really :/

TurkeyMan:
Or are you asking whether Eclipse has a mode where it will simultaneously debug a Java + C++ application by firing up 2 debuggers? That's an interesting question for someone who uses Eclipse a lot :)

starkos:
>  are you asking whether Eclipse has a mode where it will simultaneously debug a Java + C++ application by firing up 2 debuggers?

Yes, that. That's what the settings are for in Visual Studio. Seems like MonoDevelop should also have a setting like this?

TurkeyMan:
Yeah, dunno. I guess we'll have to investigate :)
I'll check it out.

starkos:
Thanks—I have Xamarin Studio installed here too, I will take a look at that if I get a chance.

starkos:
I don't see any support these switches in Xamarin Studio, presumably it just ignores these settings if it encounters them in a project. 

[This StackOverflow question](http://stackoverflow.com/questions/2762916/how-to-use-eclipse-to-debug-jni-code-java-c-c) seems to sum up the current state of Eclipse Java/JNI debugging, best that I can tell. It appears that you *can* switch between Java debugging and native code debugging with the Android stack, but there is no concept of "mixed" yet.

With that in mind, it seems reasonable to for a project to want to express "set debugging to native code, and use LDB". So I guess I'm +1 for keeping separate `debugger()` and `debuggertype()` calls, and merging this request as-is.

TurkeyMan:
Yeah, I've thought on this too, and I agree.

PR #132, opened 2015-06-25
Fix preloading of embedded modules

starkos:
os.locate() wasn't able to find the embedded scripts; added an explicit check after scanning the file system.


TurkeyMan:
Thanks for this one!

PR #135, opened 2015-06-29
Add midl compiler tags to Visual Studio vxproj files

rhuvendiek:
Prior to this premake would not generate the right tags to compile idl files. This would make projects who depend on the midl output fail.


starkos:
Looks good! Would it be possible to add a unit test for it, to make sure we don't accidentally break it in the future? You would only need to add a new test or two to `tests/actions/vstudio/vc2010/test_files.lua`.

rhuvendiek:
Sure thing. Please the added commit.

Issue #136, opened 2015-06-30
Default targetdir is incorrect.


tvandijck:
if I make a project that does not have the targetdir set.

```lua
solution 'SC2'
	configurations { 'Release', 'Debug' }
	platforms { 'x86', 'x86_64' }
	
	project 'Test'
		kind 'StaticLib'
	   
	   	files { '**.cpp' }
```

I get this in my visual studio project:

```xml
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
    <OutDir>.\</OutDir>
    <IntDir>obj\x86\Release\</IntDir>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
    <OutDir>.\</OutDir>
    <IntDir>obj\x86_64\Release\</IntDir>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
    <OutDir>.\</OutDir>
    <IntDir>obj\x86\Debug\</IntDir>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
    <OutDir>.\</OutDir>
    <IntDir>obj\x86_64\Debug\</IntDir>
  </PropertyGroup>
```

the "IntDir" properties correctly contain both platform and configuration...
but the "OutDir" are all '.\', making each build overwrite each other.
it would probably be better to make the default behave the same as "IntDir", or not set the property at all and make it use the visual studio default.

starkos:
If we do that, then we have to propagate that behavior to every other tool which Premake supports. It can't just be a VS thing.

tvandijck:
agreed, it seems that this behavior is coming from the 
```lua
function config.buildtargetinfo(cfg, kind, field)
```

so if we fix it there it'll propagate... interestringly the "objdir" is calculated in the oven process... so it's somewhat disconnected from the targetdir entirely...

starkos:
Another option (untested) might be to drop into the configuration section of `_premake_init.lua`, maybe:

```lua
targetdir "%{prj.location}/%{cfg.buildcfg}/%{cfg.platform}"
```

Would have to make sure that works correctly if platforms are not set.

PR #138, opened 2015-06-30
fix targetdir default.

tvandijck:
https://github.com/premake/premake-core/issues/136

### DIFF
--- a/src/base/config.lua
+++ b/src/base/config.lua
@@ -31,7 +31,14 @@
 	function config.buildtargetinfo(cfg, kind, field)
 		local basedir = cfg.project.location
 
-		local directory = cfg[field.."dir"] or cfg.targetdir or basedir
+		local targetdir
+		if cfg.platform then
+			targetdir = path.join(basedir, 'bin', cfg.platform, cfg.buildcfg)
+		else
+			targetdir = path.join(basedir, 'bin', cfg.buildcfg)
+		end
+
+		local directory = cfg[field.."dir"] or cfg.targetdir or targetdir
 		local basename = cfg[field.."name"] or cfg.targetname or cfg.project.name
 
 		local prefix = cfg[field.."prefix"] or cfg.targetprefix or ""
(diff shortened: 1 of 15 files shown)


starkos:
LGTM but tests are failing?

tvandijck:
yeah, a lot of tests are actually testing against "./".
I'll submit fixes...

TurkeyMan:
Aye, LGTM (except CI fail). I've experienced this issue recently.
Nice one!

tvandijck:
So I fixed all the unit-tests I can fix... the xcode module is so annoying though.... I don't have that, and to try and fix that I would need to fork it, update the submodule file to correctly point to it, and then make the branch use it.... which isn't a mergable action.... Any suggestions?? this whole sub-module business is a PITA.... I hate it...

starkos:
Hmm…so now I'm thinking the CI server shouldn't be unit testing the modules, otherwise we're going to end up with a chicken-and-egg problem where you can't update core because the modules don't reflect the necessary changes, and we don't want to change the modules because they will break until core is updated.

The right development sequence would be to make the changes to core, and then to update the modules to catch up.

We don't currently have CI enabled for the module repositories—that would actually be super useful. I didn't do the CI setup so I'm not actually sure what's involved.

> to try and fix that I would need to fork it, update the submodule file to correctly point to it, and then make the branch use it

FWIW I'm working against the same setup, and pushing to multiple repositories. It is certainly less convenient but it hasn't been anything major. I don't update submodules or any of that. I have my own fork of premake-xcode set up as a second remote on the submodule. I make my changes on a local branch in the submodule, push it to my second remote, and then submit a pull request on it. You just have to take care *not* to update the submodule reference in premake-core; the core maintainers can take care of doing that when they (okay, mostly me) accept the request.

I'm not completely opposed to moving everything into core, but let's open a discussion somewhere else about that.

starkos:
I've merged this and fixed up all the tests.

Issue #86, opened 2015-05-28
path.translate always translates to \


tvandijck:
if (lua_gettop(L) == 1) {
		sep = "\\";
	}

Shouldn't that be dependent on the OS that path.translate is called on?

starkos:
Since %99 of the time I'm translating to Windows, I made it the default. If there's a way to check the OS that maintains the current behavior everywhere its used in the core code, I'm good with that. Otherwise maybe another input like "*" or something?

tvandijck:
> Since %99 of the time I'm translating to Windows,
in path.translate you did... but all other path functions use '/' instead... path.normalize, join, etc.

I think honestly using a / instead of a \ would be an option, since that works on all platforms including windows. but I can figure out if we're on posix or not, and use a / only then..

starkos:
Sorry, I didn't really understand that last comment?

PR #139, opened 2015-06-30
fix path.translate.

tvandijck:
https://github.com/premake/premake-core/issues/86


starkos:
From [the path.translate() docs](https://github.com/premake/premake-core/wiki/path.translate): "If `newsep` is not specified a Windows-style backslash is assumed." 

Either you'll need to pass in a wildcard to indicate you really want the platform-specific separate (i.e. "*"), or you will need see which OS is being targeted (i.e. `os.get()`) and update the docs to the new behavior (and pass the tests ;)

PR #145, opened 2015-07-08
Fix #39, working tokens and wildcards in the same path

### LINKED ISSUE #39: files with tokens and "*"
**prerequisites**
```batch
mkdir testfiles
touch testfiles/{x64,win32}{bla,foo}
touch testfiles/{foo,baa}
```

premake5.lua:
```lua
solution "myTestSln"
  location "testfiles/sln"

  platforms { "x64", "win32" }
  configurations { "debug", "release" }

  project "myTestPrj1"
    kind "Utility"
    files "testfiles/%{cfg.platform}*"
```

**Problem**
No files are in the project.

**Cause**
The `files` function calls later in progress the `store` function (api.lua:765).
This function find a "*" and call `os.matchfiles`, which didn't find something. This is okay, because at this time we cannot detoken all tokens but we have to do this in such a case.

```Lua
		store = function(field, current, value, processor)
			print(value)
			if value:find("*") then
				print("glob detected")
				value = os.matchfiles(value)
				for i, file in ipairs(value) do
					value[i] = path.getabsolute(value[i])
				end
			else
				print("no glob detected")
				value = path.getabsolute(value)
			end
			return value
		end,
```

**Fix**
a) detokenize in the function above (in my opinion not a good idea)
b) make os.matchfiles later in the oven. But I have no clue if a complete fileliste is important before the oven runs.

I will do the fix if someone can answer me which solution is the best.

Meoo:
```
Add a "translate" accessor to field kinds, allowing to perform a custom
processing (after detoken has happened)
Detoken is now applied to the "remove" elements
Detoken and translate are now performed in configset.fetch
```
This commit adds the new
`translate = function(field, current, _, processor)`
to field kinds. This function is called after detoken has been performed, and must return a table containing the result of the translation. For file and paths, it performs a wildcard lookup if required.

This code create a new regression error on vstudio_vs2010_rule_vars.onBooleanVar, but I really have no idea how this can happen...

It would be great if some people could test this patch on their solutions (if possible with many wildcards and tokens).

### DIFF
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -743,23 +743,20 @@
 	premake.field.kind("directory", {
 		paths = true,
 		store = function(field, current, value, processor)
-			if string.sub(value, 1, 2) == "%{" then
-				return value
-			elseif value:find("*") then
-				value = os.matchdirs(value)
-				for i, file in ipairs(value) do
-					value[i] = path.getabsolute(value[i])
-				end
-				return value
-			else
-				return path.getabsolute(value)
-			end
+			return path.getabsolute(value)
 		end,
 		remove = function(field, current, value, processor)
 			return path.getabsolute(value)
 		end,
 		compare = function(field, a, b, processor)
 			return (a == b)
+		end,
+
+		translate = function(field, current, _, processor)
+			if current:find("*") then
+				return os.matchdirs(current)
+			end
+			return { current }
 		end
 	})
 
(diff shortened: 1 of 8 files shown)


starkos:
Nice, not many people are willing to delve that deep into the code! I don't see anything that jumps out as an issue, but the CI tests are failing. If you can get those fixed up I will pull this locally and see how it fares against my set of projects.

Meoo:
I just found the bug, it took me more time than expected though.

In detoken.lua:107, there is a "return" where it should be "return value"
The bug is not really related to this issue, but I'm going to add it to the pull request anyway

starkos:
Look at this now, and wondering if we shouldn't move all wildcard expansion to your new code path, and drop it from the setters entirely. When you implemented this, did you run into any cases where the upfront wildcard expansion was necessary?

tvandijck:
removing it from the setters and make it lazy would definitely help with performance in a lot of cases where filters are used to add/remove files for certain platforms. I would entirely support a change like that.

Meoo:
I can't really think of any case where upfront wildcard expansion would be necessary... and with this PR, wildcards are always expanded after the token expansion, but I'm still a bit worried about files and paths starting with a token...
See 75fcb73b which added a special condition for tokens starting with a token, but removed the call to path.getabsolute for them...
I see too many cases; a path could start with a "target.directory" token, in that case the path.getabsolute wouldn't be needed, but if the token is just a directory name (e.g. "cfg.buildcfg"), then we need to use path.getabsolute... the problem being we only know the correct base working directory in the setter.

About performances, I'm not sure it would really help because you would still have to expand the wildcards before the files are removed.

starkos:
Thanks for this!

PR #154, opened 2015-07-14
Added new module that outputs the raw hierarchical structure of a solution

tvandijck:
This is extremely handy for debugging...

### DIFF
--- a/modules/raw/_manifest.lua
+++ b/modules/raw/_manifest.lua
@@ -1,0 +1,5 @@
+return {
+	"_preload.lua",
+	"raw.lua",
+	"raw_action.lua",
+}
--- a/modules/raw/_preload.lua
+++ b/modules/raw/_preload.lua
@@ -1,0 +1,16 @@
+newaction
+{
+    trigger         = "raw",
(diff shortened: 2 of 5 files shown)


benvanik:
Super useful, thanks for doing this! (I was thinking of doing it myself tonight :)

starkos:
I'm currently working on a tutorial for authoring new actions which will show how I think "best practices" (shiver) should be done and a few improvements. Once that's done (should be soon) I'll loop back to this.

starkos:
As I mentioned, I'm working on a similar but more comprehensive module to demonstrate how to create new actions. I had been exporting to JSON, though it just occurred to me now that if I exported to Lua tables instead you could then use Premake itself to manipulate the result.

If something like that were available, would this module still be useful? I think I'd rather decline this one for now?

tvandijck:
We used it for debugging, but yeah, any other thing that would become available that we can use to that extend would work for us.... we just figured it might help someone...

tvandijck:
Closing this, if we want this in the future, I'll reopen...

PR #155, opened 2015-07-14
add zip.extract(src, dst) method.

tvandijck:


tritao:
Looking good, I'm excited to finally be able to use these APIs for package management.

tvandijck:
Yeah, I would love for someone else to extend this further... we're only really using the extract here, but I can imagine a "compress" method would be nice too.. one that takes a folder name, and a zip name.

tvandijck:
ok, finally... all tests pass... I was screwing up the recursive mkdir...

starkos:
This is great, thanks!

Issue #126, opened 2015-06-24
how to add exist props and rules


leeonix:
Premake5 add [Custom Rules](https://github.com/premake/premake-core/wiki/Custom-Rules) and [rules](https://github.com/premake/premake-core/wiki/rules), it's great.

but i have exist file. how to add into my project?

for example [ragel.props](https://github.com/mozy/mordor/blob/next/mordor/ragel.props)

whish you add. Thx.

starkos:
See [externalrule()](https://github.com/premake/premake-core/wiki/externalrule)

leeonix:
That wiki's example is little.
I copy the example code and modify this code to use ragel's rules
```Lua
    rules { 'ragel' }
    externalrule 'ragel'
        location 'ragel'
        filename 'ragel'
        fileextension '.rl'
```

But it generate rule files.
```
D:\workshop\ragel\lua-csv-ragel>premake5 vs2010
Building configurations...
Running action 'vs2010'...
Generating build/csv.sln...
Generating build/csv.vcxproj...
Generating build/test.vcxproj...
Generating build/test.vcxproj.user...
Generating ragel/ragel.props...
Generating ragel/ragel.targets...
Generating ragel/ragel.xml...
Done.
```

leeonix:
And I can't search `externalrule` code. where is it ?

starkos:
It is generated in api.callback()

leeonix:
en, plz test my code. why  externalrule() cannot use.

leeonix:
@starkos I wish an exactly externalrule() usage.

starkos:
The code in the example is what I'm using in a client project here, almost exactly, and it works. When I get a chance I will try to reproduce the issue you're seeing.

leeonix:
@starkos I'll create a branch, name `use-ragel-rules` and copy [rule files](https://github.com/leeonix/lua-csv-ragel/tree/use-ragel-rules/ragel) and modify [Premake5.lua](https://github.com/leeonix/lua-csv-ragel/blob/use-ragel-rules/Premake5.lua)
You can clone one and check it out.

leeonix:
this `externalrule` is real great.
two years ago, I want to add masm.rules for my project.
I modify `premake-4.2` source code add the rules for myself.
If `externalrule` can use, that solve many problem.

starkos:
Haven't forgotten about this by the way, just haven't had a chance to pull your branch yet.

leeonix:
[...]

PR #157, opened 2015-07-15
Fix handling of external rules

starkos:
As reported in issue #126. Added check to prevent generation of external rules and solutions the same way we skip external projects. Also noticed that rules were not turned on for Visual Studio versions later than 2010; fixed that too.


tvandijck:
lgtm

PR #161, opened 2015-07-16
add 'entrypoint' API.

tvandijck:


TurkeyMan:
LGTM. We should probably leave an open issue to remind supporting the other compilers, but this fixes the existing problem case.

TurkeyMan:
Can rebase this? I'd like to see it merged.

tvandijck:
done, I also fixed up a whole bunch of incorrect whitespace... I don't know what happened to that file, but it definitely made this merge a lot more complex than needed.

TurkeyMan:
Are you aware of `.editorconfig` files? (http://editorconfig.org/)
There's one in premake root. Assuming you're a VS user, there is an EditorConfig plugin which can be installed by the in-game extension manager (Tools -> Extension Manager); search for EditorConfig.
Most editors have a parallel solution.
You shall never have whitespace problems again! :P

starkos:
I have the Sublime Text plugin installed…though maybe it isn't on all of my development VMs? I'll check and make sure.

tvandijck:
Yeah, I got it installed for both sublime and visual studio too...
but it doesn't really help in all cases unfortunately... for example it doesn't automatically fix scoping of "end" etc...

PR #163, opened 2015-07-21
Updated extensions

TurkeyMan:
Updated MonoDevelop and CodeLite with new detection mechanism.
...also wrote bonus comprehensive unit tests.


TurkeyMan:
seems my tests don't work on linux. are there tricks i've missed to making tests work properly across platforms? it's doing things like stressing over '.exe', or the direction of the slashes...
I tried `_OS = "linux"` in my test's prepare functions, but that didn't seem to have any effect.

TurkeyMan:
Used `system="windows"` instead, works fine now.

PR #166, opened 2015-07-24
3rd party libraries are now options

dcourtois:
I removed the hardcoded ENABLE_CURL and ENABLE_COMPRESSION values and replaced them by options.
By default Premake will generate a project without any 3rd party library.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -17,13 +17,6 @@
 
 	premake.api.deprecations "off"
 
---
--- Enable contrib support for 3rd party libraries
--- Curl includes support for http / https downloads
--- Compression will eventually include support for ZLib / Zip
-
-	local ENABLE_CURL = false
-	local ENABLE_COMPRESSION = true
 
 --
 -- Register supporting actions and options.
@@ -69,6 +62,18 @@
 	}
 
 
+	newoption {
+		trigger = "withcurl",
+		description = "Enable Curl 3rd party lib"
+	}
+
+
+	newoption {
+		trigger = "withzip",
+		description = "Enable Zlib/Zip 3rd party lib"
+	}
+
+
 
 --
 -- Define the project. Put the release configuration first so it will be the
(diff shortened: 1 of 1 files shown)


leeonix:
yeah! you already do it!
@tvandijck your job.
@starkos what do you think?

tvandijck:
awesome, LGTM ;) I can't click on 'Pull' though, so we gotta wait for @starkos to do that ;)

starkos:
Sorry for the late response, super busy ATM. Can we make the the options `--no-zlib` and `--no-curl` instead? I would rather build these in by default so local builds match the official binary packages, as and so I don't need to write a bunch of caveats into the docs.

Thanks for tackling this!

dcourtois:
Hi, no problem for the late response. And yes, I could change the option, although I would prefer to leave it as "by default, Premake only" as it is. I tried looking for uses of Curl or Zlib through the source code of Premake, but I couldn't find any. Wouldn't it make more sense to keep the default Premake build to the bare minimum ?

What caveats are you refering to ? (if it's just documentation to update, I would gladly do it)

starkos:
> I would prefer to leave it as "by default, Premake only" as it is

The version that I am going to ship to everyone else is going to have those features enabled; the default build that you get from the source also needs to have those features enabled or we're going to get a steady stream of "hey I built from source and now [related APIs] have stopped working" questions.

> What caveats are you refering to ? (if it's just documentation to update, I would gladly do it)

I mean that for every feature we implement on top of zlib/curl/whatever other library we decide to add in the future we're going to have to say "oh, if you built from source you first need to make sure you enabled flag X or this feature you're reading about that is in the official builds won't be available".

Please enable them by default, thanks!

tvandijck:
Also, there is uses of the curl API already in the commit.... and I have two pull requests outstanding that add zlib, and fix some of the curl stuff... so they are 'available'... but it was definitely a... get this commit in first, so we have the framework setup, and then do the API's, since those typically get more feedback as well...

dcourtois:
ok ok I'l change that :)
Just a quick question about the syntax of the option: should I use no-curl or nocurl ? (the other options seem to follow the coding convention of "all lower cases" ?)

starkos:
I think it should be `--no-curl` and `--no-zlib`. Thanks!

dcourtois:
done

starkos:
Thanks! The CI is currently failing, so I want to give this a quick local test before I merge it. I'll try to do that ASAP.

starkos:
Not sure what the error was, I can't reproduce it here. So thanks, merged!

PR #168, opened 2015-07-25
Added some missing SSE versions that I've wanted a few times

TurkeyMan:
VS doesn't support these directly, but GCC/Clang does, and I have missed them.
I have crafted it such that VS will fall back to SSE2 for the versions between SSE2 and AVX.


TurkeyMan:
Not keen on this?

starkos:
Sorry, just hadn't had a chance to look. Thanks!

PR #171, opened 2015-07-26
Add capture support to UTF8 BOM output

starkos:
If `io.utf8()` is called while running unit tests, it would be output directly to the console since there is no active file output while testing. Replace call with premake.utf8() as a companion to premake.w() and ignore while capturing to prevent this spurious output.


TurkeyMan:
I made another PR with the submodule update included (so the build didn't fail)

PR #173, opened 2015-07-27
Llvm toolset - alt

TurkeyMan:
This is an alternative approach to the VS version issue.
This one has simpler logic, but removes support for vs versions without 'v';
ie, `msc-100` must be `msc-v100`.

I'm just offering options. Feel free to suggest other solutions.
Personally, I felt a combination approach just bloated the logic too much. If you want to keep `msc-100` support, I'd stick with the other PR.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -970,11 +970,11 @@
 		end,
 	}
 
-  api.register {
-    name = "customtoolnamespace",
-    scope = "config",
-    kind = "string",
-  }
+	api.register {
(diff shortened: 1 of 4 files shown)


starkos:
I'd really prefer to keep the current syntax if at all possible. We have a lot of scripts using these calls and they aren't all mine, so we'd need to support it for backward compatibility even if deprecate it anyway.

TurkeyMan:
Cool, I prefer the other PR too. I'll close this one.

PR #176, opened 2015-07-27
Fixed path normalization issue

akaStiX:
Fix for the issue discussed here - https://github.com/premake/premake-core/pull/64
I know it's really weird, but it took me a lot of time to understand, what this code is actually doing.
And it also passes tests :)

### DIFF
--- a/src/host/path_normalize.c
+++ b/src/host/path_normalize.c
@@ -5,6 +5,7 @@
  */
 
 #include "premake.h"
+#include <ctype.h>
 #include <string.h>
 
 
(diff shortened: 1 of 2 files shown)


tvandijck:
Can we add some tests to regress this particular problem? it would document the corner case this is trying to solve.... otherwise this looks fine to me...

akaStiX:
While adding the test, I've found a lot of other issues in this function.
It's necessary to rewrite it, instead of trying to fix it.
Example tests that fail, and their outputs:
```lua
test.isequal("d:/", path.normalize("d:/./"))
```
will output "d:".
Leading white spaces are not handled at all. However this is easily fixable.

Any trailing things in a path that's part of the string containing multiple paths is not removed. i.e.
```lua
test.isequal("d:/game ..a/b", path.normalize("d:/game/ ..a/b"))
```
will output "d:/game/ ..a/b"

I've also found bugs in my my patch.

starkos:
What's the status of this one? The comments make it sounds like it might have some issues and should *not* be merged?

akaStiX:
It should **NOT** be merged. I'll post another pull request, when I'll fix it. Hopefully, I'll do that today or tomorrow

akaStiX:
I've uploaded new changeset, that fixes all the issues I could come up with except one. If you'll take a look at a log of Travis, you'll see, that it fails on a simplest case:
```lua
path.normalize("d:/")
```
will output **d:**.
So I wanted to hear your opinion, @starkos on this issue, before fixing it myself. How should I fix it? Do you really have to remove all the trailing '/' or we can leave one and remove extra ones? Or, maybe, I should write a special corner case code, that will try to handle paths that match this pattern **X:/**, where _X_ is a drive letter?

starkos:
Can we just not trim a `:/` sequence?

akaStiX:
I've fixed it. Check the code and answer this questions:
> P.S. What is the naming convention? I am using camelCase in my code, should it be old_school_c_convention? And should this code be C89 compliant?

starkos:
> What is the naming convention?

camelCase is fine. We've been using an underscore "_" where a dot "." might appear in the Lua script (i.e. path_normalize()). For consistency maybe use normalizeSubstring() or normalize_substring()? Otherwise coding style looks fine.

> And should this code be C89 compliant?

Yes, or the older MS compilers will choke.

src/host/path_normalize.c
```diff
@@ -33,55 +24,107 @@ int path_normalize(lua_State* L)
 
 		/* filter out .. */
 		if (ch == '.' && last == '.') {
-			ptr = dst - 3;
-			while (ptr >= buffer) {
+			last = 0;
+
+			const char* ptr = writePtr - 3;
```
starkos: MSC will fail this; variables must be defined at the start of the block.

akaStiX: That's why I was asking about C89 compatibility

akaStiX:
I've created new pull request https://github.com/premake/premake-core/pull/205

PR #177, opened 2015-07-28
Fixed GCC prefix, which was all messed up somehow

TurkeyMan:
Sanitised the behaviour of the gcc tool name.


jimon:
By default cfg.gccprefix is nil (or am I missing something ?), so this code will return nil even if tool name is correct.

TurkeyMan:
Yup. This code is only supposed to return a value in the case of a non-default tool name.

Issue #37, opened 2015-04-10
Change 'solution' to 'workspace'


TurkeyMan:
As discussed on the forum, 'workspace' seems a more natural term for more users.
It would be good to switch this over at some point.

aggsol:
:+1: I agree, solution is as very Visual Studio centric term.

starkos:
This is now complete: the official term in all code and on the wiki is now "workspace". Support for "solution" has been left in place to avoid breaking pretty much everyone; we can talk about deprecating it someday.

TurkeyMan:
What an effort! Congratulations on this one! :)

starkos:
Thanks! I had a drink or seven…

TurkeyMan:
Good plan!

PR #180, opened 2015-07-29
Begin renaming solution to workspace

starkos:
Begin the switch from "solution" to "workspace" (see https://github.com/premake/premake-core/issues/37)

This change is going to touch a *lot* of code so I'm going to try and do it in phases and hopefully not mess anyone up too badly. This request changes the most visible names (the `solution()` API call and the `onSolution()` action callback) so that I can use the new names as I'm writing the action development tutorial. I also hit a few easy renames while I was in there.

### DIFF
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -28,9 +28,9 @@
 			dotnet = { "mono", "msnet", "pnet" }
 		},
 
-		onWorkspace = function(wrk)
+		onWorkspace = function(wks)
 			premake.escaper(make.esc)
-			premake.generate(wrk, make.getmakefilename(wrk, false), make.generate_solution)
+			premake.generate(wks, make.getmakefilename(wks, false), make.generate_solution)
 		end,
 
 		onProject = function(prj)
@@ -43,8 +43,8 @@
 			end
 		end,
 
-		onCleanWorkspace = function(wrk)
-			premake.clean.file(wrk, make.getmakefilename(wrk, false))
+		onCleanWorkspace = function(wks)
+			premake.clean.file(wks, make.getmakefilename(wks, false))
 		end,
 
 		onCleanProject = function(prj)
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -1,7 +1,7 @@
 --
 -- _make.lua
 -- Define the makefile action(s).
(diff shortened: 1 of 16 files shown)


starkos:
Oh—I should mention that all existing code should continue to work without modification. I have no near term plans to deprecate the use of "solution", but all of the documentation and examples will be switched to the new names. Maybe we can consider deprecating the old names (and lots of other stuff, for that matter, in Premake 6).

src/actions/make/_make.lua
```diff
@@ -28,9 +28,9 @@
 			dotnet = { "mono", "msnet", "pnet" }
 		},
 
-		onSolution = function(sln)
+		onWorkspace = function(wrk)
```
TurkeyMan: I fear there may be gravity in this line. There is a tendency to propagate conventions like this...

I'm not sure I like `wrk` here. I'm not sure why, but it makes me feel awkward :)
Considering it's going to appear hundreds or thousands of times throughout premake...
Alternatives that come to mind are `wks` or `wsp`. Perhaps we can get a sample on preference?

I know this is a massive bikeshed (sorry!), But it's best to make a call on it before it proliferates.

tests/testfx.lua
```diff
@@ -244,17 +246,19 @@
...
+	function test.createWorkspace()
+		local wrk = workspace("MySolution")
 		configurations { "Debug", "Release" }
 
-		local prj = project "MyProject"
-		language "C++"
-		kind "ConsoleApp"
+		local prj = project("MyProject")
+		language("C++")
+		kind("ConsoleApp")
```
TurkeyMan: Curious change in pattern; are you moving away from api's without brackets in terms of styling?

TurkeyMan:
This is awesome!

tests/testfx.lua
starkos: I usually use parens when coding, and no parens when project scripting, if that makes sense. This sort of falls between but I didn't like this:

```lua
local prj = project "prj"
```

…and once I changed that it made since to at least be locally consistent.

starkos:
Changed "wrk" to "wks".

tests/testfx.lua
TurkeyMan: Okay, sounds like a good pattern.

TurkeyMan:
Loving it! I'll start migrating code too while working on other things as I encounter it.

PR #182, opened 2015-07-31
File queries

tvandijck:


TurkeyMan:
Looks okay to me. Can you tell where this is being used?

Bekenn:
The main thing here is that file "configs" (fcfg in the code) are really now file "contexts"; files can hold properties just like any other context without needing to be referenced as part of a configuration.  That is, I can query a property on main.cpp without also filtering for debug or release.

Xcode targets don't have separate lists of files for each configuration; instead, each target has a single list of files and that list is shared for all configurations.  Individual files in the list can have attributes associated with them, most commonly special command-line parameters that need to be passed to the compiler (for instance, to disable optimizations when compiling a particular file).

Prior to this change, the Xcode module would have had to check the properties for every configuration of every file and then apply the union of those property lists when generating the settings for the Xcode file reference.  Now, it can just query the file directly.  Any properties set on a file are also inherited by the file configuration (fsub in the code), so nothing is lost to other back-ends.

starkos:
When I introduced config sets and context I had intended to use them to make configuration fetching more generic and flexible and to make project and file configurations behave more consistently. I never got a chance to finish the work, but this definitely looks like a step in the right direction.

PR #183, opened 2015-07-31
Fully bake solution configs.

tvandijck:


TurkeyMan:
Sorry, I'm not quite sure what I'm looking at. What is the practical difference here?

Bekenn:
This change makes solution configs full-fledged configuration objects, just like project configs.

Prior to this change, although premake allowed you to specify properties for configurations at the solution level, the property values were stored at the project level.  That's fine for Visual Studio, where build settings are only applied at the project level, but Xcode has a tiered approach wherein a build setting for a target ("project" in premake-speak) can be inherited from the enclosing Xcode project ("solution" in premake-speak).

A common approach in Xcode projects is to specify generic settings at the Xcode project level and then only override particular values for individual targets.  This change allows the Xcode module to figure out where a property was specified in the lua script and apply it at the corresponding level in the generated Xcode project.

starkos:
I think this is good (and another step toward more generic configuration fetches), I'd just like to give it a little exercise before I merge it. Will try to get to it today or tomorrow.

PR #184, opened 2015-07-31
Local queries

tvandijck:
This allows our 'xcode' module to query for config values only for the local scope. So only things that were set specifically in a filter { configurations:debug }... but also allows you to query for values local to the solution...

This allows for solution, project & config independent queries, when generating the xcode project settings.

### DIFF
--- a/src/base/configset.lua
+++ b/src/base/configset.lua
@@ -63,23 +63,31 @@
 --    the context filtering case-insensitive.
 -- @param ctx
 --    The context that will be used for detoken.expand
+-- @param origin
+--    The originating configset if set.
 -- @return
 --    The requested value.
 ---
 
-	function configset.fetch(cset, field, filter, ctx)
+	function configset.fetch(cset, field, filter, ctx, origin)
 		filter = filter or {}
 		ctx = ctx or {}
 
 		if p.field.merges(field) then
-			return configset._fetchMerged(cset, field, filter, ctx)
+			return configset._fetchMerged(cset, field, filter, ctx, origin)
 		else
-			return configset._fetchDirect(cset, field, filter, ctx)
+			return configset._fetchDirect(cset, field, filter, ctx, origin)
 		end
 	end
 
 
-	function configset._fetchDirect(cset, field, filter, ctx)
+	function configset._fetchDirect(cset, field, filter, ctx, origin)
+		-- If the originating configset hasn't been compiled, then the value will still
+		-- be on that configset.
+		if origin and origin ~= cset and not origin.compiled then
+			return configset._fetchDirect(origin, field, filter, ctx, origin)
+		end
+
 		local abspath = filter.files
 		local basedir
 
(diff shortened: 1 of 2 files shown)


src/base/configset.lua
```diff
@@ -61,22 +61,30 @@
 --    blocks with terms fully contained by this list will be considered in
 --    determining the returned value. Terms should be lower case to make
 --    the context filtering case-insensitive.
+-- @param onlylocal
+--    If true, don't fetch values from parent configs.
```
TurkeyMan: Nitpick; It looks like this comment might have meant `origin` instead? Either that or it's mismatched? The description doesn't look like what the `origin` param appears to be.

TurkeyMan:
LGTM, but it's quite a bit to digest. I've run into cases in generators where this would be useful too.

src/base/configset.lua
tvandijck: Yeah, I noticed that one too... it's from one of the earlier commits, when it was in fact "onlylocal", it got renamed, and the meaning changed, but the comment not updated... I'll have to ask the author about that, and fix this one.

starkos:
I'm unable to apply this one due to the changes from pull request #183.

tvandijck:
I'll rebase ;)

tvandijck:
Rebased... 

@Bekenn can you verify I did this correctly?

starkos:
Any feedback on this one?

BTW, I'm putting together a proposal for a more comprehensive solution to the whole configuration querying thing. At least formalizing what it needs to do, even if I'm not sure how it will actually work yet.

Bekenn:
Sorry, I was away on vacation until a couple days ago; I'll check it out now.

src/base/configset.lua
```diff
@@ -88,40 +96,48 @@
...
-			local value = block[key]
 
-			-- If the filter contains a file path, make it relative to
-			-- this block's basedir
-			if value ~= nil and abspath and not cset.compiled and block._basedir and block._basedir ~= basedir then
-				basedir = block._basedir
-				filter.files = path.getrelative(basedir, abspath)
-			end
+			if not origin or block._origin == origin then
+				local value = block[key]
```
Bekenn: Remove keyword local.

src/base/configset.lua
```diff
@@ -88,40 +96,48 @@
...
 			end
 		end
 
 		filter.files = abspath
 
 		if cset.parent then
-			return configset._fetchDirect(cset.parent, field, filter, ctx)
+			return configset._fetchDirect(cset.parent, field, filter, ctx, origin)
 		end
 	end
```
Bekenn: Maybe add an explicit return nil here.

Bekenn:
This looks fine at a glance; I can actually test it tomorrow at work.

Bekenn:
Ack, I didn't have time.  I'll try to set aside some time for testing it tomorrow, but this looks good to me.

tvandijck:
@starkos if this works for you and doesn't break any tests, go ahead and merge this one... I reviewed it again, and @Bekenn had a look at it too... we're all good...

Bekenn:
I agree with @tvandijck, this looks good.

tvandijck:
jeej ;)

PR #185, opened 2015-07-31
Buffered I/O, and only write generated file if it changed.

tvandijck:
OK, so this adds Buffered I/O to premake, basically generating any file in memory first, then comparing it to existing data on disk, and only writing the file if it actually changed..

This isn't so much about performance, as it is about memory usage as well as making it such that if no changes were made, no files on disk were touched, and hence no reloads are caused in Visual Studio or xcode. 

I made this a native C type, because the existing lua tables are insufficiently memory efficient to support this at reasonable memory usage levels on a project the scale of SC2/Heroes.

Anyway, unfortunately this is going to break the unit-tests, due to a small error in the existing xcode back-end that emits a "\r\n" in the unit-tests where at runtime it would not... It's a single line fix.

### DIFF
--- a/src/host/buffered_io.c
+++ b/src/host/buffered_io.c
@@ -1,0 +1,105 @@
+/**
+ * \file   buffered_io.c
+ * \brief  provide buffered io.
+ * \author Copyright (c) 2014
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include "premake.h"
+
+typedef struct struct_Buffer
+{
+	size_t capacity;
+	size_t length;
+	char*  data;
+} Buffer;
+
+
+int buffered_new(lua_State* L)
+{
+	Buffer* b = (Buffer*)malloc(sizeof(Buffer));
+	b->capacity = 0;
+	b->length = 0;
+	b->data = NULL;
+	lua_pushlightuserdata(L, b);
+	return 1;
+}
+
+
+static void do_write(Buffer* b, const char *s, size_t l)
+{
+	char* data;
+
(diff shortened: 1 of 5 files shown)


TurkeyMan:
This looks really cool to me.
Do you know that you can submit a PR for a submodule, and at the same time, submit a PR (like this one) to the parent module with the submodule reference updated to the revision of the pending PR in the submodule?
It's kinda nice to do that since it will please the unit tester.

tvandijck:
Yeah, I know, the problem is that it is in the xcode module, which I don't have a fork for.... and I know I can just create one, but it would overwrite our new xcode module, and it's just a big PITA, because everytime I try I get all kinds of conflicts and have to manually delete folders and do git reset, etc to fix things.. it's a bit of a big timesink...

starkos:
Fix Xcode bug and merged, with `--no-ff` this time so it shows up on GitHub. Thanks!

src/host/buffered_io.c
```diff
@@ -0,0 +1,105 @@
...
+
+static void do_write(Buffer* b, const char *s, size_t l)
+{
+	char* data;
+
+	if (b->length + l > b->capacity)
+	{
+		size_t cap = (b->capacity * 3) / 2;
+		if (cap <= 65536)
+			cap = 65536;
```
starkos: Why are we limiting the buffer size? I would like to remove this.

starkos: Oh shoot never mind—I get it. Stupid brain.

PR #188, opened 2015-08-01
Chomp newlines in os.outputof()

### LINKED ISSUE #45: os.outputof carries over newlines
Having custom buildoptions (and linkoptions) often require some *-config tools to be run for creating proper flags. I think this is supposed to be done with os.outputof(). However the final newline character gets carried over in the final Makefile. For example:

```
buildoptions { os.outputof("pkg-config --cflags jansson"), os.outputof("wx-config --cppflags") }
```

will result in an unwanted newline in the Makefile:

```
ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -g -I/usr/local/Cellar/jansson/2.7/include
 -I/usr/local/Cellar/wxmac/3.0.2/lib/wx/include/osx_cocoa-unicode-3.0 -I/usr/local/Cellar/wxmac/3.0.2/include/wx-3.0 -D_FILE_OFFSET_BITS=64 -DwxDEBUG_LEVEL=0 -DWXUSINGDLL -D__WXMAC__ -D__WXOSX__ -D__WXOSX_COCOA__
```

causing gmake to fail with ```*** missing separator.  Stop.```

moomalade:
Fixes #45

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -392,6 +392,11 @@
 			exitcode = -1
 		end
 
+		-- chomp trailing newlines
+		if result then
+			result = string.gsub(result, "[\r\n]+$", "")
+		end
+
 		return result, exitcode
 	end
 


TurkeyMan:
I'm gonna merge this since it fixes a bug and makes some common things easier, but I wonder if there's ever a desire for the output verbatim?

moomalade:
Yeah, I wondered too, though it may raise a flag.
I figure it's used for calling sdl-config, wx-config, pkg-config and the like, and hopefully stripping trailing newlines won't affect anyone further parsing the output, but reduce the surprise when it doesn't.

TurkeyMan:
Indeed, that was my judgement when I clicked the merge button :)

PR #189, opened 2015-08-01
Improve module loadfile() error messages

### LINKED ISSUE #143: No error reported when failed to parse extension
If premake fails to parse an "required" extension for some reason, for example :

    if True then print("b") else print("c") else -- here we wrote else instead of end

Then no error reported, but "required" fails with "module not found".
This behavior makes hard to develop extensions.

PS. not sure if it's premake or lua default behavior.

moomalade:
Have the module package loader return the loadfile() error for file system modules,
and report the error from the calling script, rather than globals.lua.

This give marginally better errors when require()'ing modules with syntax errors.
The error will appear in the package loader list with the correct call site.

     Error: /Users/blah/premake5.lua:7: module 'wee' not found:
         no field package.preload['wee']
         load error /Users/blah/modules/wee/wee.lua:3: '=' expected near '<eof>'
	 ....

Fixes #143

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -68,6 +68,9 @@
 				if chunk then
 					return chunk
 				end
+				if err then
+					return "\n\tload error " .. err
+				end
 			end
 		end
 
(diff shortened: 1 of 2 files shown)


TurkeyMan:
Nice. I've run into this problem lots.

PR #191, opened 2015-08-01
fix for path.join problem when both string start with a "..", a unit …

### LINKED ISSUE #190: path.join not working as expected anymore with ../
according to the wiki:
path.join("MyWorkspace", "MyProject")
returns "MyWorkspace/MyProject"

The problem:
p = path.join("../", "../../test")
expected would be "../../../test"
but I get the exact opposite: "../test"

This worked back in alpha1.

Edit: Seems to be caused by this performance improvement:
https://github.com/premake/premake-core/commit/6f20d213288d83a23cd87a8a85f6b71addf2aef6

Lusito:
…test has been added

fixes #190

### DIFF
--- a/src/host/path_join.c
+++ b/src/host/path_join.c
@@ -71,6 +71,10 @@ int path_join(lua_State* L)
 				break;
 			}
 
+			if (start[0] == '.' && start[1] == '.' && start[2] == '\0'){
+				break;
+			}
+
 			/* otherwise trim segment and the ".." sequence */
 			if (start != buffer) {
 				--start;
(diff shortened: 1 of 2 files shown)


TurkeyMan:
LTGM, but @starkos or @tvandijck should give this the thumbs up. I was watching, but wasn't involved in those changes.

tvandijck:
I think we should pull this... BUT I believe we ran into similar issues when doing the os.match changes, which is in another pull request, so I'll have to rebase that one once this one is merged. This one comes with tests, so needs to get priority over my changes for sure...

PR #192, opened 2015-08-02
os.pkgconfig() to allow easy system pkg-config dependencies

moomalade:
It's useful to allow scripts to specify system dependencies using the native package manager.
It's non trivial to implement it as nicely as other premake directives.

Not sure about the API, not completely sold it should be in core, but having the functionality out of the box feels like a good thing, and is super nice to have on platforms where you have it. Happy to rework, module or discard.

     filter "system:macosx or linux"
       os.pkgconfig { "zlib", "openssl", "sdl" }

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -582,3 +582,74 @@
 		end
 		return id
 	end
+
+--
+-- Call system pkg-config to populate build and link fields.
+-- Does nothing on Windows.
+--
+-- @param packages
+--    String or table of packages to configure
+-- @param extra_flags
+--    Optional extra flags to pass to pkg-config
+-- @paths
+--    Optional table of paths for PKG_CONFIG_PATH
+-- @return
+--    None
+-- @example
+--    os.pkgconfig( { "zlib", "openssl" }, "--static", "./vendor/stage"  )
+
+	function os.pkgconfig(packages, extra_flags, paths)
+
+		if os.get() == "windows" then
+			return
+		end
+
+		if type(packages) == "table" then
+			packages = table.concat(packages, " ")
+		end
+
+		local flags = extra_flags or ""
+
+		if type(paths) == "table" then
+			paths = table.concat(paths,":")
+		end
+
+		local function execute_pkg_config( flags, prefix )
+
+			local cmd = "pkg-config "
+
+			if paths then
+				cmd = "PKG_CONFIG_PATH=" .. paths .. " " .. cmd
+			end
+
+			cmd  = cmd .. flags .. " " .. packages
+
+			local pipe = assert(io.popen(cmd, 'r'))
+			local result = assert(pipe:read('*a'))
+			local b, exitcode = pipe:close()
+			if not b then
+				exitcode = -1
+			end
+			if exitcode ~= 0 then
+				error("pkg-config command failed: '" .. cmd .. "'", 3)
+			end
+			result = string.gsub(result, "[ \n]+$", "")
+
+			if prefix then
+				local a = string.explode(result, " ")
+				result = {}
+				for i,v in ipairs(a) do
+					result[i] = string.gsub(v, prefix, "")
+				end
+			end
+
+			return result
+		end
+
+		includedirs  ( execute_pkg_config( flags .. " --cflags-only-I", "-I" ) )
+		buildoptions ( execute_pkg_config( flags .. " --cflags-only-other" ) )
+		links        ( execute_pkg_config( flags .. " --libs-only-l", "-l" ) )
+		libdirs      ( execute_pkg_config( flags .. " --libs-only-L", "-L" ) )
+		linkoptions  ( execute_pkg_config( flags .. " --libs-only-other" ) )
+	end
+


TurkeyMan:
I think it's valuable, and I also think it should probably be in core.
Can you add some unit tests to demonstrate the api usage, and the expected output?

tritao:
I agree this would be really good to have on core, though I think we should provide some way to check if a package is provided by the system so that a fallback can be provided in case it does not.

moomalade:
@TurkeyMan Will do.
@tritao Anything thoughts on the API, maybe pkgconfigopt() to match the other 'opt' functions, which returns a true/false condition. Could also have pkgconfigtest() that just returns if it's possible, but would prefer to keep it smaller:

    project "whatever"
        os.pkgconfigopt("sdl2") or error("sdl2 is required for this project")
        os.pkgconfigopt("zlib") or defines("WITHOUT_ZLIB")

tritao:
I think it would be much more flexible to provide at least one API for finding a package (say  `os.findpackage`) and returning its info as a Lua table and another to use the package (like you already provide).

The first API would be much more flexible and could also return extra information provided by pkg-config, like version and dependencies.

It might also be a good idea to allow a second parameter for supporting different versions of packages, something like: `os.findpackage('zlib', '1.2.8')`.

tvandijck:
it would also be nice if we could come up with something that actually works across platforms. This patch appears linux/unix only...

moomalade:
pkg-config runs on windows, but for a package configuration tool to be useful, it needs to be widely used by software deployments. Most 'make install' steps on OS X and Linux provide pkg-config files that describe their dynamic/static dependencies, include locations etc, which you can source from the system installs, or from whatever local prefix you're building them to.

On Windows I don't think it's used at all, I've certainly never encountered it outside of Linux/OSX.
Are you aware of any alternatives?

starkos:
I would prefer if this were implemented as a proper project API which simply set values into a list to be processed later, rather than trying to do all the work at call time. So…

```lua
project "whatever"
   pkgconfig { "zlib", "openssl", "sdl" }
```

…should simply create a list of `{ "lib", "openssl", "sdl" }` in the configuration, like `defines()` or `includedirs()` or any other project API. Other bits of the script can add new values to the list or call `removepkgconfig()` to remove values.

Then the list can be processed either in the exporter or in the tool adapters—not sure which would be best—to actually populate the right flags.

And I'm with @tvandijck: I'd like to see a fallback code path for systems which don't provide `pkgconfig`. Failing the script isn't a good solution, as it requires all call to this API to be wrapped in `if _OS == "linux" then…` or similar.

tvandijck:
I don't think: 
```lua
project "whatever"
   pkgconfig { "zlib", "openssl", "sdl" }
```

would be enough though.... It's definetely better to do it as an API, but pkgconfig has no real particular meaning... what do you want from the package? a library that depends on a package may only need the header files, and so only wants the includes setup... but an application will need to link against it, but may not need the includes setup, since all it does is link...

furthermore you could have packages that contain tools, with post build steps or prebuild steps, such as the google proto buffers, which comes with a compiler that processes .proto files. it would be nice if the protobuf package provided both the filters for this, as well as the header files and libraries.. but different parts of a program may need different parts of that package..

tritao:
The point of this pull was to support pkg-config, which is just natively supported on Linux (and other UNIX-y systems) as far as I know. So I think we should develop a good API for just that, and leave other features for later.

@starkos I think we can and should develop a general-purpose cross-platform package API (building on top of the HTTP and Zip work) but I see that as a different thing and it could optionally support native pkg-config packages when available using the API introduced in this pull.

@tvandijck I agree with you, which is why I suggested an API returning a table with the information so that build scripts can use whatever they need. When we figure out the main use cases then we can re-visit and introduce some higher level APIs if they make sense.

moomalade:
@starkos I think doing it during baking would work well, then any modules inspecting defines/includes will see them before the exporter. Currently it no-ops on Windows and I agree I'd prefer to see an error there instead, if it's used in an active configuration for that platform (otherwise we don't care).

It's a bit more complex because there's a bunch of common options to pkg-config to consider.

The --static flag controls static libraries or dynamic libraries.
PKG_CONIFG_PATH env var is the search path for pkg-config files.
PKG_CONFIG_DISABLE_UNINSTALLED controls if uninstalled packages are allowed.
PKG_CONFIG_LIBDIR set the system pkg-config directory, useful to set to nothing if you want to ensure you're using only package you've built in your own sandbox/prefix.

I'm not keen on re-wrapping existing well defined tools with new API, so I was thinking of something like:

`pkgconfig( { "pkg1", "pkg2" }, { "--static", ["PKG_CONFIG_LIBDIR"] = "whatever" } )`

If it begins with '--' it's added to the command line. Otherwise it's added to the environment for the call.

@tritao Brings up a use case where you want to get a table of options and select/manipulate or fallback, which a bake time solution wouldn't support, but perhaps that could be a separate API.

moomalade:
Versioning wise `pkgconfig { "sdl>=2.1.0" }` would work well.

moomalade:
I'll push a new version over the weekend so there's something more to see. If it feels too unix specific, we can pop it into a module and talk about a more generic system separately, which may well hook pig-config on unix systems. Or if we're all in agreement we can iterate on the proposal.

starkos:
I'm going to decline this, but I still think it would be a nice feature to have. If you could publish it as a module for review, I'm open to including it in core once the kinks are worked out!

moomalade:
Great stuff, makes sense to me.
I'll rework as little module when I next get some free cycles.

tarruda:
If desired, I can rework #355 to be included in the `os` module as a core function.

PR #201, opened 2015-08-09
CI for linux, osx, mingw

TurkeyMan:
Tags are built and uploaded to Github Releases.
MinGW build can't run unit tests (Travis.CI doesn't have WINE present).

### DIFF
--- a/contrib/libzip/premake5.lua
+++ b/contrib/libzip/premake5.lua
@@ -6,25 +6,26 @@ project "zip-lib"
 	flags       { "StaticRuntime" }
 	location    "build"
 
-	files 
+	files
 	{
 		"**.h",
 		"**.c"
 	}
 
 	configuration "linux"
-		defines {"HAVE_SSIZE_T_LIBZIP", "HAVE_CONFIG_H"}
+		defines { "HAVE_CONFIG_H" }
 
 	configuration "windows"
 		defines {"_WINDOWS"}
 
-	configuration "macosx"
-		defines { 'HAVE_SSIZE_T_LIBZIP' }
+	-- HAX: do we have a way to filter on the MSC toolset?
+	filter "action:not vs*"
+		defines { "HAVE_SSIZE_T_LIBZIP" }
 
 	configuration "Release"
 		defines {"NDEBUG"}
 		flags   { "OptimizeSize" }
 
 	configuration "Debug"
-		defines {"_DEBUG"}		
-		flags   { "Symbols" }	
+		defines {"_DEBUG"}
+		flags   { "Symbols" }
--- a/scripts/package.lua
+++ b/scripts/package.lua
@@ -20,16 +20,18 @@
 -- Check the command line arguments, and show some help if needed.
 ---
 
-	local usage = 'usage is: package <branch> <type>\n' ..
+	local usage = 'usage is: package <branch> <type> [noprompt]\n' ..
 		'       <branch> is the name of the release branch to target\n' ..
-		'       <type> is one of "source" or "binary"\n'
+		'       <type> is one of "source" or "binary"\n' ..
+		'       [noprompt] will skip the confirmation prompt\n'
 
-	if #_ARGS ~= 2 then
+	if #_ARGS < 2 or #_ARGS > 3 then
 		error(usage, 0)
 	end
 
 	local branch = _ARGS[1]
 	local kind = _ARGS[2]
+	local noprompt = _ARGS[3] == "noprompt"
 
 	if kind ~= "source" and kind ~= "binary" then
 		error(usage, 0)
@@ -40,7 +42,7 @@
 -- Make sure I've got what I've need to be happy.
 --
 
-	local required = { "git", "make", "gcc", "premake5", "zip" }
+	local required = { "git", "make", "gcc", "zip" }
 	for _, value in ipairs(required) do
 		local z = execQuiet("%s --version", value)
 		if z ~= 0 then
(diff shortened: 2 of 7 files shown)


TurkeyMan:
Note: OSX failed because I'm still waiting for a Travis.CI guy to manually enable OSX builds for premake/premake-core.

TurkeyMan:
If anyone wants to help me out getting this across the line, there are 2 things that need attention before this can merge which I'm having trouble with.

First, look at the mingw build; you'll see it appears to have succeeded, but it didn't actually link!
```
Linking Premake5
../../contrib/libzip/build/bin/Release/zip-lib.lib: could not read symbols: Archive has no index; run ranlib to add one
collect2: ld returned 1 exit status
make[2]: *** [../../bin/release/premake5.exe] Error 1
make[1]: *** [Premake5] Error 2
make[1]: Leaving directory `/home/travis/build/premake/premake-core/build/bootstrap'
make: *** [mingw] Error 2
```
I'm at a loss to explain this. If I build it locally on a linux machine it's fine. Do we have any linux experts or mingw experts floating around that may be able to offer some hints?
The only other problem with the CI is that we can't run unit tests for a windows binary on a linux host.


I have also tried to make an auto-build-and-release-tags feature work nicely.
There are a couple of outstanding issues. I'd like some help with it.
You can see it in action here: https://travis-ci.org/TurkeyMan/premake-core/builds/74773385

Check out the linux and/or osx build, it complains `Error: premake-core/src/base/os.lua:388: 'popen' not supported` when packaging. Again, this doesn't happen when running locally. >_<

MinGW has a few problems built in to package.lua; it uses `os.get()`/`os.is()` a lot to decide if it should do something for windows or not windows. This is all broken when the host build environment is linux. @starkos, can you think of an elegant way to approach this, or shall I just push more args into package.lua which the buildscript supplies?

contrib/libzip/premake5.lua
```diff
@@ -6,25 +6,26 @@ project "zip-lib"
...
+		defines { "HAVE_CONFIG_H" }
 
 	configuration "windows"
 		defines {"_WINDOWS"}
 
-	configuration "macosx"
-		defines { 'HAVE_SSIZE_T_LIBZIP' }
+	-- HAX: do we have a way to filter on the MSC toolset?
+	filter "action:not vs*"
+		defines { "HAVE_SSIZE_T_LIBZIP" }
```
TurkeyMan: This is awful, but I couldn't think of a good way to do it. We need to filter on toolset, and I don't think we can do that currently, no?

scripts/package.lua
```diff
@@ -40,7 +42,7 @@
 -- Make sure I've got what I've need to be happy.
 --
 
-	local required = { "git", "make", "gcc", "premake5", "zip" }
+	local required = { "git", "make", "gcc", "zip" }
```
TurkeyMan: This is strange, but if premake5 is present here, package.lua never works for me. Problem exists in all environments I try.

starkos:
Trying to get a look at this, hopefully today…

scripts/package.lua
starkos: Is `premake5` on the path? It is calling through `os.execute()` which is spinning up a new shell environment. That's the only thing that immediately comes to mind, but you probably already considered that.

contrib/libzip/premake5.lua
starkos: You can `filter "toolset:msc"` but that's not set for Visual Studio builds (probably should be though!). Or maybe `filter { "not gcc", "not clang" }`? Still not great.

starkos:
> MinGW has a few problems built in to package.lua; it uses os.get()/os.is() a lot to decide if it should do something for windows or not windows. This is all broken when the host build environment is linux. @starkos, can you think of an elegant way to approach this, or shall I just push more args into package.lua which the buildscript supplies?

How about adding one more "target system" argument that, if not set, defaults to the current system? And maybe output that value in the "Here's what I'm about to do" message? I have no objection to that.

I'm stumped on the MinGW issues but if I get a chance I'll try my Google-fu and see if I can come up with anything. Also a heads up: I'm in the process of assembling an alpha5 release now, which I'm just going to do manually.

starkos:
Oh, maybe post something on the forums about the MinGW issues?

starkos:
So, oh yeah, I also use MinGW to make the Windows builds for the releases. So I'm running into the same issues. Working on it…

scripts/package.lua
TurkeyMan: It is in the path, but in the case of the build scripts, I set the path to include it before calling premake5 package.
I guess what happens is the new shell environment isn't inheriting the current state of the environment, rather creating a new environment.

contrib/libzip/premake5.lua
TurkeyMan: Ah, I didn't know filter supported 'toolset'... should probably update the manual to mention that: https://github.com/premake/premake-core/wiki/filter

starkos: > should probably update the manual to mention that

Done.

starkos:
Spent a little time looking into this but don't have any great answers so far. Ended up manually building the Windows binary with VS just to get alpha5 out the door. Still on my to-do list to keep working on it.

TurkeyMan:
No worries, this is pending on my short-list too. I'm not gonna get to it tonight though, I have about 20 other emails to sift through ;)

tvandijck:
I just made a pull request to add the Mac OS X build... no mingw yet...

tvandijck:
Can we close this one? we got mac osx and linux live now... so only mingw/windows would be nice, but Travis does not yet support either... Internally we got a jenkins cluster doing all three platforms, but I don't think I can access it's status externally... I guess we could spin up a windows vm on Azure or AWS or something if we really want the coverage... I'm sure we can get a free vm for this?

starkos:
Defer to Manu. It would really be helpful to me to get MinGW builds working, but haven't had a chance to investigate.

tvandijck:
http://www.appveyor.com/

it's basically Travis for Windows, no idea how hard it would be to setup...

d-led:
I've recently tried appveyor - not much harder to set up than Travis. Here's an example config with a call to premake https://github.com/d-led/nanomsg_example/blob/master/appveyor.yml

samsinsane:
Closing this off since we have Linux, OSX and Windows building now. If we need MinGW builds, a PR can be opened specifically for that.

PR #202, opened 2015-08-10
If action contains no valid language or kind lists, assume everything is supported

starkos:
This makes it easier to write and maintain small utility or reporting actions that don't actually care about project kinds of languages. If you really don't want your action to support any kinds or languages (which would be weird, but okay) you can do:

```lua
valid_languages = {}
valid_kinds = {}
```


tvandijck:
lgtm

PR #210, opened 2015-08-16
added support for preloading user modules

dcourtois:
With this modification, if we add a "_user_modules.lua" file (or more than one) somewhere in the _MAIN_SCRIPT_DIR or its children, the modules listed will be preloaded like the modules in "_modules.lua".

This means that when you create a version of Premake which embeds your custom modules, their help (if any) will be displayed when running "premake5 --help" (which wasn't the case until now, and which proved a bit annoying :p)

The code is fairly simple, so feel free to check and ask me questions if something seems wrong, or if you want me to do it another way. I tested it using only one _user_modules.lua file in my addons folder, but also with one _user_modules.lua in each of my addons.
All the stuff in _preload.lua got correctly loaded and the help for my actions / options were correctly displayed.

I don't know if there is a way to unit test this since it's something only usefull when embedded custom stuff ...

### DIFF
--- a/scripts/embed.lua
+++ b/scripts/embed.lua
@@ -90,6 +90,11 @@
 	local mask = path.join(_MAIN_SCRIPT_DIR, "**/_manifest.lua")
 	local manifests = os.matchfiles(mask)
 
+-- Find all of the _user_modules.lua files within the project
+
+	local userModuleFiles = {}
+	userModuleFiles = table.join(userModuleFiles, os.matchfiles(path.join(_MAIN_SCRIPT_DIR, "**/_user_modules.lua")))
+	userModuleFiles = table.join(userModuleFiles, os.matchfiles(path.join(_MAIN_SCRIPT_DIR, "_user_modules.lua")))
 
 -- Generate an index of the script file names. Script names are stored
 -- relative to the directory containing the manifest, i.e. the main
@@ -109,17 +114,6 @@
 			filename = path.getrelative(baseDir, filename)
 			table.insert(result, '\t"' .. filename .. '",')
 		end
-	end
-
--- Try to look for _user_modules.lua files in the _MAIN_SCRIPT_DIR folder
--- and it subfolders. If we find at least one, add the _embedded_user_modules.lua
--- entry.
-
-	local userModuleFiles = {}
-	userModuleFiles = table.join(userModuleFiles, os.matchfiles(path.join(_MAIN_SCRIPT_DIR, "**/_user_modules.lua")))
-	userModuleFiles = table.join(userModuleFiles, os.matchfiles(path.join(_MAIN_SCRIPT_DIR, "_user_modules.lua")))
-	if #userModuleFiles > 0 then
-		table.insert(result, '\t"src/_embedded_user_modules.lua",')
 	end
 
 	table.insert(result, '\t"src/_premake_main.lua",')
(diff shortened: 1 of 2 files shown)


starkos:
If all we're trying to do is load a few modules from the current project's source tree, as this request currently does, I think it be better to just `require()` them from the project script instead. That's more straightforward and since we're probably only talking about a few modules the performance hit should be small.

Preloading is intended as a (perhaps premature) optimization to keep the core executable startup time quick. We are going to be adding a lot of modules and we don't want to be fully executing every one of them each time we start up.

If we are going to expand that outside of the core modules, it really needs to work with a user's system script. That is, I should be able to download a collection of modules that I need for my work environment, place them on my Premake path, and have them all work automatically. Right now you can do that by adding a `require()` call to each one to your system script.

To solve this more general case, we can search for a `user_modules.lua` script like you've done here, but check the entire search path instead of just the current project directory.

Another solution might be to add a new call like `register()` which causes the preload script, if available, to be run. Then you could just add a bunch of `register()` calls to your system script.

Thinking more long term, I'd love to see the recent HTTP and .zip additions used to build a more general purpose module manager a la npm which could download a module and register it in a single step.

dcourtois:
Hi, sorry if my description was a bit misleading, I should have kept it simple. The problem I had is the following : I can embed my custom modules just fine, but their help (the actions and options registered by those modules) is never shown when running Premake5 --help.
This is what this pr fixes, by applying the same logic as for the _modules.lua file.

And I didn't search the whole search paths for _user_modules.lua files because currently the manifest files that are used are only the ones located in the main script dir hierarchy. So I didn't want to add modules that would not end up being embedded.

Is it clearer ?

dcourtois:
Any news on this ? If the pull request is not the correct way to fix the issues related to embedded modules (non-official ones) just tell me, I'm willing to do the job.

But right now, it's really frustrating to not see the actions/options documentation of my embedded modules when I run Premake5 --help ... I really wish that custom modules could be embedded just the way the official ones are (using the preload optimisation, having support for action/option help, etc.)

scripts/embed.lua
```diff
@@ -136,6 +147,16 @@
...
 	end
 
+-- Write he user modules.
+
+	if #userModuleFiles > 0 then
+		local userModules = {}
+		for _, userModuleFile in ipairs(userModuleFiles) do
+			userModules = table.join(userModules, dofile(userModuleFile))
+		end
+		appendScript(result, "return {" .. table.implode(userModules, "\\\"", "\\\"", ",\\n") .. "}")
```
starkos: Confused by this line. Why shouldn't `_user_modules.lua` look exactly like `_modules.lua` (already a table, also like `_manifest.lua` and `_tests.lua`) and be loaded the same way? Not a big deal but would prefer the consistency?

dcourtois:
Hi, `_modules.lua` will always be unique, but because `_user_modules.lua` files are user files, their might be several of them, that then need to be merged.

For the moment, the only way to embed non-official scripts into the Premake executable is by having the following hierarchy:

    Premake\
       |--- premake-core\             <- clone from Github, don't want to change anything in here
       |--- addons\
       |      |--- Qt\
       |      |--- CompilationUnit\
       |      |--- _user_modules.lua  <- contains 'return { "Qt", "CompilationUnit" }'
       |--- build\                    <- a build folder containing build scripts (not really)
       |--- premake-addons.lua        <- contains 'dofile("premake-core/premake5.lua")'

When I want to build the Premake5 executable which embeds my modules, I run the embed action on the `premake-addons.lua` at the root, or else it won't find my addons' `_manifest.lua` files (embed action only looks for manifest files in the main script dir)

Now I thought that if I want to embed someone else's scripts, I will have to modify the `_user_modules.lua` file. Or, if I support multiple user modules files, just do this:

    addons\
       |--- MyModules\
       |        |--- Qt\
       |        |--- CompilationUnit\
       |        |--- _user_modules.lua    <- contains 'return { "Qt", "CompilationUnit" }'
       |--- SomeoneElseModules\
                |--- ModuleA
                |--- ModuleB
                |--- _user_modules.lua    <- contains 'return { "ModuleA", "ModuleB" }'

Maybe it's not a good idea and I should stick to the way `_modules.lua`. I don't mind, I just thought that supporting multiple user modules files could be usefull in some cases :) (and anyways, when we have a npm like system, this user modules file will probably be automatically managed by it)

So just tell me what you think, I can make the change during the day ! (and thanks for taking the time to deal with me :p)

starkos:
Oh, never mind, I misunderstood what the code was doing, you were already doing what I suggested.

Though since you are already joining all the user module lists together in the embed script, what about also joining in `_modules.lua`? Then the core code could just load a single list that has all the modules—it doesn't really care if they are "core" or added by you. That puts all the complexity into the embedding script and keeps the core code simple.

Does that make sense? Something like:

```
local modules = dofile("_modules.lua")
for _, userModuleFile in ipairs(userModuleFiles) do
   modules = table.join(modules, dofile(userModuleFile))
end
appendScript(result, "return {" .. table.implode(modules, "\\\"", "\\\"", ",\\n") .. "}")
```

You could embed that with the name `src/_modules.lua` and then the core code wouldn't need to change at all?

dcourtois:
Of course, you're totally right ! Why didn't I think about that !
I'll do that right away.

starkos:
This is great, thanks!

PR #219, opened 2015-08-24
Empty projects no longer ignore platform toolset when there are no "cpp" files present

### LINKED ISSUE #93: premake5 alpha 3 generate vs2010 project on vs2013 solution
Hello,

I've just created 3 project calling "premake5 vs2013" 

in the solution I have 2 project which works perfectly fine and one which have the platform toolset option set to visual 2010 (not installed on my machine).

Has anyone seen this issue? Is there a work around?

samsinsane:
I'm unsure of the intended functionality of this code, but to me, if you specify the VS2013 action, it should make a VS2013 project regardless of the files you're adding.

Fixes #93

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -143,6 +143,7 @@
 		if cfg.kind == p.UTILITY then
 			return {
 				m.configurationType,
+				m.platformToolset,
 			}
 		else
 			return {
@@ -1515,12 +1516,15 @@
 			version = action.vstudio.platformToolset
 		end
 		if version then
-			-- should only be written if there is a C/C++ file in the config
-			for i = 1, #cfg.files do
-				if path.iscppfile(cfg.files[i]) then
-					m.element("PlatformToolset", nil, version)
-					break
+			if cfg.kind == p.NONE or cfg.kind == p.MAKEFILE then
+				for i = 1, #cfg.files do
+					if path.iscppfile(cfg.files[i]) then
+						m.element("PlatformToolset", nil, version)
+						break
+					end
 				end
+			else
+				m.element("PlatformToolset", nil, version)
 			end
 		end
 	end
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -348,7 +348,7 @@
 	end
 
 	function m.resourceCompile(cfg)
-		if cfg.system ~= p.XBOX360 and config.hasResourceFiles(cfg) then
+		if cfg.system ~= p.XBOX360 and p.config.hasFile(cfg, path.isresourcefile) then
 			local contents = p.capture(function ()
 				p.push()
 				p.callArray(m.elements.resourceCompile, cfg)
@@ -1517,11 +1517,8 @@
 		end
 		if version then
 			if cfg.kind == p.NONE or cfg.kind == p.MAKEFILE then
-				for i = 1, #cfg.files do
-					if path.iscppfile(cfg.files[i]) then
-						m.element("PlatformToolset", nil, version)
-						break
-					end
+				if p.config.hasFile(cfg, path.iscppfile) then
+					m.element("PlatformToolset", nil, version)
 				end
 			else
 				m.element("PlatformToolset", nil, version)
--- a/src/base/config.lua
+++ b/src/base/config.lua
@@ -359,14 +359,15 @@
 
 
 
---
--- Determine if a configuration contains one or more resource files.
---
+---
+-- Returns true if any of the files in the provided container pass the
+-- provided test function.
+---
 
-	function config.hasResourceFiles(self)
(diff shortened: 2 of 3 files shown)


TurkeyMan:
Is it that VS doesn't emit this XML element if the project has no cpp source? Premake typically tries to do what VS would do under the same conditions.

This PR looks reasonable to me. I'm inclined to pull, but I'm surprised this code existed in the first place, so I think @starkos should comment incase there is a reason for it.

samsinsane:
I just did a quick check with VS2013, it definitely adds the XML element for an empty project. So I'm still unsure what the initial idea was for this functionality.

TurkeyMan:
No, me either. I'll let starkos pull, but lgtm.

starkos:
I was trying to match the VS behavior for utility and custom build projects that don't include C++ files, yes. Let me take a look at this again, perhaps I can narrow down the condition in which I apply it. I have a set of test projects here I can run it against.

TurkeyMan:
You often talk about this large set of test projects; how is it that they are so much more comprehensive than the unit tests? Perhaps we should we be including such tests alongside the unit test suite? Build a few complex projects in full, and then diff them against expected output or something...

tvandijck:
The platform toolset also has to be written for Utility/Makefile projects btw.... we end up getting the "upgrade" dialog in visual studio 2013 if we leave those xml elements out.....

starkos:
> You often talk about this large set of test projects; how is it that they are so much more comprehensive than the unit tests?

They are just big, real world codebases. It is not that they are more comprehensive, just that they do sometimes catch edge cases that haven't been covered by the tests yet. Oh, and its kind of my job to make sure that new iterations of Premake don't break them ;) I try to write unit tests when I find gaps…except when I don't, like the EOL issues I just submitted a pull request for. Perhaps I should fix that.

samsinsane:
@starkos How'd you go with your tests? Should I bring this branch up to latest?

starkos:
Looking at this one now, trying to remember why I made that change. I *did* have a reason…

tvandijck:
I locally merged this... ran into the same issue.... if you make a Visual Studio 2013/2015 project with utility/makefile or none "kind", you end up with a visual studio 2010 vcxproj that the IDE asks you to upgrade... pretty annoying ;)

tritao:
I've also been affected by that problem about the IDE asking to upgrade some projects... it's really annoying. I didn't know why that happened so I added this hack which seemed to fix it:

```
-- Premake hacks

local p = premake
local m = p.vstudio.vc2010

premake.override(premake.vstudio.vc2010, "platformToolset", function(base, cfg)
    local tool, version = p.config.toolset(cfg)
    if version then
      version = "v" .. version
    else
      version = action.vstudio.platformToolset
    end
    m.element("PlatformToolset", nil, version)
end)
```

samsinsane:
@tvandijck are you saying that this PR still has the same problem? :(

tvandijck:
No, I'm saying I merged it because I ran into the same issue, and now I no longer do...
although I needed an extra fix for the utility project type:

```lua
@@ -143,6 +143,7 @@
 		if cfg.kind == p.UTILITY then
 			return {
 				m.configurationType,
+				m.platformToolset,
 			}
 		else
```

samsinsane:
Ahh, sorry, completely misunderstood you. I might add that fix into this PR too, when I get a chance :)

starkos:
It looks like we exclude that element when `cfg.kind == "None" or cfg.kind == "Makefile"`. Would it possible to add that conditional around the check for C++ files? So if kind is one of those types it does the C++ file check, otherwise it can go ahead and always output the element?

samsinsane:
@starkos I've added the check you requested, as well as the fix for utility projects that @tvandijck provided. Let me know if I've missed anything. :)

starkos:
Yes, this looks good, thanks so much! While I'm being a total PITA, would you mind adding a few tests like…

```lua
	function suite.output_onConsoleAppAndNoCpp()
		kind "ConsoleApp"
		removefiles "hello.cpp"
		prepare()
		test.capture [[
<PlatformToolset>v110</PlatformToolset>
		]]
	end

	function suite.skipped_onNoMakefileAndNoCpp()
		kind "Makefile"
		removefiles "hello.cpp"
		prepare()
		test.isemptycapture()
	end

	function suite.output_onNoMakefileAndCpp()
		kind "Makefile"
		prepare()
		test.capture [[
<PlatformToolset>v110</PlatformToolset>
		]]
	end
```

Also, for bonus points, it would be really nice to replace that file search loop with `if p.config.hasFile(cfg, path.iscppfile)`…

```lua
function config.hasFile(self, testfn)
    local files = self.files
    for i = 1, #files do
        if testfn(files[i]) then
            return true
        end
    end
end
```

starkos:
(BTW, you can say no, I won't take it personally…)

samsinsane:
@starkos I've made the requested changes, but I've managed to cause a conflict somewhere. Are you happy for me to rebase this branch on to latest master?

starkos:
Perfect, thanks so much!

PR #220, opened 2015-08-24
The links API, for visual studio, no longer appends .lib if an extension already exists.

### LINKED ISSUE #87: Unable to use "links" to link with .obj files in VS2010
I'm using premake 5.0-alpha3 to create a VS2010 project where I need to link to an .obj file.
I'm unable to use the "links" command as premake adds the ".lib" suffix, eg:
`links { "something.obj" }`
will add "something.obj.lib" to the project dependencies.

Is it possible to make the links function check for .obj and not add the lib suffix?
(The workaround for the moment is to use "linkoptions" to link with the obj)

samsinsane:
I believe this is the desired functionality mentioned in the issue.

Fixes #87

### DIFF
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -340,7 +340,13 @@
 		-- Then the system libraries, which come undecorated
 		local system = config.getlinks(cfg, "system", "fullpath")
 		for i = 1, #system do
-			table.insert(links, path.appendextension(system[i], ".lib"))
+			-- Add extension if required
+			local link = system[i]
+			if not p.tools.msc.getLibraryExtensions()[link:match("[^.]+$")] then
+				link = path.appendextension(link, ".lib")
+			end
+
+			table.insert(links, link)
 		end
 
 		return links
--- a/tests/actions/vstudio/vc2010/test_link.lua
+++ b/tests/actions/vstudio/vc2010/test_link.lua
@@ -165,6 +165,40 @@
 
 
 --
+-- Any system libraries specified in links() with valid extensions should
+-- be listed with those extensions.
+--
+
+	function suite.additionalDependencies_onSystemLinksExtensions()
+		links { "lua.obj", "zlib.lib" }
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+	<GenerateDebugInformation>false</GenerateDebugInformation>
+	<AdditionalDependencies>lua.obj;zlib.lib;%(AdditionalDependencies)</AdditionalDependencies>
+		]]
+	end
+
+
+--
+-- Any system libraries specified in links() with multiple dots should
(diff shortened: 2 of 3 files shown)


src/actions/vstudio/_vstudio.lua
```diff
@@ -340,7 +340,13 @@
 		-- Then the system libraries, which come undecorated
 		local system = config.getlinks(cfg, "system", "fullpath")
 		for i = 1, #system do
-			table.insert(links, path.appendextension(system[i], ".lib"))
+			-- Add extension if required
+			local link = system[i]
+			if link:match("[^.]+$") == link then
```
TurkeyMan: It doesn't look like this regex would work if there's more than one `.` in the filename? Libs often have a version number or whatever in them too.

samsinsane: I did a quick check, it works fine with more than one `.` in the filename. Should I add another test for this case?

TurkeyMan: I thought I did a quick test showing the opposite. Maybe VS regex behaves different than Lua's... It'd be nice to add the unit test showing it works for multiple dot's.

TurkeyMan: Actually, I can see it would work for `x.y.lib`, but does it work for `x.y`? I would this code not add `.lib` in that case, but it should, no?

TurkeyMan: Assuming `"x.y":match("[^.]+$") == "y"`, right? In that case it will never add `.lib` when there are multiple periods.

tests/actions/vstudio/vc2010/test_link.lua
```diff
@@ -165,6 +165,40 @@
...
+	end
+
+
+--
+-- Any system libraries specified in links() with multiple dots shouldn't
+-- have .lib appended to the end
+--
+
+	function suite.additionalDependencies_onSystemLinksExtensionsMultipleDots()
+		links { "lua.5.3.lib" }
```
TurkeyMan: Try: `links { "lua.5.3.lib", "lua.5.4" }`

src/actions/vstudio/_vstudio.lua
samsinsane: To have that functionality, we'd have to have a list of acceptable library extensions. Which seems like this issue would just come up again later for another extension?

TurkeyMan: Yup, and that's probably the reason a fix hasn't been submit before ;)
This is just for windows, so I think it's safe to assume `.lib` and `.obj` for now.

samsinsane: No worries, I'll add this in.

samsinsane:
@TurkeyMan This now does as we discussed earlier, let me know if there's anything else required. :)

TurkeyMan:
Looks good, thanks!

PR #225, opened 2015-08-24
Added ignoreDefaultLibraries API for Visual Studio 2010+ projects.

### LINKED ISSUE #89: Unable to set "Ignore Specific Default Libraries" in VS2010 project
There doesn't appear to be a way to set the "Ignore Specific Default Libraries" field in a Visual Studio project. (Located in Project Properties -> Linker -> Input)

The workaround is to use:
`linkoptions { '/NODEFAULTLIB:"libraryname"' }`

samsinsane:
Fixes #89

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -562,6 +562,13 @@
 	}
 
 	api.register {
+		name = "ignoredefaultlibraries",
+		scope = "project",
+		kind = "list:mixed",
+		tokens = true,
+	}
+
+	api.register {
 		name = "icon",
 		scope = "project",
 		kind = "file",
--- a/tests/actions/vstudio/vc2010/test_link.lua
+++ b/tests/actions/vstudio/vc2010/test_link.lua
@@ -460,3 +460,37 @@
 </Link>
 		]]
 	end
+
+--
+-- Test ignoring default libraries with extensions specified.
+--
+
+	function suite.ignoreDefaultLibraries_WithExtensions()
+		ignoredefaultlibraries { "lib1.lib", "lib2.obj" }
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+	<GenerateDebugInformation>false</GenerateDebugInformation>
+	<ImportLibrary>bin\Debug\MyProject.lib</ImportLibrary>
+	<IgnoreSpecificDefaultLibraries>lib1.lib;lib2.obj</IgnoreSpecificDefaultLibraries>
+</Link>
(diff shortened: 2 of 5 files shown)


src/_premake_init.lua
```diff
@@ -562,6 +562,13 @@
 	}
 
 	api.register {
+		name = "ignoreDefaultLibraries",
```
TurkeyMan: api's are lower case.

samsinsane: How did I manage to miss this... My bad, I'll fix this up.

tests/actions/vstudio/vc2010/test_link.lua
```diff
@@ -460,3 +460,20 @@
 </Link>
 		]]
 	end
+
+--
+-- Test ignoring default libraries.
+--
+
+	function suite.ignoreDefaultLibraries()
+		ignoreDefaultLibraries { "lib1.lib", "lib2.lib" }
```
TurkeyMan: This makes me wonder if should support lib names without `.lib` by auto-appending like the last issue?

tests/actions/vstudio/vc2010/test_link.lua
```diff
@@ -460,3 +460,20 @@
 </Link>
 		]]
 	end
+
+--
+-- Test ignoring default libraries.
+--
+
+	function suite.ignoreDefaultLibraries()
+		ignoredefaultlibraries { "lib1.lib", "lib2.lib" }
```
TurkeyMan: This makes me wonder if should support lib names without .lib by auto-appending like the last issue?

samsinsane: That would be consistent, I'll add it in :)

TurkeyMan:
I think this is good, but it'd be nice if `msc.lua` also gained this trick so makefiles work as well as vs projects.

samsinsane:
I'll look into that as well :)

TurkeyMan:
Thanks! Otherwise the disparity is basically just a bug waiting to be reported.

TurkeyMan:
New api's need to be listed here too: https://github.com/premake/premake-core/wiki/Project-API
But that can happen after it's pulled.

starkos:
What @TurkeyMan said re: casing and MSC support. Otherwise LGTM! Thanks!

samsinsane:
@TurkeyMan @starkos I believe that's everything. Once it's pulled in, I'll add it into the wiki :)

starkos:
Merged, thanks!

Issue #122, opened 2015-06-19
Generating secondary solutions in same premake5.lua scripts is problematic


dgregorycryptic:
I am attempting to make 2 solution files in my project.  One that contains all projects, and a second that contains a subset of the same projects.

Your wiki page on the subject indicates this is possible, but I'm not seeing a way through with the API as it is.

a) I can't specify the same project name without re-specifying all the details I already did further up in the same script.
b) The externalproject command requires a bunch of details. For instance, the uuid. And, the location.  Both of which are internally known already (or only).

There ought to be a way to refer to the existing project declaration. Do you have plans for a "projectref" command?

The upshot of what i'm asking for is to generate a second solution with references to already generated project files.

dgregorycryptic:
Oh, I think I just stumbled on what you intended to be the solution for this.

The "includeexternal" external command.

dgregorycryptic:
Hmm, it's not working completely.  Many projects I specify that are defined above in the same premake5.lua script (or included files) are not showing up in the .sln file  I will try to put together a small example the reproduces this problem.

dgregorycryptic:
**premake5.lua**
```
solution "Main"
	configurations {"Debug", "Release"}
	platforms {"Win32", "x64"}
	include "A"
	include "B"
	include "C"
	include "D"

solution "Subset"
	configurations {"Debug", "Release"}
	platforms {"Win32", "x64"}
	includeexternal "A"
	includeexternal "C"
```
**A\premake5.lua**
```
project "A"
        kind "StaticLib"
        language "C++"
        buildinputs {"a.c"}
```
**B\premake5.lua**
```
project "B"
        kind "StaticLib"
        language "C++"
        buildinputs {"b.c"}
```
**C\premake5.lua**
```
project "C"
        kind "StaticLib"
        language "C++"
        buildinputs {"c.c"}
```
**D\premake5.lua**
```
project "D"
        kind "StaticLib"
        language "C++"
        buildinputs {"d.c"}
```
Output from premake5 (alpha4) run:
```
>premake5 vs2012
Building configurations...
Running action 'vs2012'...
Generating Main.sln...
Generating A.vcxproj...
Generating B.vcxproj...
Generating C.vcxproj...
Generating D.vcxproj...
Generating Subset.sln...
Done.
```
Contents of Main.sln, A-D.vcxproj are all correct.
Contents of **Subset.sln** are incorrect (no projects listed).
```
Microsoft Visual Studio Solution File, Format Version 12.00
[...]

PR #228, opened 2015-08-25
Fixed a bug when including (includeexternal) the same project in multiple workspaces

samsinsane:
This was the suggested solution in the issue. Fixes: #122

### DIFF
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -76,12 +76,15 @@
 ---
 -- Register a general-purpose includeExternal() call which works just like
 -- include(), but marks any containers created while evaluating the included
--- scripts as external.
+-- scripts as external. It also, loads the file regardless of how many times
+-- it has been loaded already.
 ---
 
 	function includeexternal(fname)
+		local fullPath = premake.findProjectScript(fname)
 		api._isIncludingExternal = true
-		include(fname)
+		fname = fullPath or fname
+		dofile(fname)
 		api._isIncludingExternal = nil
 	end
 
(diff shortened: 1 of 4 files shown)


src/base/api.lua
```diff
@@ -76,12 +76,15 @@
 ---
 -- Register a general-purpose includeExternal() call which works just like
 -- include(), but marks any containers created while evaluating the included
--- scripts as external.
+-- scripts as external. It also, loads the file regardless of how many times
+-- it has been loaded already.
 ---
 
 	function includeexternal(fname)
+		local fullPath = os.locate(fname, fname .. ".lua", path.join(fname, "premake5.lua"), path.join(fname, "premake4.lua"))
```
starkos: Rather than copying this line, it would be better if we had a common function that could be called by `includeexternal()` and `include()` to get the full path. This could go in `_foundation.lua` and perhaps be called `premake.findProjectScript(name)` maybe?

That way if we need to change the search logic we can do it in one place. I could see this easily falling out of sync otherwise.

samsinsane: Should there be tests for this? I can't seem to find tests for the `_foundation.lua` file.

starkos: Tests are always appreciated! Maybe put them in `tests/base/test_include.lua`? If I were trying to find them, that is probably the first place I'd look.

samsinsane:
@starkos I've added the tests in, let me know if I've missed anything. :)

starkos:
The tests look good, thanks!

Issue #216, opened 2015-08-21
Option validation breaks for mixed case triggers


oyshole:
When adding new options using the newoption keyword, options defined with a mixed or upper case trigger may break during validation.

This seems to be triggered by the fact that the master option list keeps the casing of the trigger values, while the _OPTIONS table has all converted to lowercase during insertion. Thus, when validating the input options, all input options are lowercase but retrieval of options using premake.options.get is case sensitive still.

Example of broken behavior:

~~~
newoption {
  trigger =  "MyTrigger",
}
~~~

Does not work: premake5 --MyTrigger ... (Error: invalid option 'mytrigger')
Does not work: premake5 --mytrigger ... (Error: invalid option 'mytrigger')

If we change the trigger to be all lower case, making the master option list key lowercase, any input works.

~~~
newoption {
  trigger =  "mytrigger",
}
~~~

Works: premake5 --MyTrigger ...
Works: premake5 --mytrigger ...
Works: premake5 --MYTRIGGER ...

I encountered this problem trying to port lots of old premake 4 scripts to work with 5, this was the only major issue encountered.

PR #230, opened 2015-08-26
Fixed options not being added to the list as lowercase

samsinsane:
Fixes: bug #216


starkos:
Yup, looks good. Thanks!

(Though next time it would be good to have a unit test for this to make sure it doesn't get broken in the future; give a shout if you need help setting one up.)

Issue #124, opened 2015-06-23
xcopy /E also copies folders when copying files


Meoo:
I didn't believe it at first, but xcopy seems to be the worst copy utility ever made.

By default, premake5 adds the "/E" argument which sould mean "recursive, including empty folders".
The problem is when you try to copy a file, it will also copy the hierarchy of the folder this file is in (only folders, no files).
The "/S" (which should mean recursive except empty folders) is also flawed, but it is much more random. Sometimes it copies 1 or 2 folders sometimes it doesn't...
I can't think of any reason it does that, but it does (tested on win7).
To try it just xcopy /E or /S a file which have sibling folders into an empty folder.

It seems the "/E" option should never be used when copying a single file... I blame xcopy on that one.
I have no idea how this can be fixed, robocopy just denies the copy of a single file if /E or /S is used.

samsinsane:
Could this be fixed by changing the copy function defined in os.commandTokens.windows to something like this:
```
copy = function(v)
	v = path.translate(v)
	if os.isdir(v) then
		return "xcopy /Q /E /Y /I " .. v .. " > nul"
	else
		return "xcopy /Q /Y /I " .. v .. " > nul"
	end
end,
```
I'm more than happy to create a PR for this, if this is an acceptable solution to the problem?

Meoo:
Thanks for your input, but in my case (which was rather complex), the file to be copied was the output produced by the build (it didn't exist when running premake), so this wouldn't completely fix the issue.

Really, the problem just comes from xcopy...

samsinsane:
Unless I'm misunderstanding what you're saying, this would work regardless of whether the file existed or not, it has to be a directory to receive the `/E` switch.

Meoo:
Well, in my case yes it would work, but maybe someone will want to copy a folder generated (for example assets baked by another post build command), and in that case the xcopy without /E will not work, while the default behaviour with /E will always work.

To completely fix the issue the file type ought to be checked when running the command, but that would be complex and error prone just to remove some mildly annoying empty folders...

samsinsane:
You're right, I came to this conclusion too after thinking about it a bit more. There's another alternative, the check could be performed when the command is executed. I'm still trying to figure it out, but something like this:
```
return "IF EXIST " .. v .. "\NUL (xcopy /Q /E /Y /I " .. v .. " > nul) ELSE (xcopy /Q /Y /I " .. v .. " > nul)"
```
[...]

PR #231, opened 2015-08-26
Windows copy function no longer uses /E when used on a file

samsinsane:
* Added some basic tests to ensure combinations work
* Fixes: #124

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -66,7 +66,7 @@
 	function os.findlib(libname, libdirs)
 		-- libname: library name with or without prefix and suffix
 		-- libdirs: (array or string): A set of additional search paths
-				 
+
 		local path, formats
 
 		-- assemble a search path, depending on the platform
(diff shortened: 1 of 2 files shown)


Meoo:
I did some tests with the "\\NUL" trick and it seems to fail when using quotes, I had better success with a single "\" (see below).
Otherwise the code looks good to me.

```
E:\>if exist "Android"\\NUL (echo OK)

E:\>if exist "Android"\ (echo OK)
OK

E:\>if exist Android\\NUL (echo OK)
OK

E:\>if exist Android\ (echo OK)
OK

E:\>if exist "Android/Android Studio"\ (echo OK)
OK

E:\>if exist test.bat\ (echo NOK)

E:\>if exist "test.bat"\ (echo NOK)
```

samsinsane:
Good catch, I'll fix this up.

starkos:
Why should copying something be simple ;p Thanks for putting this together!

Issue #197, opened 2015-08-06
NuGet support for Visual Studio projects


aleksijuvani:
NuGet is the package manager included in the recent versions of Visual Studio. The repository contains library binaries for at least C# and C++ projects.

Right now if a Premake project is dependent on NuGet packages, the generated project files will need to be manually patched.

I would be willing to create a pull request to add support for this, but I'm unsure as to how it should work.

Should we add a `nugetlinks` function? Or should we add some kind of syntax for the existing `links` function, such as a "nuget:" prefix? Or should we do this at all? I'd like to know what you think.

aleksijuvani:
Here's an example of how this would affect the generated project files for `vs2015`. This adds a dependency to the `sdl2` and `sdl2.redist` NuGet packages:

```patch
--- polygonist-shared.vcxproj   2015-08-02 18:38:26 +0300
+++ polygonist-shared.vcxproj   2015-08-02 18:38:35 +0300
@@ -729,7 +729,29 @@
       <Project>{3096E6B0-9C00-9B27-E53E-4CCD51E8787C}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
+    <Import Project="packages\sdl2.redist.2.0.3\build\native\sdl2.redist.targets" Condition="Exists('packages\sdl2.redist.2.0.3\build\native\sdl2.redist.targets')" />
+    <Import Project="packages\sdl2.2.0.3\build\native\sdl2.targets" Condition="Exists('packages\sdl2.2.0.3\build\native\sdl2.targets')" />
   </ImportGroup>
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
+    </PropertyGroup>
+    <Error Condition="!Exists('packages\sdl2.redist.2.0.3\build\native\sdl2.redist.targets')" Text="$([System.String]::Format('$(ErrorText)', 'packages\sdl2.redist.2.0.3\build\native\sdl2.redist.targets'))" />
[...]

PR #232, opened 2015-08-26
vs2010_vcxproj.lua: add onImportExtensionTargets

aleksijuvani:
As discussed in #197. Adds an empty function for overriding that
allows us to create new extension targets in overrides without
breaking rule generation.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -43,6 +43,7 @@
 			m.assemblyReferences,
 			m.files,
 			m.projectReferences,
+			m.importLanguageTargets,
 			m.importExtensionTargets,
 		}
 	end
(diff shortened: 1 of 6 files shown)


starkos:
Thanks for this contribution, but I don't think this is the right way to about it. I could accept it for now, but then it would break your code later when I changed it.

I think `m.importExtensionTargets()` should be changed to something like this:

```lua
m.elements.importExtensionTargets = function(prj)
    return {
        m.importRuleTargets,
    }
end

function m.importExtensionTargets(prj)
    p.push('<ImportGroup Label="ExtensionTargets">')
    p.callArray(m.elements.importExtensionTargets, prj)
    p.pop('</ImportGroup>')
end
```

The `Microsoft.Cpp` import should be moved to a new function; I'm thinking:

```lua
function m.importLanguageTargets(prj)
    p.w('<Import Project="$(VCTargetsPath)\\Microsoft.Cpp.targets" />')
end
```

And the current contents of the loop become:

```lua
function m.importRuleTargets(prj)
		for i = 1, #prj.rules do
			local rule = p.global.getRule(prj.rules[i])
			local loc = vstudio.path(prj, p.filename(rule, ".targets"))
			p.x('<Import Project="%s" />', loc)
		end
end
```

This way any module can add or remove entries to the extensions targets call array list, in any arbitrary ordering.

tvandijck:
Doing things with the callarray would definitely be the preferred method, since it is consistent with all the other places where we allow overrides... I like what you wrote up @starkos.. that would allow for future extensions too...

aleksijuvani:
That indeed makes more sense. I could see if any problems arise with the above code and update this pull request. This isn't a very critical change, so it would also be fine if you closed this pull request and implemented it properly later.

starkos:
If you use my code examples (assuming I typed everything correctly), the only other things you should need to do are update the project call array list (where `m.importExtensionTargets` was called originally) to call `m.importLanguageTargets` first, and then fix up any unit tests that get broken by the change. I think that should do it.

aleksijuvani:
How's this? It looked like `m.importExtensionSettings` could've used the same treatment so I've also changed that to use a call array.

starkos:
Looks great! Thanks!

PR #233, opened 2015-08-27
Added test case to ensure the functionality isn't broken in the future.

samsinsane:
I believe the issue is still open too. #216


TurkeyMan:
Thanks Sam! These are awesome.

PR #234, opened 2015-08-28
More workspace renames

starkos:
More name changing; starting on the test suites now too.


tests/actions/make/cpp/test_objects.lua
```diff
@@ -13,14 +13,14 @@
...
+	local wks, prj
 
 	function suite.setup()
-		sln = test.createsolution()
+		wks = test.createWorkspace()
 	end
 
 	local function prepare()
-		prj = premake.solution.getproject(sln, 1)
+		prj = premake.solution.getproject(wks, 1)
```
samsinsane: Should this be workspace too?

tests/actions/vstudio/cs2005/projectsettings.lua
```diff
@@ -13,17 +13,17 @@
...
 		_ACTION = "vs2005"
-		sln = test.createsolution()
+		wks = test.createWorkspace()
 		language "C#"
 		uuid "AE61726D-187C-E440-BD07-2556188A6565"
 	end
 
 	local function prepare()
-		prj = premake.solution.getproject(sln, 1)
+		prj = premake.solution.getproject(wks, 1)
```
samsinsane: Should this also be workspace?

samsinsane: Instead of spamming this, it looks like a lot of the test files for visual studio have this kind of thing. It appears to be intentional. :)

starkos: I'm getting there—because this is such a big change, I am trying to do it in stages. Because I am touching so many files, I want to push often so I don't end up with 100s of conflicts. And I also want to make sure that I'm not breaking things for other people as I go. 

Almost done though. One or two more pushes should do it for core, and then I will update the submodules.

tests/base/test_detoken.lua
```diff
@@ -52,8 +52,8 @@
 --
 
 	function suite.expandsNestedTokens()
-		environ.sln = { name="MySolution%{'X'}" }
-		x = detoken.expand("%{sln.name}", environ)
+		environ.wks = { name="MySolution%{'X'}" }
```
samsinsane: Should the `name` be changed to `MyWorkspace` too?

tests/actions/vstudio/cs2005/projectsettings.lua
samsinsane: Ahh true, my bad. Looks good though :)

PR #239, opened 2015-09-02
improve determinism

tvandijck:


TurkeyMan:
lgtm

starkos:
Thanks!

akaStiX:
i want to bring this back.
Unfortunately this changeset creates problems for some project types.
First of all, it's not clear for me what's wrong with the old behavior and why it wasn't deterministic?! You got the order, you've defined in your scripts - I consider that to be deterministic :)
Any way, GCC's linker is sensitive to the order of linking libraries, to avoid this you have to use linker groups, so it can cyclically resolve references.
It happens, that Android projects in VS2015 doesn't have support of linker groups, so this change breaks compilation of my project. Is there a chance to back out this changeset?

starkos:
Tom, what do you think? Can we roll this back and maybe bring it forward again in smaller pieces?

tvandijck:
The sort in the container can probably be reverted... the project order one cannot, but also should not cause any issues with link order.

we should probably add some tests to assure this kind of behavior doesn't break again as well to be honest. It would be cool if we could add some simple premake5 projects in the test folder with some expected vcxproj/gmake outputs.. I know the tests already work sort of like that, but all on a very limited set. somewhat more complete tests would be cool I think. Bugs like this, we should really just make regressions for... add a test project, proof the problem, fix the problem.

starkos:
> we should probably add some tests to assure this kind of behavior doesn't break again as well to be honest.

To be honest, I'm not sure what would cause them to break in the first place? I haven't seen this kind reordering occur in my projects.

akaStiX:
>the project order one cannot, but also should not cause any issues with link order.
The sort in project actually causes problems.
In my example I have executable that depends on a libB and libA. libA is a foundation layer for my project, libB is a layer on top of Foundation and there is an Application that sits on top of libB. No one can reference a layer, that is on top of it, only below it.
So I link everything using:
links { "libB", "libA" }
but the sort changes the order of libs in vcxproj and a linker cannot reference external symbols in libB

tvandijck:
Hmm.. so the problem is that if we reverted this entirely, the output of premake is no longer deterministic in a lot of cases... the order of your "premake" scripts isn't necessarily the order in which things are added to tables.... tables in lua are entirely non-deterministic when used as "hash tables", so when you have 

```lua
table = { ['a'] = 'a', ['b'] = 'b' }
for k,v in pairs(table)
  print(v)
end
```

it will sometime print "a, b" and sometime it will print "b, a".. the interesting thing is that within a single run it never changes... but when you run the executable multiple times, it changes from run to run.. so I'm kind of assuming that lua uses a pointer to the key as an initial sort key or something. I also noticed this is worse in Lua 5.2+ then it is in Lua 5.1

starkos:
> …tables in lua are entirely non-deterministic when used as "hash tables"

That's true, but the code in question is actually iterating an array, so it is already deterministic (but not sorted). Unless we're talking about a different part of the code?

It would still be super useful to have a sequence of steps that show the indeterminism. There might be a different way we can fix it, but I'm still not clear what is actually causing the problem.

tvandijck:
Well, I think most of this comes from our package management module, where we do:

```lua
import {
 ['a'] = '1.2.0',
 ['zlib'] = '1.2.8',
 ['pnglib'] = '1.6',
 ['b'] = 'dev'
}
```

Apart from that however, visual studio's solution definitely expects things in a particular order, although I guess that is more up to the "sln" backend, and I believe some sorting is already applied there.

anyway, I don't know what the right answer is, all I know is that if I remove this fix, we get regenerated vcxproj/slns all the time, while nothing changed.

akaStiX:
I solve this particular issue for my Android module myself or we can introduce another workaround - i.e. global hack-flag to enable\disable modules sorting, for example

tvandijck:
Yeah, I think we need to come up with a reason why "we" as in "Blizzard" get so many regenerated projects if we remove this... and instead of solving it by just sorting things, actually find the root cause and fix it there...  so I'm OK with reverting this one, and I'll find a way to get things right on our end.

akaStiX:
"You", as a "Blizzard" will benefit more from keeping this the way it is and instead I can override the m.projectReferences in my module, until MS will add support for linker groups in MSBuild

starkos:
Rolled back in PR #328.

>  I think most of this comes from our package management module

Ah, that would explain why I'm not seeing it. That probably would be the right place to address it then, since you might also want to take the link ordering issues raised here into account (i.e. a strict alpha ordering might not be sufficient).

PR #240, opened 2015-09-02
Add build script integration.

tvandijck:
This adds the script that created the project to the project tree... pretty useful.

### DIFF
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -259,6 +259,8 @@
 
 		local tr = tree.new(prj.name)
 
+		local addscript = (prj.script ~= nil)
+
 		table.foreachi(prj._.files, function(fcfg)
 			-- The tree represents the logical source code tree to be displayed
 			-- in the IDE, not the physical organization of the file system. So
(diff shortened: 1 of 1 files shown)


starkos:
I'm not a fan of this one, I really don't want Premake sticking things into my project files outside of what I specify in my script. But open to additional opinions here.

tvandijck:
I love it, because it allows me to edit the lua scripts from within my project... but I get what you are saying... maybe we can make it optional...

It's what nant and cmake do as well btw, I'm not sure if it is optional for those though..

starkos:
Can you just add…

```lua
files { "**/premake5.lua" }
```

…to your project script (or something like it)? I get what your saying, and for my personal projects I often do this too. But most (all, I think actually) of the projects I set up in other environments do not include those scripts in the generated projects.

TurkeyMan:
I've never had a problem with adding the scripts to my project explicitly. Is there a reason to do it automatically?

tvandijck:
> Is there a reason to do it automatically?

Yeah, I'm lazy.

TurkeyMan:
> Yeah, I'm lazy.

Surprising. I don't see much evidence for this ;)

starkos:
> Yeah, I'm lazy.

Me either, I can hardly keep up (or not). Despite that, this one still makes me uncomfortable, hope you don't mind (too much).

tvandijck:
;) no worries, we'll keep it as a local change..

PR #243, opened 2015-09-02
add excludefrombuild filelist api.

tvandijck:
I think in the long run we should just deprecate the ExcludeFromBuild flag, and use this API directly, however right now I bake this file list into the flags in the oven step...

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -426,6 +426,13 @@
 	}
 
 	api.register {
+		name = "excludefrombuild",
+		scope = "config",
+		kind = "list:file",
+		tokens = true,
+	}
+
+	api.register {
 		name = "flags",
 		scope = "config",
 		kind  = "list:string",
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -544,7 +544,13 @@
 				p.fileconfig.addconfig(files[fname], cfg)
 
 			end)
+
+			-- process build file exclusions for this configuration
+			oven.processExcludes(cfg, files)
 		end
+
+		-- process build file exclusions for this project
+		oven.processExcludes(prj, files)
 
 		-- Alpha sort the indices, so I will get consistent results in
 		-- the exported project files.
@@ -554,6 +560,27 @@
 		end)
 
 		return files
+	end
+
+
+--
+-- Process the exclude from build files. This assigns the ExcludeFromBuild flag to files
+-- that are listed in the excludefrombuild command.
+--
+
+	function oven.processExcludes(ctx, files)
+		if ctx.excludefrombuild and #ctx.excludefrombuild > 0 then
+			table.foreachi(ctx.excludefrombuild, function(fname)
+				local f = files[fname]
+				if f then
+					if ctx.buildcfg then
+						f.configs[ctx].flags.ExcludeFromBuild = true
+					else
+						f.flags.ExcludeFromBuild = true
+					end
+				end
+			end)
+		end
 	end
 
 


tvandijck:
This is funny... it's not tested by Travis CI, because it has the words "excludefrombuild" in the commit message ;)

starkos:
I guess this makes sense as an API, but I'm kind of uncomfortable putting in this implementation of it. If we're going to merge something, it should probably replace and deprecate the flag.

How quickly do you need this merged? I can try to loop back and have a crack at a full implementation.

tvandijck:
no quick turn around needed really... a proper fix would be nice obviously... but the change is pretty isolated, so easy to merge during a rebase...

samsinsane:
Did you guys want me to throw something together for this? I'll probably end up changing it from file list to a Boolean, I think using the filter API is more consistent, thoughts?

tvandijck:
well, if you have time for this, that be awesome for sure.... what we want however is exactly what this API provided... a "file list" that can contain a set of files that will be excluded from the build.

Right now we have a 'flag' that allows you to mark 'one' file  as exclude from build, which is nice, but you have to use the "filter { file:*.cpp }" syntax for that to make multiple files, and while that is basically exactly what we want, the syntax for multiple filters like the "file list" supports is very cumbersome... so I wanted something like this:

excludefrombuild {
    'bla*.cpp',
    'foo*.cpp',
    '*.cc'
}

and that is what this pull request does... it just kind of does it in a hacky way...

samsinsane:
Took a while for me to actually get a chance to look at this, but I've found a bit of an issue with it. `vs2005_solution.lua` uses the flag to exclude *projects* from builds, I'm not sure how we will eventually replace that usage. The API requirements that @tvandijck has suggested above make sense, but perhaps `excludefrombuild` shouldn't be the name, but instead `excludefiles`? Additionally, `removeexcludefilesfrombuild` is obnoxiously long.

This would leave `excludefrombuild` open to be utilized as a `boolean` for projects and whatever else? Thoughts on this? My immediate concern is that people might think `excludefiles` should act like `removefiles`, there might be a better name that doesn't result in a 5+ word long remove API.

ratzlaff:
Heh, I did not realize that 'excludes' got migrated over to 'removefiles'. We use that all the time here. I guess I dont really understand this change.

> Right now we have a 'flag' that allows you to mark 'one' file as exclude from build, which is nice, but you have to use the "filter { file:*.cpp }" syntax for that to make multiple files, and while that is basically exactly what we want.

We have lines like: 
`  excludes { ext .. "lib/folder/File*.*" }
  excludes { ext .. "lib/anotherFolder/**.*" }`
in our premake scripts and we have (not yet) had any issues with those files showing up in our builds.

samsinsane:
`removefiles`/`excludes` causes *all* of the actions to have 0 references to the files listed for that configuration.

`excludefiles`/`excludefrombuild` allows actions to have references to the files listed, but they don't build them. This is really only used in IDEs, it's a nice feature when you've got files that *only* compile on Linux/macOS and you're developing in Visual Studio. They just appear in your solution explorer but won't build and spew out errors.

starkos:
So, two different use cases, yes. `removefiles()` (and its compatibility alias `excludes`) removes files that were previous added with `files()`, so they don't show in the build or IDE at all, as if they were never added in the first place. Then the `ExcludeFromBuild` flag case, which allows the files to appear in the IDE, but excludes them from the build itself, potentially on a per-configuration or per-platform basis.

Getting rid of any flag is good, but I'm not sure I agree with @tvandijck on this one. I think we should continue to use `filter()`, and just change the flag to an API:

```lua
fillter { "files:bla_.cpp or foo_.cpp or *.cc" }
   excludefrombuild "Yes"
```

If listing out the files is a pain, I think it would be better to tweak `filter()` to make it easier. That way it can work with *any* settings.

tvandijck:
If filter can take a list like this:
```
        excludefrombuild {
            '**.mm',
            '**_Mac.*',
            '**_Posix.*',
            '**/Mac/*.*',
            '**/Posix/*.*'
        }
```

then yeah, I'm all in favor...
this however, looks and feels very error prone

```
fillter { "files:**.mm or **_Mac.* or **_Posix.* or **/Mac/*.* or **/Posix/*.*" }
```

besides, I already have that in a filter, so it would have to become:
```
filter { "system:windows", "files:**.mm or **_Mac.* or **_Posix.* or **/Mac/*.* or **/Posix/*.*" }
```

I guess, I would simply write that 'excludefrombuild' function to do that though, take the current filter, make a new scope and add the 'files' filter part to it... would result in the same I guess.

tvandijck:
basically, this is how it looks in one of our libs right now
```lua
    filter { 'system:windows' }
        excludefrombuild {
            '**.mm',
            '**_Mac.*',
            '**_Posix.*',
            '**/Mac/*.*',
            '**/Posix/*.*'
        }
        includedependencies {
            'mhook',
        }
        bindirs {
            'bin'
        }
filter { 'system:linux' }
        excludefrombuild {
            '**.mm',
            '**_Mac.*',
            '**_Win.*',
            '**/Mac/*.*',
            '**/Win/*.*',
            '**/WinMemory/*.*',
            '**/jemalloc/*.*',
            'Basics/Input_WinIme.cpp',
            'Basics/Input_WinImeSpecial.cpp'
        }

    filter { 'system:macosx' }
        excludefrombuild {
            '**_Posix.*',
            '**_Win.*',
            '**/Posix/*.*',
            '**/Win/*.*',
            '**/WinMemory/*.*',
            '**/jemalloc/*.*',
            'Basics/Input_WinIme.cpp',
            'Basics/Input_WinImeSpecial.cpp'
        }
```

starkos:
Understood, but I don't want to get into the situation where we are creating these one-off API variants for things deal with files. What if `filter()` looked like this instead?

```
filter { 
   system = "windows",
   files = {
      '**.mm',
      '**_Mac.*',
      '**_Posix.*',
      '**/Mac/*.*',
      '**/Posix/*.*'
   }
}
   excludefrombuild "yes"
```

tvandijck:
@starkos that would be totally acceptable and awesome yes ;)

starkos:
I have already been working with this alternate syntax with the query system rework…I'll keep pushing ahead and try to get this in soonish. I'd rather not try to make this change now and then change it again (which would be necessary) when the new system comes online. I'll see if I can think of a way to get pieces of it out sooner.

PR #244, opened 2015-09-02
Don't use deprecated LUA API's

tvandijck:


data-man:
@tvandijck
> Lua 5.2

Why not  5.3.1? :)

tvandijck:
Oh, I just haven't gotten to that one yet.... 

the biggest problem is... to move to 5.2 there is further changes required, and some of those changes are not backwards compatible, so @starkos would have to make a very conscious decision on which version to support... I think this has come up before where some of his clients are on 5.1 and unable to move up, so even moving to 5.2 would create problems.

These changes however work in both 5.1 and 5.2, so I'd like to get at least those changes merged, so the delta between 5.1 and 5.2 is smaller...

Once this is merged, I'll make a 5.2 pull request, but it is unlikely that @starkos will be merging that any time soon...

data-man:
Accidentally saw the project [GENie](https://github.com/bkaradzic/genie) - fork of Premake 4.4 beta 5 with Lua 5.3.
IMHO, there is something interesting.

tvandijck:
Oh my... we fixed so many bugs since 4.4 beta 5, it's not even funny.... they definitely took things in a different direction, by adding 'flags' for pretty much everything, instead of API's such as in premake 5. 

The guy seemed to be hurt over something @starkos told him, and forked it out of anger, instead of actually trying to help move premake along. Whatever floats your boat I guess.

It's nice they got it on lua 5.3 though, I'll have a look at that, but to be honest, even going from 5.1 to 5.2 doesn't really give us anything. There is certainly not a performance benefit if that is what everyone is expecting... So the only benefit would be that some bugs in 5.1 are fixed, or 5.2/3 has added lua features people want to use...

starkos:
Thanks for this!

PR #245, opened 2015-09-03
Don't use cfg.libdirs directly, go through config.getlinks.

tvandijck:
This is more in line with the vstudio.getLinks(), but in addition it removes the need for full paths in the links, since all paths are already added to the AdditionalLibraryDirectories element.

### DIFF
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -338,7 +338,7 @@
 		end
 
 		-- Then the system libraries, which come undecorated
-		local system = config.getlinks(cfg, "system", "fullpath")
+		local system = config.getlinks(cfg, "system", "name")
 		for i = 1, #system do
 			-- Add extension if required
 			local link = system[i]
@@ -352,6 +352,18 @@
 		return links
 	end
 
+---
+-- Assemble the list of library directories.
+--
+-- @param cfg
+--    The active configuration.
+-- @return
+--    The list of library directories, ready to be used in Visual Studio's
+--    AdditionalLibraryDirectories element.
+---
+	function vstudio.getLibraryDirectories(cfg)
+		return table.filterempty(config.getlinks(cfg, "system", "directory"))
+	end
 
 
 --
--- a/src/actions/vstudio/vs200x_vcproj.lua
+++ b/src/actions/vstudio/vs200x_vcproj.lua
@@ -849,9 +849,10 @@
 
 
 	function m.additionalLibraryDirectories(cfg)
-		if #cfg.libdirs > 0 then
-			local dirs = vstudio.path(cfg, cfg.libdirs)
-			p.x('AdditionalLibraryDirectories="%s"', table.concat(dirs, ";"))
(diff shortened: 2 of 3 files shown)


TurkeyMan:
LGTM, but It's a shame this isn't mirrored for gmake, where I find this is a much bigger problem.

tvandijck:
I think I have similar fixed for gmake, just not sure where in my changelog they are at the moment ;)

src/actions/vstudio/_vstudio.lua
```diff
@@ -338,7 +338,7 @@
 		end
 
 		-- Then the system libraries, which come undecorated
-		local system = config.getlinks(cfg, "system", "fullpath")
+		local system = config.getlinks(cfg, "system", "name")
```
starkos: This will fail if you try to link to a library relative to the project directory: `links { "../../bin/somelib" }`. Needs to be the full path.

src/actions/vstudio/_vstudio.lua
```diff
@@ -352,6 +352,18 @@
...
+--
+-- @param cfg
+--    The active configuration.
+-- @return
+--    The list of library directories, ready to be used in Visual Studio's
+--    AdditionalLibraryDirectories element.
+---
+	function vstudio.getLibraryDirectories(cfg)
+		return table.filterempty(config.getlinks(cfg, "system", "directory"))
+	end
```
starkos: If you call `vstudio.path()` here, you could get rid of the `table.filterempty()`, and you wouldn't have to call `path.translate()` on the results.

tvandijck: good call... I'll give that a try...

src/actions/vstudio/_vstudio.lua
tvandijck: hmm... but then we'd have to strip that off again for paths that are already in the libdirs. That was kind of the thing I was trying to achieve... maybe we should cover these cases with a couple more unit-tests...

starkos: If the project specified libdirs, then they wouldn't have to add path informatio to the link:

```lua
libdirs { "../../bin" }
links { "somelib" }
```

My take was that if the script went to the trouble of providing path information I should maintain it. Do you have a case where that falls over?

tvandijck: yeah, we generate a lot of this stuff through the package system, maybe I should have a look at how that is happening right now...

tvandijck: actually looking at the change a little bit more, if you provide this:
```lua links { "../../bin/somelib" }```

then this change will make the code act as if you did:

```lua
libdirs { "../../bin" }
links { "somelib" }
```

so it doesn't fail, it just works differently, but I guess that could lead to issues if you had duplicate names or something...

src/actions/vstudio/_vstudio.lua
tvandijck: oh, and tthis doesn't turn out to be correct... vstudio.path calls "project.getrelative" where config.getlinks already returns paths that are relative...

PR #246, opened 2015-09-03
fix project dependencies.

tvandijck:


starkos:
I'm looking at this one now. There's a delta from the UTF-8 BOM I need to check against a reference project (I think the newline is supposed to be there) and the ProjectDependencies section is getting removed from all of our solutions—I want to make sure that I can explain that change before it hits everyone (and that isn't breaking anything locally).

starkos:
Sorry this is taking so long—swamped. Anyway, I tried setting up a solution manually in VS 2013. I created a console executable and a DLL, and set them up to link by adding a project dependency. The resulting solution DOES have the project dependency section:

```
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Win32_ConsoleApp", "Win32_ConsoleApp\Win32_ConsoleApp.vcxproj", "{2F80D7A6-97C6-4746-8E47-DBFB2F0C0F75}"
	ProjectSection(ProjectDependencies) = postProject
		{04101836-703C-4DA4-89DA-1C4401B996FC} = {04101836-703C-4DA4-89DA-1C4401B996FC}
	EndProjectSection
EndProject
```

With this PR applied, all of those sections are getting stripped out ALL my projects. This doesn't seem right?

src/actions/vstudio/vs2005_solution.lua
```diff
@@ -32,7 +32,7 @@
 
 	function sln2005.generate(wks)
 		-- Mark the file as Unicode
-		_p('\239\187\191')
+		premake.utf8()
```
starkos: This outputs the BOM without a trailing newline, but projects generated manually with VS have that newline. This could be fixed a few ways:

```lua
p.utf8ln()   -- i.e. UTF8+newline; would have to be written

p.utf8(); p.outln()  -- I think I probably like this one best

p.w('\239\187\191')
```

tvandijck:
Maybe the rules are even more complex... I did the same but with an exe and a static lib, and then didn't get it. Which is why the PR is the way it is... I'll investigate this a little further then...

starkos:
I will try to take a look at that static lib case as well

starkos:
Okay I think I was doing something wrong on my end, I can reproduce the projects manually now without that dependency block being added. So your change is correct, and I just need to deal with a mess of deltas on this side. Merging now…thanks!

PR #248, opened 2015-09-03
Select clang as default tool on MacOSX.

tvandijck:


starkos:
A better way to do this now would be to add this to `_premake_init.lua`:

```lua
filter "system:macosx"
   toolset "clang"
```

TurkeyMan:
I've used and wondered about this pattern a few times before. Do you think it's better to explicitly assign a default, rather than to express 'default' by 'not set' as we've traditionally done?
It's a bit case sensitive I've found.

starkos:
Specifying it this way means it should "just work" in all of the exporters. Otherwise each importer has to reimplement the default (and if we need to change it…)

TurkeyMan:
That's true, and for that reason I have typically opted for this approach in my own modules too, but the problem with it is that it tends to result in explicit settings of default options. Ie, projects gain many XML entries that are redundant (explicitly stating the default value). Visual Studio visually distinguishes default values from set values.

starkos:
> projects gain many XML entries that are redundant

I'd consider those bugs that need to be fixed? Kind of like the VS PDB file one I'm chasing today.

TurkeyMan:
So redundant XML output should rather be detected and inhibited by the generator?

starkos:
I'd say it shouldn't be generating redundant XML output…is that the same thing? Perhaps we're talking past each other; do you have an example?

TurkeyMan:
I mean, the most common logic in the generators is that the presence of an option will cause an XML element to be emit. If the option is not present, the XML element is not emit and the option takes a default value.
If we populate configurations with explicit defaults, that sort of logic will no longer work, and there would need to be comparisons against observed default values (possibly unreliable?) to determine if the value is default, and then not emit the XML element.

starkos:
> If we populate configurations with explicit defaults, that sort of logic will no longer work, and there would need to be comparisons against observed default values (possibly unreliable?) to determine if the value is default, and then not emit the XML element.

Do you have an example?

tvandijck:
Ok, I'm closing this, locally for now I'll fix it with the above fix, by adding toolset 'clang' to the premake-init.... long term we'll come up with something in #266 discussion.

TurkeyMan:
Sorry again guys, been out of town for another week.
@starkos I was only raising it as a matter of prolific convention. In this particular case, I think defining default LDC is probably okay, but expressing defaults as undefined seems to be conventional, some instances:
https://github.com/premake/premake-core/blob/master/src/actions/vstudio/vs2010_vcxproj.lua#L929
https://github.com/premake/premake-core/blob/master/src/actions/vstudio/vs2010_vcxproj.lua#L972
https://github.com/premake/premake-core/blob/master/src/actions/vstudio/vs2010_vcxproj.lua#L1023

You can see the pattern develop; undefined things lead to no output (implying defaults).
As a general pattern, if we start defining explicit defaults for things, that will lead to code that's structured in this way to start emitting XML elements even when they match the defaults. There would need to be additional logic to suppress output when the value matches the default.
I also quite like the distinction between an explicit setting appearing in the XML, and an unspecified (default) not appearing, since the toolset can change the default as it sees fit, and I think that behaviour would be expected by the user.

Like I say, in this particular case where we're dealing with the toolset, I don't think it applies, I was just raising it as a general discussion point, since I've wanted to use this pattern before, but worked around it for the reasons I say.

PR #253, opened 2015-09-04
Use new premake.alias() call instead of direct variable assignment in API functions

starkos:
This allows premake.override() to work with both the aliased and canonical names.


TurkeyMan:
lgtm

PR #259, opened 2015-09-09
Roll back pull request #162

starkos:
As discussed in the conversion on the request, and on issue #151.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1610,9 +1610,8 @@
 
 
 	function m.programDataBaseFileName(cfg)
-		if cfg.debugformat ~= "c7" and cfg.flags.Symbols then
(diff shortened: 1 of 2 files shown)


tvandijck:
looks good to me, no objections here... I'll have to investigate if we can get away with it too though.

starkos:
If you have a use case for setting it, let me know. Best I could tell this was only duplicating what VS was already doing?

tvandijck:
Yeah, all I know is that we traditionally set it explicitly, but I have no idea what the science behind that is. So I just got to test this on our project and see what the results are. Worst case is that we'll keep this change, since we've always had it. Or we propose a fix where it can optionally be set with a flag, or something else.

tritao:
I know I've had problems due to PDB naming issues before and think I even sent a patch related to it in Bitbucket days.. so I'll be on the lookout to see if I hit any issues.

starkos:
Thanks! I'm happy to change this, but I think we can apply to a narrow set of circumstances.

PR #260, opened 2015-09-10
Rename "framework" to "dotnetframework".

tritao:
"framework" is a very generic term and hard to tell what it's actually about.

Keeps "framework" around for backwards compatibility.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -549,7 +549,7 @@
 	}
 
 	api.register {
-		name = "framework",
+		name = "dotnetframework",
 		scope = "config",
 		kind = "string",
 	}
@@ -1073,6 +1073,7 @@
 	api.alias("buildoutputs", "buildOutputs")
 	api.alias("cleanextensions", "cleanExtensions")
 	api.alias("configfile", "configFile")
+	api.alias("dotnetframework", "framework")
 	api.alias("editandcontinue", "editAndContinue")
 	api.alias("fileextension", "fileExtension")
 	api.alias("propertydefinition", "propertyDefinition")
--- a/src/actions/vstudio/vs2005_csproj.lua
+++ b/src/actions/vstudio/vs2005_csproj.lua
@@ -480,7 +480,7 @@
 
 	function cs2005.targetFrameworkVersion(cfg)
 		local action = premake.action.current()
-		local framework = cfg.framework or action.vstudio.targetFramework
+		local framework = cfg.dotnetframework or action.vstudio.targetFramework
 		if framework then
 			_p(2,'<TargetFrameworkVersion>v%s</TargetFrameworkVersion>', framework)
 		end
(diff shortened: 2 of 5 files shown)


TurkeyMan:
Oh, is framework a dotnet thing? I thought it was an OSX thing...

samsinsane:
I've never understood the "framework" API, is there a situation where someone would need to specify both "framework" and "toolsversion" for a project?

starkos:
> is there a situation where someone would need to specify both "framework" and "toolsversion" for a project?

I'm not sure what you mean by "toolsversion"? If you mean the `ToolsVersion` attribute on Visual Studio's `<Project>` element, that can vary from the framework. But I think I'm just not understanding the question.

I don't know if I really want to change this name though. Is it really that bad? Though I wouldn't mind merging that new unit test (thanks for that)!

tritao:
I think it's pretty bad. Especially since "framework" is a loaded term which can mean OSX frameworks.

Even Manu who is a long-term Premake user didn't know what it means.. I think that should speak for itself.

samsinsane:
@starkos sorry, I meant `toolset`. I went looking through the code, and see a reference to a "ManagedCProj". I have no idea how to create one of these in VS2013, so it seemed like `framework` could be moved into the `toolset` API. But it looks like there's a situation where both are actually required. :) Ignore me lol

TurkeyMan:
Haha, yeah I just completely ignored this api under the (false) assumption it was some Mac mischief!
I think I've given myself sufficient evidence to support this little change ;) .. It's a tad on the pointless side, but I'm not against it.

samsinsane:
>If you put yourself in the head of an user who's just seeing a Premake script for the first time... `framework` is very generic and tells you nothing about what it's related to.

Just to add to this, when I first started using premake (earlier this year), I would always search for how to do something with premake. So, putting myself in that situation again, I searched "premake framework" and the top hit is "Linking OSX .framework", and since I'm more likely to be looking up .NET, I added .NET in there which gives me the `framework` documentation as the second link (first link is about targeting the full framework instead of the client profile, the first comment is from @starkos linking to the `framework` documentation lol).

Granted both of those results take me to "industriousone.com" which was deprecated in favour of the Github wiki. So assuming that it's made abundantly clear to new users to use the Github wiki and users append "site:github.com" to the end of their searches. The `framework` documentation is the top hit for both searches, and as a result we should probably add in a reference about the OSX frameworks and point the user to the `links` documentation instead.

While I agree, the name alone is ambiguous, I find it hard to believe a new user would find the name of this particular API without seeing a use case or at the very least a reference to .NET. But I could be wrong about this? In which case, maybe we should rename it. :) (This comment is way longer than I expected, oops.)

starkos:
Merged. @tritao can you please update the wiki to match?

tritao:
Wiki updated. Thanks for the merge.

PR #261, opened 2015-09-10
Windows SSL support for HTTP

tritao:
This enables SSL support on Windows for the HTTP APIs.

Also a compile-fix that's needed for SSL to compile.

### DIFF
--- a/contrib/curl/premake5.lua
+++ b/contrib/curl/premake5.lua
@@ -2,7 +2,7 @@ project "curl-lib"
 	language    "C"
 	kind        "StaticLib"
 	includedirs {"include", "lib"}
-	defines     {"BUILDING_LIBCURL", "CURL_STATICLIB", "CURL_HTTP_ONLY", "CURL_DISABLE_LDAP" }
+	defines     {"BUILDING_LIBCURL", "CURL_STATICLIB", "HTTP_ONLY", "CURL_DISABLE_LDAP" }
 	flags       { "StaticRuntime" }
 	location    "build"
 
--- a/contrib/curl/premake5.lua
+++ b/contrib/curl/premake5.lua
@@ -14,6 +14,7 @@ project "curl-lib"
 	
 	configuration { 'windows' }
 		defines {"WIN32"}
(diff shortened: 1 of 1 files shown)


TurkeyMan:
I kinda fear that features which only work on a single operating system may actually a burden on premake rather than a feature...?

tritao:
Don't worry I plan to enable it for all other supported OS as part of my work. In fact I had already implemented and tested it on my branch a couple months ago on Bitbucket. But right now I have a lot going on so it's easier to do it incrementally.

tvandijck:
this is interesting for sure... I tried it a while ago, but could not for the life of me get it to actually work though... I might not have done things correctly on the actual https query...

One other thing that might be very interesting is the NTLM authentication on windows, if anyone has any ideas on how to get that working with curl??

tritao:
@tvandijck Do you mean you tested this pull and it didn't work?

I didn't test it exhaustively myself but it seemed to work. I even stepped through some of libcurl code just to make sure and it was going through SSL codepaths.

About NTLM, no idea, what would be the use cases?

tvandijck:
oh no, I didn't pull this one, I tried doing a similar thing a while ago, but couldn't get it to work...

Use-case for NTLM? well, if you are in a active directory domain (here in corporate life ;)), doing a request to an webserver could be guarded with authentication. Accessing such a website even with http or https will simply result in an permission denied, unless you log in... through NTLM login would be automatic.

tvandijck:
anyway, to stay on topic.... LGTM....

NTLM - we'll save for later ;)

tritao:
This is the relevant code in `curl_setup.h`:
```
/* Single point where USE_NTLM definition might be done */
#if !defined(CURL_DISABLE_HTTP) && !defined(CURL_DISABLE_NTLM) && \
    !defined(CURL_DISABLE_CRYPTO_AUTH)
#if defined(USE_SSLEAY) || defined(USE_WINDOWS_SSPI) || \
    defined(USE_GNUTLS) || defined(USE_NSS) || defined(USE_DARWINSSL)
#define USE_NTLM
#endif
#endif
```

If I'm reading it correctly, it should work automatically once we merge this.

PR #262, opened 2015-09-10
Allows an action to indicate Premake should be quiet on startup.

tritao:
This is really useful since it allows to write actions that behave like an independent tool (package manager) while using all of the Premake driver and infrastructure code.

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -247,7 +247,10 @@
 
 	function m.preBake()
 		if p.action.isConfigurable() then
-			print("Building configurations...")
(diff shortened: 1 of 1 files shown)


starkos:
The actions themselves can also output to the console. Should we perhaps replace `print()` with something like `p.console()` or the like? I'd prefer to not have to remember to wrap all the `print()` calls with conditionals.

starkos:
I'm going to decline this as it is, but if you want to resubmit using a different approach (like `p.console()` or similar) I'd certainly reconsider it.

PR #263, opened 2015-09-10
Switch 'require' to use xpcall for error handling.

tritao:
This gives the original stack trace where an error actually occurred instead of an incomplete trace where the error is handled in Premake.


starkos:
Thanks!

PR #264, opened 2015-09-10
Added JSON encoding and decoding support

tritao:
I originally threw this together as a proof-of-concept but realized that for my case just loading a Lua file directly probably works better.

Still think this might be useful in the future though so wanted to share to see feedback.

I adapted the Jansson Lua bindings from https://github.com/Twisol/lua_jansson. They look a bit complicated to me but it's because they do some special code for keeping track of nulls in JSON.

I found these bindings later which seem to be much more simple for the same task: https://bitbucket.org/timostolz/luajson/src/7ebcb034dcf2/luajson.c

But since the original seems to work fine I didn't rework the code.

### DIFF
--- a/contrib/jansson/LICENSE
+++ b/contrib/jansson/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2009-2014 Petri Lehtinen <petri@digip.org>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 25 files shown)


tvandijck:
I used this one..

http://regex.info/blog/lua/json

single lua file, does the same thing....

data-man:
For information: :)
Benchmarks for many JSON implementations: https://github.com/miloyip/nativejson-benchmark

tritao:
@tvandijck Well yes a pure Lua solution should be OK for small JSON files or serialization tasks, but for lots of data (from web services for instance) I think we'd prefer the speed of a native solution.

@data-man: Interesting, I didn't look at any benchmarks before, just went with Jansson because I had already used it from other projects and it's well maintained.

tvandijck:
Honestly, I think we should come up with a better way to include things like this into premake... I'm not opposed, but I almost feel like this should really be a module, and I feel a little bit the same with libzip and curl... it be awesome if we could add C as well as Lua functionality through modules.

TurkeyMan:
I think RapidJSON is pretty much accepted as the one true JSON lib in my neck of the woods... but that said, I kind of agree with @tvandijck.

I suspect this is a step towards package management and interacting with github?

tritao:
@TurkeyMan Yep, I originally did this for package management, but later opted to just use Lua tables for that use case.

But it could still be useful to interact with web APIs like Github.

I also agree with @tvandijck in that we need some better extensibility model and I'm not in any rush to get this in so if you guys have any ideas how that would work I'm up to discussing it.

starkos:
IMHO this one feels out of scope for premake-core. Especially since, in theory, it [can be done in plain Lua](http://hg.prosody.im/trunk/file/tip/util/json.lua).

Issue #257, opened 2015-09-08
Invalid field prefix 'toolset' in filter


JohannesMP:
According to the <a href=https://github.com/premake/premake-core/wiki/filter>filter documentation</a> the `toolset` field prefix **is** supported in filters.

However when my `premake5.lua` contains the following:

```
-- compiling for c++14
filter { "toolset:clang" }
  buildoptions { "-std=c++14" }
```

I get an error stating `Invalid field prefix 'toolset'` for the line with the filter.

Is this a bug or is the documentation outdated?

tritao:
Hey, I also hit this and tried to fix it in the referenced pull but it's still not working.

jeremyong:
If you look at the code for various toolsets like gcc.lua or clang.lua, you can see that the flags C++11 and C++14 are automatically mapped in the correct place. If you replace what you have with

```lua
flags { "C++14" }
```

things should work.

JohannesMP:
I appreciate the attempt to help, but that was just an example; not what I actually was hoping to use it for in my actual projects. The point is that a filter which is stated in the documentation to work, doesn't.

starkos:
@tritao has tried to fix this; feedback on PR #265 would be appreciated.

samsinsane:
Closing this since #265 was merged.

PR #265, opened 2015-09-10
Support toolset prefix in filters.

tritao:
Tries to fix issue #257.

This is not passing the tests locally, I've spent some time trying to track it down, I guess I messed up the code somewhere but I'm not very familiar with this part of Premake.

Any ideas @starkos?

### DIFF
--- a/src/base/criteria.lua
+++ b/src/base/criteria.lua
@@ -32,6 +32,7 @@
 		options = true,
 		platforms = true,
 		system = true,
+		toolset = true,
 	}
 
 
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -432,10 +432,12 @@
 
 		local system = p.action.current().os or os.get()
 		local architecture = nil
+		local toolset = nil
 
 		if platform then
 			system = p.api.checkValue(p.fields.system, platform) or system
 			architecture = p.api.checkValue(p.fields.architecture, platform) or architecture
+			toolset = p.api.checkValue(p.fields.toolset, platform) or toolset
 		end
 
 		-- Wrap the projects's configuration set (which contains all of the information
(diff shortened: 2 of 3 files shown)


starkos:
I'm not seeing anything immediately wrong. I'll have to pull it later and take a look. Could be a side effect from other tests that needs to be cleared out by the test framework between runs?

samsinsane:
I think it's [this line](https://github.com/premake/premake-core/blob/master/tests/testfx.lua#L149), `#expected` will be `1` while `#actual` will be `0`, then it falls into the next block which will iterate over the pairs in `expected` and assume that `actual` is of the same size, which results in `nil` being passed down into the next function. The line I linked just needs to be changed to `~=` and it should pass the tests again. :)

samsinsane:
Ignore my previous comment, I misunderstood what was wrong. I think there's still an issue in the function I linked, this test case should have failed instead of producing an error in the testfx file. Regardless, the issue is that `prepare` doesn't update `prj` to the filtered version, it's still the original. You'll need to either add the filter into `p.project.bake` in `oven.lua` (if it's a valid thing to do @starkos?) or iterate through the `wks.configs` and find the config that was baked in `oven.bakeConfigs(wks)`.

Hopefully this sorts the PR out and we can merge it in soon :)

starkos:
> Regardless, the issue is that prepare doesn't update prj to the filtered version

That's close. The project settings don't take the toolset into account, only configurations do. So you will have to modify the setup to look like:

```lua
	local wks, prj, cfg

	function suite.setup()
		wks, prj = test.createWorkspace()
	end

	local function prepare()
		cfg = test.getconfig(prj, "Debug")
	end
```

And then replace `prj.defines` with `cfg.defines` everywhere.

Trass3r:
Nice, works for me so far!

JohannesMP:
Seems to work. Looking forward to hopefully seeing it in future builds.

samsinsane:
@tritao have you had a chance to make the change that @starkos mentioned? Once that test is working, this should be good to merge.

PR #272, opened 2015-09-16
Changed dependency order for ${CUSTOMFILES}

### LINKED ISSUE #251: gmake and ${CUSTOMFILES} is kinda awkward
I'm building a Qt app, the Moc steps on the Qt headers are among `${CUSTOMFILES}`, and the output of which (.cpp files) are compiled among `$(OBJECTS)`.

Here is the target rule as written by premake:
```
$(TARGET): $(GCH) $(OBJECTS) $(LDDEPS) $(RESOURCES) ${CUSTOMFILES}
```

If I manually rearrange such that:
```
$(TARGET): $(GCH) ${CUSTOMFILES} $(OBJECTS) $(LDDEPS) $(RESOURCES)
```
This builds fine.

I have other custom build rules that should come before compiling too like this one for example:
```
	filter "files:shaders/**"
		buildmessage 'text2c.sh %{file.relpath} %{cfg.objdir}/shaders/%{file.basename}.inc'
		buildcommands { path.getabsolute('.') .. '/text2c.sh "%{file.relpath}" "%{cfg.objdir}/shaders/%{file.basename}.inc"' }
		buildoutputs { '%{cfg.objdir}/shaders/%{file.basename}.inc' }
```
Since these generate `.inc` files that are are used by the compiler, they must be produced beforehand.

Simply moving the `${CUSTOMFILES}` dependency would fix this for me, but it might break someone else... I wonder what we can do instead to control ordering of custom build rules...?

I'd suggest that we do change the order in the short term though, I think it's far more likely to break peoples projects the way it is currently.

TurkeyMan:
${CUSTOMFILES} is often used to produce dependencies for the build, so it needs to be first.

Fixes #251

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -352,7 +352,7 @@
 
 
 	function make.cppTargetRules(prj)
-		_p('$(TARGET): $(GCH) $(OBJECTS) $(LDDEPS) $(RESOURCES) ${CUSTOMFILES}')
+		_p('$(TARGET): $(GCH) ${CUSTOMFILES} $(OBJECTS) $(LDDEPS) $(RESOURCES)')
 		_p('\t@echo Linking %s', prj.name)
 		_p('\t$(SILENT) $(LINKCMD)')
 		_p('\t$(POSTBUILDCMDS)')


TurkeyMan:
Merging this, apologies in advance if this breaks anyone's build!
Doesn't show any issues is ony of my projects. @starkos gave his blessing ;)

PR #273, opened 2015-09-16
Group sibling libs for GCC/Clang

### LINKED ISSUE #255: gmake link order failing
Just putting this here so it doesn't get lost.

I'm having troubles with gmake, where LIBS is populated with libs in a bad order that doesn't actually link (ld is notoriously fussy about lib link order).
I haven't looked into this yet, but I wonder if others have encountered this?
Any known workarounds or solutions?

In my case, I have a stack of libs that depend on each other. It seems to me that the libs should appear in reverse order of the build order; libs that want symbols from another lib need to appear before that lib on the command line. Libs with mutual dependencies often require that they appear twice to avoid mutual exclusion problems. Eg, `libx.a` and `liby.a` both refer to symbols in each other: `LIBS += out/libx.a out/liby.a out/libx.a`
These relationships are awkward to express...

TurkeyMan:
Wrap sibling dependencies in `--start-group`, `--end-group` (for GCC and Clang)

This approach is very simple, but pretty crude.
It actually works surprisingly well, I haven't seen any IDE's that reorder the dependencies as stated in their project files, so treating these like libs in the link list actually works fine.
Alternative is that the function that flattens the libs out into a string is contained in the tool, or this logic is duplicated in each action and further predicated by GCC/Clang.

Fixes #255

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -296,6 +296,11 @@
 			end
 		end
 
+		if #result > 0 then
+			table.insert(result, 1, "-Wl,--start-group")
+			table.insert(result, "-Wl,--end-group")
+		end
+
 		-- The "-l" flag is fine for system libraries
 
 		local links = config.getlinks(cfg, "system", "fullpath")
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -296,7 +296,7 @@
 			end
 		end
 
-		if #result > 0 then
+		if #result > 1 then
 			table.insert(result, 1, "-Wl,--start-group")
 			table.insert(result, "-Wl,--end-group")
 		end
--- a/tests/actions/make/cpp/test_make_linking.lua
+++ b/tests/actions/make/cpp/test_make_linking.lua
@@ -145,6 +145,30 @@
     end
 
 --
+-- Check a linking multiple siblings.
+--
+
+	function suite.links_onSiblingStaticLib()
+		links "MyProject2"
+		links "MyProject3"
+
+		test.createproject(wks)
+		kind "StaticLib"
+		location "build"
+
+		test.createproject(wks)
+		kind "StaticLib"
+		location "build"
+
+		prepare { "ldFlags", "libs", "ldDeps" }
+		test.capture [[
+  ALL_LDFLAGS += $(LDFLAGS) -s
+  LIBS += -Wl,--start-group build/bin/Debug/libMyProject2.a build/bin/Debug/libMyProject3.a -Wl,--end-group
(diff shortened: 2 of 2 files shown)


tvandijck:
Looks good to me... but again, doesn't work on clang for MacOSX.

here is another project where they had to:
https://github.com/eliben/llvm-clang-samples/issues/9


Which brings up a much bigger issue actually....
for example -Og to disable debugging is only supported with GCC 4.6 onward, we still have projects on GCC 4.4 (boohoo but reality)... and so we need to use -O instead...

just selecting 'gcc' or 'clang' as a toolset is not enough for all these options to be properly selected in premake. Some options have only been introduced in later versions, and will basically lead to incompatible makefiles. gmake "should" detect version, and make decisions based on that.

and in this particular case... clang 3.8 seems to accept these options on my linux box, but with xcode on mac is errors out with "unknown option".

TurkeyMan:
Hmm. I don't have any test environment for that platform...
How do you suggest I fix this in the short term, while we think on that toolset versioning issue?
Would you mind tweaking this PR for you OSX environment?

starkos:
The whole toolset adapter approach needs a rethink. The current system worked fine for the small use cases I had for it, but it isn't going to scale up. Versioning is one issue we need to address. Another is the ability to mix and match compilers and linkers and and other tools together within a project.

TurkeyMan:
@starkos I agree completely. But in the meantime, we're getting broken makefiles so I'd like to see this patch merged for the short-term in some form.

starkos:
Understood. Go ahead and approve it.

TurkeyMan:
Tom has final say on this. As is, this breaks OSX. I'm hoping he can suggest a patch, it seems the args are different on that platform.

tvandijck:
No just merge it... we have this as a local change as well, so I'll just merge that on top of this and make a new pull request for that... I'm not going to block progress...

TurkeyMan:
kk

starkos:
I'm having some issues with this change, see [issue #351](https://github.com/premake/premake-core/issues/351).

TurkeyMan:
Yup, expected. Tom said he has a further patch on this, where does that sit?

starkos:
Mentioned this on the linked issue, but for completeness: the Mac OS X patch was applied.

tvandijck:
Yeah, it's not very well isolated into it's own commit... 

but it is in here:
https://github.com/Blizzard/premake-core/commit/076e018db179e1653df0e8bee26bb469e06cd051

TurkeyMan:
Seems like a reasonable stop-gap. Still seems like a bit of a hack though :P

PR #274, opened 2015-09-16
context.mergeFilters doesn't need to do a deepcopy, because table.merge already copies everything.

tvandijck:


tvandijck:
duplicate, already merged.

PR #275, opened 2015-09-16
Optimize getvpaths for memory

tvandijck:
Also sort by length of pattern so the more specific matches match first.
We had discussed this previously but I never got to make a pull request for it..

### DIFF
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -314,55 +314,99 @@
 		-- for tables, each table continuing the path key, which looks up the
 		-- array of paths with should match against that path.
 
-		for _, vpaths in ipairs(prj.vpaths) do
-			for replacement, patterns in pairs(vpaths) do
-				for _, pattern in ipairs(patterns) do
-					local i = abspath:find(path.wildcards(pattern))
-					if i == 1 then
+		local function sort_by_length(a, b)
+			local lena = string.len(a)
+			local lenb = string.len(b)
+			if (lena == lenb) then
+				return a > b
+			end
+			return lena > lenb
+		end
 
-						-- Trim out the part of the name that matched the pattern; what's
-						-- left is the part that gets appended to the replacement to make
-						-- the virtual path. So a pattern like "src/**.h" matching the
-						-- file src/include/hello.h, I want to trim out the src/ part,
-						-- leaving include/hello.h.
+		-- cache patterns.
+		local keys         = prj.cached_vpath_keys
+		local vpaths       = prj.cached_vpaths
+		local wildcards    = prj.cached_wildcards
 
-						-- Find out where the wildcard appears in the match. If there is
-						-- no wildcard, the match includes the entire pattern
+		if not vpaths then
+			vpaths = {}
+			keys = {}
+			wildcards = {}
 
-						i = pattern:find("*", 1, true) or (pattern:len() + 1)
-
-						-- Trim, taking care to keep the actual file name intact.
-
-						local leaf
-						if i < max then
-							leaf = abspath:sub(i)
-						else
-							leaf = fname
-						end
-
-						if leaf:startswith("/") then
-							leaf = leaf:sub(2)
-						end
-
-						-- check for (and remove) stars in the replacement pattern.
-						-- If there are none, then trim all path info from the leaf
-						-- and use just the filename in the replacement (stars should
-						-- really only appear at the end; I'm cheating here)
-
-						local stem = ""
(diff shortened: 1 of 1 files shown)


src/base/project.lua
```diff
@@ -314,55 +314,99 @@
...
+				-- leaving include/hello.h.
 
-						-- check for (and remove) stars in the replacement pattern.
-						-- If there are none, then trim all path info from the leaf
-						-- and use just the filename in the replacement (stars should
-						-- really only appear at the end; I'm cheating here)
-
-						local stem = ""
-						if replacement:len() > 0 then
-							stem, stars = replacement:gsub("%*", "")
```
tvandijck: I think we can optimize this further, because this 'gsub' will always give the exact same answer..

src/base/project.lua
```diff
@@ -314,55 +314,99 @@
...
-							if stars == 0 then
-								leaf = path.getname(leaf)
-							end
-						else
-							leaf = path.getname(leaf)
-						end
+				-- Find out where the wildcard appears in the match. If there is
+				-- no wildcard, the match includes the entire pattern
+
+				i = pattern:find("*", 1, true) or (pattern:len() + 1)
```
tvandijck: this also is always the same value, and the only use of "pattern" so we could potentially precalculate that, and not store the pattern, but only the wildcards expanded version.

tvandijck:
@starkos can you review this one and test it on your customer projects... its not a very risky change, but the sorting by match length is definitely a behavior change.

starkos:
Yes, I will try to do it tonight if I can.

starkos:
Sorry I haven't been able to get to this yet, hopefully before the end of the week.

starkos:
Apologies again, I haven't forgotten about this, I'm just tracking way behind right now and trying to get caught up with the latest master branch changes. As soon as I've got everything working on the latest revision I'll start clearing PRs again!

tvandijck:
No worries... I've not been able to put much time in premake either.. I started on a pretty big task here myself.

starkos:
*Finally* got a chance to try this out (sorry about that) and it doesn't seem to be behaving as expected. Most projects are fine. One project has changed vpaths from "source/intrinsics/xxx" to just "source", which would seem to be the opposite behavior? Once changed from "headers" to "source". I will strip it down to a simple test case if I can.

starkos:
Okay, I see what's going on. I made a change quite a while back to make vpaths deterministic, so that the project script could control the order of application. People have been taking advantage of this behavior to write patterns like:

```.lua
vpaths { ["Headers"] = "*.h" }
vpaths { ["Source"] = { "**.h", "**.cpp" }
```

Which reads as "put header files in this directory into 'Headers', and put all the headers from subdirectories into 'Source'."

I have to admit, having seen this, I would prefer to leave control of application order to the script. Would it be possible to decouple the memory improvements from the longest-path matching?

starkos:
Actually, looking at this one again, `*.h` and `**.h` should be considered the same "length", right? In that case, the only issue there is that this PR chooses the last one and it needs to choose the first instead. That would actually fix all the issues I saw in my own projects.

Perhaps though it could be modified to only apply the longest pattern rule against vpaths that have been specified in the same table? So in this case it would apply the rule, because everything is listed in one block:

```lua
vpaths { 
    ["Headers"] = "*.h" }
    ["Source"] = { "**.h", "**.cpp" }
}
```

In this case it would respect the ordering provided by the script and *not* apply the rule:

```lua
vpaths { ["Headers"] = "*.h" }
vpaths { ["Source"] = { "**.h", "**.cpp" }
```

Does that make sense?

tvandijck:
I'm sure we can do that.... probably just be removing the sort_by_length call...
If you have the change locally for testing, please test it without the sorting, and see if that gives you the old behavior... otherwise more work is needed...

tvandijck:
Oh, and to your second comment... I thing that would be possible, if we did the sorting at a different stage in the baking process, but yes not impossible.. We could also add special rules to the sorting callback to treat the * as better then **, although that seems somewhat arbitrary..

We'll have to experiment a little I guess...

starkos:
I think you just have to put back the outer loop:

```lua
		for _, vpaths in ipairs(prj.vpaths) do
```

...and move all of your logic inside that loop, only flattening the values within that single block:

```lua
			for replacement, patterns in pairs(v) do
				for _, pattern in ipairs(patterns) do
					vpaths[pattern] = replacement
				end
			end
```

PR #276, opened 2015-09-16
Embed precompiled lua files, instead of source.

tvandijck:
This is somewhat of a tricky change I guess....

What it does? it adds a os.compile method to premake that takes a lua file and makes a binary lua compiled file out of it... then the embed.lua takes that and converts it into a byte array in the scripts.c.

It's pretty straightforward, and the benefits are that the embed.lua now implicitly does syntax validation. invalid lua doesn't compile, so it won't embed. where previously it just takes the source code and premake give you an error that it can't find the file, instead of a proper error.

in addition, it reduced startup times for us, its not that much, but less parsing, is less time spend, every bit helps in our case...

The biggest benefit however is, we're not stripping debug information from the bytecode, so all errors have full and correct file&line information even in a release build...

that said, the bytecode turns out to be slightly bigger then the stripped lua text, which is quite amazing to me, but so be it.


starkos:
If I remember correctly, the byte code is not portable, and a change to the VM is needed to enforce endian-ness.

tvandijck:
This is correct. Lua bytecode is not portable... a 32bit executable needs 32bit bytecode, and 64bit executable needs 64bit bytecode... It's absolutely a design flaw in my opinion, and LuaJIT actually fixed that in their lua implementation.

That said, in this case this is not a problem really... since premake is used to build premake, it is building bytecode for itself, and so it is always compatible with itself. In addition premake itself is not portable, the binary is specifically a 32-bit windows binary, or a 32-bit macosx binary.. So this change does not make it better or worse...

The only issue I ran into (since we switched to a 64bit exe) is that I have to run the bootstrap.mak process from a command line where the 64bit compiler is setup in visual studio.

starkos:
There is also a byte ordering issue: byte code generated on a little-endian machine will not execute on a big-endian system. Once upon a very long time ago Premake did in fact embed byte codes, and I had a fix for the endian-ness issues. There is also a patch for Lua floating around the internets somewhere, but I couldn't locate it in my quick search.

tvandijck:
I see... I still don't see how that could pose an issue though.... a little-endian executable would not run on a big-endian machine either right?? 

it's not like the machine code in the binary is going to run on a machine where the endianess suddenly changes? or are there architectures out there that do? I don't know can you make a premake binary for a little endian mips, that then executes on a big endian version of the same mips? does such a chip even exist??

I guess the only problem would be cross compiling... so for example creating a premake binary on windows for a big endian machine. Or in windows 32bit creating a 64 bit binary for linux... That would indeed be broken with this change... but as long as you create the premake binary on the platform where it is going to be used, this should never be a problem.

tvandijck:
I agree it is a little bit of a controversial change maybe..., that's why I made it so that if you embed the source code it still works... We could make a switch for the embed action that disables it, or have it be disabled by default, but allow a switch to enable it.. whichever is preferred...

for us it has helped a lot tracking down a few problems in release, since the error messages make a lot more sense now...

starkos:
That makes sense. I guess I was running into trouble back then more because of the 32/64-bit fat binaries for Mac. I agree that the accurate debugging information would be great, and that byte code > original source code is weird. :)

tvandijck:
Just to be clear... I think the priority of this is pretty low.... there is a lot of other pull requests that are probably more important... so don't waste your time on this one yet ;) we'll keep testing it, and add the switch to it, so it can be turned off and on whenever we need to.

tvandijck:
OK, so I revised this PR to be optional...

"premake5 embed" will do exactly as it did before, embedding the source code.
"premake5 embed --bytecode" will compile the code into bytecode and embed that.

Both work, but the caveat with --bytecode is that it is platform specific, and so the bootstrap executable HAS to be build with the same arch as the final executable. The upside is that you get proper callstack information.

starkos:
Based on the earlier discussion, it sounds like bytecode is probably the way to go. There are no more weird fat-binaries-with-different-endian-architectures any longer. The switch isn't a bad idea, but would it make sense to reverse it, and embed bytecode by default?

starkos:
(Oh, and thanks so much for all the progress today!)

tvandijck:
yeah, we can default it to enabled and then have a --no-bytecode switch instead I guess.. sure, I'll work on that tomorrow... gotta run for a meeting now.

tvandijck:
better this way?

starkos:
Yes, LGTM.

PR #282, opened 2015-09-22
if premake table is not initialized use original loadfile

jsfdez:
We are customizing premake and using a different lua state for other requirements. We don't always initialize premake so it fails when trying to call loadfile.

### DIFF
--- a/src/host/lua_auxlib.c
+++ b/src/host/lua_auxlib.c
@@ -32,9 +32,12 @@ LUALIB_API int luaL_loadfile (lua_State* L, const char* filename)
 
 	int bottom = lua_gettop(L);
 	int z = !OKAY;
+	int premake_nil = 0;
 
 	lua_getglobal(L, "premake");
-	if (lua_isnil(L, -1)) {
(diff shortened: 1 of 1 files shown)


starkos:
I'm not super comfortable with this. Can you just exclude `src/host/lua_auxlib.c` from your builds, and use the original `src/host/lua-5.1.4/src/lauxlib.c` directly instead?

jsfdez:
The problem is we have a tool which controls repositories, changes, builds, ... and we want to integrate it directly in premake using different embedded scripts. We need premake available for generating projects using a unique binary.

starkos:
Thanks for this contribution, but I'm going to decline it. Those overrides are in place for a reason, and I'm not comfortable providing an "official" code path that doesn't use them.

Since it sounds like you are using a modified version of Premake anyway, you are of course free to continue using your modifications locally.

Finding a way to link with native (i.e. C) code is definitely on our radar. If we can come up with a way to do it we absolutely will. No great solutions yet.

PR #284, opened 2015-09-25
Added configuration actions array to allow easy extension of actions

samsinsane:
I'm not sure if this is the best solution, and I'm more than happy to change this implementation. I just wanted an easier way to add the `Deploy` action than what I did in my [WinRT module](https://github.com/LORgames/premake-winrt/blob/master/winrt.lua#L125-L232). I'm sure there's a better way of handling the problem in my module, but I was stuck on it for a couple days before finding that solution.

### DIFF
--- a/src/actions/vstudio/vs2005_solution.lua
+++ b/src/actions/vstudio/vs2005_solution.lua
@@ -13,6 +13,13 @@
 	local tree = p.tree
 
 
+---
+-- Add namespace for element definition lists for premake.callArray()
+---
+
+	sln2005.elements = {}
+
+
 --
 -- Return the list of sections contained in the solution.
 --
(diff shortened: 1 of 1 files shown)


starkos:
Could you provide a little more context? I'm not entirely clear on the problem you're solving with this one?

samsinsane:
Currently this particular function isn't module friendly (ignoring the issues I had overriding it) if you do manage to override this particular function, it's already written out the entire block to the buffer. So, if you override the function you can insert your lines before or after this:
```
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|x86 = Debug|x86
		Release|x86 = Release|x86
	EndGlobalSection
	GlobalSection(ProjectConfigurationPlatforms) = postSolution
		{guid}.Debug|x86.ActiveCfg = Debug|x86
		{guid}.Debug|x86.Build.0 = Debug|x86
		{guid}.Release|x86.ActiveCfg = Release|x86
		{guid}.Release|x86.Build.0 = Release|x86
	EndGlobalSection
```

The problem is, I need to add `{guid}.Debug|x86.Deploy = Debug|x86` and similarly for release after the Build.0 lines. The Deploy line was added in VS2012 to facilitate WinRT development, but is also required by the [Cross Platform Mobile Development](https://www.visualstudio.com/en-us/features/mobile-app-development-vs.aspx) feature in VS2015.

I'm pretty sure this solution isn't the ideal way, so I'm more than happy to go back and change it. Either way, I'll be needing something like this as I'm currently trying to extend the premake-android module to support VS2015 and the mobile development feature.

starkos:
Ah, thanks for that explanation. The solution code needs to be broken up to use call arrays, like we did for the projects, but I haven't had a chance to get to it. That would be a better way to break this up, as it would be more consistent with the other code, and easier to extend with other entries later if needed. The catch being that I generated a lot of intermediate values in that function. Ugly.

What if we start by ripping the whole block after the `-- For each project in the solution…` comment out into it's own function, something like (untested):

```lua
sln2005.elements.projectConfigurationPlatforms = function(cfg)
	return {
		sln2005.activeCfg,
		sln2005.build0,
	}
end

function sln2005.projectConfigurationPlatforms(prj, sorted)
		p.w("GlobalSection(ProjectConfigurationPlatforms) = postSolution")
		local tr = p.workspace.grouptree(prj.workspace)
		tree.traverse(tr, {
			onleaf = function(n)
				local prj = n.project
				table.foreachi(sorted, function(cfg)
					p.callArray(sln2005.elements.projectConfigurationPlatforms, cfg)
				end
			end
		})
		p.w("EndGlobalSection")
end
```

…and then implement `activeCfg()` and `build0()` and add your own in there? Does that make sense? I don't know if it would be worth trying to package all of these intermediate values up into a table and passing them as an additional argument (`context`?), or if it is better to just recalculate them where they are needed?

samsinsane:
@starkos I've made the changes, I decided to go down the path of using the `context` variable. I think it makes more sense to do that than recalculate the other variables over and over again. They're all required, in a way, to generate the output, except for `excluded`.

TurkeyMan:
You know, I think I have a patch almost identical to this floating around somewhere. IIRC, I rejigged this the same way some time back for Deploy on Android maybe >_<
Might have been back in the bitbucket days when it seemed like a lot of effort to submit a pull request ;)

Anyway, LGTM :)

src/actions/vstudio/vs2005_solution.lua
```diff
@@ -139,6 +146,62 @@
...
+		local tr = p.workspace.grouptree(wks)
+		tree.traverse(tr, {
+			onleaf = function(n)
+				local prj = n.project
+				table.foreachi(sorted, function(cfg)
+					local context = {}
+					-- Look up the matching project configuration. If none exist, this
+					-- configuration has been excluded from the project, and should map
+					-- to closest available project configuration instead.
+					context.prj = prj
```
starkos: While there's no harm in doing it this way, you can also access the project in the handler functions via `cfg.project`.

starkos:
Great, thanks for this!

PR #285, opened 2015-09-28
use rpaths & soname/install_name to link sibling shared libraries

noresources:
…on OS X and Linux

new api runpathdirs().
	<toolset>.getrunpathdirs() adds user defined rpath set using runpathdirs().
	handle runpathdirs in gcc, clang and snc.

make.list() can now add quotes to list elements

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -746,6 +746,13 @@
 	}
 
 	api.register {
+		name = "runpathdirs",
+		scope = "config",
+		kind = "list:path",
+		tokens = true,
+	}
+
+	api.register {
 		name = "runtime",
 		scope = "config",
 		kind = "string",
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -171,9 +171,20 @@
 -- Format a list of values to be safely written as part of a variable assignment.
 --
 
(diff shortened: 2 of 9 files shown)


TurkeyMan:
Given a quick once-over, this LGTM, but I haven't given this enough time to say that with confidence.
I need this patch at work, so if nobody reviews this before me, I'll give it a shot at work when I get a moment to work on our linux build...

starkos:
If you could that would be great, most of my projects are VS and Xcode these days, I don't have many good examples to try it on.

tritao:
This is also biting us on https://github.com/mono/CppSharp.

ddobrev:
Can we download the binaries from the build for this pull - https://travis-ci.org/premake/premake-core/builds/82604919 ? We really need this fix.

tritao:
Re-discovered this issue again today after some shared library loading problems on OSX.

Thinking again about this it seems this pull would be an improvement over the current behavior, can this be merged? It can always be reverted if it breaks something.

tritao:
Updated the pull for recent Premake master: https://github.com/premake/premake-core/pull/568

Tested on my project and works fine.

tvandijck:
Merged: #568 instead.

Issue #286, opened 2015-09-29
Running custom commands before compiling (with make -j)


kankaristo:
I started using Premake yesterday and I'm hoping to start using it instead of qmake. It's been great so far, thanks for an awesome tool!

I've got most things working, but I'm still having a problem running custom commands before any compiling takes place. I had similar problems with qmake, which is one of the reasons I'm switching to Premake.

One of the custom commands can take over 10 seconds to finish and it *must* be run before compiling, because it generates some of the code (actually, most of the code, nearly 200 000 lines).

So, I'm trying to use Premake to run [SWIG](http://www.swig.org/) to generate code before `make` attempts to compile it.

If I put the commands in a project's `prebuildcommands`:
```lua
project "MyProject"
    prebuildcommands{
        "./swig.sh";
    }
    
    [...]
```

...it works fine if I only use `make -j1` (a single job). This is exactly the same as it was with qmake; they're only "prebuild" for one of the `make` jobs, not the whole build process.

If I put the commands in a separate project and make it a dependency:
```lua
project "MyProject"
    links "SWIG"
    
    [...]

project "SWIG"
    kind "Utility"
    
    prebuildcommands{
        "./swig.sh";
    }
```

This time, the command is run before anything in "MyProject" is compiled (great!), but then I get an error about `g++` having no input files in the "SWIG" project. There's a warning about "Utility" being and unsupported "kind". Is it unsupported for `gmake` or is it completely unimplemented (no mention in the documentation)?

Is there some way to get this to work? I'd hate to resort to the same kind of hacks I had to use with qmake...

kankaristo:
Also, I'm using the release version from Premake's website. I'll try the Git version tomorrow to see if that changes anything.

kankaristo:
I started looking around in the code and figured out how to get this to work. You guys have a fantastic introduction on how to extend Premake, it was really easy to get into.

I'm just getting started with Premake, so I might be doing something evil, but it seems to work well. At least I'm probably not breaking anything, because the "Utility" kind was not supported with gmake before.

Pull request coming soon.

kankaristo:
This is working well enough for me, so I'll close the issue.

TurkeyMan:
I'm having this problem where -j# is causing build ordering issues too. Did you solve this by moving those jobs into a separate utility project? That seems a bit lame. I wonder if there's a better solution possible.
[...]

PR #287, opened 2015-10-01
Add basic support for the 'Utility' kind in gmake

kankaristo:
This adds (very) basic support for the "Utility" kind in gmake.

This is a very quick and dirty implementation and works in my use case, but may break in some others. I'm just getting started with Premake and haven't even read through all of the wiki yet, so please let me know if I should do something differently.

The main goal with this is to be able to use `prebuildcommands`, etc. in a project, without needing to compile something. See https://github.com/premake/premake-core/issues/286 for an example use case.

### DIFF
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -20,7 +20,7 @@
 		shortname       = "GNU Make",
 		description     = "Generate GNU makefiles for POSIX, MinGW, and Cygwin",
 
-		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib" },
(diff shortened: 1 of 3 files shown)


tritao:
Thanks for the contribution.

Some testing would be useful to make sure we do not regress this functionality in the future.

starkos:
If you need help adding a test, give a shout. You really one need to add a new test function to `tests/actions/make/cpp/test_make_linking.lua`, and can just copy one of the existing tests.

kankaristo:
Alright, I'll add a test. :)

By the way, the file you mentioned (`test_make_linking.lua`) has 2 functions with the name `suite.onExternalLibraryWithPath`. Won't the second one override the first one?

kankaristo:
Test added. :)

PR #292, opened 2015-10-05
fix path.translate in gmake backend.

tvandijck:


tvandijck:
hmm.. this one breaks a few tests... :( I'll revisit...

samsinsane:
Does #307 cover this fix? It should provide the target OS separator instead of assuming the Windows separator.

starkos:
It looks like #307 is complementary, yes. We'll probably have to specify an OS in the Gmake test suite setup functions to get consistent behavior then (perhaps already covered by #307).

tritao:
For reference, the failing test is:

```
make_cs_response.listResponseRules: ...premake-core/tests/actions/make/cs/test_response.lua:65: (6) expected:
	$(SILENT) if exist $(RESPONSE) del $(OBJDIR)\MyProject.rsp
...but was:
	$(SILENT) if exist $(RESPONSE) del $(OBJDIR)/MyProject.rsp
stack traceback:
	...premake-core/tests/actions/make/cs/test_response.lua:65: in function <...premake-core/tests/actions/make/cs/test_response.lua:51>
running time : 1.13
1363 tests passed, 1 failed
```

Can we just disable this test for now and merge this? The C# backend has been broken on non-Windows platforms for at least two months now, this is affecting (breaks compilation) of all of our projects using Premake.

tritao:
I've reworked this code, please check pull request #347.

PR #295, opened 2015-10-05
Move Lua into folder without version, so it's easier to update to a newer version.

tvandijck:


starkos:
You've been busy this morning!

I actually thought having in a separate, versioned folder made it easier to update, since you could place them side-by-side during the migration. But I don't really have a strong opinion either way.

tvandijck:
Yeah, the problem is that we have hardcoded paths in our source code and build scripts... so putting them side by side doesn't really give you much... since we'd still have to modify a lot of other pieces...

for our merge history it's just easier if there was no version number in the path.
more importantly, we could potentially even import lua as a module from a git repository this way.

tvandijck:
anyway, I wanted to discuss this one first, before merging, so on approval for this one.... the rest I merged this morning were all just bugfixes or very simple non-intrusive changes...

starkos:
LGTM

tritao:
Since you're already moving it, why not put Lua into contrib/ with the rest of the other dependencies?

tvandijck:
hmm, there is an idea...  the next question that then comes to mind it... do we want to build it as a library, just the other 'contrib' ones?

tvandijck:
hmm, that has some unfortunate side effect on linux apparently :( for now I think I'm just going to pull the rename locally, and will work on moving into the contrib folder at a different time if you don't mind.

PR #297, opened 2015-10-07
Fix malloc-defining header for clang compilation on Mac.

codemaster:
This fixes the use of a non-standard/platform-specific malloc.h by replacing it with the standard stdlib.h, which defines malloc.

Error encountered during compilation:
`src/host/path_wildcards.c:9:10: fatal error: 'malloc.h' file not found`

### DIFF
--- a/src/host/path_wildcards.c
+++ b/src/host/path_wildcards.c
@@ -6,7 +6,7 @@
 
 #include "premake.h"
 #include <string.h>
-#include <malloc.h>
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


samsinsane:
:+1: LGTM. I actually encountered this problem last night, but thought it might be related to running OSX 10.11, or some misconfiguration on my end.

samsinsane:
Oops, forgot to click merge earlier. Thanks for the fix!

PR #300, opened 2015-10-07
add os.copydir

tvandijck:


src/base/os.lua
```diff
@@ -578,3 +578,57 @@
...
+-- @param filter
+--    Optional, defaults to "**". Only filter matches will be copied. It can contain **(recursive) and *(filename).
+-- @param single_dst_dir
+--    Optional, defaults to false. Allows putting all files to dst_dir without subdirectories.
+--    Only useful with recursive (**) filter.
+-- @returns
+--    True if successful, otherwise nil.
+--
+	function os.copydir(src_dir, dst_dir, filter, single_dst_dir)
+		if not os.isdir(src_dir) then error(src_dir .. " is not an existing directory!") end
```
samsinsane: Is this consistent? I feel like it's supposed to be over three lines. Additionally, there should probably be an empty line between this check and the code below.

src/base/os.lua
```diff
@@ -578,3 +578,57 @@
...
+		filter = filter or "**"
+		src_dir = src_dir .. "/"
+		print('copy "' .. src_dir .. filter .. '" to "' .. dst_dir .. '".')
+		dst_dir = dst_dir .. "/"
+		local dir = path.rebase(".",path.getabsolute("."), src_dir) -- root dir, relative from src_dir
+ 
+		os.chdir( src_dir ) -- change current directory to src_dir
+			local matches = os.matchfiles(filter)
```
samsinsane: Was this intentionally indented an extra level? I can kind of see a reason for it, but it's probably more consistent to have it at the same level as the surrounding lines.

src/base/os.lua
tvandijck: you are absolutely right... will fix.

src/base/os.lua
tvandijck: nope... will fix.

akaStiX:
IMHO, it would be better to use os.execute("{COPY} arg1 arg2) due to the fact that os.copyfile requires you to provide existing path.
And reusing existing functionality is a better option, than building something absolutely the same

tvandijck:
I kind of agree.... it's been sitting in our repo for a while, I don't even think anyone is still using it...

PR #301, opened 2015-10-07
Make version based on information from git.

tvandijck:
This actually works really well, and removed the need to set the VERSION define manually when making a tag/branch or release...

it does currently break the package.lua script though I think... We don't use that for internal releases here, since we just use the yaml file to make a build and we deploy the output from the buildfarm, so there this automation works really nicely.

Anyway, just a proposal at this point I guess..

### DIFF
--- a/scripts/embed.lua
+++ b/scripts/embed.lua
@@ -92,6 +92,19 @@
 	end
 
 
+	local function makeVersion(result)
+		local version = os.outputof('git rev-parse --short HEAD')
+		local branch  = path.getname(os.outputof('git describe --all'))
+
+		buffered.writeln(result, "// Version")
+		buffered.writeln(result, "#ifndef PREMAKE_NO_BUILTIN_SCRIPTS")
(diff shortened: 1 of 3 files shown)


samsinsane:
I like the idea, but I have concerns about using git from the source code, it would probably be safer to have CI provide that information.

starkos:
There should probably be a fallback in case `git` can't be found on the path (i.e. someone using a GUI client)?

tvandijck:
Yeah, it would probably be better if we could somehow get it through CI, maybe by setting up some kind of Environment variable in the yaml file or something... I don't really intent to get it merged, but it was one of those things where I ran into having to modify the VERSION define all the time, and in my local git history that just became somewhat unmanageable "20 commits all with "update version""..

this automagically gets the tag name from a release and puts in the exe, so now all I have to do is on github click "draft new release" and our buildfarm makes a build for all platforms with the right tagname in it...

But yeah, definitely up to debate and up for improvements. this does indeed heavily rely on having git.

PR #302, opened 2015-10-07
parse for known options in the additional options

tvandijck:


samsinsane:
I like alerting the user to the new APIs, but maybe this should be done in the API layer instead, similar to `api.deprecateValue`? This would allow for calling to the correct API, which avoids code duplication and potential tag duplication in the output.

starkos:
I'm with Sam on this one: it is more work to set up, but an `api.deprecateValue` style approach could be reused for any exporter. And it would declutter the exporter code, since the option would have been translated to the corresponding API earlier in the process.

Parsing build options directly, rather than providing a proper API, is a slippery slope. Are there equivalent switches to `/MACHINE` and `/LARGEADDRESSAWARE` for other toolsets? I'm not entirely opposed to the approach, just think it should be used sparingly.

tvandijck:
/MACHINE should really just be automated based on the "architecture" already... the default often just doesn't work...
As for the /LARGEADDRESSAWARE, I think that is the most stupid property ever, since the only thing it does is set 1 bit in the executable header output. I don't think there is any other compiler that has this flag, nor even the concept of this particular feature. It is 100% windows only, and for 32 bit processes in particular... 

It's a pretty old change at this point, I remember making a pull request on bitbucket for this one, and since things have changed, which is where the warnings came in, since we had /wd everywhere in our scripts, and needed a way to figure out where they came from. I'm pretty indifferent to this change to be honest, although I like how it can effectively put things in the right box in visual studio, without the need for API's/Flags in premake.

starkos:
> I like how it can effectively put things in the right box in visual studio

Yes, that's kind of my concern, that people will submit PRs that only handle the flag, because it is much easier than creating the full API. You know…"people"… :)

tvandijck:
at the same time however... creating full API for something that is 100% MSVC and Windows only, do we really need to do that? I guess I could just add a "LargeAddressAware" flag, and do it that way, the MACHINE, we can just add properly based on cfg.architecture.

starkos:
> creating full API for something that is 100% MSVC and Windows only, do we really need to do that?

You can make a case for parsing the option there. Though adding a boolean `largeaddressaware` call will likely be less work than building an `api.deprecateValue` style system for handling implicit translations (though I understand you have a bunch of projects already using the build option…)

tvandijck:
wha, just one project... I think I'll just take a stab at it... it'll clean up and get rid of this change.

PR #303, opened 2015-10-07
Mac OSX doesn't have lib64 dirs.

tvandijck:


noresources:
To be complete, OS X does not have any architecture-specific folder since all system libraries are universal binaries

lib32 should be removed too

tritao:
LGTM

PR #304, opened 2015-10-07
More fixes for detoken.

tvandijck:


src/base/detoken.lua
```diff
@@ -49,6 +49,7 @@
 			-- convert the token into a function to execute
 			local func, err = loadstring("return " .. token)
 			if not func then
+				print("load error:", err)
```
samsinsane: Should this use `error` instead?

src/base/detoken.lua
```diff
@@ -56,16 +57,26 @@
 			setfenv(func, e)
 
 			-- run it and get the result
-			local result = func() or ""
+			local success, result = pcall(func)
+			if not success then
+				err    = result
+				result = nil
+			else
+				err    = nil
```
samsinsane: Wouldn't this already be `nil` since this function returns on `loadstring` failure?

samsinsane:
Someone with a deeper understanding of this functionality should definitely look over this, but it LGTM.

src/base/detoken.lua
tvandijck: You are correct.. we can remove the else I guess...

src/base/detoken.lua
tvandijck: no... that would abort immediately, and not go back to the caller (detoken) which does the error instead but with a little more context...

starkos: We shouldn't need to print the error here though right? Because it falls through to an `error()` in the caller?

tvandijck: correct, although the error() would just display whatever is in "err" and give no context as to what caused that err... so I added a print here so I did get some information about whether it was the load that failed, or something else.

tvandijck: I guess we could do 
```lua 
return nil, "load error: " .. err
```
instead...

samsinsane: Personally, I prefer having one error message, but since it's only two messages it's probably not a big deal :)

src/base/detoken.lua
samsinsane: Your call, there's merits both ways, less code is always good, but it's very explicit that `err` is `nil` with the else there which is good.

tvandijck: ok, leaving this one in place then.

src/base/detoken.lua
tvandijck: I kinda like the idea I posted... I'll make that change....

TurkeyMan:
Oh no! Not good! Did break everything! >_<

src/base/detoken.lua
```diff
@@ -49,23 +49,33 @@
...
-				return nil, err
+				return nil, "load error: " .. err
 			end
 
 			-- give the function access to the project objects
 			setfenv(func, e)
 
 			-- run it and get the result
-			local result = func() or ""
+			local success, result = pcall(func)
```
TurkeyMan: This change in behaviour; `result` is no longer coerced to `""` if `func()` returns `nil`, seem to cause problems...

PR #306, opened 2015-10-08
Fix sysincludedirs for GCC toolset

saberhawk:
A stray space was causing GCC to look for paths like " ../library/include" instead of "../library/include"

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -404,7 +404,7 @@
 
 
 	function make.includes(cfg, toolset)
-		local includes = premake.esc(toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs))
+		local includes = toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs)
 		_p('  INCLUDES +=%s', make.list(includes))
 	end
 


samsinsane:
I don't know much about this, but I looked up the [documentation for GCC](http://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html) and it appears that what was originally there is correct.
> -isystem dir
Search dir for header files, after all directories specified by -I but before the standard system directories. Mark it as a system directory, so that it gets the same special treatment as is applied to the standard system directories. If dir begins with =, then the = will be replaced by the sysroot prefix; see --sysroot and -isysroot.

saberhawk:
The problem here is that the generated gmake INCLUDES line looks like "-isystem\ ../library/include" and not "-isystem ../library/include". This isn't what the documentation describes. Regular include folders (handled 4 lines above) are documented the same way, but the code looks like '-I' and not '-I '.

samsinsane:
Interesting, you raise a good point about the documentation being inconsistent with the actual usage of GCC. I did more research but was unable to find an example of `-isystem` that used a relative path. I wonder if that might be why it wasn't working and if absolute paths still work with your changes?

saberhawk:
Absolute paths are still working with both gcc and clang. I suspect that the parsing code handles all include folder types the same way. In any case, escaping a forced space after -isystem seems wrong since " .." is a weird but valid filename.

samsinsane:
Thanks for doing those tests for me, sorry for being a pain lol.

This LGTM, but I'll let one of the other devs go over this before it's merged in. Just in case there's something I've missed. Thanks for the fix though!

starkos:
I think the real cause of the problem is at `make_cpp.lua:407`:

```lua
local includes = premake.esc(toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs))
```

We are escaping the return value of the `getincludedirs` call; this is what is adding the backslash. Can we fix it here instead?

saberhawk:
Fixed it there instead. I also added support for this function in the Xcode module.

starkos:
Thanks! If you modified the Xcode module you will need to submit a separate pull request as that code is (currently) in a separate repository.

saberhawk:
I submitted a pull request for premake-xcode already, but I'm unsure of what else needs to be done for submodule pull requests.

PR #307, opened 2015-10-12
path.translate fix

### LINKED ISSUE #86: path.translate always translates to \

samsinsane:
path.translate now uses the target OS separator by default instead of Windows separator. Fixes #86

The wiki will need to be updated to reflect this change when/if it's merged.

### DIFF
--- a/src/host/path_translate.c
+++ b/src/host/path_translate.c
@@ -32,7 +32,20 @@ int path_translate(lua_State* L)
 	char buffer[0x4000];
 
 	if (lua_gettop(L) == 1) {
-		sep = "\\";
+		// Get target OS
+		lua_getglobal(L, "os");
+		lua_getfield(L, -1, "get");
+		lua_call(L, 0, 1);
+		const char* os = luaL_checkstring(L, -1);
+		lua_pop(L, 2);
+
+		// Use target OS separator
+		if (strcmp(os, "windows") == 0) {
+			sep = "\\";
+		}
+		else {
+			sep = "/";
+		}
 	}
 	else {
 		sep = luaL_checkstring(L, 2);
--- a/tests/actions/make/cs/test_response.lua
+++ b/tests/actions/make/cs/test_response.lua
@@ -15,6 +15,7 @@
 	local wks, prj
 
 	function suite.setup()
+		premake.action.set("vs2010")
 		wks = test.createWorkspace()
 	end
 
--- a/tests/actions/vstudio/cs2005/projectelement.lua
+++ b/tests/actions/vstudio/cs2005/projectelement.lua
@@ -29,7 +29,7 @@
 --
 
 	function suite.on2005()
-		_ACTION = "vs2005"
+		premake.action.set("vs2005")
 		prepare()
 		test.capture [[
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
@@ -38,7 +38,7 @@
 
 
 	function suite.on2008()
-		_ACTION = "vs2008"
+		premake.action.set("vs2008")
 		prepare()
 		test.capture [[
 <Project ToolsVersion="3.5" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
@@ -47,7 +47,7 @@
 
 
 	function suite.on2010()
-		_ACTION = "vs2010"
+		premake.action.set("vs2010")
 		prepare()
 		test.capture [[
 <?xml version="1.0" encoding="utf-8"?>
@@ -57,7 +57,7 @@
 
 
 	function suite.on2012()
-		_ACTION = "vs2012"
+		premake.action.set("vs2012")
 		prepare()
 		test.capture [[
 <?xml version="1.0" encoding="utf-8"?>
@@ -67,7 +67,7 @@
 
 
 	function suite.on2013()
-		_ACTION = "vs2013"
+		premake.action.set("vs2013")
 		prepare()
 		test.capture [[
 <?xml version="1.0" encoding="utf-8"?>
--- a/tests/actions/vstudio/cs2005/projectsettings.lua
+++ b/tests/actions/vstudio/cs2005/projectsettings.lua
@@ -16,7 +16,7 @@
 	local wks, prj
 
 	function suite.setup()
-		_ACTION = "vs2005"
+		premake.action.set("vs2005")
 		wks = test.createWorkspace()
 		language "C#"
 		uuid "AE61726D-187C-E440-BD07-2556188A6565"
@@ -51,7 +51,7 @@
 
 
 	function suite.OnVs2008()
-		_ACTION = "vs2008"
+		premake.action.set("vs2008")
 		prepare()
 		test.capture [[
 	<PropertyGroup>
@@ -70,7 +70,7 @@
 
 
 	function suite.OnVs2010()
-		_ACTION = "vs2010"
+		premake.action.set("vs2010")
 		prepare()
 		test.capture [[
 	<PropertyGroup>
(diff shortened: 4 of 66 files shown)


samsinsane:
I wasn't expecting so many tests to fail, I'm going to go through and update the tests to use `path.translate` where possible, I should have this done by the end of tomorrow. @starkos does this sound ok, or is there a better way to do this?

starkos:
Why are the tests failing?

samsinsane:
The tests are expecting `\` in paths but are getting `/` because the tests are being run on Linux. I was thinking of setting `_OPTIONS["os"] = "windows"` but I feel like that would other unintended side effects.

starkos:
Are they tests for Visual Studio? If so, you can also set `_ACTION = "vs2010"` (or an appropriate version) in setup.

samsinsane:
They're not just for Visual Studio, I believe some of them already set that value and are still failing. Unless I'm mistaken, setting that shouldn't change the way paths are output?

starkos:
> wouldn't it be easier to introduce a "os.separator" property, and do this string compare only once at startup?

You can't use the local system separator, because you might be generating for a different platform, i.e. generating a VS project from Linux.

> They're not just for Visual Studio...Unless I'm mistaken, setting that shouldn't change the way paths are output?

Setting the action to Visual Studio implicitly sets the OS to "windows" (and `_ACTION = "xcode4"` should set the OS to "macosx").

If you can point me to a specific test that you're having trouble with I can take a look and see if I have any better suggestions.

tvandijck:
> You can't use the local system separator, because you might be generating for a different platform, i.e. generating a VS project from Linux.

correct, but I was suggesting to base it off of os.get() still... just once at startup.
instead of for every path.translate... I know you guys don't really care much about performance, but it's a big deal for us...

samsinsane:
Oh ok, I'll have a look when I get a chance, maybe I misread the test setup. Thanks for the tip! I'll let you know how I go.

starkos:
> correct, but I was suggesting to base it off of os.get() still... just once at startup.

Sorry, it looked like you were suggesting looking at `package.path`. Yes, I agree, keeping this function quick is important, it gets called a *lot*. You will have to be sure the action's `os` value has been picked up—I think this happens in `action.set()` but I'm not sure where in the bootstrapping that is called, and whether that is before or after the first call to `path.translate()`.

samsinsane:
I'll look into moving the check so it isn't run all the time, it definitely isn't in the most ideal location.

samsinsane:
@starkos it looks like setting `_ACTION` directly doesn't change the OS, but using `premake.action.set()` does, should I use this instead?

starkos:
I think `action.set()` makes more sense than `_OPTION["os"] = "windows"` for sure.

samsinsane:
Alright, I'll go through and update the tests to use `action.set()` instead of `_ACTION`.

samsinsane:
@starkos I've got all the premake-core tests passing, just left with ones in codelite, d and monodevelop modules. The `action.set()` function sets `_OS`, should I just set that to `windows` for those test suites?

starkos:
I would check with Manu on how to handle those.

TurkeyMan:
Yeah, just spoke with Sam; the question is, do we use `path.translate` on the expected result in unit tests, or do we set the OS explicitly and continue expecting a literal?
We were both leaning towards `path.translate`-ing the expected unit test output... @starkos what do you reckon?

tvandijck:
I think as long as path.translate itself is properly tested independently, using path.translate in other tests should not be a problem...

starkos:
What tests are we talking about here?

samsinsane:
The four tests are:
- `codelite_workspace.onNestedProjectPath` in `test_codelite_workspace.lua`
- `codelite_workspace.onExternalProjectPath` in `test_codelite_workspace.lua`
- `mono_d.OnProject_objectsDirectory` in `test_monodevelop.lua`
- `monodevelop_project_config.OnProject_OutputPath` in `test_config.lua`

starkos:
Thanks! Is there any reason for CodeLite or MonoDevelop to be using Windows-style separators? Since they are cross-platform tools it seems like the more generic forward slash would be better here?

If that's not correct and they are supposed to be using backslashes, is that for all platforms or just Windows? And is that for projects that are generated on Windows, or when a configuration's `system` is set to `"Windows"`?

tritao:
I asked my co-worker (works on MonoDevelop) and he told it's due to compatibility with MSBuild.

samsinsane:
I've got PRs open on the premake-codelite, premake-dlang and premake-monodevelop modules, once they're merged in, I'll update this PR and that should fix up all the tests.

TurkeyMan:
And they're merged! Thanks guys!
Sorry again for basically disappearing recently! >_<

samsinsane:
Thanks @TurkeyMan! Looks like there was another update to the codelite project that added another test, I'll need to fix that one up too.

samsinsane:
Ok, the update didn't add another test it actually broke the test. Can someone with codelite experience/knowledge take a look at this broken test? The `WorkingDirectory` attribute in the `General` tag is expecting `bin/Debug` by default and we're emitting an empty string, is this valid?

tritao:
Any news about this pull? It would be great if this fix could be merged soon.

samsinsane:
I just need a codelite user to give me a bit of info on what to do with the codelite module. I've been pretty busy lately and haven't had much time to follow up on this issue.

TurkeyMan:
Me too! Sorry again. My overtime counter is off the charts.

Just looking at this; not quite sure what the workingdir should be... currently it is:
```
local targetpath = project.getrelative(prj, cfg.buildtarget.directory)
local workingdir = iif(isExe, targetpath, "")
```

That doesn't seem right... and apparently this change causes that logic to change somehow? Seems weird.
We have the `debugdir` api, do we reckon that should be used to set `WorkingDir` in this context? Otherwise I think I leave it blank, and let the user set it manually if they want to.
I've never been quite clear on what `debugdir` is meant to mean. Anyone know the difinitive word on that?

tritao:
`debugdir` is supposed to set the working directory used when launching the debugger.

TurkeyMan:
Right, 'used when launching the debugger', but not 'in general', and that's what makes it awkward... we don't have a specification for working dir when not debugging.

tritao:
I'm not sure what you mean about "specification for working dir when not debugging", what do you have in mind?

TurkeyMan:
I mean, 'working dir' is a concept that exists both while debugging and not... but we only control one of the cases. The line in question here is the not-while-debugging case.
So I'm thinking resolution here would be to emit a blank entry?

tritao:
OK now I understand what you mean. I think I originally introduced the `debugdir` and called it that way because that's the terminology used by VS. If other IDEs support setting the working directory for non-debugging scenarios we could generalize the naming of this API.

TurkeyMan:
Even visual studio has both concepts (i think?).
So, votes for renaming `debugdir` to `workingdir`, or adding a second api such that both cases are specified independantly?
Perhaps a solution where `workingdir` is applied to both cases by default, and `debugdir` may override `workingdir` in the debug case if specified...?

starkos:
> So, votes for renaming `debugdir` to `workingdir`

samsinsane:
@starkos was this supposed to get closed?

starkos:
No, no idea what happened there. Reopening!

samsinsane:
@starkos thanks for sorting out the issues with the codelite module in core. I think that was the only thing holding this PR up, should be good to merge now. Let me know if there's any issues that I need to fix up.

samsinsane:
Has anyone had a chance to go over this PR? It'd be nice to finally have this sorted out, and merged into core.

tvandijck:
This might be nitpicking, but even though the "premake.action.set" change is in a separate commit, could we do that in a separate PR... I find that a really good change that I would like to see irrespective of the path.translate fix..

starkos:
I'm a little nervous about this one as that function is used all over the place, but okay, let's see how it goes!

PR #310, opened 2015-10-14
sln2005 Override Fix

samsinsane:
Fixed bug in `sln2005` that prevented p.override from working on the  `sln2005.elements` functions.


samsinsane:
Haha, I'm not having much luck with CI lately, I'm guessing it failed because the submodule reference is only in my fork?

PR #313, opened 2015-10-19
Solution files need a '\n\r' after the utf8 marker.

tvandijck:


starkos:
Been meaning to get to that—thanks!

tvandijck:
Yeah, it's weird that without it the 'VS Version Selector' doesn't identify the right version, but for vcxproj the \n\r are not needed.... not very consistent.... oh well... this works...

Issue #316, opened 2015-10-22
path.normalize issues


akaStiX:
It has issues again :(
There is a convention in Premake, that all paths should use **/** for all separators.
Apparently that doesn't work on Windows, when you try to execute some os commands like del\ren, etc.
Fixing this will be trivial, I just cannot do the pull request at the moment, so I'd like to get the opinions first

tvandijck:
well, that's why there is the path.translate method, to make the path windows compatible.
path.normalize always gives paths with '/'... that's not really an issue with path.normalize though, that's just what it does..

just do "path.translate(path.normalize(...))" for os command arguments.

akaStiX:
Well, than this is even simpler.
This is how os.execute looks like
```lua
	premake.override(os, "execute", function(base, cmd)
		cmd = path.normalize(cmd)
		cmd = os.translateCommands(cmd)
		return base(cmd)
	end)
```
it does normalization, but forgets about translation :(

starkos:
We can't just blindly translate the command though, because DOS commands use the forward slash to signify a switch, i.e. `dir /b`.

Suggestions welcome!

akaStiX:
well... Than we should get rid of path.normalize in os.execute at all...

starkos:
Perhaps. The way `path.normalize()` used to work it would clean up the command but not any arguments. Perhaps that has been broken by the recent changes though.

akaStiX:
Some time ago I've fixed path.normalize. Before that it was totally broken and it messed up the entire command

akaStiX:
@starkos  can you take a look on my pull request that solves this issue? https://github.com/premake/premake-core/pull/320

samsinsane:
I believe this issue has been resolved by #570.

Issue #279, opened 2015-09-21
os.execute dont called as expected.


friendwu:
I'm now using os.execute to retrieve 3rd packages, code like below:

```
        local xx = "wget https://github.com/redis/hiredis/archive/v0.13.3.tar.gz -O ./deps/hiredis_v0.13.3.tar.gz"
        printf(xx)
        local z = os.execute(xx)
        if z ~= 0 then
            error("download pkg failed", 0)
        end 
```

when I run premake5, the url are translated into a wrong value, log message are as below:

```
$ premake5 deps
Running action 'deps'...
getting deps hiredis
wget https://github.com/redis/hiredis/archive/v0.13.3.tar.gz -O ./deps/hiredis_v0.13.3.tar.gz
--2015-09-21 16:20:23--  ftp://https//github.com/redis/hiredis/archive/v0.13.3.tar.gz
           => "./deps/hiredis_v0.13.3.tar.gz"
Resolving https... failed: No address associated with hostname.
wget: unable to resolve host address “https”
Error: download hiredis failed
```

PS: my own lua interpreter can call os.execute as expected:

```
$ lua
Lua 5.2.3  Copyright (C) 1994-2013 Lua.org, PUC-Rio
> require "os"
> os.execute("wget https://github.com/redis/hiredis/archive/v0.13.3.tar.gz -O /vagrant/work/updns/deps/hiredis_v0.13.3.tar.gz")
--2015-09-21 16:22:03--  https://github.com/redis/hiredis/archive/v0.13.3.tar.gz
Resolving github.com... 
```

it really confuse me.

friendwu:
and just now I found premake support http.download, this API is not documented, and when I try it, the response message shows me that https is an Unsupported protocol.

tritao:
I've now documented the HTTP APIs on the wiki. About HTTPS, there's a pending pull request to support it, we should have it soon.

samsinsane:
This bug got me at work, I figured out that it's because premake performs `path.translate` on the string before passing it to the original Lua `os.execute`. Is there any reason why we need to translate the string before executing it? The translate seems unnecessary, and mostly destructive?

TurkeyMan:
Aye, does seem weird...

tritao:
Hmm... @samsinsane... that got me thinking.. that's probably why I've been getting some bugs when invoking commands with HTTP URIs.. they're passed down to the OS missing some slashes...

samsinsane:
[...]

PR #320, opened 2015-10-23
os.execute windows path separator fix

akaStiX:
Fix for a https://github.com/premake/premake-core/issues/316

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -11,7 +11,6 @@
 --
 
 	premake.override(os, "execute", function(base, cmd)
-		cmd = path.normalize(cmd)
 		cmd = os.translateCommands(cmd)
 		return base(cmd)
 	end)
@@ -539,7 +538,7 @@
 			local token = cmd:match("^{.+}")
 			if token then
 				token = token:sub(2, #token - 1):lower()
-				local args = cmd:sub(#token + 4)
+				local args = path.normalize(cmd:sub(#token + 4))
 				local func = map[token] or os.commandTokens["_"][token]
 				if func then
 					cmd = func(args)


akaStiX:
I am looking at the tests that failed and I don't understand, why do we need those trailing spaces in tokenized commands?!

samsinsane:
If I recall correctly, I added those tests for an implementation that was then proven to not work in all situations. An alternative was provided and I must've forgotten to remove the tests, you can remove the tests with `_ExtraSpace` in them as they don't test anything.

samsinsane:
I was slightly incorrect with my previous comment, those tests actually do something. They ensure the `IF EXISTS {dir}\\` check doesn't have a space between `{dir}` and `\\`, you can probably just adjust the `xcopy` parts to not contain the extra space.

tritao:
Ping, what's the status on this fix?

tvandijck:
1362 tests passed, 5 failed

tritao:
I can see there's some failures, but as far as I can tell it's just meaningless whitespace?

```
before: IF EXIST a\ (xcopy /Q /E /Y /I a  > nul) ELSE (xcopy /Q /Y /I a  > nul)
after: IF EXIST a\ (xcopy /Q /E /Y /I a > nul) ELSE (xcopy /Q /Y /I a > nul)
```

starkos:
Failing tests are failing tests. Either we need to fix the implementation to pass the tests, or adjust the tests to match the implementation (if the whitespace really is meaningless).

akaStiX:
I am sorry guys for a huge delay, I'll fix my pull requests

tritao:
Thanks @akaStiX.

samsinsane:
As I mentioned above, these tests are valid but the xcopy parts don't need to have the extra space. It only ensures that `IF EXISTS a\` isn't `IF EXISTS a \`.

Additionally, it seems it would be better if each token command dictated whether or not the arguments needed to be normalized, for example, this would break `echo https://github.com` (basically what was reported in #279 for `os.execute`) which would emit `echo https:/github.com`. Also, while we only have file system commands in here, if we (or a module) added support for things such as `wget`/`curl`, this will definitely break those as they rely on URLs.

Other than that, this LGTM once the tests expected values are fixed to not have the extra whitespace. Thanks @akaStiX!

tvandijck:
@akaStiX what is the status on this PR? do we want it merged, in which case we would really like to see it fixed... otherwise it's just going to sit here and wait for fixes.

samsinsane:
Closing this in favour of #570.

PR #324, opened 2015-10-24
OSX and Linux SSL support and OSX fixes

mversluys:
Small changes to get premake-core compiling.
Support for SSL added using Darwin SSL.
Tested on OSX Yosemite with Xcode 7.1

### DIFF
--- a/contrib/curl/lib/curl_config.h
+++ b/contrib/curl/lib/curl_config.h
@@ -2,7 +2,7 @@
 /* lib/curl_config.h.in.  Generated from configure.ac by autoheader.  */
 
 /* Location of default ca bundle */
-/* #define CURL_CA_BUNDLE "/etc/pki/tls/certs/ca-bundle.crt" */
+/* #undef CURL_CA_BUNDLE */
 
 /* Location of default ca path */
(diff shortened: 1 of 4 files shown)


starkos:
These look okay to me—Tom, do these changes cause any issue at your end?

tritao:
LGTM except the `CURL_CA_BUNDLE` change which should probably be limited to non-Linux platforms.

tvandijck:
> Tom, do these changes cause any issue at your end?

These changes come from here at Blizzard actually, so I think we're good ;)
But @tritao has a good point...

mversluys:
SSL wasn't enabled on Linux, so wasn't worried about the CURL_CA_BUNDLE.

Made the necessary changes to Linux so that it will use OpenSSL for SSL, if it's installed (it looks for the presence of the ssl library).

The CURL_CA_BUNDLE was changed to do the same test that's in the configure step for the curl library and looks in several locations for the certificates. The existing curl code was configured on CentOS 5 which uses a different location than other distros.

Tested on Ubuntu 15.10.

tritao:
Awesome, LGTM.

starkos:
Thanks everyone!

PR #327, opened 2015-10-26
Rollback PR #246, see discussion on ticket

starkos:


starkos:
D'oh, got the original PR wrong. Will create a new PR with the correct information.

PR #336, opened 2015-10-30
getrelativepath specific rule 'return absolute when only filesystem root in common'

matthieu-blin:
return the relative path Even if the file system root is the only thing in common

fix problems when generating a vs2012 project referencing another project, both in i the samefile system root
C:\Lib
C:\Project referencing Lib stuff in libdirs
>> generate vcproj with absolute "C:\Lib" inside
Unexpected since generating with C:\Git\Lib and C:\Git\Project generate only relative stuff

### DIFF
--- a/src/host/path_getrelative.c
+++ b/src/host/path_getrelative.c
@@ -48,9 +48,9 @@ int path_getrelative(lua_State* L)
 		++i;
 	}
 
-	/* if I end up with just the root of the filesystem, either a single
-	 * slash (/) or a drive letter (c:) then return the absolute path. */
-	if (last <= 0 || (last == 2 && src[1] == ':')) {
+	/* if they have nothing in common ( different drive by example ) 
+	 * return absolute path */
+	if (last <= 0) {
 		dst[strlen(dst) - 1] = '\0';
 		lua_pushstring(L, dst);
 		return 1;
--- a/tests/base/test_path.lua
+++ b/tests/base/test_path.lua
@@ -213,8 +213,12 @@
 		 test.isequal("c", path.getrelative("/a/b/","/a/b/c"))
 	end
 
-	function suite.getrelative_returnsAbsPath_onContactWithFileSysRoot()
-		test.isequal("C:/Boost/Include", path.getrelative("C:/Code/MyApp", "C:/Boost/Include"))
+	function suite.getrelative_returnsRelativePath_onContactWithFileSysRoot()
+		test.isequal("../../Boost/Include", path.getrelative("C:/Code/MyApp", "C:/Boost/Include"))
+	end
+	
+	function suite.getrelative_returnsAbsPath_onContactWithDifferentFileSysRoot()
+		test.isequal("E:/Boost/Include", path.getrelative("C:/Code/MyApp", "E:/Boost/Include"))
 	end
 
 


starkos:
The problem is that sometimes you really do want the system root, i.e. `/usr/lib`. And in those cases converting to a relative path, which often ends up look like `../../../../../usr/lib`, makes it impossible to relocate or share your project without regenerating.

I'll leave this up a bit for comment, but I would be inclined to *not* merge it for those reasons. Maybe someone can come up with a clever way to distinguish between the two that doesn't involve rewriting Premake's file path handling.

matthieu-blin:
Well i got only a small vision about premake, but for me, the /usr/lib sample is some kind of specific behavior and i should have specified it (with specific declaration or $libpathwhatever)
I agree it totally make senses on linux or maybe osx where we have some consensus about where to put lib, but on windows...
Anyway i surely don't want to break all projects here, and maybe we could solve the problem basic and straight :
-Having a way to force relative or absolute on specific root(s) on declaration
OR even stupid 
-Handle '<Letter>:' differently than '/' (don't see the case where i don't want relative when i reach 'E:')

that discussion aside, if someone has similar problem, and does not want to move their project folder into subfolder and can't recompile premake with a patch or anything, it can still override getrelative on extend.lua like this

premake.override(path, "getrelative", function(base, src, dst)
	-- normalize the two paths
	src = path.getabsolute(src)
	dst = path.getabsolute(dst)

	-- same directory?
	if (src == dst) then
		return "."
	end
	
	-- dollar macro? Can't tell what the real path is; use absolute
	-- This enables paths like $(SDK_ROOT)/include to work correctly.
	if dst:startswith("$") then
		return dst
	end
	
	src = src .. "/"
	dst = dst .. "/"

	-- find the common leading directories
	local idx = 0
	while (true) do
		local tst = src:find("/", idx + 1, true)
		if tst then
			if src:sub(1,tst) == dst:sub(1,tst) then
				idx = tst
			else
				break
			end
		else
			break
		end
	end
	
	-- if they have nothing in common return absolute path
	if idx == 0 then
		return dst:sub(1, -2)
	end
	
	-- trim off the common directories from the front 
	src = src:sub(idx + 1)
	dst = dst:sub(idx + 1)
	
	-- back up from dst to get to this common parent
	local result = ""		
	idx = src:find("/")
	while (idx) do
		result = result .. "../"
		idx = src:find("/", idx + 1)
	end

	-- tack on the path down to the dst from here
	result = result .. dst

	-- remove the trailing slash
	return result:sub(1, -2)
end)

samsinsane:
Could you please provide more information on these problems that VS2012 has with absolute paths? In my experience, Visual Studio uses absolute paths when creating a new project, regardless of location, so I find it hard to believe that it would have problems with absolute paths.

matthieu-blin:
Visual has no problem with absolute path, except that i have the exact same problem as you mention with /usr/lib : can't relocate or share the project if teammate just put their project on another drive.

samsinsane:
Oh ok, you said it fixes problems with generating VS2012 projects?

Anyway, your solution to the problem works for your specific case, but you can apply the example @starkos provided to a Windows environment as well. Having `C:\Lib` makes more sense than `..\..\..\..\..\..\Lib`, it also has the same relocation and sharing problems that he mentioned. In fact, it's even worse on Windows because there's multiple drives, which means if you work in `C:` but your team works in `C:` and `D:`, your generated projects are completely useless because they're 100% relative within `C:`.

Ultimately, I agree with @starkos, this shouldn't be merged.

matthieu-blin:
And i agree with both of you, so i'm closing this pull request (we'll stick with our hacky solution for our specific context)
Anyway i think we need more flexibility since both cases should be resolved, but can't think about more clever solution right now as @starkos  mentioned.
Thanks for your toughts on that anyway

PR #341, opened 2015-11-03
Add targetMachine property.

tvandijck:


tvandijck:
this is the last one so we can get rid of that "recognize linkoptions" pull request I had earlier... which I removed entirely from our premake code now...

starkos:
Great, thanks!

starkos:
Hey, if you get a chance could you create pages for these on the wiki? (Just get them into the Project API and What's New pages at least.)

tvandijck:
I'll put it on my todo list... probably after BlizzCon (http://us.battle.net/blizzcon/en/) though... busy week this week.

starkos:
Hi Tom, sorry to resurrect yet another old pull request, but why was this change needed? If I manually create a mixed architecture (Win32 & x64) project in Visual Studio it doesn't add these elements.

(Gradually returning from the dead here, trying to get all of my projects caught up to the latest code.)

Thanks!

tvandijck:
It doesn't, but once you rename configurations to "DebugSlow", "DebugFast", etc, the automatic properties from visual studio seem to not be picked up all the time... We ran into link warnings where a lib was build as 32bit, but the exe was trying to link as 64bit... so we had to explicitly set this property.

starkos:
Interesting…I'll play around and see if I can reproduce it.

starkos:
I've thrown everything I can think of at this and I can't reproduce this. Would it be okay if I rolled it back and introduced a `targetmachine` API for this instead? Or, if you could send over one or two of the affected projects perhaps I can deduce what is causing VS to get confused?

tvandijck:
Yeah, I tried removing it just now, to see what would happen:

ScintillaWin.cxx
ResourceCompile:
  
Lib:
  LINK : warning LNK4068: /MACHINE not specified; defaulting to X86
..\..\win32-x86_64-vc100-DebugFast\obj\Scintilla\LexA68k.obj : fatal error LNK1112: module machine type 'x64' conflicts with target machine type 'X86'

Build FAILED.


So it's definitely something we need... Now the interesting part is that this is on a pretty standalone library without any external dependencies, so I should in theory be able to isolate this and make a reproduction for you.... give me a few hours though.

tvandijck:
I send you a repro by email.

PR #342, opened 2015-11-04
Update to Curl 7.45.0

tvandijck:
Updating to latest stable curl, and fixing some of the premake settings we setup for curl.


tritao:
I noticed the new config files for our supported platforms. Did you generate them yourself with CMake or are they bundled by upstream now?

Also is there anything in particular you need from this new Curl version or just keeping things up to date?

Either way, LGTM.

tvandijck:
I generated them with cmake on Mac OSX 10.10, Centos 7 and Win32, and then hand modified them to fix some things here and there for windows in particular.. So it is not 100% the same as the curl distribution.

Nothing in particular I needed, just keeping things up to date..
I'm working on libzip, but that things is somewhat more complex.. I may give up and rip it out and use minizip instead, it's a lot smaller, and unzips a zip just as good...

Issue #196, opened 2015-08-04
What is the 'correct' way to add Frameworks for gmake projects on macos


JohannesMP:
Summary
-

I was wondering what the 'correct' way would be to include non-system *.framework files on macos in Premake5, with the ultimate goal of using gmake with clang or gnu compiler.

My assumptions about compiling with frameworks 
-

Correct me if I'm wrong, but as far as I've been able to find out, when compiling a program with a non-system framework under clang or gcc/g++ you need to:

1. specify what name of framework to use: `-framework <FRAMEWORK NAME>`
2. specify location of framework(s) to use: `-F <PATH TO FRAMEWORK>`

So for example, if I'm compiling a program using SDL2, where SDL2.framework is located at `/Library/Frameworks/` then I would add `-framework SDL2  -F /Library/Frameworks` to my compile string.

From some experimentation - see snippet: https://gist.github.com/JohannesMP/6ff3463482ebbdc82c2e - I've established that, at least on my machine, step **2.** is required and not including it causes the compiler to not be able to find the header file in the framework when compiling the object files, and also results in an error when linking them.

-------

Question regarding Premake5
-

When using Premake, I've seen some projects only do the following:

    ...
    configuration "macosx"
        links {"OpenGL.framework","CoreFoundation.framework","SDL2.framework"}
    ...

But, at least when using gmake, this seems to create a make file that does not include step **2.** above, and so I get the same error as I did in <a href =https://gist.github.com/JohannesMP/6ff3463482ebbdc82c2e>my snippet above</a> (lines 15 and 27)

To be able to account for this I've so far done the following:

    ...
    configuration "macosx"
        links {"OpenGL.framework","CoreFoundation.framework","SDL2.framework"}    
    configuration {"macosx", "gmake"}
        buildoptions {"-F /Library/Frameworks"}
        linkoptions {"-F /Library/Frameworks"}
    ...

However that has the problem of not being portable - for example some users don't have their frameworks located in the global frameworks directory and instead have it in their user directory, ie `~/Library/Frameworks`. Furthermore if I try to include both locations, there is a good chance that one of them may not yet exist on a user's machine (they're not created by default) and so they may get an error when running make:

    ld: warning: directory not found for option '-F/Users/Jo/Library/Frameworks'

So instead of using `buildoptions` and `linkoptions`, is there a better, portable way to include frameworks in Premake5?

[...]

PR #344, opened 2015-11-09
Added frameworkdirs option.

ardneran:
This should now support FRAMEWORK_SEARCH_PATHS in xcode_common.lua. This will allow frameworks to exist in non sdk paths.

configuration "macosx"
	frameworkdirs { "Library/MacOS/SDL2-2.0.3" }
	linkoptions { "-framework SDL2" }

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -672,6 +672,13 @@
 	}
 
 	api.register {
+		name = "frameworkdirs",
+		scope = "config",
+		kind = "list:directory",
+		tokens = true,
+	}
+
+	api.register {
 		name = "linkoptions",
 		scope = "config",
 		kind = "list:string",


samsinsane:
I'm not sure this is the best solution to the problem, there's a discussion going on here: #196. Additionally, you can use `links { "SDL2.Framework" }` instead of the `linkoptions` API, which is what makes this change problematic.

ardneran:
I have been wanting this feature for a while now so sorry for jumping the gun. However the problem here according to you is not the code but the use case. Xcode already supports framework search directories and all this does is give the user a way to populate those. You can ignore the way I'm using this change as I know it is probably not the right way. I just don't want users to have to install sdl in specific paths. See https://github.com/premake/premake-xcode/pull/22 for use case. Let me know if there is anyway I can help alleviate this problem.

starkos:
FWIW, if the `links{}` is the only objection, I'm okay with this.

samsinsane:
Sorry, I forgot to mention in my comment that the `links` thing is brought up in the discussion. I tend to agree with the opinion that we shouldn't have `links` support frameworks if we have `frameworkdirs`. I'm fine for this to go through, but it doesn't affect me as much as the others in the discussion. :)

Minor objection/question, should this PR update the premake-xcode submodule reference to the related PR over there?

ardneran:
I see your point. This is much more of a placeholder solution if you go the links way however one way of using links can be in conjuction with frameworkdirs. That way links can contain a list of frameworks and frameworkdirs can contain a list of directories to search for those frameworks. I don't see this as throwaway in the future if you go that way.

samsinsane:
Since the PR in the xcode module has been accepted and the submodules have been updated. I'm going to merge this in. @ardneran are you able to create an entry in the wiki for this new API. Thanks!

ardneran:
Thanks @samsinsane. I will add to the wiki and send a pull request.

PR #347, opened 2015-11-27
Fixed C# GNU Make backend response files to deal with platform path separators

tritao:
Obsoletes pull request #292.

Aditionally, we should be able to remove the `local sep = os.is("windows") and "\\" or "/"` line once pull request #307 is merged.


starkos:
Thanks! (I'll respond to your email shortly)

Issue #348, opened 2015-11-30
Is there a builtin API for generating platform configuration headers?


tarruda:
I'm trying to port a cmake script to premake but got stuck in the platform configuration bits:

```cmake
check_include_files (fcntl.h HAVE_FCNTL_H)
check_include_files (unistd.h HAVE_UNISTD_H)
check_function_exists (close HAVE_CLOSE)
check_function_exists (getpid HAVE_GETPID)
check_type_size (int64_t INT64_T)
configure_file (src/config.h.in src/config.h)
```

Is there a builtin API(or provided via modules) that allow me to do the same?

tvandijck:
Unfortunately premake does not currently have something similar to the "configure_file" API from cmake. I can certainly imagine a module providing similar functionality, but I don't think such a module exists today...

tarruda:
> Unfortunately premake does not currently have something similar to the "configure_file" API from cmake. I can certainly imagine a module providing similar functionality, but I don't think such a module exists today...

I could give it a shot if no one is working on this already. Should I open a PR?

tarruda:
I've been looking into how cmake implements these checks, and its a bit more complicated than I initially considered:

- cmake creates temporary cmake projects
- these temporary projects can inherit settings from the parent project
- each platform probably needs special handling to call the appropriate build tool.

I created a simplified proof of concept implementation for premake, it is crude but shows the idea:

```lua
local premake_c_template = [[
workspace 'try-compile'
  configurations {'test'}
  language 'C'

project 'try-compile'
  kind 'ConsoleApp'
  files 'test.c'
]]
local system_include_template = [[
#include <%s>

int main(void)
{
  return 0;
}
]]
local function_exists_template = [[
char %s();
int main(int argc, char*argv[]){
  %s();
  if(argc> 1000) {
    return *argv[0];
  }
  return 0;
}
]]
local type_size_template = [[
int main(void) {
  printf("%%zu", sizeof(%s));
}
]]

function os.try_compile(code, run)
  local dir = '/tmp/premake-check-if-compiles'
  os.mkdir(dir)
  if not os.chdir(dir) then
    error('cannot change to temporary directory')
  end
  -- write code to a test file
  local f = io.open('test.c', 'w')
  f:write(code)
  f:flush()
  f:close()
  f = io.open('premake5.lua', 'w')
  f:write(premake_c_template)
  f:flush()
  f:close()
[...]

PR #355, opened 2015-12-05
Add pkgconfig and platform-test modules

tarruda:
Hi

I have recently developed these two modules which IMHO provide useful features to properly configure packages on unix systems. While ATM they can be used as third party modules, I'd love to see them integrated into premake core.

#348 contains more information about the platform-test module


starkos:
Nice work! This could be a good core feature, but it needs to support more exporters and be used and tested by more people first.

PR #356, opened 2015-12-05
Allow http.get and download to follow redirections

Meoo:
Downloading from Github redirects to amazon S3, so follows must be enabled for http.download to work.
I added it to http.get too, but it may be less useful.


Meoo:
For some reason it compiled on VS, closing and reopening to have a single commit

PR #357, opened 2015-12-05
Allow http.get and download to follow redirections

Meoo:
Downloading from Github redirects to amazon S3, so follows must be enabled for http.download to work.
I added it to http.get too, but it may be less useful.

### DIFF
--- a/src/host/http.c
+++ b/src/host/http.c
@@ -193,6 +193,7 @@ int http_get(lua_State* L)
 		return 1;
 	}
 
+	curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
 	code = curl_easy_perform(curl);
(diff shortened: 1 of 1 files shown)


tritao:
lgtm

Meoo:
I got the "Unsupported procotol" error for both get and download with premake5 alpha6, so it seems to be required for get too if you want to follow redirections.

tritao:
On a second look... we are already seting up `CURLOPT_FOLLOWLOCATION` to 1 in `curl_request` without your changes... so it looks like this pull should not be needed, unless there's a bug in our implementation.

Could you double-check by re-compiling a Premake build from master without your changes and testing again?

Meoo:
You are right, everything works well with the latest version, my bad...

PR #367, opened 2015-12-17
make "premake.checkVersion" use Semantic Versioning

tvandijck:
http://semver.org/


tvandijck:
SemVer is the new fad, so lets comply ;)

samsinsane:
LGTM, I'll leave this for @starkos to check over, the change from .dev to -dev, same for alpha# and beta#, could be a problem?

starkos:
Sure, since we're breaking stuff left and right anyway ;) Thanks!

Issue #318, opened 2015-10-22
bootstrap.mak should exclude or remove scripts.c before a the bootstrap build.


tvandijck:
it's basically just including *.c, but if there is an old dangling scripts.c the bootstrap build may fail.

akaStiX:
removing scripts.c should be enough - it solved my problem

tvandijck:
This is actually rather easy to fix, we just did this locally...
instead of putting scripts.c in src/host/scripts.c, we put it into src/scripts.c

The bootstrap.mak file does not include any files in the src/ folder, but only src/host/*.c, so that way it will get automatically excluded.. I'll try to make a pull request for this, soon.

PR #369, opened 2015-12-17
Move scripts.c one up, so bootstrap.mak doesn't include it.

tvandijck:
fixes: https://github.com/premake/premake-core/issues/318


samsinsane:
LGTM

PR #370, opened 2015-12-18
Allow the pathVars to return a table that specifies explicitly that a token represents a relative or absolute path.

tvandijck:
Not all paths that start with a $ are absolute paths in Visual Studio, so the path.isabsolute is incorrect in many cases.


tvandijck:
@starkos If there are no objections, I'd like to merge this one.

starkos:
Haven't had a chance to look, go ahead and merge.

PR #375, opened 2015-12-18
adding /Debug:FastLink support for VS2015

tvandijck:


src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1231,7 +1231,15 @@
 
 
 	function m.generateDebugInformation(cfg)
-		m.element("GenerateDebugInformation", nil, tostring(cfg.flags.Symbols ~= nil))
+		if cfg.flags.Symbols and _ACTION >= "vs2015" then
+			if cfg.flags.FastLink then
+				m.element('GenerateDebugInformation', nil, "DebugFastLink")
+			else
+				m.element('GenerateDebugInformation', nil, "Debug")
```
tritao: Why the change in behaviour in this case? Are `Debug` and `DebugFastLink` the only allowed values in VS 2015?

starkos:
I've been trying to shift away from flags in favor of API calls. We currently have `debugformat()` which may not be a good fit. Perhaps we should have `symbols()` with values of "Off", "On", "Fast"?

src/actions/vstudio/vs2010_vcxproj.lua
samsinsane: The change is due to VS2015 having 3 valid options, while everything prior supported `true`/`false`, VS2015 supports `No`, `Debug` and `DebugFastLink`. The default value is `Debug`, so there should be support here to specify `No`. I agree with what @starkos was saying, this should be a proper API but I'm not sure if `symbols` is the correct name?

tvandijck: What @samsinsane is saying.... I'm all for a proper API, but the problem is, much of this only applies to VS2015, so the "symbols" api (or whatever the name is) should only allow certain values for certain actions... and there currently does not exist such a feature in premake.

samsinsane: @tvandijck I believe the solution is for all the actions to translate the values to their respective set of "allowed" values. For example, `optimize` supports six values and [gcc.lua](https://github.com/premake/premake-core/blob/b4bc07f4ed6ccdb8f94b4695283fda11203df4b7/src/tools/gcc.lua#L61-L68), [snc.lua](https://github.com/premake/premake-core/blob/b883c7f45f7e08b39effcf7d174833a67ee1133e/src/tools/snc.lua#L21-L28), [dotnet.lua](https://github.com/premake/premake-core/blob/b883c7f45f7e08b39effcf7d174833a67ee1133e/src/tools/dotnet.lua#L210-L214) and [msc.lua](https://github.com/premake/premake-core/blob/b883c7f45f7e08b39effcf7d174833a67ee1133e/src/tools/msc.lua#L56-L63) all translate the values differently and not just to toolset specific values, `Off` and `Debug` are the same in msc, 'Full' and 'Speed' are the same for gcc, only three are supported in dotnet but they're all the same and they're all different for snc.

@TurkeyMan was telling me to do this for the premake-android module PR I opened and it makes sense as it allows premake user scripts to be OS, toolset and action agnostic as much as possible. All we, as users, need to do is give premake a single value and those poor premake developers need to figure out what that means for each generator. :stuck_out_tongue:

tvandijck: It's a bummer that OS, toolset and action agnostic is somewhat of a pipedream... this is not a problem of premake really, CMake and many other systems have similar problems... all those build systems are just too damn different.

Anyway, I'm for proper API... I'll see if I can do that.... "symbolinformation" seems like a proper name no?
And then valid values are: "Default", "On", "Off", "FastLink", where "Default" will not emit the element, "On" is explicitly on, off explicitly off, and FastLink==On on all platforms but VS2015.

starkos: Sorry, jumping in without having done my homework, just wanted to point out the existing `debugformat()` API, in case they can somehow be merged here. And I think that just `symbols()` could be good for the API name; it would be great to get that flag deprecated.

tvandijck: I think the debugformat is very specific towards the actual format of the debug information. With symbols "off" there would be no debugformat, but with symbols "on" the format would be whatever you specified... rolling those into one API, would require every format to be rolled into the "on" and "FastLink" options, so you would get:

"Default", "DefaultC7", "Off", "On", "OnC7", "FastLink", "FastLinkC7"... I don't think that works really well, certainly not if additional formats would become available.... I'm actually surprised that only C7 is exposed in the "debugformat" API, since there is 3 other supported formats for MSVC.
- None
- Program Database (/Zi)
- Program Database for Edit And Continue (/ZI)

but interestingly to enable "EditAndContinue" you use an entirely different API
```lua
editandcontinue "On"
```

tvandijck:
For what it's worth... I just tried to add the 'symbols' API as proposed... that breaks about 109 unit-tests so far, which no real conclusive answer as to what it should really do...

I did this:

```lua
	api.register {
		name = "symbols",
		scope = "config",
		kind = "string",
		allowed = {
			"Default",
			"On",
			"Off",
			"FastLink",
		}
	}

	api.deprecateValue("flags", { "Symbols" }, nil,
	function(value)
		symbols "On"
	end,
	function(value)
		symbols "Default"
	end)
```

this changes the behavior quite a bit, for example in 

```lua
function m.debugInformationFormat(cfg)
```

When do you output the element? only when it is "on"/"fastlink"? no, that is not correct, because if it is "Default" it should also output the element, since we don't really know what the default is for Visual Studio. The problem however is, that now a lot of tests get this element added, since symbols are neither on or explicitly off. 

Should we even have "Default"? if not, why do some other API's have it? If we only have "On", and "Off"... then what is the default? So many questions....

starkos:
I don't think "Default" makes any sense for this one. The deprecation logic should be:

```lua
 api.deprecateValue("flags", { "Symbols" }, nil,
    function(value)
        symbols "On"
    end,
    function(value)
        symbols "Off"  -- this is what happens if you removeflags { "Symbols" }
    end)
```

Thanks for tackling this!

tvandijck:
Ok, so I removed "default" and set symbols to "Off" by default.... that results in 0 fails in tests for the core libraries, but unfortunately it breaks xcode/codelite since I don't have those branches locally...

It seems to be a relatively minimal change this way... but I'm not 100% convinced this is right.

tvandijck:
Oh, and I removed the 'location' calls from the contrib libs... this was intended for a seperate pull request... I'll see if I can isolate... it fixes that '--to' problem in the publish.

starkos:
> unfortunately it breaks xcode/codelite

This shouldn't be happening, the whole point of `api.deprecate()` is that it allows you to switch to a new behavior while not breaking the old one. So `cfg.flags.Symbols` should set get set, etc. I'm not seeing any obvious errors in your code though?

I do have the other modules in my working copy and will try to take a look at this.

tvandijck:
abandoning this, we'll keep this as a local change... I have no time to make this into a proper API setup.

PR #376, opened 2015-12-18
Add utility and makefile type in gmake.

tvandijck:


src/actions/make/make_makefile.lua
```diff
@@ -0,0 +1,93 @@
+--
+-- make_makefile.lua
+-- Generate a C/C++ project makefile.
```
tritao: This needs to be updated.

src/actions/make/make_utility.lua
```diff
@@ -0,0 +1,67 @@
+--
+-- make_utility.lua
+-- Generate a C/C++ project makefile.
```
tritao: Same.

src/actions/make/make_cpp.lua
```diff
@@ -98,6 +99,14 @@
 	end
 
 
+	function make.exePaths(cfg)
```
tritao: Is this related to the Utility and Makefile kinds changes?

starkos:
These would be great to have. The only thing that jumps out at me is we have separate calls for things like `makefileTargetRules()` and `utilityTargetRules()` while in the VS actions that would just be `targetRules()` with a if-else statement for the different project types. Not sure if one approach is really better than another of it they are just different.

src/actions/make/make_makefile.lua
tvandijck: updated to what?

src/actions/make/make_cpp.lua
tvandijck: unrelated, but I kind of had a hard time extracting it... it basically just adds support for "binpath", just like in MSVC.

tvandijck:
I have no opinion on what is better, I don't understand enough of the gmake backend to make a ruling on that.... we just found it easier to separate and be explicit...

src/actions/make/make_makefile.lua
tritao: At the time it seemed to me like a copy pasted description, the Utility and Makefile types should not be C/C++ project related right?

Trass3r:
What is the status of this?

tvandijck:
It's working for our use-cases.... it may not for many other use-cases... but it really just needs to be merged or declined... as I said, I have no opinion so I'm not clicking on the "merge" button.

Trass3r:
Well I guess having it in is way better than not as long as it exits if any assertions you made are violated.

starkos:
I haven't merged because I want to think about the API issues I mentioned in my earlier comment, and I was hoping someone else might chime in. I will go ahead and merge it.

tvandijck:
a cleanup/refactor pass is definitely in order to make this more similar to the visual studio action.. I'm swamped in work though, so I just don't have the time for this right now. anyone else is obviously welcome to contribute...

PR #377, opened 2015-12-18
Add new API characterset(); make Premake match Visual Studio default behavior

starkos:
This one might be contentious...

Visual Studio has been using Unicode as the default character set since at least 2008 (and possibly longer, I haven't checked back that far). During that time, Premake has been stubbornly defaulting to MBCS instead, to maintain backward compatibility with older versions of itself, and because we only had the "Unicode" or "not Unicode" flag to work with.

This PR proposes a clean break and switches the default to Unicode.

Chances are you either a) don't care what character set you're using, didn't even noticed it was MBCS, and wouldn't notice if it changed or b) are already setting `flags "Unicode"` in every one of your projects like me. Or you c) really do need MBCS and are going to be totally pissed when this PR lands and all your projects break.

So...thoughts? How do folks feel about merging this one?


TurkeyMan:
I strongly approve of this change, and batted for it in the past ;)

The set of people who really do need MBCS, and will be surprised that their project doesn't build/link will likely be very small. They should probably also tolerate this sort of change from a build clearly labelled 'alpha'. The errors this flag produces are noisy and impossible to miss; I've never heard of silent breakage from this change.

tritao:
:+1:

samsinsane:
:+1: I tend to agree with what @TurkeyMan said in regards to this. Additionally, [this MSDN article](https://msdn.microsoft.com/en-us/library/5z097dxa.aspx) says:
> If you are doing new development, you should use Unicode for all text strings except perhaps system strings that are not seen by end users. MBCS is a legacy technology and is not recommended for new development.

and
>In Visual Studio 2013 and later, the MFC libraries for multi-byle character encoding (MBCS) will be provided as an add-on to Visual Studio and will be available free of charge to Visual Studio customers (Professional and Enterprise editions only) from the MSDN download site. 

Microsoft is removing support for this technology, as well as openly recommending developers not use it, so it only makes sense that premake keeps up with the times and deprecates the use of this technology.

tvandijck:
well, they are only really removing it from MFC, which in itself is technology that is not recommended really ;) But yeah, I wholhartely agree with this change... even though it's already been merged ;)

Hey, I was on holiday break, skiing in Aspen... so I had more important things to do ;)

Zylann:
I have an old solution with loads of projects in it that I want to recompile but now it doesn't works in VS2015 with Unicode...
I don't want to migrate to Unicode now because it's an old project and I don't plan to work on it much. However there is something in it I want to check and run, and for that I need MBCS. What is the flag to setup in premake?

tvandijck:
```lua
characterset 'MBCS'
```

Zylann:
Thanks

PR #388, opened 2015-12-26
Fix: Write pchheader to Makefile as reletive path

stream009:
Because cfg.pchheader is "string" instead of "path", we need to resolve actual path before write it to the Makefile.

Without this, PCH variable in Makefile won't point actual file when Makefile is generated into different directory by the effect of "location" API function.


starkos:
Thanks!

PR #390, opened 2015-12-28
Fix VS warnings

starkos:
Clean up the build warnings from Visual Studio, so I can turn in warnings-as-errors for my local builds.


samsinsane:
LGTM, should we look at converting over to use the Unicode functions instead of the ANSI ones though?

starkos:
I'm fine with converting things to Unicode, but this PR is just to get things building cleanly.

PR #391, opened 2015-12-28
Split validation checks into individual tests and call arrays

starkos:
Breaks up the monolithic validation functions into individual test functions joined by call arrays, as we do in the exporters. This allows modules to add new tests and override existing ones in a more granular way.


samsinsane:
LGTM

PR #395, opened 2016-01-06
Fix some C# lang about BUGs

lailongwei:
Fix in vs c# project, 'debugdir' option invalid BUG;
FIx in linux c# project, generated makefile format error BUG(All path separator is back-slash[\], not adapted system type);
Fix vs c# project generate .user file format error BUG( The platform type is "Any CPU", but in .csproject, platform type is configured to "AnyCPU";

### DIFF
--- a/src/actions/make/make_csharp.lua
+++ b/src/actions/make/make_csharp.lua
@@ -92,11 +92,12 @@
 --
 
 	function cs.getresourcefilename(cfg, fname)
+		local sep = os.is("windows") and "\\" or "/"
 		if path.getextension(fname) == ".resx" then
 		    local name = cfg.buildtarget.basename .. "."
 		    local dir = path.getdirectory(fname)
 		    if dir ~= "." then
-				name = name .. path.translate(dir, ".") .. "."
+				name = name .. path.translate(dir, ".", sep) .. "."
 			end
 			return "$(OBJDIR)/" .. premake.esc(name .. path.getbasename(fname)) .. ".resources"
 		else
@@ -191,16 +192,17 @@
 		local makefile = path.getname(premake.filename(prj, ext))
 		local response = make.cs.getresponsefilename(prj)
 
+		local sep = os.is("windows") and "\\" or "/"
+
 		_p('$(RESPONSE): %s', makefile)
 		_p('\t@echo Generating response file', prj.name)
 
 		_p('ifeq (posix,$(SHELLTYPE))')
 			_x('\t$(SILENT) rm -f $(RESPONSE)')
 		_p('else')
-			_x('\t$(SILENT) if exist $(RESPONSE) del %s', path.translate(response, '\\'))
+			_x('\t$(SILENT) if exist $(RESPONSE) del %s', path.translate(response, sep))
 		_p('endif')
 
-		local sep = os.is("windows") and "\\" or "/"
 		local tr = project.getsourcetree(prj)
 		premake.tree.traverse(tr, {
 			onleaf = function(node, depth)
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -271,7 +271,7 @@
 
 		local arch = architecture(cfg.system, cfg.architecture)
 		if not arch then
-			arch = iif(isnative, "x86", "Any CPU")
+			arch = iif(isnative, "x86", "AnyCPU")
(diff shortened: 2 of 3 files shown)


src/actions/vstudio/vs2005_csproj_user.lua
```diff
@@ -57,6 +57,7 @@
 				if #contents[cfg] > 0 then
 					p.push('<PropertyGroup %s>', m.condition(cfg))
 					p.outln(contents[cfg])
+                    m.localDebuggerWorkingDirectory(cfg)
```
samsinsane: This should be put inside the call array `m.elements.userConfigPropertyGroup` defined on line 21 of this file.

src/actions/vstudio/vs2005_csproj_user.lua
```diff
@@ -65,7 +66,11 @@
 		end
 	end
 
-
+    function m.localDebuggerWorkingDirectory(cfg)
```
samsinsane: This function should probably be placed in a logical place, with the other user configuration property group functions, either above or below `m.localDebuggerCommandArguments` depending on the order of the call array `m.elements.userConfigPropertyGroup` mentioned in my comment above.

src/actions/vstudio/_vstudio.lua
```diff
@@ -271,7 +271,7 @@
 
 		local arch = architecture(cfg.system, cfg.architecture)
 		if not arch then
-			arch = iif(isnative, "x86", "Any CPU")
+			arch = iif(isnative, "x86", "AnyCPU")
```
samsinsane: I'm not exactly sure what the issue you're trying to resolve is, but this has introduced another bug, the configuration is now `Debug Any CPU` instead of `Debug`.

src/actions/make/make_csharp.lua
```diff
@@ -191,16 +192,17 @@
...
+		local sep = os.is("windows") and "\\" or "/"
+
 		_p('$(RESPONSE): %s', makefile)
 		_p('\t@echo Generating response file', prj.name)
 
 		_p('ifeq (posix,$(SHELLTYPE))')
 			_x('\t$(SILENT) rm -f $(RESPONSE)')
 		_p('else')
-			_x('\t$(SILENT) if exist $(RESPONSE) del %s', path.translate(response, '\\'))
+			_x('\t$(SILENT) if exist $(RESPONSE) del %s', path.translate(response, sep))
```
samsinsane: There's a test case for this particular output located in tests/actions/make/cs/test_response.lua on line 70. You should create two versions of this function, one for Windows and one for Linux, to ensure that it's correctly emitting the OS separator.

tritao: This particular branch should only executed on Windows so why did you need to change this separator?

src/actions/vstudio/_vstudio.lua
lailongwei: This issus is in visual studio c# project, when i use debugdir, debugargs option to setting my debug directory and debug arguments, bug this two options are invalidate,  open .user file and .csproj file to compare, I found in .csproj file the platform is "AnyCPU", but in .user file the platform is "Any CPU", I open the project settings in visual studio, try to config a debug directory(settings->debug) and then save the project settings, and open the .user file again, I found visual studio append my settings above, and the Platform is "AnyCPU", so I guess this is a premake bug in C# project, so I trying to modify this line code, In other language type, I have not tested it.

src/actions/make/make_csharp.lua
lailongwei: I think the premake code must ignore platform differences, so the <slash> and <backslash> should be hide, so modify this line code.

src/actions/vstudio/vs2005_csproj_user.lua
lailongwei: En, when I modify this code, I'm in company, and I also know this fix code can not place here... I'm very sorry, Is in a hurry, for this change, I will modify the code again and request pull again,

src/actions/vstudio/vs2005_csproj_user.lua
lailongwei: Ok, thanks

samsinsane:
There hasn't been any updates to this PR in over 2 months. If you fix up the issues raised in this PR, please feel free to reopen the PR for review.

PR #396, opened 2016-01-10
Fix the oven to store the global container in the project and configs.

tvandijck:
I wanted to do:
```lua
premake.api.register {
	name = "packagecache",
	scope = "global",
	kind  = "path",
}
```

but that crashes in the validateScopes on this line:
```lua
okay = okay or p.field.compare(f, self[scope][f.name], self[f.name])
```
since self['global'] did not exist for the 'project' and 'config' scopes.

### DIFF
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -121,6 +121,8 @@
 		verbosef('    Baking %s...', self.name)
 
 		self.solution = self.workspace
+		self.global = self.workspace.global
+
 		local wks = self.workspace
 
 		-- Add filtering terms to the context to make it as specific as I can.
@@ -456,6 +458,7 @@
 		ctx.project = prj
 		ctx.workspace = wks
 		ctx.solution = wks
+		ctx.global = wks.global
 		ctx.buildcfg = buildcfg
 		ctx.platform = platform
 		ctx.action = _ACTION
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -43,7 +43,13 @@
 ---
 
 	function oven.bake()
-		p.container.bake(p.api.rootContainer())
+		-- reset the root _isBaked state.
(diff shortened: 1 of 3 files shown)


tvandijck:
This is a work in progress on something I'm doing at home, so I'm a bit limited to working on this during the weekends where I have time... not sure yet why things broke though.

src/base/oven.lua
```diff
@@ -43,7 +43,7 @@
 ---
 
 	function oven.bake()
-		p.container.bakeChildren(p.api.rootContainer())
+		p.container.bake(p.api.rootContainer())
```
samsinsane: I think this is what broke your build, the function `bake` is in the `p.global` namespace.

tvandijck:
This works now, the last fix basically is really only a fix to get the unit-tests going, and it does not in any way affect the normal execution of things...

in the unit-tests we create and destroy workspaces and projects, and bake several times to test for the expected output... The "global" object however, is only created once, at startup... so with these changes the baking no longer just baked the workspaces, but starts with baking the global object... which means that in the unit-tests this broke, since the _isBaked value would simply think that the global object was already baked... while true, in the unit-tests that prevents anything else from ever bake...

In normal execution however... the action only bakes everything once, and hence the global object gets correctly baked once. We just never call bake twice in that situation, so this work-around if you will just never really happens on normal execution...

The effect of this change however is that configsets in the global space get correctly baked as well.
So you can do:

```lua
premake.api.register {
    name = "packagecache",
    scope = "global",
    kind  = "path",
}
```

and then in the action or anywhere after the baking you can access the value as such:

```lua
wks.global.packagecache
```

This could have obviously been achieved with just a global variable.. but the nice thing of the API system is that is allows path expansion, file lists, tokens, etc... so I found this really useful.

samsinsane:
:+1: LGTM

PR #398, opened 2016-01-11
add premake extension data.

tvandijck:
I added this data to the output solution, which allows this visual studio extension to then run the premake step from within visual studio.

https://github.com/tvandijck/PremakeExtension

The extension currently only supports Visual Studio 2015, but I'm sure other people can help getting that to work for other versions if they so please..

### DIFF
--- a/src/actions/vstudio/vs2005_solution.lua
+++ b/src/actions/vstudio/vs2005_solution.lua
@@ -29,6 +29,7 @@
 			"ConfigurationPlatforms",
 			"SolutionProperties",
 			"NestedProjects",
+			"ExtensibilityGlobals"
 		}
 	end
 
@@ -293,6 +294,32 @@
 
 
 --
+-- Write out the ExtensibilityGlobals block, which embeds some data for the
+-- Visual Studio PremakeExtension. In two seperate methods for override-ability.
+--
+	function sln2005.PremakeGlobals(wks)
+		-- we need to filter out the --file argument, since we already output
+		-- the script separately.
+		local args = {}
+		for _, arg in ipairs(_ARGV) do
+			if not (arg:startswith("--file") or arg:startswith("/file")) then
+				table.insert(args, arg);
+			end
+		end
+
+		_p(2, 'PremakeBinary = %s', _PREMAKE_COMMAND)
+		_p(2, 'PremakeScript = %s', p.workspace.getrelative(wks, _MAIN_SCRIPT))
+		_p(2, 'PremakeArguments = %s', table.concat(args, ' '))
+	end
+
+	function sln2005.ExtensibilityGlobals(wks)
+		_p(1, 'GlobalSection(ExtensibilityGlobals) = postSolution')
+		sln2005.PremakeGlobals(wks)
+		_p(1, 'EndGlobalSection')
+	end
+
+
+--
 -- Map solution sections to output functions. Tools that aren't listed will
 -- be ignored.
 --
@@ -301,7 +328,8 @@
 		return {
 			sln2005.configurationPlatforms,
 			sln2005.properties,
-			sln2005.NestedProjects
+			sln2005.NestedProjects,
+			sln2005.ExtensibilityGlobals,
 		}
 	end
 
--- a/src/actions/vstudio/vs2005_solution.lua
+++ b/src/actions/vstudio/vs2005_solution.lua
@@ -295,27 +295,48 @@
 
 --
 -- Write out the ExtensibilityGlobals block, which embeds some data for the
--- Visual Studio PremakeExtension. In two seperate methods for override-ability.
+-- Visual Studio PremakeExtension.
 --
 	function sln2005.premakeExtensibilityGlobals(wks)
-		-- we need to filter out the 'file' argument, since we already output
-		-- the script separately.
-		local args = {}
-		for _, arg in ipairs(_ARGV) do
-			if not (arg:startswith("--file") or arg:startswith("/file")) then
-				table.insert(args, arg);
+		if wks.editorintegration then
+			-- we need to filter out the 'file' argument, since we already output
+			-- the script separately.
+			local args = {}
+			for _, arg in ipairs(_ARGV) do
+				if not (arg:startswith("--file") or arg:startswith("/file")) then
+					table.insert(args, arg);
+				end
 			end
-		end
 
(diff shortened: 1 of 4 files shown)


src/actions/vstudio/vs2005_solution.lua
```diff
@@ -293,6 +294,32 @@
...
+		end
+
+		_p(2, 'PremakeBinary = %s', _PREMAKE_COMMAND)
+		_p(2, 'PremakeScript = %s', p.workspace.getrelative(wks, _MAIN_SCRIPT))
+		_p(2, 'PremakeArguments = %s', table.concat(args, ' '))
+	end
+
+	function sln2005.ExtensibilityGlobals(wks)
+		_p(1, 'GlobalSection(ExtensibilityGlobals) = postSolution')
+		sln2005.PremakeGlobals(wks)
```
samsinsane: It would probably be good to use a call array here instead, it allows modules to add more extensions and what not. :)

samsinsane:
I'm quite excited to try this out, but I'm wondering if this should be locked behind an API call. `editorintegration` or something, which allows the end user to choose whether they want to use this extension or not. Maybe even emit a warning if this API is used on non-VS actions, maybe even non-VS2015 actions but still emit the extension data?

starkos:
I would definitely at least want the option to turn if off, even if it is on default.

Also, if possible, I'm trying to migrate to the indentation aware `p.push(); p.w(); p.pop()` instead of `_p()`.

Curious to give it a try!

samsinsane:
I've tried it out, it's awesome! Only issue is it requires the premake script entry to work, which means you need these changes but that's acceptable really. For testing purposes, I altered it locally to only require the binary, defaulting the script to premake5.lua. I am definitely going to push this extension at work, we currently just have batch files beside our premake5.lua files to quickly regenerate the solutions. Thanks Tom!

tvandijck:
> I would definitely at least want the option to turn if off, even if it is on default.
Sure, I can certainly add that I guess... I didn't because the extra data in the solution is harmless if you don't have the plugin... It can really contain any kind of data..

> Also, if possible, I'm trying to migrate to the indentation aware p.push(); p.w(); p.pop() instead of _p().

I copy pasted from "sln2005.NestedProjects", which used the API I'm using, but I'm sure we can fix that..

> For testing purposes, I altered it locally to only require the binary, defaulting the script to premake5.lua

I'm sure we can make the "Extension" have some sane defaults that it detects if these tags are not available in the solution... There is already a property page in the "Tools / Options" dialog in visual studio that allows you to specify a premake binary and a checkmark to override the binary tag in the solution. Possibly we can make the Extension somewhat smarter for sure... feel free to submit pull requests in that repository, I'd be happy to take such input.

The reason however why the "PremakeScript" tag is in the solution, is that when you specify the --to option to premake, is that the .sln file can really be anywhere, and in our case on Starcraft we specifically default it to be elsewhere, so the Extension wouldn't know where to find it otherwise, unless I hardcoded the Starcraft 2 rules into the Extension, which I didn't want to do.

The same went for the 'PremakeBinary" tag, I added that after I had the propertypage setting which I though was going to be enough, but Starcraft 2 has several branches, and each of those branches sometimes have a different version of premake committed, and these may or may not always be backwards compatible... so I made the property page only be a backup or when the checkbox is checked an override...

But again, as far as the Extension goes, it's a public repository, and I'd be happy to review and integrate pull requests if they are sane. And if people like we can also move it in to the "premake" organization here on github (since it is now on my personal page).

starkos:
> Sure, I can certainly add that I guess... I didn't because the extra data in the solution is harmless if you don't have the plugin... It can really contain any kind of data..

Two justifications for my comment: a) some environments can be very sensitive about diffs landing upstream due to regulation/validation, and b) it is not uncommon to pre-generate project files to ship to customers (who do not use Premake) and want everything to look neat and tidy.

tvandijck:
Oh I understand the desire ;) and the additional commit here add the flag... as well as fixes the p.push/pop usage... which we should btw fix globally in that file...
also added a test, to make sure the flag works properly... it's enabled by default, but I can go either way.

samsinsane:
I'm pretty happy with most of these changes. I still think it should use a call array, allowing a module to override the call array instead. Currently, a module would either emit another `GlobalSection` or it would need to check `editorintegration` before overriding the functions. Additionally, I think the check for `editorintegration` should be in the `premakeExtensibilityGlobals` function. 

Slightly off topic: Just to add to the last part, in the future it might good to add a conditional function for the call array elements. So instead of just specifying a list of functions, you specify a condition function and the actual function. I believe there's a few sections in the vcxproj generator where you can emit nothing at all, but instead an empty section tag appears in the file.

tvandijck:
> @samsinsane  I still think it should use a call array

I thought about that as well... I wasn't really sure if that was desired though...
So instead I split things up into two methods, one that sets up the GlobalSection, and one that emits the actual data that goes inside... so you could simply override premakeExtensibilityGlobals, emit your tags, and then call base.

> I think the check for editorintegration should be in the premakeExtensibilityGlobals function.

Thought about that too... I think emitting an empty GlobalSection however is also undesirable.. so if that is the case, the if statement is in the right place... or we would have to capture whatever is emitted inside, then see if that is empty or not and emit the result... we do that in other places too I beleive... it just get more and more complex for emitting the 3 lines of code I wanted to emit.

As for the rest, I'll leave that up to @starkos to comment on..

samsinsane:
@tvandijck I agree with you, and I feel like I'm being difficult, so I apologize for that. But `premakeExtensibilityGlobals` won't be called if `editorintegration` is false, which means modules would need to override both functions and do different things depending on the value of `editorintegration` which will result in multiple tags for `GlobalSection` when it's false. I'll admit, if VS can handle it, it's not the worst case, but it does put complexity onto module developers.

Anyway, overall, I'm happy with this PR :+1:, thanks again Tom. I've got a couple of ideas for the extension, but I'll need to find some time to put aside to work on it. I'm really excited to work on it and see where it goes.

tvandijck:
@samsinsane OK, so callarray it is, and I moved the if statement, and now use the p.capture thing to see if there is any content, if not then the tag isn't made... so this should satisfy all you wishes ;) hehe. I agree it's better too...

When this passes, I'll merge it, if there are no further objections.

Trass3r:
I'd prefer a solution more like CMake with a base VS project of type makefile/utility/whatever that contains the premake script and the commandline used to generate the solution.

Issue #400, opened 2016-01-15
Specifying link mode preference for external libraries on Linux


yuyoyuppe:
So I saw [some discussion](https://groups.google.com/forum/#!searchin/premake-development/linking/premake-development/szEuw8GRjpY/4RNOOG1gBQAJ) about implementing this feature, but it's not in the alpha 7 yet. Any updates planned on that? Also `*:static"`/`*:shared"` is still prefered syntax, right?

samsinsane:
Closing as #401 was merged.

PR #401, opened 2016-01-15
Link mode preference

yuyoyuppe:
Possible #400 implementation. Now we can set a flag to control linker's behavior like this:
```lua
  links {"boost_system:shared"}
  links {"boost_thread:static"}
  links {"boost_regex"} -- flag is optional
```
Default link mode is shared, i.e. `-Wl,-Bdynamic`.

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -331,7 +331,7 @@
 		if #static_syslibs > 1 then
 			table.insert(static_syslibs, "-Wl,-Bdynamic")
 			move(static_syslibs, result)
-	  end
+	    end
 		move(shared_syslibs, result)
 
 		return result
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -297,7 +297,6 @@
 		end
 
 		-- The "-l" flag is fine for system libraries
-
 		local links = config.getlinks(cfg, "system", "fullpath")
 		for _, link in ipairs(links) do
 			if path.isframework(link) then
@@ -305,7 +304,20 @@
 			elseif path.isobjectfile(link) then
 				table.insert(result, link)
 			else
-				table.insert(result, "-l" .. path.getname(link))
+				local endswith = function(s, ptrn)
+					 return ptrn == string.sub(s, -string.len(ptrn))
+				end
+				local name = path.getname(link)
+				-- Check link mode preference and set flags for linker accordingly
+				if endswith(name, ":static") then
+					name = string.sub(name, 0, -8)
+					table.insert(result, "-Wl,-Bstatic -l" .. name .. " -Wl,-Bdynamic")
+				elseif endswith(name, ":shared") then
+					name = string.sub(name, 0, -8)
+					table.insert(result, "-Wl,-Bdynamic -l" .. name)
+				else
+					table.insert(result, "-l" .. name)
+				end
 			end
 		end
 
@@ -360,4 +372,3 @@
 		end
 		return nil
 	end
-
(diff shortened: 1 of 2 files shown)


starkos:
Looking good. It would be nice it would group the libraries by link mode to reduce the number of `-Wl` flags on the command line. Mostly aesthetic perhaps?

tvandijck:
LGTM.

samsinsane:
LGTM, I'm not 100% sure how the `getlinks` function works, but if you link against one of your own projects does it auto-detect the link mode? This was mentioned in the forum link, but what you added doesn't seem to do any detection, is it handled in the code that was already there?

yuyoyuppe:
>if you link against one of your own projects does it auto-detect the link mode?

Do you mean if dependent library projects(`SharedLib`/`StaticLib`) are still linked correctly? This decorator is only used for 3rdparty and system libraries. So yes, it's handled elsewhere and wasn't affected by the change.

A few words about aesthetics. From gcc documentation:

```
VxWorks Options
<...>
-Bstatic
-Bdynamic
  These options are passed down to the linker.  They are defined for compatibility with Diab.
```

Is there another native way? Yes, _undocumented_ `-l:` flag which works like this:
```
g++ <...> -l:libboost_program_options.a
```
It seems that static linking isn't popular on Linux.

tvandijck:
Just looking at this, "This decorator is only used for 3rdparty and system libraries"
it would be nice if the system gave an error if you used the decorator on any 'of your own projects'.
that way at least it would catch it.

other then that, this looks good to me, and I'd like to merge it asap.

IstuntmanI:
I have a problem when linking static libraries, I guess that it is pretty much related to this pull request, so I won't create an issue for now:
```lua
links { "lib/mysql6.1.10/linux/libmysqlclient.a:static", "lib/mysql6.1.10/linux/libmysqlcppconn-static.a:static" }
```
it outputs this line to the `.make` file:
```cmake
LIBS += -Wl,-Bstatic -llibmysqlclient.a -llibmysqlcppconn-static.a -Wl,-Bdynamic -lrt
```
and it gives me these errors when it comes to the linking stage:
```
/usr/bin/ld: cannot find -llibmysqlclient.a
/usr/bin/ld: cannot find -llibmysqlcppconn-static.a
```

To fix this I need to change the `.make` file:
1. I need to specify the full path.
2. I need to remove the `-l` part before the path

so this works just fine:
```cmake
LIBS += -Wl,-Bstatic lib/mysql6.1.10/linux/libmysqlclient.a lib/mysql6.1.10/linux/libmysqlcppconn-static.a -Wl,-Bdynamic -lrt
```

Is there any way to fix this by using premake, or it is actually a bug ? I'm not experienced with this tool. I'd like to keep using the `links` rule

samsinsane:
@IstuntmanI Try using the following instead:
```lua
libdirs { "lib/mysql6.1.10/linux" }
links { "libmysqlclient.a:static", "libmysqlcppconn-static.a:static" }
```

IstuntmanI:
I'm already using
```lua
libdirs { "lib/**" }
```

samsinsane:
@IstuntmanI so why are you specifying the full path to the libraries in `links`? Does it work when you remove the path from `links` and just have what I said above? Also, I don't think you need the `:static` qualifier in this particular case, as you're linking a static library?

IstuntmanI:
This is why:
```make
g++: error: libmysqlclient.a: No such file or directory
g++: error: libmysqlcppconn-static.a: No such file or directory
```

Doing `libdirs { "lib/**" }` is just fine, this is the line it generates (simplified a bit):
```lua
ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib32 -Llib/mysql6.1.10 -Llib/mysql6.1.10/linux -Llib/mysql6.1.10/windows -m32 -shared -Wl,-soname=NAME.so -s
```

There's just something wrong with linking a static library in a local path. I described it in my first comment above. I tried various alternative but none works without having to edit manually the `.make` file.

PR #404, opened 2016-01-17
add macro to bootstrap to override the visual studio version.

tvandijck:


samsinsane:
Does this need to be manually manipulated, or can it be specified from the command line?

tvandijck:
well the default is set to vs2012, but you can now specify this macro on the command line to override it...
so you can do:
```
nmake -f Bootstrap.mak MSDEV=vs2015 windows
```

and it will run with vs2015 instead... I don't do any validation of the macro however, so if you put garbage in it, you'll get garbage... anyway, the new "jenkins" windows box uses this, since I didn't want to install both vs2012 and vs2015 just to support the upgrade path.

samsinsane:
Are you able to update all the documentation to reflect this change? I'm not sure how many things there are, but I guess `BUILD.txt` and the wiki entry should do. :)

tvandijck:
sure thing.

samsinsane:
Thanks :)

PR #409, opened 2016-01-20
Add support for DOS-style environment variables, e.g. %JAVA_HOME%

starkos:
We currently recognize paths that start with an environment variable like `$(HOME)` and treat them as absolute. This treats DOS-style variables like `%JAVA_HOME%` the same way.


samsinsane:
:+1: LGTM

tvandijck:
This change broke the tokenizer::
       targetdir = path.join(_MAIN_SCRIPT_DIR, "%{cfg.buildcfg}")

will now just return "%{cfg.buildcfg}"

PR #424, opened 2016-02-05
clean up the build settings a little.

tvandijck:


tvandijck:
This just removes useless tags from each project, since we can globally set these at the solution level, which makes scripts smaller and easier to maintain.

samsinsane:
:+1: LGTM

samsinsane:
@tvandijck when you get a chance, can you fix up the conflicts that have occurred? I think you're good to merge it in too. Cheers :)

PR #429, opened 2016-02-22
Remove debug settings from core build scripts

starkos:
These debug settings cause the absolute path to the local premake-core project to be written into the Visual Studio .user file. This isn't a portable path, and shouldn't be generated by the distributed project scripts.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -136,8 +136,6 @@
 			targetdir   "bin/debug"
 			defines     "_DEBUG"
 			flags       { "Symbols" }
-			debugargs   { "--scripts=" .. path.translate(os.getcwd()) .. " test"}
-			debugdir    ( os.getcwd() )
 
 		configuration "Release"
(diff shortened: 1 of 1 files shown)


samsinsane:
:+1: LGTM

tvandijck:
can't we make it relative instead? this is such a bummer to loose.... so much convenience.

starkos:
I believe it was original relative and then a recent PR made it absolute, let's see…look like [this one](https://github.com/premake/premake-core/pull/59)?

tvandijck:
that's weird, things worked fine for me in VS2010,2012,2013 and 2015...
oh well, I'll put this back in as a local change... sucks...

starkos:
I'm fine with rolling back this change if things work…haven't had a chance to investigate it. Maybe open an issue for it and I can look into it?

tvandijck:
No worries, I'll have a look at it...

PR #430, opened 2016-02-23
Add Build Customizations option

lwjason:
Support Build Customization settings for projects. Now Premake5 can
recognize Build Customizations File that 3rd party API (e.g. CUDA)
maded.

Use this like : 
buildcustomizations{ "CUDA 7.5" } in the project

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -78,6 +78,12 @@
 	}
 
 	api.register {
+		name = "buildcustomizations",
+		scope = "project",
+		kind = "list:string",
+	}
+
+	api.register {
 		name = "builddependencies",
 		scope = { "rule" },
 		kind = "list:string",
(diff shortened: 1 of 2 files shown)


samsinsane:
:+1: LGTM.

My only concern is the hardcoded location of `$(VCTargetsPath)\\BuildCustomizations`, [vs-tool](https://github.com/juj/vs-tool) which is used by another Premake module, doesn't appear to use this folder instead places all customizations beside Microsofts files. It seems like the folder is arbitrary and, potentially, unique to CUDA?

starkos:
I agree that `BuildCustomizations` should not be part of the hardcoded path. I *think* `$(VCTargetsPath)` is okay though? I try to check some sample projects…

samsinsane:
Oops, sorry, yes it's the `BuildCustomizations` folder that I'm concerned about, `$(VCTargetsPath)` should be fine. Let me know how it goes with the sample projects. :)

starkos:
Will do—in the meantime I would be okay with merging with `$(VCTargetsPath)`.

starkos:
@lwjason — if you make that change and push it to your repository/branch this pull request will automatically update, and then we can go ahead and get it merged in. Thanks!

samsinsane:
Thanks @lwjason!

PR #433, opened 2016-02-25
Fixed issue with incorrect UUID being generated for groups

### LINKED ISSUE #100: problem with group containing project name
Hello,

I run into this issue, well describe here
https://bitbucket.org/premake/premake-dev/issue/260/group-produces-incorrect-results-in-visual

I don't know if you want me to stop duplicating issues from bitbucket?

I'm running premake5 alpha 3

samsinsane:
Fixes #100


starkos:
With this approach, you could end up with different UUIDs for a project if it is included in two different solutions using two different groups. We need the project UUID to always be the same in the scenario.

The *group* UUID can be different in each solution though. Perhaps we can prepend a string at workspace.lua:128 instead?

```lua
node.uuid = os.uuid("group:" .. node.path)
```

samsinsane:
That makes sense, should it only prepend that string when it's a group?

starkos:
I'm pretty sure it only uses that value for groups. For projects, it would look at `prj.uuid` instead.

samsinsane:
Alright, I've made those changes, also fixed up the tests. Just waiting on the build to pass, but should be good to go otherwise. :)

tvandijck:
LGTM

PR #435, opened 2016-02-25
debugargs & debugdir defaults for convenient debugging.

tvandijck:


tvandijck:
For what it is worth.... the fact that visual studio does not accept relative paths in LocalDebugDirectory is total BS... This works in all version of Visual Studio I have installed, which includes 2010, 2012, 2013 and 2015.

Now... the "LocalDebugDirectory" property however does work relative from the vcxproj on disk, so if you output those to some location other then your source folder, you will have to adjust for that obviously.

As for the "DebugArgs", those are passed into the application "as-is" and hence paths are relative to the working folder... using the $(ProjectDir) token kind of gets around that in this case, so that it always points to an "absolute" location.. but other then that, there are again no restrictions made by Visual Studio itself.

starkos:
My bad then, as I think I'm the one who approved the other PR. We'll need to back that one out too…I'll go do that now.

tvandijck:
Well, before we do that though... we should ask the maker of that PR what his issues were, and what problem he was trying to fix...

I'm not against his changes at all, if they actually fix something that is broken, but I'm not seeing the behavior he describes, other then that the paths are indeed somewhat "hard" to setup correctly, since they are relative to the location of the vcxproj... but that is something we can adjust for...

I'd like things to go forward, and I'm here to help wherever I can, but I don't feel reverting things solves much.

starkos:
Sorting this out now. In the meantime, this change would be cleaner as:

```lua
debugargs   { "--scripts=%{prj.location} test"}
debugdir    "%{pr.location}"
```

This version still expands to `$(ProjectDir)`, but will also work with other toolsets.

starkos:
Ah okay I see, it was the `path.translate(os.getcwd())` in the previous call that was messing things up and causing the absolute path. I will merge the tweaked version of this now…

PR #436, opened 2016-02-25
disable warnings in contrib libs.

tvandijck:


tvandijck:
Just missed these little tags for the contrib library settings.... we're not going to fix them, so lets just disable them... this was already the case for curl.

PR #438, opened 2016-02-25
Move lua into 'contrib' folder, and build as any other 3rd party lib.

tvandijck:
This has been a long standing request.... I'm not sure what the value is really, so I'm really not married to this, but it does look more 'consistent', and separates "3rd-party" from "premake" code very clearly.


tvandijck:
Oops, forgot to update the bootstrap.mak file ;)

samsinsane:
:+1: LGTM

PR #439, opened 2016-02-29
Added support for schema files in C# projects

### LINKED ISSUE #416: Schema file *.xsd not set correctly in *.csproj
The schema files ending with 'xsd' will not set correctly in csproj xml file. These files should be embedded with the Tag DependUpon.
E.g.:
```
<Compile Include="Schema\FILE.cs">
    <DependentUpon>FILE.xsd</DependentUpon>
</Compile>
```

It perfectly works with *.resx and

samsinsane:
Fixes #416


starkos:
LGTM—thanks!

PR #440, opened 2016-03-01
Cleaned up vstudio linker elements

### LINKED ISSUE #150: linkoptions also being passed to librarian in vs2012
Is this intentional, because it seems like a bug?

For instance, I'm passing:

```
	linkoptions "/NODEFAULTLIB:libc.lib"
```

samsinsane:
Fixes #150

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -371,11 +371,17 @@
 --
 
 	m.elements.linker = function(cfg, explicit)
-		return {
-			m.link,
-			m.lib,
-			m.linkLibraryDependencies,
-		}
+		if cfg.kind == p.STATICLIB then
+			return {
+				m.lib,
+				m.linkLibraryDependencies,
+			}
+		else
+			return {
+				m.link,
+				m.linkLibraryDependencies,
+			}
+		end
 	end
 
 	function m.linker(cfg)
@@ -386,30 +392,22 @@
 
 
 	m.elements.link = function(cfg, explicit)
-		if cfg.kind == p.STATICLIB then
-			return {
-				m.subSystem,
-				m.generateDebugInformation,
-				m.optimizeReferences,
-			}
-		else
-			return {
-				m.subSystem,
-				m.generateDebugInformation,
-				m.optimizeReferences,
-				m.additionalDependencies,
-				m.additionalLibraryDirectories,
-				m.importLibrary,
-				m.entryPointSymbol,
-				m.generateMapFile,
-				m.moduleDefinitionFile,
-				m.treatLinkerWarningAsErrors,
-				m.ignoreDefaultLibraries,
-				m.largeAddressAware,
-				m.targetMachine,
-				m.additionalLinkOptions,
-			}
-		end
+		return {
+			m.subSystem,
+			m.generateDebugInformation,
+			m.optimizeReferences,
+			m.additionalDependencies,
+			m.additionalLibraryDirectories,
+			m.importLibrary,
+			m.entryPointSymbol,
+			m.generateMapFile,
+			m.moduleDefinitionFile,
+			m.treatLinkerWarningAsErrors,
+			m.ignoreDefaultLibraries,
+			m.largeAddressAware,
+			m.targetMachine,
+			m.additionalLinkOptions,
+		}
 	end
 
 	function m.link(cfg, explicit)
@@ -428,15 +426,13 @@
 
 
 	m.elements.lib = function(cfg, explicit)
-		if cfg.kind == p.STATICLIB then
-			return {
-				m.treatLinkerWarningAsErrors,
-				m.targetMachine,
-				m.additionalLinkOptions,
-			}
-		else
-			return {}
-		end
+		return {
+			m.subSystem,
+			m.optimizeReferences,
+			m.treatLinkerWarningAsErrors,
+			m.targetMachine,
+			m.additionalLinkOptions,
+		}
 	end
 
 	function m.lib(cfg, explicit)
--- a/tests/actions/vstudio/vc2010/test_link.lua
+++ b/tests/actions/vstudio/vc2010/test_link.lua
@@ -105,10 +105,9 @@
 		kind "StaticLib"
 		prepare()
 		test.capture [[
-<Link>
+<Lib>
 	<SubSystem>Windows</SubSystem>
-	<GenerateDebugInformation>false</GenerateDebugInformation>
-</Link>
+</Lib>
 		]]
 	end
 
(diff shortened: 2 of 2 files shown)


starkos:
I'm a little concerned that you had to modify existing unit tests to pass this…there's a good chance that I wrote those tests based on real-life projects. I'll try to do some testing.

samsinsane:
I tested this with VS2013 and VS2015 a while ago, based on what I can remember, the only time a StaticLib project contained the `Link` element was when it was changed to a StaticLib, additionally, it never seemed to read from or write to the `Link` element, as I mentioned in the issue.

Based on [this](https://blogs.msdn.microsoft.com/vcblog/2016/02/24/stuck-on-an-older-toolset-version-move-to-visual-studio-2015-without-upgrading-your-toolset/), it seems more likely that those tests were incorrect. It's possible that VS2015 will detect that it's a VS2010 project and read from and write to the `Link` element for certain situations, or that VS2010 reads and writes to both? Let me know what your tests find! I'd love to understand this a bit better :)

starkos:
It may be used by VSI add-ons (third-party or proprietary toolsets). If so I can submit a follow-up PR, or just rewrite the necessary code as Premake extension. So LGTM

starkos:
Finally getting a chance to look at this, and the `GenerateDebugInformation` element has disappeared from many of our projects. I think it is all static library projects, but out of time to investigate further today.

samsinsane:
Yeah, I wasn't able to find that element in the Librarian, does it exist in earlier versions of VS? Or does VS use it, even though there's no GUI for it? Sorry about that!

starkos:
I've [attached samples]((https://github.com/premake/premake-core/files/169897/VS-LinkLib-Samples.zip) generated manually (*not* with Premake) from VS 2010–2013; this is what VS outputs when you choose File > New Project… and create a Win32 static library project.

They all use `<Link>` elements  and contain the `GenerateDebugInformation` element. 

Can we please revert this PR and try to fix #150 in a more focused way?

samsinsane:
Interesting, does it use it? Additionally, does it use it if it's in the Lib element? If the PR needs to be reverted, please don't hesitate to revert it. I'm kind of hoping that the templates aren't quite right from MS, even though it seems extremely unlikely, haha.

starkos:
Unless there is a requirement to do otherwise, we try to match the default Visual Studio output exactly.

samsinsane:
That's fair enough, but in this particular case it doesn't look like VS even uses these elements? Should we be emitting elements that aren't used just to match the default output of an IDE? I feel taking this approach will lead to errors and inconsistencies. I'll use the `SubSystem` element as an example for this.

In VS2015, `SubSystem` has 9 possible values, Premake supports 2 currently. In the event that Premake supports the other 7 values, and that they have an impact on the static lib in anyway. Setting the SubSystem to `WindowsCE` on a static lib will have Premake (prior to this PR) emit the following:
```
<Link>
  <SubSystem>WindowsCE</SubSystem>
</Link>
<Lib>
  <SubSystem>WindowsCE</SubSystem>
</Lib>
```
This isn't matching the VS output, but to do so would require more code for a (seemingly) purely aesthetic output.

I'm not entirely sure what the correct solution would be, looking at your provided examples and a VS2015 project, I noticed the following:
- VS2010 emits a `Lib` element, warnings as errors.
- VS2012/VS2013 appear to be the same and are the same as VS2010 except for the `Lib` element.
- VS2015 seems the same as VS2012/VS2013, except it doesn't have the `GenerateDebugInformation` element.

tvandijck:
In my opinion, I rather have premake output a project that works, rather then trying to be 100% equal to the IDE. The IDE has in the past done things in odd ways, and besides that, encoding the different behaviors of each version of visual studio seem more work and for no other reason than to make things look the same.

starkos:
> Should we be emitting elements that aren't used just to match the default output of an IDE? 

Yes.

> I feel taking this approach will lead to errors and inconsistencies.

If we match the behavior of the tool, we get the same behavior as the tool. If we create different markup, we're more likely to get different behavior. This has been a development principle of Premake for a long time and it has worked well.

> to do so would require more code for a (seemingly) purely aesthetic output.

Whose aesthetic?

Matching the behavior of the tool is simple and objective. It is easy to verify, easy to debug, provides consistent and expected results, and completely avoides debates about aesthetics.

> I'm not entirely sure what the correct solution would be

Match the output of the tool. Then you don't have to wonder.

> [TomF] In my opinion, I rather have premake output a project that works, rather then trying to be 100% equal to the IDE.

Other than #150, which is obviously a Premake bug, what wasn't working? (serious question)

tvandijck:
> Other than #150, which is obviously a Premake bug, what wasn't working? (serious question)

I was making more of a general statement... I like things to work, rather then follow the IDE 100%. I have no known bugs on our end in this regard, so it was more of an opinion, rather then a "we can't revert this because it'll break things".

samsinsane:
I feel like I've inadvertently made this out to be a huge deal for me, it definitely isn't. I was mostly just trying to clarify what we should be doing, how we should achieve it, and in what circumstances we don't emit the default IDE output.

Personally, I see this as one of those times to not emit the default IDE output. It doesn't provide any functionality, changing the values doesn't appear to impact VS. Matching the output isn't as simple as the solution we have, as I pointed out above with the `SubSystem` example, so I guess what would you expect to happen there? And to a lesser extent, can you provide an extremely brief example of what you'd expect our solution to be like?

I'm picturing something like:
```
m.elements.link = function(...)
  if staticlib then
    return {
      m.subSystemAlwaysWindows,
      m.debugInfoAlwaysOn,
      ...
    }
  end
end
```

I'm happy to go down this path, just want to make sure I'm on the same page. Unfortunately, I don't think I'm quite there, sorry to drag this out a bit longer.

starkos:
> Personally, I see this as one of those times to not emit the default IDE output.

If you are referring to the lib vs. link element issue, I disagree.

I'm afraid I don't understand the issue with subsystem…what does VS emit, and why is that a problem? And is this something you are actually trying to implement, or just a hypothetical?

samsinsane:
It was a hypothetical situation, it is flawed because we don't have a concept of changing the subsystem. So, let's assume we can set the subsystem manually in Premake. Here's where the current solution falls apart.

VS2013 emits the following for a static lib, by default:
```
    <Link>
      <SubSystem>Windows</SubSystem>
      <GenerateDebugInformation>true</GenerateDebugInformation>
    </Link>
```
Going into Librarian->General and changing the SubSystem value to `WindowsCE` results in the following:
```
    <Link>
      <SubSystem>Windows</SubSystem>
      <GenerateDebugInformation>true</GenerateDebugInformation>
    </Link>
    <Lib>
      <SubSystem>WindowsCE</SubSystem>
    </Lib>
```
Note: SubSystem in Link is still Windows.

Here's how Premake would handle `subsystem "WindowsCE"`, if it had such an API:
```
    <Link>
      <SubSystem>WindowsCE</SubSystem>
      <GenerateDebugInformation>false</GenerateDebugInformation>
    </Link>
    <Lib>
      <SubSystem>WindowsCE</SubSystem>
    </Lib>
```
Note 1: SubSystem in Link is WindowsCE.
Note 2: GenerateDebugInformation in Link is false. Requires `flags { "Symbols" }` to emit true

Premake isn't matching the default output of VS. It looks like it would, but it actually changes values that would normally not change when using the IDE. I feel like what is currently happening isn't what you're saying should be happening? Is this expected?

starkos:
We handle a similar situation with `AdditionalOptions` with multiple functions: `additionalCompileOptions()`, `additionalLinkOptions()`, etc. So here it would make sense to split `subsystem()` up into `linkSubsystem()` and `libSubsystem()`.

Side note in case anyone decides to implement this: add a new value to `system()` instead of creating a whole new API: `system("WindowsCE")`.

To the original issue raised in #150 by @dgregorycryptic: we might want to have `linkOptions()` and `libOptions()`. But I'd like to see an actual use case for that first, as it might make more sense to fill in a missing API or two instead, or to simply filter those options in the project script.

samsinsane:
I'm not sure that splitting the linker into two paths is the solution here. What would be wrong with something like:
```
m.elements.linker = function(cfg, explicit)
  if cfg.kind == p.STATICLIB then
    return {
      m.staticLink,
      m.lib,
      m.linkLibraryDependencies,
    }
  else
    return {
      m.link,
      m.linkLibraryDependencies,
    }
  end
end

function m.staticLink(cfg)
  p.push('<Link>')
  m.element("SubSystem", nil, "Windows")
  ... -- this should contain the other elements and checks against _ACTION
  p.pop('</Link>')
end
```
The Link block in a static lib should never change, since VS doesn't allow the user to change it without changing the kind of project that it is. (I'm guessing this decision was made to allow devs to easily swap between generating Static and Shared libs?)

Is having this as a hardcoded output a bad thing? Doesn't it meet your requirements of matching the tool output, while not increasing the complexity of using and extending Premake?

In regards to #150, unless I'm misunderstanding what was said, @dgregorycryptic is saying that having Premake emit both Link and Lib blocks for a static lib feels wrong. Which is more or less what we've been discussing here. But I might've misunderstood the entire issue from the start?

starkos:
I'd rather see the code follow the conventions used everywhere else instead of trying to be clever with this one special case. And Visual Studio can be and is used to develop for platforms other than Windows.

> In regards to #150, unless I'm misunderstanding what was said, @dgregorycryptic is saying that having Premake emit both Link and Lib blocks for a static lib feels wrong.

Do what the tool does. Think of all the time you'll save.

PR #441, opened 2016-03-01
Fix a new Visual Studio unused variable build warning

starkos:
We have warnings-as-errors turned on here; need these fixed.


samsinsane:
:+1: LGTM

PR #445, opened 2016-03-12
Add NuGet submodule

aleksijuvani:
As discussed in issue #197. NuGet is the package manager included in recent versions of Visual&nbsp;Studio. This pull request will add a new function called `nuget` that can be used to specify these package dependencies for C++ and C# projects.

Syntax is as follows:

```lua
nuget { "package:version" } -- in project scope
```

Please review. I've tested this functionality with a personal project of mine, but it wouldn't hurt to have some more eyes on this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "modules/d"]
 	path = modules/d
 	url = https://github.com/premake/premake-dlang.git
+[submodule "modules/nuget"]
+	path = modules/nuget
+	url = https://github.com/aleksijuvani/premake-nuget.git
--- a/src/_modules.lua
+++ b/src/_modules.lua
@@ -9,4 +9,5 @@
 		"monodevelop",
 		"codelite",
 		"d",
+		"nuget",
 	}


samsinsane:
The module looks nice, but I'm not sure if it should be added to core. This is just more code that needs to be maintained, and there's already enough that needs to be maintained. I think your best bet would be to add it into the [list of 3rd party modules](https://github.com/premake/premake-core/wiki/Modules) for now.

@starkos what are your thoughts?

tritao:
Pretty much every .NET project nowadays needs Nuget packages so it would be nice if this could be integrated in the core distribution, at least until we have some better way to distribute / depend on 3rd party modules.

starkos:
NuGet is a feature of Visual Studio now, and since the VS exporter is still in core, it make sense that this would be as well. But we would need a few changes first…

If we are going to keep this code as a module, the repository needs to be transferred to the Premake team account, and permissions (admin access to Premake team + write access to original author) set accordingly.

BUT, as there isn't too much code involved, and since this is a feature of VS and not an add-on, it might make more sense to submit a PR that integrates the changes directly into the VS exporter, rather than bolting it on top.

Also…should we consider using a more generic API name instead of `nuget()`, as future-proofing against the addition of a toolset with a different package manager?

tritao:
I don't think so. If/when we add a more generic package management system to Premake, Nuget will still need to be supported somehow, I see them as orthogonal concepts.
In my projects I can see myself both using Premake and Nuget packages in the same project.

By the way, once this gets in, I'd like to figure out a way to support it under the Make backend for C# projects. I think the way to go is to get Premake to drive nuget.exe and get the needed packages, either at runtime or during project generation.

What do you think?

aleksijuvani:
> BUT, as there isn't too much code involved, and since this is a feature of VS and not an add-on, it might make more sense to submit a PR that integrates the changes directly into the VS exporter, rather than bolting it on top.

I agree. I could take a look if this is the way to go.

> Also…should we consider using a more generic API name instead of `nuget()`, as future-proofing against the addition of a toolset with a different package manager?

I'd say go with YAGNI and change it in the future if/when Premake supports an another package manager.

starkos:
> If/when we add a more generic package management system to Premake, Nuget will still need to be supported somehow, I see them as orthogonal concepts.

Sorry for the confusion: I wasn't referring to Premake packages but rather, for example, Eclipse packages.

>  I'd like to figure out a way to support it under the Make backend for C# projects. I think the way to go is to get Premake to drive nuget.exe and get the needed packages, either at runtime or during project generation.

Premake generated Projects/Makefiles need to be able to stand on their own. I'd suggest adding an `install` rule to the makefile that handles the call to NuGet directly.

starkos:
> I'd say go with YAGNI and change it in the future if/when Premake supports an another package manager.

I'm good with this.

samsinsane:
> I'd say go with YAGNI and change it in the future if/when Premake supports an another package manager.

I feel like you've missed the situation where a module developer, like yourself, might create a module for package management, like you did. Why don't we just create the system now? `getpackage { "manager:package:version" }` or `getpackage { "nuget:sdl2:2.0.3" }` seems like a better idea than having every module developer that wants to implement a package manager having to recreate the "get package" concept. It also means that modules that extend the `getpackage` concept, will be easier to merge in, there won't be a discussion, like this one, required to merge it in.

Personally, I don't see the place for YAGNI in premake-core, if that approach was taken from the start, we would never have `premake.override`. Which is a pretty core component to module development, but not to core development, we have the code and we can just change it.

aleksijuvani:
> I feel like you've missed the situation where a module developer, like yourself, might create a module for package management, like you did. Why don't we just create the system now? `getpackage { "manager:package:version" }` or `getpackage { "nuget:sdl2:2.0.3" }` seems like a better idea than having every module developer that wants to implement a package manager having to recreate the "get package" concept. It also means that modules that extend the `getpackage` concept, will be easier to merge in, there won't be a discussion, like this one, required to merge it in.

I get where you're coming from, but I disagree.

What if we need to support specifying custom package repositories for a package manager? Version ranges? I think it's impossible to try and create a generic interface for package dependencies if your only point of reference is a single package manager.

tritao:
I've thought a bit about this overnight and can definitely see some value in abstracting package management functionality in Premake.

My only concerns is that different package systems will have different requirements, so we either design a very simplified system like the syntax suggested by @samsinsane (and end up with a limited system with only the common aspects between them) or end up with a very complex API to support all the use cases that it would have been better going on with different API for each of them.

I seem to remember some libtool support that has been proposed in the past, and from what I recall that would not map very well to systems like NuGet or npm, so a unified syntax might not work too well. I can't seem to find that PR now, so my memory might be foggy..

Anyway, it's been months of waiting just to get this finally ready for merging, we (and I'm sure other people) have had to develop a lot of workarounds for supporting NuGet packages, getting this in core ASAP would be most welcome. Just this week I've had to develop some NuGet related workarounds in one of our projects.

Let's not hold it out for a unified interface that honestly I think we do not yet have a good idea how to design without knowing which kinds of package managers Premake will need to support in the future.

samsinsane:
> What if we need to support specifying custom package repositories for a package manager? Version ranges? I think it's impossible to try and create a generic interface for package dependencies if your only point of reference is a single package manager.

`getpackage { "manager:package:version:version" }`? I don't see why we couldn't have a variable number of arguments here. Maybe, instead of a delimited string, we do something similar to vpaths (might require work in the API system)?
```
getpackages {
  ["manager"] = {
    { package="package", version="version", repository="repository" }, -- custom repositories
    { package="package", version="start:end" } -- version ranges
  }
}
```
This might get a bit unwieldy, but if it's extensible and the different implementations can require additional __documented__ parameters, I don't see why this wouldn't work.

> Anyway, it's been months of waiting just to get this finally ready for merging, we (and I'm sure other people) have had to develop a lot of workarounds for supporting NuGet packages, getting this in core ASAP would be most welcome. Just this week I've had to develop some NuGet related workarounds in one of our projects.

I'm not quite sure what you're trying to say here, I don't see how having this module in premake-core changes anything? You can already utilize the module, so either your team has decided to not use any premake modules, which shouldn't impact the decisions made in this PR, or the module is broken, which makes me hesitant to see this code come into core in its current state.

-----------

I'm really concerned about taking the YAGNI approach with "pretty much every .NET project" requiring this feature. If people have been waiting months for this feature, surely we can spend a fraction of that time discussing a better solution that won't cause incompatibilities later?

Ultimately, I don't have that much experience with package managers, so I'm happy to defer to those with the experience. If it's not possible to create a generic solution that allows for different package managers because they're significantly different, that's totally fine, but if we can create a generic solution, we should.

tvandijck:
I'm somewhat with the other on the "maybe there is a more generic approach"...
That said... nuget is quite a system, and I wouldn't mind this as is, but maybe as Jason said make it part of the VS action itself, instead of a bolt on...  although admittedly, when the topic of nuget came up initially, everyone was "make it a module"... so I guess to reverse course on that is also somewhat unfair...

Either way... Just my two things on package management:
* version ranges, are NP complete. 
* package dependencies, and "getpackage/import" are and need to be two separate things.
** for example a package can't claim that it depends on A version 2.0, since that will make that package entirely unusable... the mathematical probability that a set of packages works together is 0 if versioned dependencies are specified in each package. (feel free to email me for a better explanation)
* version constraints are OK, but when using "semver" that is already explicitly encoded, so those do not have to be encoded using ranges. 

Nuget deals with versions very explicitly... the solution specifies the versions used, packages only mention that they depend on another package by name, and a minimum version. The dependency however has nothing to do with the "getpackage/import" api.

In my mind the getpackage/import API is something you do on the solution, and nowhere else..
So if that is the case all you really need is a way to 'register a package' with a common system, and the package 'manager' whatever that is can simple do whatever it wants to download that package onto the system.

and that can be:
* "nuget.import("package", version)"
* "git.clone("package", revision)"

as long as each of those calls correctly add the resulting information to a table that is shared in premake, all packages, wherever they come from can be depended on through common API.

starkos:
To clarify my previous comment, all I was really looking for was a more generic name for the API than "nuget". I think the current call itself is fine and can be easily extended if we need to grow.

If and when we need to support additional package systems (which may or may not ever happen, and who knows how Premake itself will have changed by then), we can implement it with real requirements, rather than trying to design an idealized system in the absence of real use cases.

> although admittedly, when the topic of nuget came up initially, everyone was "make it a module"... so I guess to reverse course on that is also somewhat unfair...

I'm trying to be consistent in what I'm telling people who want to contribute, and I think asking people to develop and share a module for entirely new features is reasonable. @aleksijuvani: If you don't want to prepare a PR to merge directly into the VS exporter, let me know and I'll do it for you.

aleksijuvani:
> I'm trying to be consistent in what I'm telling people who want to contribute, and I think asking people to develop and share a module for entirely new features is reasonable. @aleksijuvani: If you don't want to prepare a PR to merge directly into the VS exporter, let me know and I'll do it for you.

Oh no, it's fine. I can take care of it. I'll try to get it done this week.

aleksijuvani:
@starkos: I've prepared pull request #453 to merge directly into the VS exporter.

PR #450, opened 2016-03-17
Added targetplatformversion API

### LINKED ISSUE #315: Visual Studio - Target Platform Version
Hi,
Is there a way of setting the VS2015/Project Properties/General/Target Platform Version? Its always set to 8.1 on Windows 10

Many Thanks.

samsinsane:
Fixes #315

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -967,6 +967,12 @@
 	}
 
 	api.register {
+		name = "systemversion",
+		scope = "project",
+		kind = "string",
+	}
+
+	api.register {
 		name = "targetdir",
 		scope = "config",
 		kind = "path",
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -127,6 +127,7 @@
 			m.ignoreWarnDuplicateFilename,
 			m.keyword,
 			m.projectName,
+			m.targetPlatformVersion,
 		}
 	end
 
@@ -1921,6 +1922,14 @@
 
 	function m.targetName(cfg)
 		m.element("TargetName", nil, "%s%s", cfg.buildtarget.prefix, cfg.buildtarget.basename)
+	end
+
+
+	function m.targetPlatformVersion(prj)
+		local min = project.systemversion(prj)
+		if min ~= nil and _ACTION >= "vs2015" then
+			m.element("WindowsTargetPlatformVersion", nil, min)
+		end
 	end
 
 
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -506,3 +506,14 @@
 		return pairing
 	end
 
+
+--
+-- Given a project, returns requested min and max system versions.
+--
+
+	function project.systemversion(prj)
+		if prj.systemversion ~= nil then
+			local values = string.explode(prj.systemversion, ":", true)
+			return values[1], values[2]
+		end
+	end
--- a/tests/actions/vstudio/vc2010/test_globals.lua
+++ b/tests/actions/vstudio/vc2010/test_globals.lua
@@ -191,3 +191,64 @@
 </PropertyGroup>
 		]]
 	end
+
+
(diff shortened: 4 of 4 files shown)


tritao:
lgtm

starkos:
We currently use `system()` to choose the target OS, i.e. `system("Windows")`. Would it make sense to change this API to `systemversion()`?

OS X has a similar-ish concept where you can target min and max version ranges of core SDKs—Premake itself uses them via `linkoptions()` currently. It would be nice to unify the ideas if possible.

samsinsane:
`systemversion` makes sense, I'll try to push the change when I get a chance.

Unifying the ideas sounds great, but I don't really know how this would work. `systemversion("min:max")`? Could that be confusing when the action or system doesn't support ranges?

samsinsane:
@starkos updated to be `systemversion` and allows `systemversion("min:max")` for the OSX version system.

tvandijck:
LGTM

starkos:
Looks great, thanks! Can you throw together a doc page for the wiki to capture the "min:max" parameter format?

samsinsane:
[Done!](https://github.com/premake/premake-core/wiki/systemversion) Let me know if I need to add more information.

redorav:
I have a problem with this API. I've tried to use it as per the docs, but I'm getting a warning and the property doesn't get enabled. I have a simplest case that goes wrong.

```
workspace "Workspace"
	platforms { "A", "B" }
	location ("Workspace")
	
	systemversion("8.1")
	
	filter { "platforms:A" }
		system("windows")
		
	filter { "platforms:B" }
		system("windows")
		systemversion("10.0.15063.0")

project ("Project1")
	kind("WindowedApp")
	language("C++")
```

If I try to put that through premake alpha12, it complains that 

`** Warning: 'systemversion' on config 'Debug|B' differs from project 'Project1'; may be set out of scope`

Perhaps I'm doing something wrong? My expectation is that the filter in configuration B would apply to both Debug and Release but somehow it gets confused and only tries to apply it to one of them (in practice it doesn't apply it to either).

samsinsane:
@redorav Unfortunately, Visual Studio only supports one version per project and they don't make that very clear. It's only obvious when you attempt to set it in two different configurations and it doesn't work.

redorav:
I see. I tried to do it manually and it's just like you describe... what a shame. I wanted to be able to have a DX12+Win10 configuration without affecting other configurations which that actually compile in previous versions of Windows.

redorav:
@samsinsane I tried playing with this a little bit more. From the Visual Studio options menu you can't change this property per project because it goes into the "Globals", but it seems that it can be done by tweaking the .vcxproj file. Typically properties go in the "Configuration" label, and conditionally setting the **WindowsTargetPlatformVersion** fixes this issue and allows different configurations to targets different Windows SDKs. I tested this in a small example project that included d3d12.h, which compiles when targeting Windows 10 and doesn't when targeting Windows 8 (which is my use case)

I don't know how premake creates the vcxprojs but it looks like it should be relatively easy to add. Here is what I did in the vcxproj to make this work.

```xml
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
    <ConfigurationType>Application</ConfigurationType>
    <UseDebugLibraries>true</UseDebugLibraries>
    <PlatformToolset>v141</PlatformToolset>
    <CharacterSet>MultiByte</CharacterSet>
    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
</PropertyGroup>
```

samsinsane:
@redorav Nice, glad there's an alternative!

> Should I create a new bug for this? I kind of feel like it's well suited here since it goes hand in hand with the targetplatformversion API

Creating a new bug would be great! This is actually a merged pull request, so we can't really "reopen" it or anything.

redorav:
Ah. you're totally right, I've created https://github.com/premake/premake-core/issues/936 for this.

PR #452, opened 2016-03-17
Modified project link dependencies to be recursive. 

tvandijck:
Linking projects will now link the link dependencies of all their link dependencies. To keep this in check, static libs now ignore their link dependencies.

This should make it so libraries can specify their dependencies freely while the generated projects do the right thing and specify them at the linking level.

### DIFF
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -176,31 +176,54 @@
 			return result
 		end
 
-			local function add_to_project_list(cfg, depproj, result)
-				local dep = p.workspace.findproject(cfg.workspace, depproj)
-					if dep and not table.contains(result, dep) then
-						table.insert(result, dep)
-					end
+		local function add_to_project_list(cfg, depproj, result)
+			local dep = p.workspace.findproject(cfg.workspace, depproj)
+			if dep and not table.contains(result, dep) then
+				table.insert(result, dep)
 			end
+		end
+
+		local function getExpandedLinkDependencies(cfg, projectName, result)
+			local projectObj = premake.workspace.findproject(cfg.workspace, projectName)
+			if projectObj and projectObj.links then
+				for cfg in project.eachconfig(projectObj) do
+					for _, dependencyName in ipairs(cfg.links) do
+						getExpandedLinkDependencies(cfg, dependencyName, result)
+						table.insertkeyed(result, dependencyName)
+					end
+				end
+			end
+		end
 
 		local linkOnly = m == 'linkOnly'
 		local depsOnly = m == 'dependOnly'
 
 		result = {}
-			for cfg in project.eachconfig(prj) do
-			if not depsOnly then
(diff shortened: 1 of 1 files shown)


samsinsane:
LGTM, can you add some tests for this though?

starkos:
This touched almost every project we have. It removed dependencies which were explicitly set in the project script. It added dependencies that we don't want, with no (easy) way to remove them.

It also increased our project generation time by more than a factor of four.

I'm closing this one. I appreciate the contribution and understand the need, but there needs to be a way to explicitly control which dependencies are propagated—no "magic".

tvandijck:
I understand.. that said:

> It removed dependencies which were explicitly set in the project script.

but linking a library against a library makes no sense.. even more so, this ends up in the <link> section of the vcxproj, which is ignored by msbuild for static libs... so it basically removed dead weight... 

> It added dependencies that we don't want, with no (easy) way to remove them.

arguably, dependencies you don't want, shouldn't have been in the link chain, but instead should have properly used the "dependson" construct where build order was intended, using links to setup build order would be kind of 'dirty'.

> but there needs to be a way to explicitly control which dependencies are propagated—no "magic".

Our understanding was that that was exactly what "dependson" and "links" was for.. dependson is for build order dependencies... and links is for link dependencies. The change here is that links now propagate... as in, if a library links against another library, the resulting executable now only has to link against what it actually used, and "automatically" gets what that lib used..

Say for example I have:

"graphicsEngine", which uses DirectX.
and I have an application that uses "graphicsEngine", I now no longer need to specify "DirectX" as well for the application, and hence the lua file for the "application" is now platform independent, since all it has to do it "links { "graphicsEngine"}" and it's done... instead of:

"links { "graphicsEngine"} filter {windows} links {"DirectX"}"

It basically means the responsibility of correctly "exporting" what a library uses is with the library, instead of with the final application script, our script got half the size because of this.

I'm all for making this more explicit, any feedback towards the idea is very welcome, instead of just shutting it down.. obviously the project generation time is certainly a concern, I get it... but on our massive project, it literally had no measurable effect at all.

starkos:
> I'm all for making this more explicit, any feedback towards the idea is very welcome

The uses/usages approach that has been discussed earlier, where a library project can describe what settings are required to use it, has been the best approach I've seen proposed. I'm pretty sure you've been in on some of those discussions, hopefully you know what I'm talking about.

As a whole, that is a big feature to implement, but it could perhaps be started with only linkages?

> obviously the project generation time is certainly a concern, I get it... but on our massive project, it literally had no measurable effect at all.

Ours went from ~1.5 min. to over seven.

Sorry to be so terse with the feedback lately…I'm trying very hard to make progress on fixing the Xcode exporter, the new queries, and 5.0 in general, but all of what little time I have is being used up just trying to keep up with the PR changes.

PR #453, opened 2016-03-19
Add NuGet support to the Visual Studio exporter

aleksijuvani:
As previously discussed in pull request #445 and in issue #197.

How is this? Let me know if anything needs to be changed.


src/_premake_init.lua
```diff
@@ -1096,6 +1096,13 @@
 		kind = "boolean",
 	}
 
+	api.register {
+		name = "nuget",
+		scope = "project",
+		kind = "list:string",
+		tokens = true,
+	}
+
```
starkos: Small nitpick, but I'm trying to keep these API definitions alphabetized. I'd like to add some automated JavaDoc style documentation at some point, and this way the APIs will appear alphabetized in the generated docs.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -315,6 +315,45 @@
 			end
 		end)
 
+		if _ACTION >= "vs2010" and prj.nuget then
```
starkos: This whole block should be split out into its own function so that it can be overridden in modules. Try to follow the model used in the vcxproj code: a function that uses `premake.callArray()` to output the individual elements. This whole C# exporter needs to be reworked to follow that model at some point; for now let's at least stick to it for any new code.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -315,6 +315,45 @@
...
+				-- This is kind of shit because we will need to add new
+				-- versions of the .NET Framework here.
+
+				local frameworks = {}
+				if targetFrameworkVersion >= 11 then table.insert(frameworks, "net10") end
+				if targetFrameworkVersion >= 20 then table.insert(frameworks, "net11") end
+				if targetFrameworkVersion >= 30 then table.insert(frameworks, "net20") end
+				if targetFrameworkVersion >= 35 then table.insert(frameworks, "net30") end
+				if targetFrameworkVersion >= 40 then table.insert(frameworks, "net35") end
+				if targetFrameworkVersion >= 45 then table.insert(frameworks, "net40") end
```
starkos: This should be a lookup table  (maybe `vstudio.frameworkVersions = { ["11"] = "net10", ["20"] = "net11", ... }` so it can be modified or extended from modules.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -315,6 +315,45 @@
...
+				table.insert(frameworks, targetFramework)
+
+				for _, framework in pairs(frameworks) do
+					local assembly = vstudio.path(prj, p.filename(prj.solution, string.format("packages\\%s\\lib\\%s\\%s.dll", vstudio.nuget2010.packageName(package), framework, vstudio.nuget2010.packageId(package))))
+					_x(3, '<HintPath Condition="Exists(\'%s\')">%s</HintPath>', assembly, assembly)
+				end
+
+				_p(3, '<Private>True</Private>')
```
starkos: We're phasing out `_p()` and `_x()` in favor of `p.w()` and `p.x()`. If you can use the newer functions it will save us some rework later. If the new functions cause indentation issues then it is okay to stick with the old ones.

src/actions/vstudio/vs2010.lua
```diff
@@ -35,6 +35,28 @@
 
 
 
+--
```
starkos: I'd prefer not to have two `generateSolution()` functions to juggle. Can this logic go in the 2005 version, with an `if _ACTION >= "vs2010"` if necessary?

src/actions/vstudio/vs2005_csproj.lua
aleksijuvani: I had originally used `p.w()` and `p.x()`, but I changed them to `_p()` and `_x()` exactly because of indentation issues.

src/actions/vstudio/vs2010_nuget.lua
```diff
@@ -0,0 +1,88 @@
...
+					end
+				end
+			end
+		end
+	end
+
+	function nuget2010.packageFramework(wks, package)
+		local prj = packageProject(wks, package)
+
+		if prj.language == "C++" then
```
starkos: Use `project.iscpp()` and `project.isdotnet()` here instead of string literal compares.

src/actions/vstudio/vs2005_csproj.lua
starkos: Okay, that's fine then…we'll get them when we rework the surrounding code.

src/actions/vstudio/vs2005_csproj.lua
aleksijuvani: I think something like `vstudio.frameworkVersions = { "net10", "net11", "net20", "net30", "net35", "net40", "net45" }` would make more sense. A little more code would be needed in the function, but the table itself would be more generic and could be utilized elsewhere if needed.

aleksijuvani:
@starkos: Is this better?

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -314,7 +319,61 @@
 				_x(2,'<Reference Include="%s" />', name)
 			end
 		end)
+	end
+
+	function cs2005.nuGetReferences(prj)
+		if _ACTION >= "vs2010" and prj.nuget then
```
starkos: Minor nitpick for future reference: the baking system will ensure `prj.nuget` always has a value. It will just be an empty list `{}` if not set by the user's script.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -314,7 +319,61 @@
...
+								frameworkVersion,
+								vstudio.nuget2010.packageId(package)
+							)
+						)
+					)
+
+					_x(3, '<HintPath Condition="Exists(\'%s\')">%s</HintPath>', assembly, assembly)
+				end
+
+				for k, frameworkVersion in ipairs(vstudio.frameworkVersions) do
```
starkos: It would be good to pull this logic out into its own function, both to make it more clear what it is doing and also so we can reuse it elsewhere if needed. Maybe something like `vstudio.identifyFrameworkVersion(prj)`?

aleksijuvani: What we're doing here is building a list of .NET Framework versions up to and including the project's target version.

Maybe it would make more sense to move this logic into a function in `vstudio.cs2005` instead of in `vstudio`?

If we're going the generic route, should we also change `vstudio.frameworkVersions` to store the versions like `{ "1.0", "1.1", "2.0", "3.0", "3.5", "4.0", "4.5" }` instead of `{ "net10", "net11", "net20", "net30", "net35", "net40", "net45" }`?

This would be more consistent with the way the `dotnetframework` API works. We need to do some string transformations to convert between the two formats in either case, so it's purely a style choice.

starkos: I'm okay with that approach, sure.

aleksijuvani:
@starkos: Is this good?

samsinsane:
Can you squish those commits down into one? Most of your commits either fix errors or move code around, they don't really need to be separate commits.

starkos:
FWIW, I prefer several small commits to one big one—it makes it easier to tie a specific change to the intent. The [Files Changed](https://github.com/premake/premake-core/pull/453/files) tab does a good job of showing the big picture.

Not everyone agrees with me (hi @TurkeyMan), so I tend to let contributors follow the style that works for them, and try not to overdo it myself.

aleksijuvani:
I would prefer to keep this as several small commits for the reasons listed above, but either way is fine for me. Please let me know if anything else needs to be changed in order to get this merged.

samsinsane:
You both seem to be keen on keeping all those commits, I personally don't see value in most of them. I'll leave it here since I don't think it's worth arguing over, but FWIW, I probably side with @TurkeyMan on this issue.

TurkeyMan:
Haha, how come I'm being roped into this? :P

In the interest of sharing ideas, my personal sense of a good PR lies somewhere in the middle; I see that a PR should represent a single context, and then I like the series of patches in the PR to represent significant tasks or meaningful segments of that context (ie, moments to be considered from a communication/reviewing perspective).
I don't find this PR too bad, but I do agree there are a few more patches than necessary, for instance, `Alphabetize nuget API definition`, which just amends a patch prior in the same PR. In cases like that, I would amend the patch that introduced that code in the first place, rather than introducing incorrect code and then correcting it in a follow-up patch. A trivial rebase (a single squish in the right sequence) would have smoothed it over here. There are other similar cases present here too.

It's all about making your PR easier to review. People often review patch-by-patch from the bottom up, so if you correct for errors in future patches, it creates a state that reviewers need to keep in their heads, ie, they may encounter an issue already corrected by a follow-up they haven't gotten to yet; do they comment as soon as they encounter the problem, or remember the complete set of feedback for comment at the end? The former is more convenient for the reviewer, and the latter quickly becomes impractical in larger PR's.

Don't mistake that this was a request to squash the whole thing into one, I don't think that's what @samsinsane was suggesting. Keeping the number of patches that introduce new concepts is fine, but the patches that corrected prior patches could have been amendments instead.

Just my 2c :) .. something you might like to keep in mind. It's definitely subject to personal style. I understand the idea of correcting mistakes at the point of entry is foreign to people who haven't used git too much (I'm not suggesting this is you, but something I see a lot), but I think it's generally agreed to be good juju with many OSS projects on Github.

Ultimately, I feel the same as @starkos, people are responsible for their own PR presentation, and improving on that is a continuous personal learning experience. From my experience on other projects, quality PR presentation does strongly increase your patches chance of being merged.

TurkeyMan:
Oh yeah, and LGTM :P .. I'm excited for this feature, thanks!

samsinsane:
Thanks @TurkeyMan! You pretty much nailed what I was getting at, I just didn't really have the time to explain it like that.

@aleksijuvani thanks for putting the effort in to get this ready for premake-core. I'm always excited by new things in the VS exporter, hopefully I get to try this out when it's merged in.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -314,8 +330,44 @@
...
+								cs2005.formatNuGetFrameworkVersion(frameworkVersion),
+								vstudio.nuget2010.packageId(package)
+							)
+						)
+					)
+
+					_x(3, '<HintPath Condition="Exists(\'%s\')">%s</HintPath>', assembly, assembly)
+				end
+
+				_p(3, '<Private>True</Private>')
```
ClxS: One thing to note about this which we've had to fix on our internal version - this doesn't respect NoCopyLocal. On our version we replaced this line with:
if premake.config.isCopyLocal(prj, package, cfg) then
      _p(3, '<Private>True</Private>')
 else
       _p(3, '<Private>False</Private>')
end

PR #457, opened 2016-03-29
Fix getextension to handle paths w/ dot correctly

eyalsk:


src/base/path.lua
```diff
@@ -230,7 +231,7 @@
 			return p
 		end
 
-		if not newext:findlast(".", true) then
+		if newext ~= "" and not newext:findlast(".", true) then
```
tvandijck: if #newext > 0

would be a little faster then the string compare...
but I also think checking for 'nil' would be nice, since then you can write

path.replaceextension(s)

So I'd change this to:

if newext and #newext > 0 and not newext:findlast(".", true) then

tests/base/test_path.lua
```diff
@@ -413,6 +421,10 @@
 			test.isequal("/nunit/framework/main.foo", path.replaceextension("/nunit/framework/main",".foo"))
 	end
 
+    function suite.getabsolute_replaceExtensionWithEmptyString()
```
tvandijck: agreed, the getabsolute is a type, and must go here..

eyalsk:
It's nice but I don't know whether it makes _sense_ because when you replace something you generally expect to pass two arguments so maybe we can create a new function for it like clearextension? something like this:
```
function path.clearextension(p)
    return replaceextension(p, "")
end
```

starkos:
FWIW, I think `replaceextension(p, "")` is sufficient. Thanks for contribution!

tvandijck:
Yeah, I can go either way.... although I do think  
```
#newext > 0
```

is faster then 

```
newext ~= ""
```

but that is just nitpicking...

eyalsk:
@tvandijck you're right, I've changed it. :)

PR #461, opened 2016-03-31
Update require to resolve modules that are placed in default location

eyalsk:
Provides a strategy for module resolution to modules that are placed within the default module's directory that is `.premake/<module>` or `.premake/<module>.lua`

The search starts at the current working directory of the running premake script and walks up to the root.

### DIFF
--- a/src/base/globals.lua
+++ b/src/base/globals.lua
@@ -74,13 +74,24 @@
 ---
 
 	premake.override(_G, "require", function(base, modname, versions)
-		local result, mod = pcall(base,modname)
+        local result, resmod = pcall(premake.moduleresolver.resolveModule, modname)
+        
+        if result then
+            modname = resmod
+            resmod = ""
+        end
+        
+        local mod
+        result, mod, versions = pcall(base, modname, versions)
+        
 		if not result then
-			error( mod, 3 )
+			error(mod .. resmod, 3)
 		end
+        
 		if mod and versions and not premake.checkVersion(mod._VERSION, versions) then
 			error(string.format("module %s %s does not meet version criteria %s",
 				modname, mod._VERSION or "(none)", versions), 3)
 		end
+        
 		return mod
 	end)
--- a/src/base/moduleresolver.lua
+++ b/src/base/moduleresolver.lua
@@ -1,0 +1,90 @@
+--
+-- moduleresolver.lua
+-- 
+-- Provides a strategy for module resolution to modules that are placed within the default module's directory. 
+-- 
+-- Copyright (c) 2002-2014 Jason Perkins and the Premake project
+--
+
+    local p = premake
+
+    p.moduleresolver = {}
+
+    local moduleresolver = p.moduleresolver
+
+---
+-- Check whether the module is either a path to a module or a module's name.
+--
+
+    function moduleresolver.isModulePath(module)
+        return string.contains(module, "/") or string.contains(module, "\\")
+    end
+    
+--
+-- Get the name of the default module's directory.
+--
+
+    function moduleresolver.getDefaultModuleDirectory()
+        return ".premake"
+    end
+    
+---
+-- Get the path to the default module's directory.
+--
+
+    function moduleresolver.getDefaultModulePath(modulePath)
+        return modulePath and path.join(moduleresolver.getDefaultModuleDirectory(), path.appendExtension(modulePath, ".lua")) or modulePath
+    end
+    
+--
+-- Search the module recursively, 
+-- it starts at some unknown path and walks up to the root.
+--
+
+    local function searchModuleFile(searchPath, modulePath, trackedPaths)
+        trackedPaths = trackedPaths or {}
+        local fullpath = path.join(searchPath, modulePath)
+        table.insert(trackedPaths, fullpath)
+        if not os.isfile(fullpath) then
+            searchPath = path.getdirectory(searchPath) 
+            fullpath = searchPath ~= "." and searchModuleFile(searchPath, modulePath, trackedPaths) or nil
+        end
+        return fullpath, trackedPaths
+    end
+
+    moduleresolver.searchModuleFile = searchModuleFile
+    
+--
+-- Get the relative module's path.
+--
+
+    function moduleresolver.getRelativeModulePath(currentWorkingDir, modulePath)
+        return modulePath and path.getrelative(currentWorkingDir, modulePath) or modulePath
+    end
+    
+--
+-- The strategy for module resolution is as follow:
+--  1. When the module is a name then:
+--      1. It gets the default location that is .premake/<module> or .premake/<module>.lua
+--      2. It searches the module, starting at the current working directory and walks up to the root.
+--      3. It gets the relative path for the module.
+--      4. Finally, if there's an error it fail-fast.
+--  2. When the module is a path then it does nothing.
+--  3. Finally, returns with the module.
+--
+
+    function moduleresolver.resolveModule(module)
+        local originalModule = module
+        if not moduleresolver.isModulePath(module) then 
+            local trackedPaths
+            local currentWorkingDir = os.getcwd()
+            module = moduleresolver.getDefaultModulePath(module)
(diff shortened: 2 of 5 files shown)


tvandijck:
LGTM, I think this is an nice addition.

src/base/moduleresolver.lua
```diff
@@ -0,0 +1,90 @@
+--
+-- moduleresolver.lua
+-- 
+-- Provides a strategy for module resolution to modules that are placed within the default module's directory. 
+-- 
+-- Copyright (c) 2002-2014 Jason Perkins and the Premake project
```
starkos: Can you fix this to "2016", or remove the date?

src/base/globals.lua
```diff
@@ -74,13 +74,24 @@
 ---
 
 	premake.override(_G, "require", function(base, modname, versions)
-		local result, mod = pcall(base,modname)
+        local result, resmod = pcall(premake.moduleresolver.resolveModule, modname)
```
starkos: I can guess that "mod" is short for "module". What is "resmod"?

src/base/moduleresolver.lua
```diff
@@ -0,0 +1,90 @@
...
+
+    p.moduleresolver = {}
+
+    local moduleresolver = p.moduleresolver
+
+---
+-- Check whether the module is either a path to a module or a module's name.
+--
+
+    function moduleresolver.isModulePath(module)
```
starkos: Maybe we should have a `path.hasdir()` or something more general? I know we do a similar test elsewhere, and there is nothing module-specific about this?

src/base/moduleresolver.lua
```diff
@@ -0,0 +1,90 @@
...
+    function moduleresolver.getDefaultModuleDirectory()
+        return ".premake"
+    end
+    
+---
+-- Get the path to the default module's directory.
+--
+
+    function moduleresolver.getDefaultModulePath(modulePath)
+        return modulePath and path.join(moduleresolver.getDefaultModuleDirectory(), path.appendExtension(modulePath, ".lua")) or modulePath
```
starkos: This looks like it would hit on `.premake/module.lua` which is fine, but the convention we've set is `.premake/module/module.lua`.

src/base/globals.lua
eyalsk: resmod stands for _res_olved _mod_ule but maybe it's too clever, I'll update it with something more readable.

src/base/moduleresolver.lua
eyalsk: Sure, I can probably remove isModulePath and add a new general function for it to path.

src/base/moduleresolver.lua
eyalsk: I'll add it, would it be okay if I'll add it in addition to .premake/module.lua?

src/base/moduleresolver.lua
eyalsk: Sure! :)

starkos:
So first off, kudos, this is definitely a well-written contribution. Thanks for that!

In general, a little more explanation would be nice. If I'm reading the code right, you are scanning from the current working directory up looking for a folder called ".premake"?

I've made a couple of comments inline, but here are a few more thoughts:

* It is nice to have the location strategy broken out like this, but it would also be nice to have all of the module locating logic in one place (see `m.moduleLoader()` in `_premake_main.lua`). I'm not sure if this code should go there or that code should go here or something else.

* I'm a *little* uncomfortable about Premake just bopping its way up the file system looking for stuff. I'm not a fan of magic behaviors in general; I like to specify things explicitly. I'm trying to keep an open mind on this, but I would kind of prefer to specify my module search paths myself. What's your mental model for this (i.e. what are you picturing as your development setup that needs this feature?)

* If we do accept this module, I'm not sure that a hidden folder is the best choice for a default? Maybe it should be just "premake" or "modules" (we already use "modules" in the existing loader).

eyalsk:
@starkos thank you very much! 

> In general, a little more explanation would be nice. If I'm reading the code right, you are scanning from the current working directory up looking for a folder called ".premake"?

Yeah exactly, can you elaborate on the parts where more explanation is needed?

> It is nice to have the location strategy broken out like this, but it would also be nice to have all of the module locating logic in one place (see m.moduleLoader() in _premake_main.lua). I'm not sure if this code should go there or that code should go here or something else.

I'd look it up and write my opinion about it later.

> I'm a little uncomfortable about Premake just bopping its way up the file system looking for stuff. I'm not a fan of magic behaviors in general; I like to specify things explicitly. I'm trying to keep an open mind on this, but I would kind of prefer to specify my module search paths myself. What's your mental model for this (i.e. what are you picturing as your development setup that needs this feature?)

I kinda agree with you on the part where it goes all the way up to the root although I don't really think it's much of a big deal, I tend to like convention over configuration and module resolution needs to have some predefined behaviour that we agree on out of the box.

We can certainly have an option to provide some _stopper_ mechanism such as `premake.moduleBase` where the default would be `_MAIN_SCRIPT_DIR`.

> If we do accept this module, I'm not sure that a hidden folder is the best choice for a default? Maybe it should be just "premake" or "modules" (we already use "modules" in the existing loader).

Well, I generally prefix directories that 3rd-party tools use with with a period, it also makes these directories stays at the top and prevents conflicts between the application directories and tools directories  although in this case Premake is unique enough so it's probably a non-issue.

I really don't like `module` because it's too generic, if anything I'd go with premake or _premake as the default.

starkos:
> Yeah exactly, can you elaborate on the parts where more explanation is needed?

A summary of what the PR was attempting to do. We are all limited on time (you too I'm sure)—I'd rather spend it providing feedback, rather than trying to figure out what the heck it does.

>  I tend to like convention over configuration and in my opinion module resolution needs to have some predefined behaviour that we agree on out of the box.

We [already have that](https://github.com/premake/premake-core/wiki/Locating-Scripts), as [introduced here](https://github.com/premake/premake-core/wiki/Using-Modules). 

It would still be helpful to know what kind of project setup you are trying to achieve.

eyalsk:
@starkos 

> A summary of what the PR was attempting to do. We are all limited on time (you too I'm sure)—I'd rather spend it providing feedback, rather than trying to figure out what the heck it does.

You're right but I thought it's clear enough, I guess it wasn't so sorry.

> We already have that, as introduced here. 

Yeah, the issue is that these paths exist in a central place on the system as opposed to being local to the application, not all modules supposed to be shared.

I tend to put all my modules in a specific directory as part of the application itself and having to deal with relative paths or paths at all when all I want is to load a module it can be quite bad experiences.

So you do have an existing behaviour but it seems like you expect modules to be shared across applications.

> It would still be helpful to know what kind of project setup you are trying to achieve.

That's the structure of the project.
```
.
|-- .premake
|-- src
|   |-- launcher
|   |-- library
|   |   `-- utils
|   `-- platform
|       |-- abstract
|       `-- win32
`-- tests
    `-- platform
        `-- win32
```

starkos:
> Yeah, the issue is that these paths exist in central places on the system as opposed to being local to the application, not all modules supposed to be shared.

I put my per-project modules into a folder called "modules" at the same location as my top-level, solution/workspace script (`_MAIN_SCRIPT_DIR`).

```
.
|-- premake5.lua
|-- modules
|-- src
```

This location is on the existing search path and works out of the box. It sounds like you want to do the same thing, but don't like the name "modules". 

The ugly way to do this in your solution/project script would be:

```lua
premake.path = premake.path + ";" + path.join(_MAIN_SCRIPT_DIR, ".premake")
```

A way to add a new search location more cleanly would be a nice addition, maybe something like (totally off the top of my head):

```lua
requirepath(".premake")

function requirepath(p)
   premake.path = premake.path + ";" + path.join(os.getcwd(), p)
end
```

After that, you should be able to `require()` from any of your scripts and have it resolve to that path.

eyalsk:
@starkos Nice! so that's how you do it haha... thanks!

requirepath or configpath or something like this would be pretty great in my opinion.

PR #463, opened 2016-04-04
migrated "configuration" to "filter"

neico:
There's the requested PR.

Excuse me for also fixing quote characters (and the few spaces) for consistency, but my OCD ticked off when I opened those files... (not that previous versions of those files were any better...)

Edit: oh wow, travis actually didn't like the escape sequence?... (that's what happens when you rely on editor syntax highlighting...)

Edit2: yay, fixed, after 2 force pushes, poor travis...

### DIFF
--- a/contrib/lua/premake5.lua
+++ b/contrib/lua/premake5.lua
@@ -21,8 +21,8 @@ project "lua-lib"
 		"etc/*.c"
 	}
 
-	configuration "linux or bsd or hurd or aix"
+	filter "system:linux or bsd or hurd or aix"
 		defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
 
-	configuration "macosx"
+	filter "system:macosx"
 		defines     { "LUA_USE_MACOSX" }
--- a/premake5.lua
+++ b/premake5.lua
@@ -99,22 +99,22 @@
 			defines { "CURL_STATICLIB", "PREMAKE_CURL"}
 		end
 
-		configuration "Debug"
+		filter "configurations:Debug"
 			defines     "_DEBUG"
 			flags       { "Symbols" }
 
-		configuration "Release"
+		filter "configurations:Release"
 			defines     "NDEBUG"
 			optimize    "Full"
 			flags       { "NoBufferSecurityCheck", "NoRuntimeChecks" }
 
-		configuration "vs*"
+		filter "action:vs*"
 			defines     { "_CRT_SECURE_NO_DEPRECATE", "_CRT_SECURE_NO_WARNINGS", "_CRT_NONSTDC_NO_WARNINGS" }
 
-		configuration { "windows", "Release" }
+		filter { "system:windows", "configurations:Release" }
 			flags       { "NoIncrementalLink", "LinkTimeOptimization" }
 
-		configuration { "macosx", "gmake" }
+		filter { "system:macosx", "action:gmake" }
 			buildoptions { "-mmacosx-version-min=10.4" }
 			linkoptions  { "-mmacosx-version-min=10.4" }
 
@@ -146,44 +146,44 @@
 			"contrib/**.*"
 		}
 
-		configuration "Debug"
+		filter "configurations:Debug"
 			targetdir   "bin/debug"
 			debugargs   { "--scripts=%{prj.location}/%{path.getrelative(prj.location, prj.basedir)} test"}
 			debugdir    "%{path.getrelative(prj.location, prj.basedir)}"
 
-		configuration "Release"
+		filter "configurations:Release"
 			targetdir   "bin/release"
 
-		configuration "windows"
+		filter "system:windows"
 			links       { "ole32", "ws2_32" }
 
-		configuration "linux or bsd or hurd"
+		filter "system:linux or bsd or hurd"
 			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
 			links       { "m" }
 			linkoptions { "-rdynamic" }
 
-		configuration "linux or hurd"
+		filter "system:linux or hurd"
 			links       { "dl", "rt" }
 
-		configuration "linux"
+		filter "system:linux"
 			if not _OPTIONS["no-curl"] and os.findlib("ssl") then
 				links       { "ssl", "crypto" }
 			end
 
-		configuration "macosx"
+		filter "system:macosx"
 			defines     { "LUA_USE_MACOSX" }
 			links       { "CoreServices.framework" }
 			if not _OPTIONS["no-curl"] then
 				links   { "Security.framework" }
 			end
 
-		configuration { "macosx", "gmake" }
+		filter { "system:macosx", "action:gmake" }
 			toolset "clang"
 
-		configuration { "solaris" }
+		filter { "system:solaris" }
 			linkoptions { "-Wl,--export-dynamic" }
 
-		configuration "aix"
+		filter "system:aix"
 			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
 			links       { "m" }
 
--- a/contrib/curl/premake5.lua
+++ b/contrib/curl/premake5.lua
@@ -1,43 +1,43 @@
-project 'curl-lib'
-	language    'C'
-	kind        'StaticLib'
-	includedirs { 'include', 'lib' }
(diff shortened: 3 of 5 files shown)


tvandijck:
LGTM

contrib/lua/premake5.lua
```diff
@@ -21,8 +21,8 @@ project "lua-lib"
 		"etc/*.c"
 	}
 
-	configuration "linux or bsd or hurd or aix"
+	filter "system:linux or system:bsd or system:hurd or system:aix"
```
samsinsane: Specifying the prefix on each item isn't required, `filter "system:linux or bsd or hurd or aix"` should work just fine. Personally, I prefer the less verbose way, but that is just a personal preference.

neico: Oh good to know, I don't like that line that much either, it's shown on the wiki but not actually "documented" as a feature...

samsinsane: I found [this](https://github.com/premake/premake-core/wiki/filter) entry to document that feature pretty well, should there be something more explicit, or obvious?

While I agree that a smaller list makes more sense to manage, those code blocks are off by default in Lua and we should follow that behaviour. Additionally, I'm going to assume that means the code that is compiled without those defines will work on all platforms, including Linux, BSD, Hurd and AIX. That's probably a terrible assumption to make though.

neico: Thtat's why I didn't do it.

And yeah the examples only show kind: being used like you described, but it's not explicitly mentioned like the other stuff that's happening, unless  

> If any keyword pattern fails to match the current context, the entire filter is skipped over.

is supposed to be that hint...

Sometimes I miss a talk page for each wiki page like with mediawiki...

PS: seems like my intuitive "not" thinking is wrong, do you really need to put "not" between the scope and the keyword? (after looking at the code, you apparently need to, still I think that you shouldn't separate the keyword from it's scope trough a not, but that's just my opinion...)

Actually I just looked the filter code up and noticed that [if there's no scope then scope "configurations" is assumed](https://github.com/premake/premake-core/blob/master/src/base/criteria.lua#L66), should I reflect this in the PR, unless this detail is discouraged like the comments mention.

Also seems like "iif" isn't documented in the wiki (it's not even in the "lua additions" list), did it go missing in the transition between bitbucket and github?

starkos: > do you really need to put "not" between the scope and the keyword

Currently yes. I'm not opposed to supporting `"not system:windows"` in addition to the current syntax though.

> and noticed that if there's no scope then scope "configurations" is assumed, should I reflect this in the PR

No, that's a convenience for simple projects. Your current approach is better in this case.

> seems like "iif" isn't documented in the wiki

Quite possible—feel free to add it!

premake5.lua
```diff
@@ -99,22 +99,22 @@
...
 			defines     "NDEBUG"
 			optimize    "Full"
 			flags       { "NoBufferSecurityCheck", "NoRuntimeChecks" }
 
-		configuration "vs*"
+		filter "action:vs*"
 			defines     { "_CRT_SECURE_NO_DEPRECATE", "_CRT_SECURE_NO_WARNINGS", "_CRT_NONSTDC_NO_WARNINGS" }
 
-		configuration { "windows", "Release" }
+		filter { "system:windows", "action:Release" }
```
neico: Whoops? Should be "configurations:" instead of "action:" I guess...

starkos: Ha, nice catch.

contrib/lua/premake5.lua
neico: > Currently yes. I'm not opposed to supporting "not system:windows" in addition to the current syntax though.
I wonder though, shouldn't [this](https://github.com/premake/premake-core/blob/master/tests/base/test_criteria.lua#L69) fail in that case then? Either I read the code for "not" wrong or the test isn't correct...

> Quite possible—feel free to add it!
done, also added printf and verbosef, someone might need to fill those pages with updated info which I lack (like removing the old doc styles and specifying when they were added)

starkos: > I wonder though, shouldn't this fail in that case then?

Oh hey, look at that, I guess I already wrote it. :)

neico: Why not use it then? I mean it makes more sense imo.

starkos: Go for it.

TurkeyMan: > > do you really need to put "not" between the scope and the keyword

> Currently yes. I'm not opposed to supporting "not system:windows" in addition to the current syntax though.

Sorry I'm barely keeping up lately, but I spotted this one while skimming... I'm not sure I can get behind this.
It's not clear to me what `"not system:windows"` means, whereas I find `"system: not windows"` to be quite obvious.
I fear that supporting both possibilities will lead to confusion, and the latter is much more intuitive to my mind.
What's the advantage of supporting the former expression?

neico:
Here's an overview of usages of "not" and "configuration" lines that are left (most are in premake4.lua which is fine, but there's a few on the modules that need adjusting as well):
Notepad++ Regex

I was playing with the thought of unifying the single argument lines ex. `configuration { "solaris" }` to `configuration "solaris"` but then thought, if I were to do that I might as well do it in a separate PR (which I probably won't because that's probably taking it too far...)

contrib/lua/premake5.lua
samsinsane: I tend to agree with @TurkeyMan here, what would `"not configurations:windows* or windows or architectures:x64"` even do? Does the `not` apply to the whole string, the whole scope, or the first term?

samsinsane:
I feel like having `configuration "solaris"` instead of `configuration { "solaris" }` would be inconsistent with other uses of `configuration`. I'm not a fan of things like that, but if I'm the only one, then I'm more than happy to let it go as we do support doing it.

contrib/lua/premake5.lua
neico: Since filters are seperated by " or " it only applies to the first term, maybe being able to use ( and ) inside could help categorize things in a better way (same as it does in actual lua code)

samsinsane: I find the fact that it only applies to the first term to be a bug.

I honestly don't know if adding support for parenthesis is a good idea, it increases the complexity of the code that needs to be maintained. Also, it's just waiting for someone to request `and` support, which will make having a list of strings with an implicit `and` kind of redundant. I feel like we should just use a function at that point, but that's probably just going to make it harder to use filter.

TurkeyMan: It's not obvious to me how a 'term' is defined. Is a scope a term? Am I applying `not` to the whole expression beneath the scope, or the first item listed after the scope, or the scope itself in some way?

What I asked though, was what advantage does supporting that syntax offer? I feel it can only lead to confusion, can you make a case for that expression to exist?

TurkeyMan: Sorry, I just realised what you meant by the sequence of `or`'s. `not x or y or z`, I see `not` doesn't gel with `or` very well. The feeling I get from my comment above still stands though, that `not` to the left of the scope doesn't make sense to me.

Working `and` into those expressions doesn't seem too unreasonable, but that inevitably leads to parentheses... that starts to seem like too much, but I have always found it a bit strange that `or` is stated within the string, and `and` is expressed by comma separation outside the string.

I think there's probably way to much water under the bridge to change anything now,  but I do wonder if a more regular or intuitive expression might be possible.

TurkeyMan: Keep  in mind, this is probably the single most thoroughly used mechanism in premake, so I don't think it hurts to absolutely nail this.

neico: I'm still impressed that a single line (and suggestion) started quite a discussion worth of a separate "issue"...

And yeah for the longest time I didn't even know that "and" is actually possible because I only ever saw "or" but never "and" (I mean this syntax was also present with "configuration" so it's been there for quite a while)

TurkeyMan: It's like I say above, I think this is one of premake's central core mechanic's, so it's definitely worthy of special attention. We could spin off a separate thread if you like. I'd ask that, at least in the mean time, you don't write code like `not scope:thing`, and we can decide what sort of expressions we want to support on this elsewhere?

starkos: Okay everyone, we're getting off topic here—this PR is only to bring `premake5.lua` up to date. Let's get it merged with `filter "system:linux or bsd or hurd or aix"`, and open a new issue or discussion for the criteria formatting.

samsinsane:
Thanks!

PR #465, opened 2016-04-06
add 'nested' kind for premake.api.register

tvandijck:
This allows you to do this:

```lua
premake.api.register {
    name     = "schemacompiler",
    scope    = "config",
    kind     = "nested",
    fields   = {
        template = {
            kind     = "path",
            tokens   = true,
            pathVars = true
        },
        inputs = {
            kind     = "list:path",
            tokens   = true,
            pathVars = true
        },
        outputs = {
            kind     = "list:path",
            tokens   = true,
            pathVars = true
        }
    }
}
```

### DIFF
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -1092,6 +1092,31 @@
 	})
 
 
+--
+-- nested data kind; wraps simple values into a table, returns others as-is.
+--
+	premake.field.kind("nested", {
+		store = function(field, current, value, processor)
+			-- create meta-table.
+			if not current then
+				current = p.nested.create(field)
+			end
+
+			-- copy values.
+			for name, v in pairs(value) do
+				current[name] = v
+			end
+
+			-- return result.
+			return current
+		end,
+		compare = function(field, a, b, processor)
+			-- TODO: is there a reliable way to check this?
+			return true
+		end
+	})
+
+
 
 ---
 -- Start a new block of configuration settings, using the old, "open"
--- a/src/_manifest.lua
+++ b/src/_manifest.lua
@@ -44,6 +44,7 @@
 		"base/config.lua",
 		"base/fileconfig.lua",
 		"base/rule.lua",
+		"base/nested.lua",
 
 		-- project script processing
 		"base/oven.lua",
--- a/src/base/field.lua
+++ b/src/base/field.lua
@@ -51,37 +51,41 @@
 ---
 
 	function field.new(f)
-		-- Translate the old approaches to data kind definitions to the new
-		-- one used here. These should probably be deprecated eventually.
+		local function upgradeKind(f)
+			-- Translate the old approaches to data kind definitions to the new
+			-- one used here. These should probably be deprecated eventually.
 
-		if f.kind:startswith("key-") then
-			f.kind = f.kind:sub(5)
-			f.keyed = true
+			if f.kind:startswith("key-") then
+				f.kind = f.kind:sub(5)
+				f.keyed = true
+			end
+
+			if f.kind:endswith("-list") then
+				f.kind = f.kind:sub(1, -6)
+				f.list = true
+			end
+
+			local kind = f.kind
+
+			if kind == "object" or kind == "array" then
+				kind = "table"
+			end
+
+			if f.list then
+				kind = "list:" .. kind
+			end
+
+			if f.keyed then
+				kind = "keyed:" .. kind
+			end
+
+			-- Store the translated kind with a new name, so legacy add-on code
+			-- can continue to work with the old value.
+
+			f._kind = kind
 		end
 
-		if f.kind:endswith("-list") then
-			f.kind = f.kind:sub(1, -6)
-			f.list = true
-		end
-
-		local kind = f.kind
-
-		if kind == "object" or kind == "array" then
-			kind = "table"
-		end
-
-		if f.list then
-			kind = "list:" .. kind
-		end
-
-		if f.keyed then
-			kind = "keyed:" .. kind
-		end
-
-		-- Store the translated kind with a new name, so legacy add-on code
-		-- can continue to work with the old value.
-
-		f._kind = kind
+		upgradeKind(f)
 
 		-- Make sure scope is always an array; don't overwrite old value
 		if type(f.scope) == "table" then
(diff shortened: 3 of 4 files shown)


starkos:
I've wanted something like for a while; we really need to rework the custom build commands to do something similar. I've never liked how it looks in the script though:

```lua
schemacompiler {
   path = "some/path/value",
   inputs = { "input1", "input2" }
}
```

In particular, I don't like have to switch from `field("value")` to `field = "value"`. But I've never been able to come up with a better syntax that can reliably group the nested fields together into one object.

Any chance of getting some tests in `tests/api`? Also, does `kind = "list:nested"` work? We'd want that for the custom build commands (where we really need a way to group together inputs, outputs, and commands to allow for multiple build commands within a configuration).

tvandijck:
> In particular, I don't like have to switch from field("value") to field = "value". But I've never been 

I see what you mean, yes... that is unfortunately something you can't get away from, since that is just the limitation of lua syntax. The thing I don't really like is that the content is a hash table, and so you get things in an order that lua sees fit.. between execution those properties are pretty much in a random order, and never consistent...

> Also, does kind = "list:nested" work?

I think it does, but I haven't tested it at all... I'm trying to get detoken in 'nested' items to actually work first.. Right now it is not detokenizing any of the fields... The path stuff works, so I get correct absolute paths etc...

samsinsane:
Very interesting. Could this be used for something like:
```
libraries {
  ["SDL2"] = {
    includedirs = { "3rdParty/SDL2/includes" }
    libdirs = { "3rdParty/SDL2/bin" }
    libs = { "SDL2.lib" }
  },
  ["WindowsSockets"] = {
    libs = { "ws2_32.lib" }
  }
}
```

Would that be `kind = "list:keyed:nested"`? Which could also be used for that dependencies API people want:
```
libraries {
  ["MyLibrary"] = {
    includedirs = { "all", "the", "dirs" },
    libs = { "MyLibrary" }
  }
}
```
And then you could have `dependencies { "MyLibrary", "SDL2", "WindowsSockets" }`. Sorry that got a little off topic, I got excited by the idea of reducing the size of my Premake scripts haha.

starkos:
This raises a good point: what happens if one of the nest field names is the same as a non-nested field name? It will probably be necessary to mangle the nested field names like "schemacompiler.template".

@samsinsane examples have me wondering if we could introduce another kind of…well, I guess it can't be containers, not for the current implementation anyway, but something like:

```lua
library "SDL"
    includedirs { "3rdParty/SDL2/includes" }
    libdirs { "3rdParty/SDL2/bin" }
    links { "SDL2" }

library "WindowsSockets"
   links { "ws2_32" }
```

This would probably require reworking the way that values are fetched from the configuration sets which…oh hey, look at that…happens to be what I'm currently working on. I'm not sure how it work with Tom's original example though. Maybe just…

```lua
schemacompiler  -- hmm, does Lua allow this? Is really schemacompiler()
   path "some/path/value"
   input { "input1", "input2" }
```

tvandijck:
> This raises a good point: what happens if one of the nest field names is the same as a non-nested field name? 

Actually, the nested fields are not exposed as API calls... only the top level becomes a "method" 
So nested fields will not collide with top level API's.

> -- hmm, does Lua allow this? Is really schemacompiler()
nope, that is not lua... you'll get a compiler error.

> @samsinsane examples have me wondering if we could introduce another kind of…well, I guess it can't be containers,

I experimented with custom containers once, but that was really not very successful..
That said though, I really wish we could make these containers, and hence rework the container stuff to support this... I'd like to be able to introduce arbitrary containers, but also have API's allow name collisions if they are not targeting the same container...

For example, in sams example... the "links" API inside the "library" container, really has a very different meaning, and maybe I would even want it to be a different kind... So I'd like to be able to do this:

maybe not the best example, but you get my drift I hope.
```lua
premake.api.register {
    name     = "links",
    scope    = "library",
    kind     = "list:string"
}

premake.api.register {    
    name     = "links",
    scope    = "project",
    kind     = "list:path"
}
```

Basically, if the scope is different, it should not lead to a collision.
I think it would even be better if "containers" had some form of polymorphism to it, and hence a project and workspace "inherit" from a 'config' container, then 90% of the API's that currently have multiple scopes assigned, could simply use the 'config' base...

in which case the API might become more like

```lua
container.register("config",  {    
    name     = "links",
    kind     = "list:path"
})
```

anyway, I think this is maybe more a discussion for the google groups thing ;)

tvandijck:
> oh hey, look at that…happens to be what I'm currently working on.

Any details on that?? can we see it, or can we provide any early feedback?

starkos:
I should be working on something else but I'd rather be working on this.

> Actually, the nested fields are not exposed as API calls

I'll add a line comment where I think there might be issue.

> anyway, I think this is maybe more a discussion for the google groups thing

Yes, maybe…I'm not really sure what to do with this PR though. Can you hold off a bit until I've had a chance to think about how a more flexible scoping system might work? I'd hate to introduce it only to deprecate it if we can come up with a more consistent syntax.

> Any details on that?? can we see it, or can we provide any early feedback?

I was trying to sketch out a prototype module, but then I needed a way to run tests for it and so I got sidetracked pulling the test framework out into a module itself. Hopefully soon (in a relative, open source project sense).

Happy to discuss it more, but perhaps over on the forums as you suggested.

src/base/api.lua
```diff
@@ -1092,6 +1092,31 @@
...
+--
+-- nested data kind; wraps simple values into a table, returns others as-is.
+--
+
+	premake.field.kind("nested", {
+		store = function(field, current, value, processor)
+			-- set each value.
+			local result = {}
+			for name, v in pairs(value) do
+				local f = field.fields[name]
```
starkos: I could be wrong, but I think this might cause a problem if the global API has a field with the same name, or even a different nested field. I think the name might need to be decorated like `schemacompiler.template` to avoid collisions.

tvandijck: No... field.fields is the table inside the field, it's not touching any global tables.

tvandijck:
> Can you hold off a bit until I've had a chance to think about how a more flexible scoping system might work?

Sure thing... certainly if this discussion leads to something even better ;)

PR #466, opened 2016-04-06
table.deepclone should clone the metatable too.

tvandijck:


samsinsane:
LGTM, was this explicitly left out or just an oversight along the way?

starkos:
LGTM, go for it.

tvandijck:
> just an oversight along the way?

probably that... the example in the lua documentation actually has it too, and I just ran into an issue with the detokenizer on the 'nested' type loosing the meta-table in the configset._fetchDirect

PR #468, opened 2016-04-07
table fixes

tvandijck:
allow specifying the max depth at with the table.tostring needs to recurse..


starkos:
Can you do me a favor going forward and either give a more detailed description of the change in the PR, or add an entry to CHANGES.txt? Thanks!

tvandijck:
sure thing, I'm sorry... I'm moving really quickly today on some premake stuff I'm experimenting with, so hence the waterfall of small PR's with bugs I ran into...

PR #471, opened 2016-04-11
optimize copyFilters/mergeFilters

tvandijck:
As discussed on the forum...

This is the change where it became a deepcopy:
https://github.com/premake/premake-core/commit/28cfa55886ba3534bee9f02acb68770e599f5221#diff-3d9540042f8ef52dc7d1387a1ecbf71bL89

table.arraycopy however does an indexed copy of the table, which is indeed not what we want, since this is a 'keyed' table.. so instead of looping with 'ipairs', we definitely need to loop using 'pairs', but I don't think a deep copy is required.

Our projects seem to function correctly with this change.

### DIFF
--- a/src/base/context.lua
+++ b/src/base/context.lua
@@ -94,7 +94,10 @@
 --
 
 	function context.copyFilters(ctx, src)
-		ctx.terms = table.deepcopy(src.terms)
+		ctx.terms = {}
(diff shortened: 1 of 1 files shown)


samsinsane:
LGTM, I'll leave this for @starkos though as his upcoming changes to filters might rely on this?

PR #474, opened 2016-04-13
Fix optimize "Debug" for Clang

aleksijuvani:
See [issue #252 on Bitbucket](https://bitbucket.org/premake/premake-dev/issues/252/optimize-debug-generates-invalid-option).

When using the "Debug" option for `optimize`, the flag `-Og` is generated for Clang. This is not a valid option (it is valid and appropriate for GCC). When trying to compile, Clang will throw the following error message:

```
error: integral value 'g' in '-Og'
```

Since Clang doesn't have an equivalent for `-Og`, this pull request will change Clang to use `-O0` when `optimize "Debug"` is specified.

This change will [break a test in the CodeLite module](https://github.com/premake/premake-codelite/blob/master/tests/test_codelite_config.lua#L47), which will need to be changed to check for `-O0` instead of `-Og`. Should I also create a pull request there to be merged simultaneously with this one?

Let me know if anything needs to be changed, or if this isn't the way to go about doing this.

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -41,11 +41,36 @@
 --    An array of C compiler flags.
 --
 
+	clang.cflags = {
+		architecture = gcc.cflags.architecture,
+		flags = gcc.cflags.flags,
+		floatingpoint = gcc.cflags.floatingpoint,
+		strictaliasing = gcc.cflags.strictaliasing,
+		optimize = {
+			Off = "-O0",
+			On = "-O2",
+			Debug = "-O0",
+			Full = "-O3",
+			Size = "-Os",
+			Speed = "-O3",
+		},
+		pic = gcc.cflags.pic,
+		vectorextensions = gcc.cflags.vectorextensions,
+		warnings = gcc.cflags.warnings
+	}
+
 	function clang.getcflags(cfg)
 
-		-- Just pass through to GCC for now
-		local flags = gcc.getcflags(cfg)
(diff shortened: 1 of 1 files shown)


starkos:
Yes, I'd submit a PR against CodeLite if you can, that would be great. Mention this PR when you do. We can merge the CodeLite PR first, then update the module version here at the same time we approve this one.

aleksijuvani:
@starkos: I've opened premake/premake-codelite#11 to fix the unit tests.

TurkeyMan:
@aleksijuvani note, I reverted your PR that I merged because it caused unittests to fail. Premake generates CodeLite projects and unittests assuming GCC is the compiler. CodeLite has no good way to distinguish toolsets in the project files as I can see; configuring toolchain is an environment option.

TurkeyMan:
@aleksijuvani Tweak this PR to also update the submodule to the revision of your branch (which I just rebased above master), then it can be merged.

aleksijuvani:
Sure, give me a second.

TurkeyMan:
Note, the branch is called `clang_optimize`

TurkeyMan:
looks like you weren't on latest premake-core when you did that (Sam just merged my other submodule update a few minutes ago), get latest and try again :)

aleksijuvani:
Yeah, I just noticed. Give me a sec and I'll rebase.

TurkeyMan:
Right, got there. LGTM! :)

PR #475, opened 2016-04-13
Only output NuGet project element if packages were listed by the project

starkos:
This makes the project output match Visual Studio's default (for VS 2010-2013); if no NuGet projects have been added, the NuGet Target element is skipped.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1434,16 +1434,18 @@
 
 
 	function m.ensureNuGetPackageBuildImports(prj)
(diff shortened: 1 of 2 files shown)


samsinsane:
Out of curiosity, is VS2015 different or do you only have VS2010-2013?

starkos:
I only tested through 2013—VS 2015 borked itself doing an update and I've been avoiding fixing it. I'll try to get to it. Thanks for merging!

samsinsane:
No worries. That's fair enough, I've also experienced issues with the VS2015 installer. Thankfully, they're working on a new installer though!

PR #482, opened 2016-04-22
Fix default entry point for Windows console applications

okready:
By default, projects created for console applications through Visual Studio don't explicitly specify the entry point symbol; the linker automatically detects whether `mainCRTStartup` or `wmainCRTStartup` should be used based on whether `main()` or the Windows-specific `wmain()` is defined. The current behavior of Premake is to explicitly use `mainCRTStartup` in both windowed and console applications if the `WinMain` flag is not set, preventing the use of `wmain()` in console applications unless the entry point is explicitly set in the project configuration.

This change makes it so Premake does not explicitly specify the entry point symbol for Windows console applications, allowing `wmain()` to be used without the need for additional configuration settings.

This doesn't address the use of `mainCRTStartup` or `wmainCRTStartup` for windowed applications if the `WinMain` flag is not set, as that would likely break existing projects (explicitly specifying one or the other forces the use of either `main()` or `wmain()`, respectively).

### DIFF
--- a/src/actions/vstudio/vs200x_vcproj.lua
+++ b/src/actions/vstudio/vs200x_vcproj.lua
@@ -1121,7 +1121,7 @@
 
 
 	function m.entryPointSymbol(cfg, toolset)
-		if (cfg.kind == "ConsoleApp" or cfg.kind == "WindowedApp") and
+		if cfg.kind == "WindowedApp" and
 			not cfg.flags.WinMain and
 			not toolset
 		then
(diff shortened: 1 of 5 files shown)


samsinsane:
Should these code paths just be replaced by the `entrypoint` API? I'm not sure that Premake should continue to dictate the entry point if the `entrypoint` API exists.

okready:
I agree. I'd imagine most developers would understand if `WinMain` was the default start function for windowed applications on Windows, with the use of `main` being an exception (and vice-versa for console applications), since Visual Studio works like that by default anyway.

tvandijck:
Well, if the entrypoint API is set, we already make it override everything... the behavior based on the flag is still there for the cases where the entrypoint API is not set...

I guess the question is, should we just remove that behavior, and **only** set this property if the entrypoint API is set? if that is the case, this PR probably needs to be reworked to just do that.

tvandijck:
closing this... as this entire API has been removed.

PR #487, opened 2016-05-02
Gmake fixes

tvandijck:
We ran into a couple of gmake backend issues with defines and precompiled header files.

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -481,14 +481,24 @@
 
 		local pch = cfg.pchheader
 		local found = false
-		for _, incdir in ipairs(cfg.includedirs) do
-			local testname = path.join(incdir, pch)
-			if os.isfile(testname) then
-				pch = project.getrelative(cfg.project, testname)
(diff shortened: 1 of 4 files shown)


src/actions/make/make_cpp.lua
```diff
@@ -481,15 +481,26 @@
...
+				if os.isfile(testname) then
+					pch = project.getrelative(cfg.project, testname)
+					found = true
+					break
+				end
 			end
 		end
+
 		if not found then
+			print('not found', pch)
```
starkos: This should go away, but maybe use a `p.warnOnce()`?

starkos:
One comment inline, otherwise LGTM!

src/actions/make/make_cpp.lua
tvandijck: actually that should indeed go away... a leftover from me trying to figure out why it wasn't finding my pch files.

tvandijck:
fixed

samsinsane:
LGTM.

PR #488, opened 2016-05-02
Fix casing on Visual Studio rule XML encoding entity

starkos:


samsinsane:
LGTM.

tvandijck:
I'm not sure I understand this fix, it looks like it just moves a method from the callarray to an explicit call... I don't really see what the difference is... but other then me just not getting it, it LGTM..

starkos:
An argument is passed to the call array, which gets passed to that function and interpreted as a "uppercase the output " flag. Moving it out of the call array allows me to call it without any arguments.

starkos:
And also didn't see that when I wrote the code, so its a subtle one :)

tvandijck:
Ahhh.. ok, yeah that makes sense now... wow, subtle indeed ;)

Issue #483, opened 2016-04-25
"Symbols" flag not resolving properly on Visual Studio 2015


emruiz81:
Hello,

I'm having this issue with premake5 alpha8, that addind the "flags { "Symbols" }" option generates a wrong parameter on the vcxproj file, where it should say `<GenerateDebugInformation>Debug</GenerateDebugInformation>`, it says `<GenerateDebugInformation>true</GenerateDebugInformation>`, and where it should say `<GenerateDebugInformation>No</GenerateDebugInformation>`, it says `<GenerateDebugInformation>false</GenerateDebugInformation>`. 

It is really easy to reproduce, I always get the same wrong output on VS2015. I didn't try it on other visual studio versions.
Thank you in advance for your feedback! 

Ezequiel

tvandijck:
I noticed this too, but this is actually unique to VS2015... it's correct for prior version of VS2015, and is because in VS2015 there is now an option "DebugFastLink" in addition to "Debug" and "No". We need to revisit the "Symbols" flag an make proper API for it, so we can expose these additional options.

emruiz81:
You're right, the new DebugFastLink option makes this parameter more complex than just a "true or false" flag. Maybe as a quick workaround it should just resolve "No" and "Debug" for now, and in the future implement a more customizable API. Thanks for your comments!

tvandijck:
We could indeed, feel free to make a pull request for that... I don't think I have time today to get to that.

emruiz81:
I've checked out the pull request #490 . It seams to do the job properly. But it didn't make it to the alpha 9 right? 
Thanks a lot for you patch @g012 , it solved my problem instantly!

starkos:
Confirmed, it did not make it to alpha 9.

PR #490, opened 2016-05-06
Fix for new value of GenerateDebugInformation in VS2015.

g012:
PR for https://github.com/premake/premake-core/issues/483

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1304,7 +1304,10 @@
 
 
 	function m.generateDebugInformation(cfg)
-		m.element("GenerateDebugInformation", nil, tostring(cfg.flags.Symbols ~= nil))
+		local d = cfg.flags.Symbols ~= nil
+		local v
+		if _ACTION >= "vs2015" then v = d and "Debug" or "No" else v = tostring(d) end
+		m.element("GenerateDebugInformation", nil, v)
 	end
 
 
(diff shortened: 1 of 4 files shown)


src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1304,7 +1304,10 @@
 
 
 	function m.generateDebugInformation(cfg)
-		m.element("GenerateDebugInformation", nil, tostring(cfg.flags.Symbols ~= nil))
+		local d = cfg.flags.Symbols ~= nil
+		local v
+		if _ACTION >= "vs2015" then v = d and "Debug" or "No" else v = tostring(d) end
```
starkos: A nitpick: but the coding convention is to put then/else on their own lines, and it does make the code easier to follow.

starkos:
Can we get unit tests for the new output values?

g012:
Sure.
I added a vc2015 folder, I don't know if that's what you want or if you prefer to have two files for symbols in vc2010.

tvandijck:
I like it this way... LGTM...

samsinsane:
I don't think having a vc2015 folder is a good idea, it separates the tests for the same function into two files unnecessarily. In my opinion, the tests are for the functions, not for the versions of VS. The `test_symbols.lua` file should contain all the tests for VS2010-2013 and VS2015. But that's just my opinion on this, I don't know exactly what the plan was for the VS tests.

starkos:
FWIW, I agree with Sam on this one: it would be good to have all of those tests in the same place.

tvandijck:
https://github.com/premake/premake-core/pull/525

might supersede this PR.. Although adding the tests would still be welcome...

tvandijck:
I think this was all fixed in 525... I'm closing this one, feel free to reopen if I'm wrong, or missed something. As it stands it has conflicts and can't be merged, so an update would be needed anyway.

PR #493, opened 2016-05-09
Add 'Installer' kind to support external vdproj files.

tvandijck:
This allows you to add .vdproj files as externalprojects.

I would like to fully generate vdproj in the future as well, but that is a whole different project.
And maybe this should really be done in a module instead, but a few of the methods used here are not really made to be extensible, so that will require a few other changes as well.

This is mostly a PR for visibility, not sure if we should merge it.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -667,6 +667,7 @@
 			"StaticLib",
 			"WindowedApp",
 			"Utility",
+			"Installer",
 		},
 	}
 
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -473,7 +473,9 @@
 
 	function vstudio.projectfile(prj)
 		local extension
-		if project.isdotnet(prj) then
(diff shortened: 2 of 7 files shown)


samsinsane:
I like the idea, but maybe, starting by making it easier to add support for other project extensions and GUIDs would be a better MR? As for whether .vdproj should be known to core, I don't see a problem with it, it comes with VS doesn't it?

tvandijck:
> It comes with VS doesn't it

nope.. it's an extension you have to install... Which is why I don't think this is the right solution indeed. I think you are right on the money with coming up with something where is would be easier to add other project extensions.

As I said, this PR was mostly for visibility, and to indeed give this some though. In the meantime however I did need this change for one of our internal project to get along. But I really don't think we should merge this as is..

samsinsane:
Ah, I thought it was the bundled installer project. I'm in the same boat, but with Wix, if it was easier to add other project extensions (and their generate functions), modules for VS2015 Android and iOS support will benefit from that. :)

PR #496, opened 2016-05-12
Catch older unit tests up to the latest standards

starkos:
Patch up some outdated stuff in the unit tests, before the new test module lands.


tvandijck:
LGTM.

PR #497, opened 2016-05-15
Support for Authenticated http requests

mversluys:
http.get and http.download have been extended to optionally accept username[:password] which is provided to the underlying curl library. 

User agent is now also being specified as "Premake/<version>" in the http requests.

This does not interfere with the optional parameter which contains headers which is a table type and the existing code scanned the arguments of the functions looking for that table.

Also added is os_getpass which allows lua scripts to prompt for a password from the user. If accepted the relevant documentation for http and os.getpass will be added to the premake wiki.

The purpose of these changes is to facilitate authentication support in [GitHub packages](https://github.com/mversluys/premake-ghp) which allows this module to work with private repositories and with GitHub Enterprise. The addition of user agent was required, the GitHub API refuses requests which don't supply it.

The changes have been tested on OS X (10.11.4) with LLVM 7.3, Windows with Visual Studio 2015 and Linux  (Ubuntu 14.04.1) with gcc 4.8.4.

### DIFF
--- a/src/host/http.c
+++ b/src/host/http.c
@@ -157,7 +157,7 @@ static CURL* curl_request(lua_State* L, CurlCallbackState* state, FILE* fp, int 
 	curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 1);
 	curl_easy_setopt(curl, CURLOPT_FAILONERROR, 1);
 	curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, state->errorBuffer);
-	curl_easy_setopt(curl, CURLOPT_USERAGENT, "premake/5.0");
+	curl_easy_setopt(curl, CURLOPT_USERAGENT, "Premake/" PREMAKE_VERSION);
 
 	if (lua_type(L, userpwdIndex) == LUA_TSTRING) {
 		curl_easy_setopt(curl, CURLOPT_USERPWD, luaL_checkstring(L, userpwdIndex));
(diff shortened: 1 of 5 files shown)


tritao:
LGTM.

These APIs are starting to get complicated each time we add new features though.

Maybe we could extend the table approach and generalize it for all options.

That way we could get rid of the `progress` and `headers` parameters, and replace them with a `options` table, that could have optional table fields like `progress`, `headers`, `username` and `password`.

mversluys:
Agree that is at the breaking point and those signatures are desirable:

```
http.download(source, destination, { options }) 
http.get(source, { options })
```

I don't have code using headers/progress functions on hand, but will take a look at that approach.

tritao:
If you need a simple example of the progress parameter, you can find it here: https://github.com/mono/CppSharp/blob/f986963a3cef36e78320d418b4bb5eda15424fde/build/scripts/Utils.lua#L82

mversluys:
Updated to support cleaner signatures

```lua
http.download(source, destination, { options }) 
http.get(source, { options })
```

Existing signatures are still supported

```lua
http.download(source, destination, progress, headers)
http.get(source, progress, headers)
```

Verified that progress/headers work with the old and new signatures.
Verified that userpwd works as part of options.

The new signatures should allow simpler addition of any further options in the future.

tvandijck:
LGTM.

PR #498, opened 2016-05-30
Integrate testing module into premake-core

starkos:
Integrates the unit testing framework (`testfx.lua`) directly into premake-core as a new module named `self-test`. This allows module developers to run their module's unit test 
suites without having to copy `testfx.lua`, or otherwise jump through hoops. 

While working on Premake itself, nothing changes. You continue to run `premake5 test` to run the core unit test suites.

While working on external add-on modules, you can now run...

     premake5 self-test

...from any location, and Premake will run any unit test manifests it finds in that folder or any subfolders. On my local workstations I keep a folder named `Modules` for all of my add-on modules; running `self-test` from that folder verifies all those modules in one step. Super convenient.

Note that this module isn't enabled by default. I figure that this is an advanced feature that will only be used by people who are writing modules and extending Premake itself, and I didn't want the "self-test" option appear in the help text and confuse people. So you first have to turn it on by requiring it somewhere—I put it in my `premake5-system.lua` script like...

     test = require("self-test")

...so I can use it on any project.

Give it a try and let me know what you think. If approved, I'll add some text to the (currently empty) [Adding Unit Tests](https://github.com/premake/premake-core/wiki/Adding-Unit-Tests) page in the module developer section of the wiki.


samsinsane:
LGTM

PR #499, opened 2016-06-02
Improvements to http.get and http.download

mversluys:
This changes what's being returned from http.get and http.download so that it returns an HTTP result code. There was an undocumented return from http.download where the curl result was being returned. This change breaks the undocumented functionality but makes both functions consistent.

The usage of these functions is now

```lua
error_str, response_code = http.download(url, destination, { options })
result, error_str, response_code = http.get(url, { options })
```

Added a timeout option to http.get and http.download which is the maximum number of seconds before the request will timeout.

```lua
http.get(url, { timeout = 5 })
```

Documentation will be updated if accepted.

### DIFF
--- a/src/host/http.c
+++ b/src/host/http.c
@@ -173,6 +173,10 @@ static CURL* curl_request(lua_State* L, curl_state* state, const char* url, FILE
 			{
 				curl_easy_setopt(curl, CURLOPT_PASSWORD, luaL_checkstring(L, -1));
 			}
+			else if (!strcmp(key, "timeout") && lua_isnumber(L, -1))
+			{
+				curl_easy_setopt(curl, CURLOPT_TIMEOUT, luaL_checknumber(L, -1));
+			}
 
 			// pop the value, leave the key for lua_next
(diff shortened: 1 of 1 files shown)


tvandijck:
It looks good to me, but I'm going to leave it to @samsinsane to merge after review.

tritao:
LGTM also

src/host/http.c
```diff
@@ -209,13 +220,24 @@ static CURL* curl_request(lua_State* L, CurlCallbackState* state, const char* ur
 	return curl;
 }
 
+static void curl_cleanup(CURL* curl, curl_state* state)
+{
+	if (state->headers)
+	{
+		curl_slist_free_all(state->headers);
+		state->headers = 0;
```
samsinsane: Minor inconsistency here, the initialization uses `NULL` but here you used `0`.

samsinsane:
LGTM, not a fan of breaking functionality but in this particular case, returning the HTTP result code is far better than returning the CURL return code.

PR #507, opened 2016-06-10
Allow rules to generate code that subsequently gets compiled.

tvandijck:
So I made this rule:

```lua
rule 'example'
    location(path.join('..', _OPTIONS.to))

    display 'Example compiler'
    fileExtension '.example'

    propertydefinition {
        name = "output_path",
        kind = "string",
        display = "Output Path",
        description = "",
        value = "$(IntDir)",
    }

    buildmessage 'Compiling %(Filename) with example-compiler...'
    buildcommands {
        'package-example-compiler.exe [output_path] "%(Identity)"'
    }
    buildoutputs {
        '%(output_path)%(Filename).example.cc',
        '%(output_path)%(Filename).example.h'
    }
```

And wanted to get msbuild to compile the *.example.cc output files, and link them into my projects.

So two things I had to do:
- fix another bug in path.isabsolute (it would not recognize the %() tokens, and incorrectly expand)
- make the .target file generator recognize compilable files.

### DIFF
--- a/src/actions/vstudio/vs2010_rules_targets.lua
+++ b/src/actions/vstudio/vs2010_rules_targets.lua
@@ -61,6 +61,7 @@
 
 	m.elements.computeInputsGroup = function(r)
 		return {
+			m.computeCompileInputsTargets,
 			m.computeLinkInputsTargets,
 			m.computeLibInputsTargets,
 		}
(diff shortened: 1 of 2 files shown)


starkos:
LGTM!

Issue #500, opened 2016-06-03
rule: valid scope for filter and location?


youaresoomean:
Hi,

i try to make the buildcommands/outputs of a rule platform specific. But neither a filter {..} inside or around rule "..." statements has any effect.

```lua
rule "GLSLCompile"
   display "Compile GLSL file into C arrays"
   -- location("%{wks.location}/dependencies")
   fileextension ".glsl"
   buildmessage 'Packing %(Filename) into header file'
   configuration "windows or android"
        buildcommands 'call $(SolutionDir)%{wks.stvs_reltoolpath}/android/build/run_xxd.bat %(FullPath)' -- 'call $(ProjectDir)../../../../../tools/android/build/run_xxd.bat %(FullPath)'
        buildoutputs '%(FullPath).h'
    configuration "iOS or macosx"
        buildcommands 
        {
            'cd ${INPUT_FILE_DIR}',
            'xxd -i ${INPUT_FILE_NAME} | sed s/}\\;/,0x00}\\;/ | sed s/unsigned/const/g > ${INPUT_FILE_PATH}.h',
        }
        buildoutputs '${INPUT_FILE_PATH}.h'
```

Also, %{wks.location} in rule/location doesnt seem to work - errors with "attempt to index global wks (a nil value)" Is rule no valid scope for wks.xx?

PR #511, opened 2016-06-10
We need to actually compile the context of the rule, so we can propely use the "filter {}" system.

tvandijck:
I'm implementing the PBXBuildRule in our xcode backend, which is relatively simple, but due to differences in how the buildcommands etc are interpretted, I need the filter {}'s to work.

```lua
rule 'example'
    location(path.join('..', _OPTIONS.to))

    display 'Example compiler'
    fileExtension '.example'

    propertydefinition {
        name = "output_path",
        kind = "string",
        display = "Output Path",
        description = "",
        value = iif(_ACTION == 'xcode', "$DERIVED_FILE_DIR", "$(IntDir)")
    }

    filter { 'action:vs*' }
        buildmessage 'Compiling %(Filename) with example-compiler...'
        buildcommands {
            'package-example-compiler.exe [output_path] "%(Identity)"'
        }
        buildoutputs {
            '%(output_path)%(Filename).example.cc',
            '%(output_path)%(Filename).example.h'
        }

    filter { 'action:xcode' }
        buildcommands {
            'package-example-compiler "$DERIVED_FILE_DIR" "$INPUT_FILE_PATH"'
        }
        buildoutputs {
            '$DERIVED_FILE_DIR/$(INPUT_FILE_BASE).example.cc',
            '$DERIVED_FILE_DIR/$(INPUT_FILE_BASE).example.h'
        }
```

This PR allow the above to work.

### DIFF
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -58,15 +58,8 @@
 
 	p.alias(oven, "bakeWorkspace", "bakeSolution")
 
----
--- Bakes a specific workspace object.
----
 
-	function p.workspace.bake(self)
-		-- Add filtering terms to the context and then compile the results. These
-		-- terms describe the "operating environment"; only results contained by
-		-- configuration blocks which match these terms will be returned.
-
+	local function addCommonContextFilters(self)
 		context.addFilter(self, "_ACTION", _ACTION)
 		context.addFilter(self, "action", _ACTION)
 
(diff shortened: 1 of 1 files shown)


samsinsane:
This looks like it would resolve #500? LGTM.

starkos:
This looks good. Only suggestion would be to pull out the common `context.addFilter()` stuff to its own function so it can be shared with project, in case we need to modify or extend it later.

Issue #514, opened 2016-06-15
New test framework doesn't fail anymore..


tvandijck:
```
make_pch.buildRules_onC: C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:30: attempt to call a nil value (field 'pchRules')
stack traceback:
        C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:30: in upvalue 'prepareRules'
        C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:111: in function <C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:108>
make_pch.noConfig_onHeaderAndNoPCHFlag: C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:25: attempt to call a nil value (field 'pch')
stack traceback:
        C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:25: in upvalue 'prepareVars'
        C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:52: in function <C:/dev/premake/tests/actions/make/cpp/test_make_pch.lua:49>
make_flags.includeDirs: C:/dev/premake/src/base/_foundation.lua:121: Unable to find function 'includes'
stack traceback:
        [C]: in function 'error'
        C:/dev/premake/src/base/_foundation.lua:121: in function 'premake.callarray'
        C:/dev/premake/tests/actions/make/cpp/test_flags.lua:25: in upvalue 'prepare'
        C:/dev/premake/tests/actions/make/cpp/test_flags.lua:35: in function <C:/dev/premake/tests/actions/make/cpp/test_flags.lua:33>
make_clang.usesCorrectCompilers: C:/dev/premake/tests/actions/make/cpp/test_clang.lua:31: attempt to call a nil value (field 'cppConfigs')
**1228 tests passed, 0 failed in 2.37 seconds**
Done (3621ms).
```

All tests pass, even though I get about a mile of text with all kinds of failures in tests...

PR #515, opened 2016-06-15
test_runner failed to correctly count failed tests

tvandijck:
https://github.com/premake/premake-core/issues/514


starkos:
Thanks for this—looks like one of my last fixes broke more than it fixed :[

PR #516, opened 2016-06-15
Gmake refactor

tvandijck:
This is a big change, but adds rules support to gmake.


starkos:
Honestly, I'm happy for someone to take over this exporter and move it ahead. As long as it passes the tests, feel free to go ahead and merge!

tvandijck:
Yeah, this needs a lot of testing internally... I just wanted to post this as I move along with it.
So far I have 'rules' working, and the rest of the tests work in my local branch, but apparently not yet on top of master...

tvandijck:
my changes apparently break the 'd' module... I'll take a look at that...

TurkeyMan:
The other big-ticket WRT the gmake writer is support per-file options.

tvandijck:
Yeah, I welcome any PR's on my branch towards that... I'm a bit caught up in some other works here at Blizzard that is not related to anything premake.

tvandijck:
This is in production here at Blizzard now... fixed a few issues with per file configurations, but other then that no major changes... We should merge this sooner then later...

aleksijuvani:
What's the status on this? Looks like this adds support for per-file build options, I think? I'd love for this to be merged. Right now I have to resort to an ugly hack to get similar behavior.

tvandijck:
The status? hmm... well it's in production here at Blizzard.... however.... recently there have been a couple of changes to the existing gmake backend, and those changes are not included in this refactor...

and I think that causes a bunch of tests to fail right now.... I unfortunately don't really have much time to fix them right this second... any help would obviously be much appreciated.

aleksijuvani:
I think I could take a look after #628 is merged.

tvandijck:
@aleksijuvani K, I merged your changes, and rebased this against latest master... thank you for your help so far.

aleksijuvani:
No problem! Although, it looks like the merge conflict fixes got combined with commit 3bc4c430f58a2d64b5b98bc287f9c085107fd62a. The original commit looked like this: 272e2805f31b7f43d3c0143b814a3c2e77f8208f. It doesn't really matter, but it does make the commit look a bit weird... I think it would've made more sense to combine the merge conflict fixes with bbbae213fed09dd94233c54922cb4f8fa7db48c8 instead.

resetnow:
As far as I understand, the only problem here is a broken D backend. I think a proper solution would be keeping a set of per-file rules that are selected automatically based on language.

aleksijuvani:
> As far as I understand, the only problem here is a broken D backend (tests passed after I removed ones related to D)

Unless new merge conflicts have appeared since my pull request, then yes, that's right.

aleksijuvani:
I've been trying to find some time to work on this, but I'm really uncomfortable tinkering with the D backend. I have no experience with the language and have no projects to test it on. If someone else would like to move this ahead, I would much prefer that.

resetnow:
>I have no experience with the language 

Neither do I, if I ever try to fix this it'll be a Hello World project in D.

tvandijck:
neither do we unfortunately... here at Blizzard we deleted the D module entirely, we don't use it, and it just adds maintenance for us.

resetnow:
Upon further investigation, D seems to have its very own build tool (http://code.dlang.org/getting_started) — kind of looks like cargo (or vice versa). I wonder how useful D support in premake is. Since premake doesn't implement dependency management or DUB package support maybe it makes sense to drop it completely?

resetnow:
@starkos could you please weigh in on the D matter?

samsinsane:
There's plenty of reasons to keep the D module in Premake.
1. DUB doesn't support C, C++ or C# projects. Premake does.
2. From what @TurkeyMan has told me about DUB, it sounds like it's very much for small projects rather than larger projects. So per-file configurations mightn't be possible.
3. If the module can't be updated to match the gmake changes, it's a sign that the gmake changes have broken the extensibility of the action. Which means that others will have difficulty adding similar functionality into Premake.

Personally, I'd rather see the D module fixed than have it remove it.

starkos:
I've attached Manu to this issue; hopefully he can provide a little guidance.

Longer term though, if we don't have someone to maintain it, we'll have to consider moving it out of core. This was always the drawback with embedding modules into core.

tvandijck:
> If the module can't be updated to match the gmake changes, it's a sign that the gmake changes have broken the extensibility of the action.

I don't think that is the issue with the refactor really, although it certainly could be... I just renamed some methods to use the camel casing standard, and some other functions were simply removed or entirely replaced... The D module will simply have to be adjusted, and maybe the gmake backend as well to accomodate the D-modules requirements.

That said, it's why the refactor is a pull request long before merging and it being ready.. for people to review and make comments, help with exactly these issues, etc...

Here at Blizzard, this back-end is in production for our [deepmind](https://deepmind.com/blog/deepmind-and-blizzard-release-starcraft-ii-ai-research-environment/) project, and SC2/Heroes servers. So it's functional at least, but with the help of the community it can be better...

tvandijck:
BTW... if there is anyone with an fitting D example project out there, I'd be happy to give things a shot... I just know very little about D, so as much as I can make the module compile and run, I can't guarantee it'll work unless I have some test cases..

samsinsane:
> I don't think that is the issue with the refactor really, although it certainly could be... 

I didn't mean to imply there were issues with the refactor. More than if people are unable to recreate the overrides in the D module against the new gmake backend, that it's probably because the changes didn't factor in the D module. Which is pretty much what you've said here:

> The D module will simply have to be adjusted, and maybe the gmake backend as well to accomodate the D-modules requirements.

So I think we're on the same page? Overall, these changes are great, the D module just needs the same changes made to it and then a quick look over from @TurkeyMan as to whether the makefile that is generated is correct or not.

starkos:
Is it possible to patch D up enough to pass the tests and commit this change, even if it isn't necessarily correct? Once it lands, the folks that are using the D module can submit a fix to patch things back up?

tvandijck:
yes, I'll make an attempt at that.. but there have recently also been a few PR's that made changes to the gmake backend, so we need to make sure to capture those changes in this refactor... 

I'll try to make some time this week...

TurkeyMan:
Hey sorry people. I've been way AFK after I was made redundant ;)
Let me catch up...

TurkeyMan:
I'll patch up the D stuff today, I have some other changes I haven't merged (pending a new release of the D tooling, half of which was released just 2 days ago).

Just for the record, I've said many times before, I don't mind modules not being in core, in fact, it would even be more convenient in some cases, but that's not why they're there... the reason I pushed for them to be there in the first place is because we have virtually no real-world tests for when we break core. We need some canaries. The D module in particular touches a fairly large API surface area. The fact that D is a pain to maintain demonstrates why it should be around; every change that breaks core modules almost certainly breaks every other module out there, and breaking all end-users modules with every release of premake is not how you make premake popular and attract new users/developers. Obviously there is a contention with the fact that premake is extremely under-staffed, but that's a guaranteed self-fulfilling prophecy if we're breaking core every few days for all end-users.

For my money, the thought process goes like this:
- Change gmake significantly
- Breaks D
- "Oh, my core changes broke 3rd party modules... I wonder if it's possible to write my core changes in a way that are less likely (or unlikely) to break those users code?"
- Consider the situation...
- *Yes*: change solution, break less stuff. *No*: Break the code with deliberate intent (prefer core dev agreement?), perhaps publish that breaking change in a changelog?

Even if you still commit to the breaking change; you know you did it, you had the opportunity to consider other options, and you should probably take a moment to put it on the record that you broke it in that way, and how other users are supposed to correct their modules against your changes..

Obviously everything's fuzzy to some extent, but I always felt this general idea was sound.
Truefacts; the better we are able to do a good job of this stuff as a (tiny) dev team, the more likely we are to attract users and developers.

TurkeyMan:
I'm just gonna put it out there that fixing this is a massive job! >_<

tvandijck:
To be honest. I'm with @TurkeyMan on this one.... even though I made this pull request, I'm sure there is ways we can make this more backwards compatible. The problem for me however is that I can only do so much as time permits... in the end we got a game to ship, and the pace off releases on Heroes of the Storm is pretty high. So stuff like this when it needs more work, will just lag behind quite a bit, and to be honest "premake" as a product has much bigger fish to fry then this refactor.

Maybe we can take some of the ideas, and simply plug it into the existing gmake backend, and more slowly move the gmake backend forward...

resetnow:
Another crazy idea is to create a separate “staging” backend with different name.

@TurkeyMan is it true that DUB isn't intended to be used with large projects? Can we hope to see D fix from you?

@tvandijck please consider merging https://github.com/Blizzard/premake-core/pull/4 — it fixes a file name escaping problem

TurkeyMan:
@resetnow DUB is not capable of large real-world programs. It's like cargo, it just builds self-contained D projects, and doesn't know how to make working on multi-language projects practical.
The main reason I use D a lot is precisely because I want it in multi-language environments. D is binary compatible with C/C++, so if you're sick and tired of C/C++'s bullshit, you can write parts of your ecosystem in D, and marry it near seamlessly with your existing C/C++ investment.
I come from games, which are gigantic projects with fairly complex build environments. Most game shops use Visual Studio, and D just fits right in. Rebuild works as expected; only necessary parts trigger rebuilds. Plugging DUB into that environment is really clunky by contrast.

Yes, you'll see the gmake support for D rewritten from me. It's in progress... but it's a fairly big job. I'm really just writing it again from scratch after a failed effort trying to adapt the code. Much easier to just rewrite ;)

TurkeyMan:
Okay, so fixing D as a module against the new gmake has some problems.
The biggest problem is that modules requirements are resolved in `postBake()`...
This means modules can't affect the oven, which doesn't seem right.

The new gmake action calls `rule 'cpp'`, etc, so logically the D module should call `rule 'dc'` to create the rule for `.d` source files, but the rules appear to be digested in the oven, effectively meaning modules can't create rules, since they're already processed before any modules are loaded.

I can't really imagine the full impacts of moving the time that modules are loaded... why are they only loaded in `postBake()`? Is it that the conditions leading to a module being required are only discovered in the oven? I guess I can see how that's possible, but then we have a chicken/egg situation.

The `rule` processing could perhaps be moved out of the oven? All seems a bit wonky.

starkos:
Modules can define a `_preload.lua` script that runs at initialization time (that was the intention, at least). The dependency check was a bit of future-proofing intended to avoid loading unused modules. You can install the hooks in `_preload`, and then just check the configuration state before running your new logic.

TurkeyMan:
Yeah okay, I guess that's a reasonable solution. It just means that fragments of the module (the rules it defines) will be available even if the `_preload.lua`'s predicate function says not to load the module.

tvandijck:
ultimately the way I hooked up the rules, it just creates a table... there is probably a way to go around that, and simply inject stuff in that table... I think it is perfectly reasonable to make changes to the refactor to refactor it into something usable.... obviously when I first wrote this thing, I had one problem to solve, and I wanted to solve it in the most generic possible way.. that doesn't mean it is the right solution...

resetnow:
Any progress on this?

tritao:
By the way, I tested this PR last week and everything seems to work on our project.

resetnow:
It works for me too, I was just hoping using upstream software instead of fork would be better.

tvandijck:
I'll rebase it again, and merge our latest fixes into it as well, which fixes a bunch of PCH related issues, and adds support for "utility" projects with buildcommands for specific files and pre/post build steps, courtesy of the World of Warcraft team ;)

I would love to merge it, but I really need sign off from everyone... and the roadblock is still the D backend.

tvandijck:
All squashed and updated...

tvandijck:
@TurkeyMan I added the "onInitilize" on the action...

```
newaction {
        trigger         = "gmake",
...
        onInitialize = function()
			p.make.cpp.initialize()
	end
}
```

which initializes the rules for the cpp backend in this case.. you should be able to add the rule 'dc' in that callback. Actions are typically defined in the _preload.lua stage, otherwise the action would not be visible to premake at any stage.

tvandijck:
@TurkeyMan FYI... you can also insert a callback into the ```cpp.elements.makefile``` table right after the cpp.createRuleTable, and just manually add the rules for the D related files... during the backend we're really just expecting each rule to have a buildoutput/buildmessage/buildcommand and a fileExtensions property.

The only reason I went with the 'rule' logic, was that is seemed kind of logical to do, since a makefile really is just a set of rules applied in order to some set of extensions. So this new gmake backend is written that way... per file configurations are just treated as per file rules, and all 'global' rules are effectively treated exactly the same way... as per file filters (by extension) and applied and chained equally. For example if you just created a rule that has a cpp file as output, it would just automatically chain to the rule that compiles that cpp file into an .o file....

in a similar way, with this new gmake backend you could do the following:

```lua
rule 'myownlanguage'
    fileextension '.mol'
    buildoutputs '$(OBJDIR)/%{file.objname}.llvm'}
    buildcommands { 'mycompiler.exe .....'}

rule 'llvm'
    fileextension '.llvm'
    buildoutputs '$(OBJDIR)/%{file.objname}.o'}
    buildcommands { 'llvm.exe .....'}
```

And things would automatically chain from rule to rule...

tvandijck:
@TurkeyMan last one... if you have any kind of test project for D, I'd be perfectly happy to make it build with this backend and get things going... and if you have any kind of refactored but not entirely done work already for the D module... send it my way and I'll try to help out where I can...

resetnow:
@tvandijck thank you for the update! Please consider merging https://github.com/Blizzard/premake-core/pull/5

resetnow:
https://github.com/Blizzard/premake-core/pull/6 — adds more variables to custom rules.

@TurkeyMan could you at least provide an example D project with all the feature you'd like to see supported, and we'll start it from there?

resetnow:
@TurkeyMan D language seems to have 3 compilers — which one should be supported? https://wiki.dlang.org/Compilers

DrLynix:
What's missing in order to merge this? How can I help?

resetnow:
D support is missing, at this point the only way to help is to look what capabilities D backend has in master and adapt it to work with the new gmake backend.

DrLynix:
I don't know much about D, but I can compare the current makefile with those generated by this PR to check the difference with a sample D project, would that help?

resetnow:
Neither do we! (and those who know something choose to be silent, <i>cough</i>). That would help — although in this PR, D backend isn't functional at all.

starkos:
I think at this point we should seriously consider pulling the D module from core. I totally understand Manu's point about maintaining stability for external modules; it was well intentioned and a good idea. But in this case we *want* to make a breaking change and we're completely stuck because there is no one here with the domain knowledge to fix it.

Are there any objections? If not, we'll have to figure out how to cleanly pull it back out to its own repository (and who should own that repository, and where it should live).

tvandijck:
I have no objection, although I do also very much understand @TurkeyMan his point.

I also don't think this particular refactor is done, and so having anyone make a new D module that breaks another few times due to the ongoing work is just a waste of valuable "open source time" as you call that..

So my proposal... well make a premake-dlang repository here in the premake organisation with the current code in it and make it work as a module against premake5-alpha.?? and make that release. Make the readme very clearly indicate that it is supported up until that release, but due to ongoing refactoring in the gmake backend, porting will be required to pull it back up...

The premake org owns that repository, and anyone willing and able to fix it is very welcome to do so.. 
and if it ever gets fixed, we can pull it back in.... if it never gets fixed, there is no demand, or people choose to go elsewhere, which is a bummer for premake, but ultimately up to those teams. CMake after all has a lot of momentum and support in the open source world, so if they have better support for D then that may be the better choice. There is also other forks of premake out there that are in use.

And in all honesty, I also think the 'clean', 'gmake' and 'vstudio' backends should be refactored into modules.


As for the 'gmake' backend itself... I don't think it's done... I'm not really liking the 'rule' stuff I did, since it makes adding other rules really difficult... I took a stab at the D module myself last week, and basically ran into similar problems, so it makes the current code not very extendable, which was certainly one of my goals. So I'd say, we merge this, so people can test things out, and move on with per file config support etc... but I certainly want to keep churning a bit on some of the architectural changes I made, to make making a 'new' Dlang module much easier, because ultimately dlang is just another file extension, with a different toolset... so that is what the dlang module should be defining, nothing more.

samsinsane:
I'm not really sure where I stand on this. I completely agree with @TurkeyMan on keeping it in core, and that it causing issues like this is a sign of what a module developer will have to go through. But having said that, this PR has been going for awhile and it seems as if @TurkeyMan is rather busy. He also appears to be the only one with the knowledge required to point us in the correct direction.

I know this discussion is about the GMake refactor, and potentially removing the D module. But to extend on that, should we replace the D module with mobile platform modules? It might not seem like these things are related at all, but the idea of the D module in core is more as a "canary" (as @TurkeyMan puts it) than as a selling feature. I imagine people are more likely to buy into Premake because of Android or iOS support than because we support D, or any other language really.

Android will open the door to Java being added, iOS will open the door further for Swift. This covers the "adds a new language to Premake" part of the D module. iOS will also encourage refactors of the XCode back end. Both will encourage refactors of the VS and gmake back ends and the toolset system, which covers the rest of the D module parts.

No offence to the D community, but I would imagine the Android and iOS development communities are significantly larger. Which should, in theory, increase the number of potential Premake users and developers. I believe that these two systems would also make it easier to rope in the Raspberry Pi and Arduino development communities too, since they rely on GCC and Clang to target more than Linux and macOS, and x86 and x86_64.

At the end of the day, I think we will need to remove the D extension of gmake if it's blocking the gmake back end from supporting per-file configurations. But I don't think we should just remove our "canary" without any plans for a replacement.

This reply ended up being way more than I planned to write, but I've been meaning to bring up the idea of adding Android and iOS to core for awhile now and it just kind of fit here pretty nicely.

tvandijck:
I agree with @samsinsane here... but I think in order to get to a place where adding a language like D, Java or Swift is easy, we need to actually think about how we want a module for that to look...

quote honestly, in my opinion, adding another language and compiler should be as easy as:
```
newlanguage 'D'
    extensions { '.d' }
    compiler "C:\D\dmd.exe"   -- or whatever..
    <define outputs>
    <define the options and api specific to this language>
```

right now, adding a language requires someone to write a module, override 50 methods to inject itself into the right places, etc... and it requires so much inside knowledge of premake, that it almost requires you to become a core developer on the premake project to pull it off.

this gmake refactor, is an attempt at going in that direction, but I don't think I'm there yet. Since I tried something like the above for D, and couldn't do it, without injecting functions into tables, overriding some other stuff, and practically rewriting half the gmake backend through overrides.. So naturally any refactor will break modules like that... and adding android support to the visual studio action is a similar experience... 

So to become more resilient against refactors like that, we need to define proper public API's to support adding toolsets, compilers and languages.. And then refactor our current actions such that they can support those things without having to inject stuff into them.. Getting there, will require time and dedication by a number of people, I've made a first step towards this with this gmake refactor, but again, it's not good enough yet, and more work will be required to get there.

samsinsane:
@tvandijck that sounds like a really good idea. Would actions define their own option functions, or would it be completely fixed? For example, gmake will likely only need a value returned, while vstudio could require the element name to be returned too. Maybe that's as complex as it ever needs to be, but these option functions would need to work with a filtering system, so that you can setup multiple options based on the language/compiler/action/system/platform/architecture/whatever.

tvandijck:
@samsinsane yeah, I think different modules/actions whatever should be able to provide API to this API, and to 'check' if said API's are implement/filled-in, and if not give an error that this particular action doesn't have support for that 'language'...

The biggest issue is that this will just be a little bit of trial and error as we're discovering the needs of each action and how we can solve that... callbacks are nice and all, but if we can do it with just data, that be great.

resetnow:
Sorry for bugging all of you again, but what's the status of this refactor now? Did we agree to pull D out of core? If so, can this be rebased and merged? I can help with the tests, if needed.

samsinsane:
@tvandijck I was thinking about this yesterday and today. Perhaps we should make a .NET/C# module that contains all the C# based extensions to core, including the .NET toolset. This should provide a similar level of coverage that the D module did, while also being something that numerous users can provide feedback on. Thoughts?

tvandijck:
@samsinsane well, now that we have the module system in place, I actually think that both the gmake and visual studio actions should simply be modules themselves as well... but it's not all that easy... there is a lot of assumptions, mostly for historical reasons I think, about those things existing where they do, and other modules extending them... It all sort of feels like there is something missing from premake that allows these things to be easier to extent...

in the visual studio action I introduced these things:
https://github.com/premake/premake-core/blob/master/src/actions/vstudio/vs2010_vcxproj.lua#L620

which allows you to more easily add other tags... So for example, add 'D' to the vcxproj generator would be simply adding another category in that table, but it's a very different solution then what we need for the gmake module, because it's structured so differently. I don't think they technically have to be so different though.. "ClInclude" and "ClCompile" are just msbuild rules, just like any other custom defined rule.

this gmake refactor tried to basically use the "rules" api to define the default C,C++ languages as rules, which can then be interpreted the same way as any other rule, but it's not complete... the rule api doesn't cover everything we need, which can easily be seen in the m.categories tables in the Visual Studio action as well... where there is callbacks, calling very specific set of functions to generate the right visual studio XML to support a particular option. ClCompile beeing the most complicated one, for obvious reasons.

I'd like to get to a place where you specify a language with a set of file extension, and allows you to define properties and callbacks for each... so that you can say, for the C++ language any cpp file has these properties (optimization, rtti, exceptionhandling, etc).

The problem is, whichever way you go, you end up having to write action specific code in the language definition, or language specific code in the action. I don't know of a generic way to say: exceptionhandling 'on', needs to generate <ExceptionHandling>true</ExceptionHandling> on visual studio, but "--exception-handling=on" for gmake on the commandline..

Anyway, I'm rambling on... I don't know the right answer, all I know is that adding new languages requires too many overrides and too much knowledge of premake internals, which makes refactors like this almost certainly break a module or two.

samsinsane:
@tvandijck I really like the idea of having APIs to extend actions instead of using (in some cases, over using) `premake.override`. Don't need the `ClCompile` section, just call `action.removeSection("ClCompile", conditionalFunction)` or something to that effect. I think that would simplify things a bit, it doesn't matter where the code lives, it doesn't matter what the function is called, it has an external name that can be used to find it. I guess this would require some insane documentation about what exists, and it'd be much harder to just arbitrarily rename those external names.

Although, maybe the gmake code has it right? Create a table that's iterated over:
```lua
vsAndroid = {
  ["ClCompile"] = { -- Section name, VStudio looks for this, if it doesn't find it, calls default
    ["WarningLevel"] = { -- Element name, Vstudio emits this to the file "<WarningLevel>"
      ["PremakeVariable"] = "warnings", -- This or the "WarningLevel" table becomes a function with `wks`, `prj`, `cfg` passed in
      ["On"] = "...",  -- The usual value lookup conversion
    }
  }
}

action.registerExtension("vstudio", vsAndroid, function (wks, prj) return prj.system == p.ANDROID end)
```

Haven't fully thought this through, I might be missing something. I guess the first issue is that there's no way to copy the default and modify it, which you'd want to do with VStudio because that's how it works. Perhaps, if you define `ClCompile` it ignores the default, if you define `ClCompileExt` it merges the two together, overriding the default values with the `ClCompileExt` ones. Actually, maybe we want a way to say what this table extends. For example, `vsAndroid` would want to extend the `vstudio` table, but if you apply this to make, you might have `vsAndroidGCC` and you want that to extend `gcc`, and `vsAndroidClang` to extend `clang`.

I don't want to delve too far into the details yet, these ideas mightn't be close to what you're imaging, or they could just be horribly flawed. Also, should we spin this out into an issue? I feel it's relevant to this PR, but also could become extremely noisy.

starkos:
That's an interesting idea…the logical path from there is a template system.

tvandijck:
> Also, should we spin this out into an issue? 

We're 62 comments deep on this PR now ;) I think making this into an issue sounds about right, or in the github project space there is something more appropriate for this?

resetnow:
Any progress on agreeing on a solution?

tvandijck:
No agreement, and I think for the sake of progress I'm going to be moving all this code into a "gmake2" module, and make it a separate action. That way we have a cleaner way to update modules like the D module.

I think I'm going to do the same with our XCode module, which is a total rewrite of the xcode action by our dedicated mac team. Right now however that module is a replacement, rather then an addition, just like this ginormous gmake refactor. I think writing them as an addition would have been better in the first place.

I'll leave this PR open for now, but as soon as I got it working as a module, I'm going to close this.

tvandijck:
superseded by: https://github.com/premake/premake-core/pull/803

PR #518, opened 2016-06-17
Add generated files to vxcproj file.

tvandijck:
The fix we made for 'rules' where a .cpp file gets automatically compiled as part of the rule if the buildoutputs of a rule creates those files, needs to be applied to custom build steps as well.

Say for example in your project you write:
```lua
files { '*.example' }
   buildoutputs { '%{ file.basename}.cpp' }
   buildcommands { 'example-compiler %{file.relpath} %{file.basename}.cpp' }
```

then without this change, the %{file.basename}.cpp would not get compiled automatically.
And you can't do a 
```lua
files { '**.cpp' } 
```
on files that don't yet exists

This change add support in Visual Studio for those files to get added, and it makes it look pretty too:
image

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -30,6 +30,7 @@
 	m.elements.project = function(prj)
 		return {
 			m.xmlDeclaration,
+			m.updateFileTable,
 			m.project,
 			m.projectConfigurations,
 			m.globals,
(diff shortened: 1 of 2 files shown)


samsinsane:
Why does that screenshot show the same file four times? This seems like a bug?

tvandijck:
No a bug.... As you can see 3 out of 4 are 'exclude from build'... this is because in this particular setup we used the 'build/obj/x64/Debug' as the output path, and hence each configuration has it's own output. In this particular sample maybe not the best selection of tokens for the 'buildoutputs', since the output for each configuration is going to be equal. But at the same time, it did correctly test the ability for Visual Studio to setup a project correctly in the case where this **is** useful...

samsinsane:
LGTM

starkos:
This LGTM as well. Is there any logic in there that might make sense to reuse for the other exporters? Or perhaps to put somewhere in the baking step so the exporters just get it for free? I'm not entirely clear from this PR how much of it is VS-specific.

tvandijck:
I originally had this as part of the 'p.project.bake' process, indeed in a more generic way... However this seriously messes up xcode and gmake...

In xcode for example, this is already automatically done by the compiler.
In gmake, as I am adding support for rules, I also already automatically do this...

Adding this in the p.project.bake turned out to be making visual studio easier, but all other back-ends considerably harder. So in the end I opted for the VS specific approach.

starkos:
Thanks for the info!

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -57,6 +58,60 @@
...
+			local filecfg = fileconfig.getconfig(node, cfg)
+			if not filecfg or filecfg.flags.ExcludeFromBuild then
+				return
+			end
+
+			if fileconfig.hasCustomBuildRule(filecfg) then
+				local buildoutputs = p.project.getrelative(cfg.project, filecfg.buildoutputs)
+				if buildoutputs and #buildoutputs > 0 then
+					for _, output in ipairs(buildoutputs) do
+						addGeneratedFile(cfg, node, output)
```
starkos: This should first check to see if the file is actually buildable, maybe:

```lua
if path.iscppfile(output) then
```

starkos:
> I originally had this as part of the 'p.project.bake' process, indeed in a more generic way... However this seriously messes up xcode and gmake...

Sorry to suggest rework after the fact, but what about moving it back into baking (got to get my in-progress changes done and pushed, would make this easier) and marking those file cfgs with a `generated` flag? Then Xcode and Make could ignore them easily, but they would still be available to modules that needs to examine the file lists (i.e. Qt)?

TurkeyMan:
We've reverted this patch: https://github.com/premake/premake-core/pull/539
This should we re-opened and further worked.

PR #519, opened 2016-06-20
Undefine CURL_CA_BUNDLE on Linux builds

starkos:
Not sure if this is the right solution, but I had to undefine this symbol in order to Premake to build on Ubuntu for the 5.0-alpha9 release. Can someone with more Linux smarts confirm or suggest a better fix?

### DIFF
--- a/contrib/curl/lib/config-linux.h
+++ b/contrib/curl/lib/config-linux.h
@@ -4,7 +4,7 @@
 /* #undef BUILDING_LIBCURL */
 
 /* Location of default ca bundle */
-/* #undef CURL_CA_BUNDLE */
+#undef CURL_CA_BUNDLE
 
 /* Location of default ca path */
 /* #undef CURL_CA_PATH */


tritao:
I think undefining this will break HTTP SSL support. This define should point to a system path with certificates. See https://github.com/premake/premake-core/blob/ef03d2a475d7629ea1b3694d114712d3f11628fe/contrib/curl/premake5.lua#L25 for how we search for this path in Linux.

starkos:
Curious to know if 5.0-alpha9 works for you, as I did make this change as part of the binary build.

If I can't fix it this way, what would be the right way to get a build working on a stock Ubuntu install?

tvandijck:
The right way to fix this is to use the 'autoconf' module I made... Ultimately that 'config-linux.h' was auto generated on a particular Linux box, and will likely not work on other 'BSD/Linux' distributions. Ultimately the lack of autoconf in premake is causing this...

Now, with the way we bootstrap, we could add a './configure' step to the bootstrap and generate those header files for zlib/libzip and curl more correctly, but premake itself will be unable to do this right now, and hence those header files will have to be manually tweaked for each distribution or configuration. For example if you were to compile in 64-bit, a few of the defines such as "SIZEOF_SIZE_T (4)" are all wrong.

I wish software creators wouldn't rely on these kind of things for middleware so much... most of these defines are easily discoverable by using the compiler provided defines from the compiler. But that's unfortunately not done very often.

starkos:
> I wish software creators wouldn't rely on these kind of things for middleware so much... most of these defines are easily discoverable by using the compiler provided defines.

This.

I'm going to close this PR then and open an issue instead.

PR #521, opened 2016-06-22
Select clang as default tool on MacOSX.

tvandijck:


samsinsane:
LGTM, I thought this already occurred though?

tvandijck:
Well, yeah, I made a pull request for it like ages ago, but that one was declined, and I never had the time to make it the way it was suggested in that PR.

https://github.com/premake/premake-core/pull/248

tvandijck:
Apparently the monodevelop module can't handle this though :( anyone have any insight on that?

samsinsane:
I'm really not fond of how we have these "heavily used" modules holding up PRs for months.

[This block](https://github.com/premake/premake-monodevelop/blob/master/monodevelop_cproj.lua#L456-L464) seems to be the problem, but monodevelop only supports GCC which if explicitly set will cause the exact same problem since `clang.getlinks` just calls `gcc.getlinks`.

@TurkeyMan it looks like you're the main contributor (and creator) of this module, are you able to provide some insight into this?

samsinsane:
Once #607 goes through, you should be able to rebase and we can finally merge this in. :)

samsinsane:
@tvandijck can you rebase this on master? Once you've done that, and the tests pass, we can merge this into master.

tvandijck:
on it.

Issue #319, opened 2015-10-23
Configuring output .pdb file


tapika:
I had following situation: One project contained same named project but different suffix - for example:
test.dll and test.exe - both ended up with same .pdb filename - that's test.pdb.

If both projects are compiled in same folder - they conflict for same program database file.

It should be somehow possible to configure output .pdb file or specifying to use file extension in pdb filename.

I think this needs to be configured:
<ClCompile>
...
<ProgramDataBaseFileName>$(OutDir)test_exe.pdb</ProgramDataBaseFileName>

and also this:

    <Link>
...
      <ProgramDatabaseFile>$(TargetDir)$(TargetName)_exe.pdb</ProgramDatabaseFile>

tapika:
<ClCompile>
    ...
    <ProgramDataBaseFileName>$(OutDir)test_exe.pdb</ProgramDataBaseFileName>

and also this:

    <Link>
    ...
      <ProgramDatabaseFile>$(TargetDir)$(TargetName)_exe.pdb</ProgramDatabaseFile>

tapika:
See following link: 

https://github.com/premake/premake-core/issues/151

As a quick walkaround for not able to configure .pdb file.

Issue #314, opened 2015-10-21
Configuring .pdb file for project....


tapika:
Originated from: http://industriousone.com/topic/configuring-pdb-file-project
and from here: https://bitbucket.org/premake/premake-dev/issues/237/configuring-pdb-file-for-project

Moved here so this would be fixed if possible.

PR #524, opened 2016-06-22
Adding 'symbolspath' API to allow specifying where the pdb file goes.

tvandijck:
This 'finally' allows us to all eat the cake and have it...

if you specify:
```lua
symbolspath '$(OutDir)$(TargetName).pdb'
```
you get your pdb's right next to your exe/dll/lib with the same name.
if you don't specify it, it maintains the default behavior that we've gone back and forth between.

Fixes: 
https://github.com/premake/premake-core/issues/319
https://github.com/premake/premake-core/issues/314
https://github.com/premake/premake-core/issues/151

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -957,6 +957,13 @@
 	}
 
 	api.register {
+		name = "symbolspath",
+		scope = "config",
+		kind = "path",
+		tokens = true,
+	}
+
+	api.register {
 		name = "sysincludedirs",
 		scope = "config",
 		kind = "list:directory",
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1827,8 +1827,9 @@
 
 
 	function m.programDataBaseFileName(cfg)
-		-- just a placeholder for overriding; will use the default VS name
-		-- for changes, see https://github.com/premake/premake-core/issues/151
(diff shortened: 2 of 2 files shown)


samsinsane:
LGTM, can you please flag all the issues that this fixes so they're closed on merge? :)

starkos:
Nice one!

tapika:
Good, seems to be working.

But is it possible to be able also to specify empty value - according to my post on forum 2 years ago - http://industriousone.com/topic/configuring-pdb-file-project - it should be possible to generate empty <ProgramDataBaseFileName></ProgramDataBaseFileName> xml tag. But currently it's not possible - using symbol path "" will result in <ProgramDataBaseFileName>.</ProgramDataBaseFileName>
which in a turn will result in compiler error:

fatal error C1033: cannot open program database ...

I'm still bit puzzled by myself why to set this path to empty value - but I guess I want to have debugging information for static library, but when integrating further into .exe - static library would not have .pdb file defined at all.

I also suspect that my next direction would be to get rid of separate static library projects in overall - include them as compilable source code / compilable project.

tapika:
Btw - that particular linker error ignoring is finally working in vs2015. Some of my projects already reaching vs2015, but mostly vs2010, vs2012, vs2013. vs2015 is a brand new vs which still have some of drawbacks.

`            -- Works only starting from vs2015.`
`            -- warning LNK4099: PDB 'lib.pdb' was not found with 'libeay32.lib`
`            linkoptions { "/IGNORE:4099" }`

PR #525, opened 2016-06-22
Adding /Debug:FastLink support for VS2015.

tvandijck:
and explicitly handle `symbols 'Off'`

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1305,7 +1305,25 @@
 
 
 	function m.generateDebugInformation(cfg)
-		m.element("GenerateDebugInformation", nil, tostring(cfg.symbols == p.ON))
+		local lookup = {}
+		if _ACTION >= "vs2015" then
+			lookup[p.ON]       = "Debug"
+			lookup[p.OFF]      = "No"
+			lookup["FastLink"] = "DebugFastLink"
+
+			if cfg.symbols == "FastLink" then
+				m.element("FullProgramDatabaseFile", nil, "true")
+			end
+		else
+			lookup[p.ON]       = "true"
+			lookup[p.OFF]      = "false"
+			lookup["FastLink"] = "true"
+		end
+
+		local value = lookup[cfg.symbols]
+		if value then
+			m.element("GenerateDebugInformation", nil, value)
+		end
 	end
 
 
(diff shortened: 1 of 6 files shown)


tests/actions/vstudio/vc2010/test_rule_vars.lua
```diff
@@ -100,11 +100,11 @@
 
 	function suite.onPathVar()
 		createVar { name="MyVar", kind="path" }
-		myRuleVars { MyVar = "../../path/to/file" }
+		myRuleVars { MyVar = "../path/to/file" }
```
tvandijck: this is actually a separate bugfix that I hit.. if you go up to far, the path becomes absolute, and stays absolute. so ../.. actually goes up beyond the premake folder itself, and hence in our case where it is in "c:\dev\premake", the patch would actually become "C:\path\to\file", which then subsequently stays absolute.

tests/actions/vstudio/vc2010/test_excluded_configs.lua
```diff
@@ -51,7 +51,6 @@
 		test.capture [[
 <Link>
 	<SubSystem>Console</SubSystem>
-	<GenerateDebugInformation>false</GenerateDebugInformation>
```
tvandijck: Default behavior is to not output this flag at all..

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1243,10 +1243,12 @@
 			else
 				value = "EditAndContinue"
 			end
+		elseif cfg.symbols == p.OFF then
+			value = ""
```
TurkeyMan: This isn't right; `symbols 'off'` will not emit any XML. In VS, the default for symbols is `ProgramDatabase`...
It should be `value = "None"`, and if `symbols` is never called, only then should the element not be written.

TurkeyMan: #541 fixes `symbols` it more properly. This PR should be split so the different fixes can be merged independently.

TurkeyMan: And #544 implements the symbols api properly...

tvandijck:
Rebased and updated to just do the FastLink option.

samsinsane:
We should have some tests for this new output, don't want it to get broken. Also, the VS2015 license seems to have expired for AppVeyor? Did you have to provide a license?

tvandijck:
> Also, the VS2015 license seems to have expired for AppVeyor? Did you have to provide a license?

I contacted AppVeyor already about that... this is on their end to fix unfortunately...

PR #526, opened 2016-06-30
Use cached file attributes during matching

mendsley:
When _DIRENT_HAVE_D_TYPE is defined, struct dirent defines a
d_type member that will indicate if the type of the entry. This
saves an expensive syscall for each entry which can really add
up on slow filesystems (such as VirtualBox's shared folder FS).

We use the same logic as do_isfile, and return a 1 when the entry
is not a directory, and a 0 otherwise.

For our large project this reduced the overall stat syscalls from
118,865 to 39,177 and total premake generation time from 59,977ms
to 20,729ms on vboxsf.

See: https://gist.github.com/mendsley/4b56932056d5c231a94d1cdefd15c027

### DIFF
--- a/src/host/os_match.c
+++ b/src/host/os_match.c
@@ -141,14 +141,26 @@ int os_matchname(lua_State* L)
 
 int os_matchisfile(lua_State* L)
 {
-	const char* fname;
-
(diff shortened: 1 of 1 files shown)


samsinsane:
LGTM

PR #527, opened 2016-06-30
Remove extra argument to os.match

mendsley:
This value will always be ignored as os.match only takes a single
argument and is not needed to generate correct results.


samsinsane:
LGTM, is this the only place this occurs?

mendsley:
It appears to be the only instance. It was discovered while investigating performance of os.match by adding an additional argument to the function.

PR #532, opened 2016-07-08
Rearranged --start-group --end-group to enclose only the user libs.

### LINKED ISSUE #531: --start/end-group
Starting a separate topic here to debate this specific behaviour.

My original code used to produce:
```
LIBS += -Wl,--start-group workspacelib1.a workspacelib2.a -Wl,--end-group -lsyslib1 -lsyslib2
```
The logic was that cross-dependencies may exist between the users libs within the workspace, and the system libs would appear after (ie, not having cross-dependencies).

The new behaviour with `linkgroups 'On'` produces:
```
LIBS += -Wl,--start-group workspacelib1.a workspacelib2.a -lsyslib1 -lsyslib2 -Wl,--end-group
```
And this new link line causes our build to fail for some reason. In our case, Qt libs are among those system libs, and they don't seem to like being bundled into the link group. We just see unresolved externals for all Qt symbols.

I think my original behaviour was correct. Is there a reason it was changed to wrap everything?

TurkeyMan:
resolve #531

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -239,8 +239,8 @@
 --    A list of libraries to link, decorated for the linker.
 --
 
-	function clang.getlinks(cfg, systemonly)
-		return gcc.getlinks(cfg, systemonly)
+	function clang.getlinks(cfg, systemonly, nogroups)
+		return gcc.getlinks(cfg, systemonly, nogroups)
 	end
 
 
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -276,7 +276,7 @@
 -- Return the list of libraries to link, decorated with flags as needed.
 --
 
-	function gcc.getlinksonly(cfg, systemonly)
+	function gcc.getlinks(cfg, systemonly, nogroups)
 		local result = {}
 
 		if not systemonly then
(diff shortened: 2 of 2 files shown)


samsinsane:
LGTM, is the `nogroups` parameter for modules to override the grouping?

starkos:
I do remember this logic being in the code—weird that it get stepped on.

I'm also a bit confused by the purpose of `no groups` when `cfg.linkgroups` is already there?

TurkeyMan:
Well, since I'm removing the function `getlinksonly` (which was never called incidentally), I figured I should retain functionality that function provided, which is to emit a link line but without the grouping tabs in it... maybe that's not useful (I wonder why `getlinksonly` existed)? I can remove it...?

TurkeyMan:
> is the nogroups parameter for modules to override the grouping?

Sure... sounds good! :)

TurkeyMan:
Resolves #531 (<-- does this work on github?)

TurkeyMan:
fixes #531 (where do you type this on github?!)

tvandijck:
> Well, since I'm removing the function getlinksonly (which was never called incidentally), 

not true.. before this commit we had this code:
```lua
function gcc.getlinks(cfg, systemonly)		
		local result = gcc.getlinksonly(cfg, systemonly)
```

it was introduces in a previous commit to try and reuse as much code as possible between gcc and clang... clang on Mac does not like the linkgroups in xcode 6.0, so we needed it to not be there.

Now, obviously as we moved forward, and found out that it is probably better to allow a user to switch it on/off with a proper API... this kind of got reverted somewhat. I guess when I did make the changes I never really realized that system libs should not be included. Our game builds fine with it, so it never was a problem either... 

Your changes however look fine to me, we should probably merge it, as it gets us back closer to the original, while still maintaining the new linkgroups api... so that will be sufficient for our use-cases.

starkos:
Where are we at with this one? Safe to merge?

TurkeyMan:
I think that was an LGTM from tom?

PR #536, opened 2016-07-11
Fix regression in expandtoken

### LINKED ISSUE #535: detoken got broke in december
PR #304 appears to be the source of my `cfg.platforms` related issues.
We have code in our scripts like this:
```lua
targetdir "lib/%{cfg.buildcfg}_%{cfg.platform}"
```
`%{cfg.platform}` used to evaluate to an empty string if `platforms()` was never called, but now it's `nil` and we crash.
This line https://github.com/premake/premake-core/pull/304/files#diff-99f714b3251ec87d2c34c37944ca560cR59 seems to change behaviour.  
Previously,
```
local result = func() or ""
```
coerces `result` to an empty string if it is `nil`.  
Now,
```
local success, result = pcall(func)
```
`result` remains `nil`.

TurkeyMan:
Coerced result to a string, as the old behaviour did.
fixes #535

### DIFF
--- a/src/base/detoken.lua
+++ b/src/base/detoken.lua
@@ -62,6 +62,7 @@
 				result = nil
 			else
 				err    = nil
+				result = result or ""
 			end
 
 			-- If the result is an absolute path, and it is being inserted into
@@ -90,7 +91,7 @@
 					success, result = pcall(result, e)
 					if not success then
 						return nil, result
-					end					
+					end
 				end
 
 				if (type(result) == "table") then


samsinsane:
LGTM, my only concern is that this functionality might be expected elsewhere? @starkos and @tvandijck should probably look at this before it's merged in, just so we avoid anymore issues with it. :)

tvandijck:
the problem is... we're going back and forth about these things.... 

The question is, is this "broken functionality" or a "broken usage pattern"... I think in this particular case the root cause is that %{cfg.platform} returns nil, which is in my opinion correct, since cfg.platform in the table is actually nil if you don't set the platform, just like any other value will be nil if you don't specify it. (It is easy to adjust for without needing this fix by using "%{cfg.platform or ''}"

This fix in my opinion breaks consistency, since it is returning a value that is not a true representation of what is actually in the configuration... In addition, if an API is a table or list... but is not set, then this fix will return an emtpy string instead of an empty table.. so that too is inconsistent.

just my $0.02 I guess.

starkos:
> (It is easy to adjust for without needing this fix by using "%{cfg.platform or ''}"

I think that since the tokens are intended to be used inline within a string, they should return an empty string if not set, a totally reasonable default value. Especially since the alternative is a crash!

Note that the value will still be nil while the token is being evaluated, so you can provide a different default if you like, i.e. this works even with the empty string default:

     "%{cfg.platform or 'Native'}"

> LGTM, my only concern is that this functionality might be expected elsewhere?

If you are accessing `cfg.platform` directly in code, I think it is reasonable to expect that it might be nil, and to handle it there. I don't think anyone will be surprised by this behavior.

tvandijck:
> Especially since the alternative is a crash!

Isn't that desirable though, instead of having phantom values and no error message to figure out where it came from. I know on my end I fixed a ton of incorrect token issues...

TurkeyMan:
> The question is, is this "broken functionality" or a "broken usage pattern"...

I argued on this with Sam too. We could have that discussion, but the problem is, the change in behaviour broke every single one of our in-the-wild scripts. I think a change that subtly crashes end-users scripts needs very strong reason. I don't think `%{cfg.platform}` is an uncommon pattern, it appears in directory/output names constantly. I'm actually surprised it doesn't appear anywhere within premake-core, but it certainly appears in basically every premake script I've ever written.

The source of this problem came as a big surprise to me. It took me and Sam quite a while to isolate the cause. Most users would not be able to track it down short of trial-and-error.

My 2c is that most premake users probably aren't Lua programmers. Pushing `%{cfg.platform or ''}` to the end user seems a bit rough to me. Most probably think of it more like looking up an environment variable, I doubt most users realise you can write executable lua expressions in the `%{ }`.
I've always worked under the assumption that the majority of users don't realise they're writing Lua code at all. Simple Premake scripts look more like some custom declarative language (ie, by omitting parens everywhere), and we recommend indentation patterns that break sensible lua indentation rules.

I agree with @starkos; these are string tokens intended to be embedded in strings. If `nil` isn't coerced to `''`, we crash. I don't think writing a string literal should be able to crash.

> Isn't that desirable though, instead of having phantom values and no error message to figure out where it came from. I know on my end I fixed a ton of incorrect token issues...

A crash occurring from inside a string literal is not something I think end-users would/should expect.
I'm not sure what you mean by 'phantom' values? Can you give some examples that you fixed where you think this behaviour was a problem?

starkos:
> I'm not sure what you mean by 'phantom' values? Can you give some examples that you fixed where you think this behaviour was a problem?

I think he means typos in the symbols, i.e. "%{cfg.pltform}" would fail silently.

There are pros and cons to each side, FWIW I'm in the default-value-don't-crash camp.

TurkeyMan:
So... this?

starkos:
Okay, I merging this to at least get things back to where they were, and not crashing 

> I know on my end I fixed a ton of incorrect token issues...

I seem to recall adding a flag to treat deprecated API usage as an error, to help someone weed out the old calls. I don't recall if that got merged into core or not. We could do something similar to cause nil tokens to print a warning, so you would at least have a list you scan to help pin down issues like that?

Issue #537, opened 2016-07-12
Path normalization in os.execute destroys URLs


nmoehrle:
The path normalization in the [overridden](https://github.com/premake/premake-core/blob/ab1b4edcb733d547be5b0ea6a2fc1e8a722bc4b5/src/base/os.lua#L13) `os.execute` function destroys URLs by [removing](https://github.com/premake/premake-core/blob/5bbc06f4b63120718e2e54755fdbaab3065ab9a0/src/host/path_normalize.c#L68) "duplicated" slashes e.g. `https://github.com -> https:/github.com`.

In my opinion this is an unexpected and undesired behavior but I am uncertain how to address this issue.
One could either exclude urls from path normalization or modify the path normalization ([normalize_substring](https://github.com/premake/premake-core/blob/5bbc06f4b63120718e2e54755fdbaab3065ab9a0/src/host/path_normalize.c#L12)) to keep "duplicated" slashes in case of a preceding colon. I cannot think of an example where this would break the normal path normalization but only makes sense URLs are considered paths...

PR #538, opened 2016-07-12
Fix path normalization for URLs

nmoehrle:
This is a possible solution for #537

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -11,7 +11,6 @@
 --
 
 	premake.override(os, "execute", function(base, cmd)
-		cmd = path.normalize(cmd)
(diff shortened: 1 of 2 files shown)


samsinsane:
I don't think this is the correct solution to the problem, look into #320 for the preferred solution. It needs some fixes, but it's pretty much good to go.

tvandijck:
@samsinsane if this is not the right fix, should we close it?

samsinsane:
Potentially, it depends on whether @nmoehrle wants to fix this up or not.

nmoehrle:
Sorry for the mess I am in the process of cleaning it up...

PR #541, opened 2016-07-14
symbols 'off' didn't work

TurkeyMan:
It turns out that `symbols "off"` just left symbols in the default state, which in VS, is on.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1247,6 +1247,8 @@
 			if value then
 				m.element("DebugInformationFormat", nil, value)
 			end
+		elseif cfg.symbols == p.OFF then
+			m.element("DebugInformationFormat", nil, "None")
 		end
 	end
 


TurkeyMan:
I saw that #525 includes my change, and I also prefer Tom's implementation.

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1247,6 +1247,8 @@
 			if value then
 				m.element("DebugInformationFormat", nil, value)
 			end
+		elseif cfg.symbols == p.OFF then
+			m.element("debugInformationFormat", nil, "None")
```
samsinsane: `DebugInformationFormat` not `debugInformationFormat`.

Also, I believe `None` was introduced in VS2015 so this will need to check for that.

TurkeyMan: You know, you typed this patch ;)

samsinsane: Not my best work, haha

TurkeyMan:
Closed in favour of #544

Issue #473, opened 2016-04-13
Disabling -L/usr/lib64 -L/usr/lib32 on macos


JohannesMP:
When compiling a gmake project configured for 32 or 64 bit on mac (tested on Mac OS 10.10.5) I always see this warning, when using either gcc or clang:

    ld: warning: directory not found for option '-L/usr/lib32'

and

    ld: warning: directory not found for option '-L/usr/lib64'
 
which is the result of [this code](https://github.com/premake/premake-core/blob/5bbc06f4b63120718e2e54755fdbaab3065ab9a0/src/tools/gcc.lua#L238-L241).

-----

Since mac os only has `/usr/lib` and never `/usr/lib32` or `/usr/lib64` (as far as I can tell), the code should be modified to account for that so that this warning no longer occurs.

tbelaire:
Here's some sed scripts to make it easier.

```
sed -i -- 's%-Wl,--end-group%%' *.make
sed -i -- 's%-Wl,--start-group%%' *.make
sed -i -- 's%-L/usr/lib64%%' *.make
```

samsinsane:
Closing as this issue should be resolved in master due to #543.

speeder:
I am using the newest version on the site, and this bug still exists.

PR #543, opened 2016-07-17
do not add architecture-dependent library directories on macOS

noresources:
fix issues #473 and #303

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -236,8 +236,20 @@
 
 	gcc.libraryDirectories = {
 		architecture = {
-			x86 = "-L/usr/lib32",
-			x86_64 = "-L/usr/lib64",
+			x86 = function (cfg)
+				local r = {}
+				if cfg.system ~= premake.MACOSX then
+					table.insert (r, "-L/usr/lib32")
+				end
+				return r
+			end,
+			x86_64 = function (cfg)
+				local r = {}
+				if cfg.system ~= premake.MACOSX then
+					table.insert (r, "-L/usr/lib64")
+				end
+				return r
+			end,
 		},
 		system = {
 			wii = "-L$(LIBOGC_LIB)",
--- a/tests/actions/make/cpp/test_ldflags.lua
+++ b/tests/actions/make/cpp/test_ldflags.lua
@@ -48,3 +48,30 @@
   ALL_LDFLAGS += $(LDFLAGS) -L../libs -Llibs
 		]]
 	end
+	
+	function suite.checkLibDirs_X86_64()
+		architecture ("x86_64")
+		system (premake.LINUX)
+		prepare()
+		test.capture [[
+  ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib64 -m64
+		]]
+	end
+	
+	function suite.checkLibDirs_X86()
+		architecture ("x86")
+		system (premake.LINUX)
+		prepare()
+		test.capture [[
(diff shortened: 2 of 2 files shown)


src/tools/gcc.lua
```diff
@@ -236,8 +236,20 @@
...
+				end
+				return r
+			end,
+			x86_64 = function (cfg)
+				local r = {}
+				if cfg.system ~= premake.MACOSX then
+					table.insert (r, "-L/usr/lib64")
+				end
+				return r
+			end,
```
G07cha: What if there will be single check in `architecture`?
This code is little bit smaller and should work the same

```lua
architecture = function(cfg)
    local r = {}
    if cfg.system ~= premake.MACOSX then
        r =  {
            x86 = "-L/usr/lib32",
            x86_64 = "-L/usr/lib64",
        }
    end
    return r
end
```

noresources:
@G07cha I just tried your suggestion and it seems that `config.mapFlags` does not accept functions at this level

G07cha:
@noresources okay then, I'm new to Lua so that was just my guess. Looks good then, can't wait until it's will be merged.

src/tools/gcc.lua
samsinsane: Why do these return lists and not strings, like `gcc.ldflags.kind.WindowedApp`? Which would look like this:
```lua
architecture = {
  x86 = function (cfg)
    if cfg.system ~= premake.MACOSX then return "-L/usr/lib32" end
  end,
  x86_64 = function (cfg)
    if cfg.system ~= premake.MACOSX then return "-L/usr/lib64" end
  end,
}
```

samsinsane:
Can you please write some tests to go with this change? It would be good to not break this in the future.

noresources:
Tests added.

src/tools/gcc.lua
tvandijck: Agreed... consistency would be preferable... could we update this PR to use what @samsinsane is suggesting?

noresources: Hello,
Using tables here was a bit of anticipation of what will finally be necessary some days. On my linux machine I have more than one arch-dependant path (`/usr/lib32`, `/usr/libx32`, `/usr/x86_64-linux-gnu/`, `/usr/arm-linux-gnueabi/` ...) and it may be different for other distros....

Moreover, there is currently no consistency in `config.mapFlags()` use. Some calls use strings, some use tables.

That was my reason to use tables here. However, if you still prefer strings, I will change it. Tell me

samsinsane: I guess it makes more sense to be a table, it allows modules to just override the function and assume a table. Either way, we need to figure out a way to not use a function every time an OS differs to the others. While it's a bit over the top, it would be nice if there was something like this:
```lua
architecture = {
  x86 = {
    default = { "-L/usr/lib32" },
    OSX = { }
  },
  x86_64 = {
    default = { "-L/usr/lib64" },
    OSX = { }
  }
}
```

This would allow us to quite easily support Clang on Windows (I'm assuming nmake does/will support Clang).

tritao: That looks nice, but if @noresources has no time to implement it right now I'd rather see this merged as-is, it's a minor feature and it would be nice to finally get rid of these warnings in our builds :)

samsinsane: Sorry, I wasn't very clear about what I meant, what @noresources has makes more sense than what I was suggesting, given the examples they've provided. If @tvandijck is happy with this, then we should be good to merge it in as is.

samsinsane:
LGTM

PR #545, opened 2016-07-21
Add generated files to vxcproj file.

tvandijck:
This is the original change list re-applied to HEAD....

Obviously, as per discussion, this thing needs more work before we can merge... but I just wanted to re-open the pull request.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -71,13 +71,18 @@
 			local node = files[filename]
 			if not node then
 				node = fileconfig.new(filename, cfg.project)
-				node.vpath = path.join("Generated", node.name)
-				node.dependsOn = source
-				node.generated = true
-
 				files[filename] = node
 				table.insert(files, node)
 			end
+
+			-- always overwrite the vpath and dependency information.
+			if not node.vpath then
+				node.vpath = path.join("Generated", node.name)
+			end
+			node.dependsOn = source
+			node.generated = true
+
+			-- add to config.
 			fileconfig.addconfig(node, cfg)
 		end
 
@@ -88,14 +93,14 @@
 			end
 
 			if fileconfig.hasCustomBuildRule(filecfg) then
-				local buildoutputs = p.project.getrelative(cfg.project, filecfg.buildoutputs)
+				local buildoutputs = filecfg.buildoutputs
 				if buildoutputs and #buildoutputs > 0 then
 					for _, output in ipairs(buildoutputs) do
-						addGeneratedFile(cfg, node, output)
+						if not path.islinkable(output) then
+							addGeneratedFile(cfg, node, output)
+						end
 					end
 				end
-			else
-				--addRuleFile(cfg, node)
 			end
 		end
 
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -30,7 +30,6 @@
 	m.elements.project = function(prj)
 		return {
 			m.xmlDeclaration,
-			m.updateFileTable,
 			m.project,
 			m.projectConfigurations,
 			m.globals,
@@ -55,66 +54,6 @@
 		p.utf8()
 		p.callArray(m.elements.project, prj)
 		p.out('</Project>')
-	end
-
-
---
--- Update the file table with generated files.
---
-	function m.updateFileTable(prj)
-		local function addGeneratedFile(cfg, source, filename)
-			-- mark that we have generated files.
-			cfg.project.hasGeneratedFiles = true
-
-			-- add generated file to the project.
-			local files = cfg.project._.files
-			local node = files[filename]
-			if not node then
-				node = fileconfig.new(filename, cfg.project)
-				files[filename] = node
-				table.insert(files, node)
-			end
-
-			-- always overwrite the vpath and dependency information.
-			if not node.vpath then
-				node.vpath = path.join("Generated", node.name)
-			end
-			node.dependsOn = source
-			node.generated = true
-
-			-- add to config.
-			fileconfig.addconfig(node, cfg)
-		end
-
-		local function addFile(cfg, node)
-			local filecfg = fileconfig.getconfig(node, cfg)
-			if not filecfg or filecfg.flags.ExcludeFromBuild then
-				return
-			end
-
-			if fileconfig.hasCustomBuildRule(filecfg) then
-				local buildoutputs = filecfg.buildoutputs
-				if buildoutputs and #buildoutputs > 0 then
-					for _, output in ipairs(buildoutputs) do
-						if not path.islinkable(output) then
-							addGeneratedFile(cfg, node, output)
-						end
-					end
-				end
-			end
-		end
-
-		local files = table.shallowcopy(prj._.files)
-		for cfg in project.eachconfig(prj) do
-			table.foreachi(files, function(node)
-				addFile(cfg, node)
-			end)
-		end
-
-		-- we need to reassign object sequences if we generated any files.
-		if prj.hasGeneratedFiles and p.project.iscpp(prj) then
-			p.oven.assignObjectSequences(prj)
-		end
 	end
 
 
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -30,6 +30,7 @@
 	m.elements.project = function(prj)
 		return {
 			m.xmlDeclaration,
+			m.updateFileTable,
 			m.project,
 			m.projectConfigurations,
 			m.globals,
@@ -56,6 +57,60 @@
 		p.out('</Project>')
 	end
 
+
+--
+-- Update the file table with generated files.
+--
+	function m.updateFileTable(prj)
+		local function addGeneratedFile(cfg, source, filename)
+			-- mark that we have generated files.
+			cfg.project.hasGeneratedFiles = true
+
+			-- add generated file to the project.
+			local files = cfg.project._.files
+			local node = files[filename]
+			if not node then
+				node = fileconfig.new(filename, cfg.project)
+				node.vpath = path.join("Generated", node.name)
+				node.dependsOn = source
+				node.generated = true
+
+				files[filename] = node
+				table.insert(files, node)
+			end
+			fileconfig.addconfig(node, cfg)
+		end
+
+		local function addFile(cfg, node)
+			local filecfg = fileconfig.getconfig(node, cfg)
+			if not filecfg or filecfg.flags.ExcludeFromBuild then
+				return
+			end
+
+			if fileconfig.hasCustomBuildRule(filecfg) then
+				local buildoutputs = p.project.getrelative(cfg.project, filecfg.buildoutputs)
+				if buildoutputs and #buildoutputs > 0 then
+					for _, output in ipairs(buildoutputs) do
+						addGeneratedFile(cfg, node, output)
+					end
+				end
+			else
+				--addRuleFile(cfg, node)
+			end
+		end
+
+		local files = table.shallowcopy(prj._.files)
+		for cfg in project.eachconfig(prj) do
+			table.foreachi(files, function(node)
+				addFile(cfg, node)
+			end)
+		end
+
+		-- we need to reassign object sequences if we generated any files.
+		if prj.hasGeneratedFiles and p.project.iscpp(prj) then
+			p.oven.assignObjectSequences(prj)
+		end
+	end
 
 
 --
@@ -575,6 +630,15 @@
 	end
 
 
+	function m.generatedFile(cfg, file)
+		if file.generated then
+			local path = path.translate(file.dependsOn.relpath)
+			m.element("AutoGen", nil, 'true')
+			m.element("DependentUpon", nil, path)
+		end
+	end
+
+
 ---
 -- Write out the list of source code files, and any associated configuration.
 ---
@@ -598,7 +662,7 @@
 		priority   = 1,
 
 		emitFiles = function(prj, group)
-			m.emitFiles(prj, group, "ClInclude")
+			m.emitFiles(prj, group, "ClInclude", {m.generatedFile})
 		end,
 
 		emitFilter = function(prj, group)
@@ -638,7 +702,7 @@
 				end
 			end
 
-			m.emitFiles(prj, group, "ClCompile", nil, fileCfgFunc)
+			m.emitFiles(prj, group, "ClCompile", {m.generatedFile}, fileCfgFunc)
 		end,
 
 		emitFilter = function(prj, group)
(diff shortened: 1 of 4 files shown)


tvandijck:
OK, so I fixed two bugs in this change...
- object files or other 'linkable' files are no longer added.
- turns out files in the _.files table are absolute paths, so what ended up happening is that in some cases files got added twice, and paths were 'resolved' with 'getrelative' twice... which obviously breaks in all kinds of ways..
- also, I now always mark the file as 'generated' and setup the dependency, regardless of whether the file was added by this method, or already added by the normal 'files {....}' keyword.


if anyone that had trouble with this stuff could try things out, give feedback, test it... suggestions.. post them here... be blunt ;)

starkos:
Thanks, Tom! I have a good project tree to use for testing this, but I need to get it caught up to the latest master branch. I'll be working on that this week and will try to get feedback to you ASAP.

It sounds like I was seeing different issues than Manu & Sam though, so hopefully they will chime in as well.

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -57,6 +58,60 @@
...
+
+			if fileconfig.hasCustomBuildRule(filecfg) then
+				local buildoutputs = p.project.getrelative(cfg.project, filecfg.buildoutputs)
+				if buildoutputs and #buildoutputs > 0 then
+					for _, output in ipairs(buildoutputs) do
+						addGeneratedFile(cfg, node, output)
+					end
+				end
+			else
+				--addRuleFile(cfg, node)
```
dcourtois: what's this ?

dcourtois:
Hi,

What about introducing a few api commands to control those mechanisms ? As I see it, 2 commands would be perfect : one that defines the output folder for those files. It can be optional, in which case the default one would be used. But I think letting the user choose where he/she wants to put his/her generated files is important.

And the second would be a command to tell Premake to explicitly add the generated files. Without this, at least my addon won't be able to work properly since I need to take over the generated files. And also from my experience, it's always better to be explicit than implicit, even if it adds a bit of verbosity.

Does it seem reasonable ?

samsinsane:
> It sounds like I was seeing different issues than Manu & Sam though, so hopefully they will chime in as well.

I don't really know that much about the issue, my project doesn't generate files like Manu's does. I'll give him a nudge to comment on this tomorrow so it doesn't get held up for too long. I think most of his issues are covered by @dcourtois though, but I'll get him to verify this.

tvandijck:
> As I see it, 2 commands would be perfect : one that defines the output folder for those files.

```buildoutputs``` already defines where the outputs go, I don't modify that in any way.

> And the second would be a command to tell Premake to explicitly add the generated files. 

we can definitely switch this on/off. We actually used to have that here, and that caused quite a discussion here as well... people thought is was confusing, didn't have clear behavior, etc..
if anyone has a good name for this, I can certainly add that...

we had it as "buildoutputsasinputs", but this is a flag that would only affect Visual Studio, since in Xcode this is already automatic, and that is very little you can do about that. Hence I opted for similar behavior in Visual Studio.

> Without this, at least my addon won't be able to work properly since I need to take over the generated files

With my latest changes, if the generated files already exist in the files table, all I do is set the 'generated' flag and the 'dependson' flag... no other changes to the file tree are made. The "dependson" behavior is something we could potentially turn on/off, since that really only has a visual effect in the Solution Explorer, which may not always be desirable. The 'generated' flag has no functional effect on the visual studio solution.

**Comment by starkos** [comment]
> And the second would be a command to tell Premake to explicitly add the generated files. 

Couldn't your module just scan the list of files for those marked with the `generated` flag?

**Comment by dcourtois** [comment]
Hi, I've integrated this pull request in my repository to check how it behaves. Here are a quick result, taken from one of my projects. The project uses my Qt module to handle Qt specific files (*.qrc, and *.h that need to run through the `moc` precompiler, all of which generating compilable files) It also uses a few custom rules which don't generate compilable files.

Here is how it appears before, and after the PR :

pr_before

pr_after

Notice the qmldir file : it's added through the `files` command, but it doesn't generate any compilable files, it's just copied to the output folder. After the PR, qmldir is duplicated, and the Generated folder is no longer created. Here is a link to the project (rename one of the 2 Premake executables at the root level, or copy you own if you want to test) :

[MediaViewer.zip](https://github.com/premake/premake-core/files/383391/MediaViewer.zip)

I will try to see what's happening today, and report back here.

**Comment by dcourtois** [comment]
Ok, I think I know why the generated files get added twice : if I log the files in the original `oven.bakeFiles` function, all files are using an absolute path with slash path separators. But if I log `filename` in the `addGeneratedFile` function (line 65 in `vs2010_vcxproj.lua` some files (the ones which have buildoutputs) use absolute paths, but with backslashes, and some other (the ones added by my module) use relative paths, with slashes.

This is why the `addGeneratedFile` is not capable of detecting already generated files correctly. I will investigate further to see why this is happening.

**Comment by dcourtois** [comment]
Ok, so I have an almost working solution. A part of the solution is fixing the `addGeneratedFile` function :

```
			if not node then
				node = fileconfig.new(filename, cfg.project)
				files[filename] = node
				table.insert(files, node)

				-- use a default vpath if there wasn't a user provided one
				if not node.vpath then
					node.vpath = path.join("Generated", node.name)
				end

				-- add to config.
				fileconfig.addconfig(node, cfg)
			end

			node.dependsOn = source
			node.generated = true
```

In my case, it was the call to `fileconfig.addconfig(node, cfg)` that was causing problems. Also, I don't know what `dependsOn` is, but it feels really close to what the `buildinputs` api command does. Isn't it redundant ?

The other part of the fix was making my module generate `buildoutputs` with absolute paths that match the paths used in `oven.bakeFiles` (e.g. absolute and slash separator) to the test `if not node then` works correctly.

That leads me to think that it's not the most robust way of doing this test as `buildoutputs` is entirely up to the users.

Finally, there is one remaining problem with this function : if the output file path contains specials paths (like $(TargetDir), etc.) then the file gets added twice. For instance, I have this rule in my project :

```
	-- copy the qmldir file to the target dir
	filter "files:**/qmldir"
		buildmessage "Deploying %{file.name}"
		-- I need to convert this to the new copy tokens
		buildcommands { "copy \"%{file.relpath:gsub('/', '\\')}\" \"%{cfg.buildtarget.directory:gsub('/', '\\')}\\%{file.name}\"" }
		buildoutputs { "%{cfg.buildtarget.directory}/%{file.name}" }
```

which generates the following output filename : `$(TargetDir)/qmldir`. Which obviously fails the test in `addGeneratedFile`, so it's added again.

I don't have any easy fix for this last problem though ...

**Comment by starkos** [comment]
What is the reason for the "Generated" vpath again? We normally use the source file's file system location as the default path within the project tree; is there a reason that we can't do the same for generated files?

**Comment by dcourtois** [comment]
I guess that it was to cleanup the project tree : usually the output of precompilers are sent to some folder outside the source tree to avoid recursive inclusion (if you have `files { Sources }` and generate other files into `Sources`, next time you run Premake, you end up with a big problem :p) or various other problems.
And doing so usually result in some pretty uggly filter hierarchy in Visual Studio (XCode too probably)

And since vpaths only works on files added through the use of `files`, adding one in the default `addGeneratedFile` is probably the only easy way to do that for generated files.

(that's speculations based on the problems I encountered when writing my modules)

**Comment by tvandijck** [comment]
Thanks for the test case, I'll have a look at it as well....

Anyway, as for the relative paths vs absolute paths in buildoutputs... those are actually required to be absolute paths. this is how it is processed in vs2010_vcxproj.lua... if those paths are relative, then they are going to be 'double' relative after this function.... When your project happens to be generated in exactly the same folder as the lua file, this always ends up being exactly the same answer, but if you specify that the vcxproj goes to a completely different folder, this may completely break. 

	function m.buildOutputs(fcfg, condition)
		local outputs = project.getrelative(fcfg.project, fcfg.buildoutputs)
		m.element("Outputs", condition, '%s', table.concat(outputs, ";"))
	end

same is absolutely true for the _.files table in prj... files in that table should all be absolute paths, and using forward slashes.. if not, other things in premake will actually break as well...

buildoutputs containing "$(somemacro)" or "%(somemacro)", we should probably just exclude those, and just not add those... it can't work, even visual studio doesn't know what to do with it. Ultimately the best way to solve this would be using a custom props/target file, and actually do the processing of the compiled output in msbuild, and not let premake try to figure it out if that pattern is a requirement. Using the "custom rules" you can already accomplish that actually..

**Comment by tvandijck** [comment]
```lua
buildoutputs { "%{cfg.buildtarget.directory}/%{file.name}" }
```
which generates the following output filename : $(TargetDir)/qmldir.

that is because we allow the vcxproj action to replace tokens with visual studio tokens..
if you wrote:
```lua
buildoutputs { "%{ cfg.buildtarget.directory}/%{ file.name}" }
```
you don't actually get that, and premake will just resolve the path for you.
from a project point of view absolutely no difference, just not as 'pretty' I guess..

Personally I prefer the premake tokens to just resolve, and not use magic visual studio tokens, but that is obviously entirely a personal preference.

tvandijck:
The other suggestion that @starkos made was to do the "addGeneratedFiles" part in the oven instead of in the action... which might make it easier to resolve the absolute/relative paths a little easier maybe?

starkos:
> if you have `files { Sources }` and generate other files into Sources, next time you run Premake, you end up with a big problem

Premake ought to be able to figure that out though, shouldn't it? It has all the information it needs to do so, anyway.

starkos:
(…this is why I didn't tackle this feature at the time…decidedly non-trivial…many thanks for taking it on!)

dcourtois:
Isn't it possible to modify the `buildoutputs` command to always store absolute paths using forward slashes, if this is the correct behavior anyways ? That would prevent user from unintentionally using backward slashes, or relative paths, and solve the problem of the tokens.

Maybe moving the `addGeneratedFiles` logic to the oven could work to, but I have the feeling that it would be harder to ensure it works well with modules that override stuff in the oven. And you still wouldn't guarantee the users won't use backward slashes or relative paths on `buildoutputs`, and break things.

tvandijck:
```lua
	api.register {
		name = "buildoutputs",
		scope = { "config", "rule" },
		kind = "list:path",
		tokens = true,
		pathVars = true,
	}
```

it should already be that way, because the kind is "path"... unless I'm misunderstanding how that works.

dcourtois:
> it should already be that way

It is currently resolving tokens, replacing `cfg.buildtarget.directory` by `$(TargetDir)`, which is what makes thing break.

What I meant was : if we prevent this behavior (replacing tokens by vs macros) and just replace the token by the absolute path, then everything will work. But I realize that it's not a good idea.

Instead, what about checking for those macros in the `addGeneratedFile` function ? Something along the lines of :

```lua
    -- at the beginning of addGeneratedFile
    filename = vstudio.expandMacros(filename)
```

with `vstudio.expandMacros` looking something like this :

```lua
    function vstudio.expandMacros(str)
        for token, value in pairs(vstudio.pathVars) do
            -- I don't know the detoken part well enough, so I left the parameters empty,
            -- but you get the idea :)
            str = string.gsub(str, value.token, p.detoken.expand(token, ...))
        end
        return str
    end
```

dcourtois:
Small update : I managed to modify `addGeneratedFile` so that it will be able to resolve Visual Studio macros such as `$(TargetDir)`, but I then noticed that it doesn't solve my case :

- `qmldir` is included in the project using `files { "**/qmldir" }`, so its path is the absolute path to the source file.
- `buildoutputs` is the result of the copy, so its path is the absolute path of the destination file, which is different from the source one.

`addGeneratedFile` will add qmldir again :)

So, I see 2 possibilities :
- only add "buildable" generated files (e.g. compilation units) Potential problem : some projects might need to include non-buildable generated files, for embedding or some other things that I might not be aware of, so I'm not really comfortable with this solution.
- make this whole mechanism explicit. Only problem is finding a good name.

I really think the only robust way of handling this would be to have an api command to add generated files explicitly, like @tvandijck had (`buildoutputsasinputs` it was)

tvandijck:
Yeah, the more I think about this, the more the 'explicit' way appeals to me... the problem is the name. something like "compilebuildoutputs" is probably a better name, since it describes what it does. And if we do that, we can easily add this to the oven instead, which would be a lot easier to handle in the xcode/vcxproj/etc backends...

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -59,66 +58,6 @@
...
-
-			-- always overwrite the vpath and dependency information.
-			if not node.vpath then
-				node.vpath = path.join("Generated", node.name)
-			end
-			node.dependsOn = source
-			node.generated = true
-
-			-- add to config.
-			fileconfig.addconfig(node, cfg)
```
dcourtois: I think this line should go inside the `if not node then` : if the file already exists, its file config should also already be there, and adding it again might cause problems.

dcourtois:
Tested, and as expected, works fine :) There's just the little concern about the call to `fileconfig.addconfig` (see my comment) but now everything should behave nicely ! Thanks !

src/actions/vstudio/vs2010_vcxproj.lua
tvandijck: Ah yes... you are totally right... we should definitely test for that, and only add it if it's not already there... I'll look into that.

tvandijck:
@starkos are you OK merging this one now?? it's now totally optional behavior...

starkos:
Would it be possible to process the object file sequence numbers *after* the generated files are handled, so we don't have to do it twice? That can be a relative expensive operation in projects with many files.

I think you may have already investigated this, but if `compilebuildoutputs` is not set at all (nil) would it be possible to look at the outputs of the build rule, and `if project.iscpp(prj) and path.iscpp(buildOutput)` treat it as if it were set, and add the files to the build? So `compilebuildoutput` becomes an optional override for the cases where the automatic detection doesn't work?

In the same way, if the build output is linkable, we could add it to the link step.

tvandijck:
> Would it be possible to process the object file sequence numbers after the generated files are handled, so we don't have to do it twice?

I'll have a look at that... but I believe there was some kind of reason for it... can't remember off the top of my head though.

> I think you may have already investigated this, but if compilebuildoutputs is not set at all (nil) would it be possible to look at the outputs of the build rule, and if project.iscpp(prj) and path.iscpp(buildOutput) treat it as if it were set,

certainly an option, although that was mostly what we started with.. Initially I just added any file, and premake will just correctly setup cpp files as ClCompile, and all others as 'None'... I think keeping this behavior explicit is probably wisest at this stage, but I'm open to try other options.

> In the same way, if the build output is linkable, we could add it to the link step.

Interestingly, .obj files are already automatically added in visual studio,  it's very inconsistent in my opinion... The behavior of generated files from custom build steps and custom build rules is very inconsistent throughout all the IDE's I've so far encountered (codelite, xcode, vs), the only one that consistently doesn't do anything is make.

starkos:
> Initially I just added any file, and premake will just correctly setup cpp files as ClCompile, and all others as 'None'... I think keeping this behavior explicit is probably wisest at this stage, but I'm open to try other options.

You're probably right, and I'm probably the one who didn't like the magic behavior in the first place. Disregard my suggestion.

PR #546, opened 2016-07-21
Update curl and use mbedtls instead of openssl.

tvandijck:


samsinsane:
This seems like it should be good to merge. Are you able to throw some tests in for HTTPS connections? It doesn't look like we have any, and there have been a number of changes to the HTTP functions which haven't really been verified with any automated tests.

tvandijck:
I will add tests, the problem is that this is still considered behavior we can disable with the --no-http build option... so I got to figure out how to make these tests optional or something, but I will certainly do that this week..

tritao:
We could just permanently enable the HTTP support. Is there any one who actually needs Premake builds without HTTP support?

ratzlaff:
@tritao Where I work, we can only use the version of premake compiled without http support.

tritao:
@ratzlaff I see. Any specific technical reason?

tvandijck:
I think one of the reasons I've seen posted here before is that on different versions of linux different versions of libssl were linked in, and it would cause load failures saying things like "unable to load libssl.6.1.so" or something along those line... For example when you build premake on centos 5, and then try to use it on centos 7, it won't work because of OpenSSL.

This commit would actually fix that, since mbedtls is linked statically.. 

but there could be other reasons obviously...

TurkeyMan:
That's the reason for us. We just need to change the link to use a less-specific lib though and that should be fine.

ratzlaff:
@tritao our IT department is concerned with a number of things relating to approving third-party tools. A couple of items that they check for are:

1. is there a tool already approved that performs the same function.
 - has not been an issue. CMake is not widely used here - there are a number of hand crafted makefiles still in use

2. does it utilize network connections
 - has not been an issue because of --no-http

If premake were to start requiring the curl (and the associated ssl library), I would have to provide justification that a 'makefile generation tool' needs to make https connections.

For example:
notepad++ is approved because they permanently disabled the 'Update automatically' feature.
They configured a special setup.exe for installing cygwin that only downloads cygwin packages off of the shared network folders.

tritao:
Makes sense, thanks for explanation. Then I guess getting rid of the option is out of question since some folks need it.

starkos:
> I got to figure out how to make these tests optional or something

Wrap them in a `if http.get ~= nil then … end` block?

Also—great feedback everyone, thanks so much for keeping an eye on these issues, and for speaking up.

tvandijck:
Added tests...

PR #547, opened 2016-07-22
Contrib libraries as submodules.

tvandijck:
DO NOT MERGE YET..... I've only tested this on windows... not on linux & mac yet.


tvandijck:
I did not do LUA, because it seems that there is actual local changes specific to premake to lua itself... in particular the lua_pclose method has been modified... I never actually noticed this before, and we have internally upgraded to lua 5.3.2, and just have used it without modification...

Not sure what the purpose of those changes are, and why we need them... a discussion for another day.

starkos:
> Not sure what the purpose of those changes are, and why we need them

Lua's `pclose` did not return the command result to the caller previously, with this modification it does. If you don't use that feature, you wouldn't notice.

The biggest changes though are in `host/lua_auxlib.c`, where we replace `luaL_loadfile()` and do a bunch of magic to enable the script search paths, running embedded scripts, the `_SCRIPT` global, and so on. Pretty sure Premake won't run without that stuff.

tvandijck:
I think we could similarly 'patch' the pclose btw... instead of modifying the code... just let the io.pclose point to our replacement... I think that would also work better in cases where people (like us) do upgrade to a later version of lua...

There where a bunch of other changes however... so to just sum up everything significant I'm seeing against the 5.1.4 tag in the lua/lua repository:

* In lauxlib.c on line 86, there is 
```c
#ifdef _MANAGED
	#pragma managed(push, off)
#endif
```
added around luaL_error. I think this is just legacy against an old version of Visual Studio.

* In liolib.c on line 108, there is the modified io_pclose.
I think we can accomplish this by overriding it, rather then modifying it.

* In luaconf.h there is a bunch of macros that are defined in git, but undefined in premake.
```c
line 336: #undef LUA_COMPAT_VARARG
line 343: #undef LUA_COMPAT_MOD
line 358: #undef LUA_COMPAT_GFIND
line 366: #undef LUA_COMPAT_OPENLIB
```
Then on line 671, we have a couple of pclose related macros that have been modified.

That's it as far as I can tell...

starkos:
> I think we could similarly 'patch' the pclose btw... instead of modifying the code

That would be great! But I think we had to modify the code to make it return the command result, because it wasn't available to the Lua environment at all. Note that it is possible that this was changed in Lua 5.2 or 5.3, and might be a reason for an upgrade if it helps maintainability.

tvandijck:
OK, I will investigate once I get this working on linux & mac and this is merged.. 
is there a unit-test for this behavior of pclose?

starkos:
> is there a unit-test for this behavior of close?

I didn't see one in a quick search of the code

tvandijck:
OK, latest commit fixes things for linux/mac and win32, should be safe to merge if this is what we want...
Lua, can be done later...

tvandijck:
Not liking this change at all.... we should merge the autoconf first, and fix the libzip/curl libraries to correctly use it... then we can revisit this.

PR #549, opened 2016-07-22
Put back the old Symbols flag in premake5.lua

starkos:
Using brand new APIs in the `premake5.lua` script prevents people from regenerating the project files while running an older version of Premake. We need to give everyone a chance to get caught up before we use these new APIs in the build scripts.

Roll back to the old `flags { "Symbols" }` approach while we let the new `symbols()` API settle.


tvandijck:
LGTM, makes sense.

PR #550, opened 2016-07-23
Adding Experimental AutoConfig tools.

tvandijck:
Allows you to do stuff like:

``` lua
    autoconfigure {
        ['config.h'] = function (cfg)
            check_include(cfg, 'HAVE_STDIO_H', 'stdio.h')
            check_include(cfg, 'HAVE_FOOBAR_H', 'foobar.h')

            check_type_size(cfg, 'SIZEOF_SIZE_T', 'size_t')
        end
    }
```

which will generate a 'config.h' file with:

``` c
#define HAVE_STDIO_H 1
#define HAVE_SIZEOF_SIZE_T 1
#define SIZEOF_SIZE_T 8
```

on most 64bit platforms.

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -302,3 +302,34 @@
 	function clang.gettoolname(cfg, tool)
 		return clang.tools[tool]
 	end
+
+
+
+--
+-- Compile a string of text with the clang compiler.
+--
+-- @param cfg
+--    The project configuration.
+-- @param text
+--    The source text.
+-- @return
+--    on success the filename of the output object file.
+--    nil on failure.
+--
+
+	function clang.try_compile(cfg, text)
+		-- write the text to a temporary file.
+		local cppFile = path.join(cfg.autoconfdir, "temp.cpp")
+		if not io.writefile(cppFile, text) then
+			return nil
+		end
+
+		local outFile = path.join(cfg.autoconfdir, "temp.out")
+
+		-- compile that text file.
+		if os.execute('clang "' .. cppFile .. '" -o "' .. outFile ..'" &> /dev/null') then
+			return outFile
+		else
+			return nil
+		end
+	end
--- a/src/_manifest.lua
+++ b/src/_manifest.lua
@@ -56,6 +56,7 @@
 		"base/validation.lua",
 		"base/premake.lua",
 		"base/help.lua",
+		"base/autoconf.lua",
 
 		-- tool APIs
 		"tools/dotnet.lua",
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -48,6 +48,18 @@
 	}
 
 	api.register {
+		name = "autoconfigure",
+		scope = "config",
+		kind = "table:keyed"
+	}
+
+	api.register {
+		name = "autoconfdir",
+		scope = "config",
+		kind = "path"
+	}
+
+	api.register {
 		name = "basedir",
 		scope = "project",
 		kind = "path"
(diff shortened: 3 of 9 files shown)


tvandijck:
I originally made this as a module, but premake5 itself with the inclusion of curl & libzip should really use this facility to stay portable across multiple platforms, rather then just x86/x86_64 and a super recent version of ubuntu.. Some of the 'hard-coded' configs in curl and libzip are just flat out wrong for many operating systems and platform.. (centos5, arm, iphone, etc).

src/tools/clang.lua
```diff
@@ -289,3 +289,34 @@
...
+--    The project configuration.
+-- @param text
+--    The source text.
+-- @return
+--    on success the filename of the output object file.
+--    nil on failure.
+--
+
+	function clang.try_compile(cfg, text)
+		-- write the text to a temporary file.
```
tritao: This shares most of the code with GCC version below, might be a good way to share their code in a common function.

tritao:
I've always wished something like this on Premake, it might be the only feature I miss in comparison to other tools like CMake. The only suggestion I have at the moment is that maybe the code could figure out the defines name (ie `HAVE_STDIO_H`) automatically from the includes and type names (iirc Autotools works like that by default).

starkos:
My resistance to this feature has definitely softened, on the condition that is only an available feature, and not a required part of Premake's workflow.

I still hate (not too strong a word here) it when developers default to autoconf-style testing instead of just looking at the perfectly good flags provided by the compiler. Curl and zlib are no exception; pretty much all of those switches could be easily deduced without autoconf, but it would a lot of work at this point to fix them all. Boo.

tvandijck:
Yeah, I'm definitely with you on that @starkos.... in my curl/libzip/zlib -> submodule patch, you see I basically just handcrafted a config.h that does just that.... However, there is a few you can't reliably create defines for... the SIZEOF_SIZE_T for example, you can't really do it with defines, although you can get very far with detecting for 32-bit vs 64-bit and assume that sizeof(size_t) is sizeof(void*), but this is unfortunately on some weird platform not always the case.
And the same is true for checking the existance of a header file... it just can't be done without autoconf... But I agree that in many cases code written that relies on autoconf is just unneeded...

in the case of libzip & curl, I simply don't get why a define like SIZE_OF_SIZET is even needed.. just define your own types like zlib does, and **know** the sizes...

That said, this PR needs work, I somewhat followed the 'CMake' convention, i.e. not trying to automatically figure out define names, as @tritao suggests... however, I can go in any direction, and I would really like to see suggestions, and maybe even PR's on this pull request for further improvements before we consider merging this into master.

src/tools/clang.lua
tvandijck: I will have a look at that...

tvandijck:
On other thing is that this will just generate a header file with all defines you ever checked for in that configuration... just a flat list of everything... tools like autoconf/cmake actually take an input file and transform it, only outputting those defines that are in a particular "file.h.in". I didn't bother with that really, but if something like that is desired, someone else could possibly add that as far as I'm concerned.

data-man:
Very useful PR, thank you!
Just an idea: save/load a cache to/from disk.

starkos:
Not sure where to go with this one. My instinct would be to push back and say it should be packaged as a external module, at least to start?

I'm not opposed to shipping an autoconf script with Premake to help with bootstrapping, if the current bootstrap makefile isn't sufficient.

tvandijck:
I think I already published it as a module at some point.

https://github.com/Blizzard/premake-autoconf

PR #553, opened 2016-07-29
Add support for kind="list:path" in rule properties.

tvandijck:


src/actions/vstudio/vs2010_rules_props.lua
```diff
@@ -154,9 +154,9 @@
 
 
 	function m.dependsOn(r)
-    	p.w('<%sDependsOn', r.name)
-    	p.w('  Condition="\'$(ConfigurationType)\' != \'Makefile\'">_SelectedFiles;$(%sDependsOn)</%sDependsOn>',
-    		r.name, r.name, r.name)
+		p.w('<%sDependsOn', r.name)
+		p.w('  Condition="\'$(ConfigurationType)\' != \'Makefile\'">_SelectedFiles;$(%sDependsOn)</%sDependsOn>',
+			r.name, r.name, r.name)
```
samsinsane: Interesting, it looks like there is one too many `r.name`s here.

tvandijck: You are right, but I only changed whitespace here... I think my 'editorconfig' plugin fixes tabs/spaces as I save files.

luckily in lua extra arguments are pretty harmless, although not very neat ;)

PR #554, opened 2016-07-30
Add an 'uncached' sub-context to the context.

tvandijck:
OK, so this is part one of a 2-part feature I'm working on...

What this does, is it allows you to access any property like normal, but uncached...
normally when you ask for a property, for example

```lua
local value = cfg.buildoutputs
```
The buildoutputs will be expanded, and stored in the table. The next time some piece of code asks for the buildoutputs from the same context (cfg in the above example), it'll just return the previously expanded value, without running the detokenizer again...

This PR, will allow you to write:
```lua
local ctx = p.context.uncached(cfg)
local value = ctx.buildoutputs
```
which will neither ask for the 'potentially' already expanded value, not store the result of the expansion. Essentially forcing a detoken anytime you access it this way.

So why do I want this?
Well, the 2nd part of what I'm working on, will allow you to 'add' custom behavior to the detokenizer. For example, in the [Custom Rules](https://github.com/premake/premake-core/wiki/Custom-Rules) feature, 'propertydefinitions' currently have no standard way of being references as tokens, and we can't just say

```lua
propertydefinition {
    name = "StripDebugInfo",
}
buildcommand 'MyCustomCC.exe -c "%(FullPath)" -o "%(IntDir)/%(Filename).obj" %{StripDebugInfo}'
```

Instead we've somewhat made this a Visual Studio only feature and use the [StripDebugInfo] token that is native to how visual studio treats props... I however, want to use the premake token syntax, and in this particular case it is context sensitive, for the '.xml' file of the rules it needs to expand to [StripDebugInfo], but in the '.props' file it needs to expand to %(StripDebugInfo). 

This means that the 'action.pathVars' needs to be changed depending on who asks the question.

Anyway, all that is for a 2nd PR after this one gets merged...

### DIFF
--- a/src/actions/vstudio/vs2010_rules_props.lua
+++ b/src/actions/vstudio/vs2010_rules_props.lua
@@ -145,7 +145,15 @@
 
 	function m.commandLineTemplates(r)
 		if #r.buildcommands then
-			local cmds = os.translateCommands(r.buildcommands, p.WINDOWS)
+
+			-- create shadow context.
+			local env = p.rule.createEnvironment(r, "[%s]")
+			local ctx = p.context.extent(r, env)
+
+			-- now use the shadow context to detoken.
+			local cmds = os.translateCommands(ctx.buildcommands, p.WINDOWS)
+
+			-- write out the result.
 			cmds = table.concat(cmds, p.eol())
 			p.x('<CommandLineTemplate>%s</CommandLineTemplate>', cmds)
 		end
@@ -163,7 +171,12 @@
 
 	function m.executionDescription(r)
 		if r.buildmessage then
-			p.x('<ExecutionDescription>%s</ExecutionDescription>', r.buildmessage)
+			-- create shadow context.
+			local env = p.rule.createEnvironment(r, "[%s]")
+			local ctx = p.context.extent(r, env)
+
+			-- write out the result.
+			p.x('<ExecutionDescription>%s</ExecutionDescription>', ctx.buildmessage)
 		end
 	end
 
@@ -171,7 +184,14 @@
 
 	function m.outputs(r)
 		if #r.buildoutputs then
-			local outputs = table.concat(r.buildoutputs, ";")
+			-- create shadow context.
+			local pathVars = p.rule.createPathVars(r, "%%(%s)")
+			local ctx = p.context.extent(r, { pathVars = pathVars })
+
+			-- now use the shadow context to detoken.
+			local outputs = table.concat(ctx.buildoutputs, ";")
+
+			-- write out the result.
 			p.x('<Outputs>%s</Outputs>', path.translate(outputs))
 		end
 	end
--- a/src/actions/vstudio/vs2010_rules_props.lua
+++ b/src/actions/vstudio/vs2010_rules_props.lua
@@ -84,7 +84,6 @@
 		return {
 			m.propertyDefaults,
 			m.commandLineTemplates,
-			m.outputs,
 			m.executionDescription,
 			m.additionalDependencies,
 		}
@@ -144,18 +143,18 @@
 
 
 	function m.commandLineTemplates(r)
-		if #r.buildcommands then
+		-- create shadow context.
+		local env = p.rule.createEnvironment(r, "[%s]")
+		local ctx = p.context.extent(r, env)
 
-			-- create shadow context.
-			local env = p.rule.createEnvironment(r, "[%s]")
-			local ctx = p.context.extent(r, env)
+		-- now use the shadow context to detoken.
+		local buildcommands = ctx.buildcommands
 
-			-- now use the shadow context to detoken.
-			local cmds = os.translateCommands(ctx.buildcommands, p.WINDOWS)
-
-			-- write out the result.
+		-- write out the result.
+		if buildcommands and #buildcommands > 0 then
+			local cmds = os.translateCommands(buildcommands, p.WINDOWS)
 			cmds = table.concat(cmds, p.eol())
-			p.x('<CommandLineTemplate>%s</CommandLineTemplate>', cmds)
+			p.x('<CommandLineTemplate>@echo off\n%s</CommandLineTemplate>', cmds)
 		end
 	end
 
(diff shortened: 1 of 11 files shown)


tvandijck:
With the uncached context, we can modify the rules 'executionDescription' method like this:

```lua
		if r.buildmessage then
			-- add this rules properties to the environ.
			local env = {}
			for _, def in ipairs(r.propertydefinition) do
				env[def.name] = '%(' .. def.name .. ')'
			end

			-- create shadow context.
			local ctx = p.context.uncached(r, env)

			-- write out the result.
			p.x('<ExecutionDescription>%s</ExecutionDescription>', ctx.buildmessage)
		end
```

So what this does, is we create a local environment that we provide to the uncached context, effectively overriding the 'environ' of 'r' in this case... Then we add all the propertydefinitions as 'global' accessible variabled into that environment, and use the uncached context to detokenize the buildmessage...

```
propertydefinition {
    name = "StripDebugInfo",
}
buildmessage '%{StripDebugInfo}'
```

will now correctly resolve to 
```xml
<ExecutionDescription>%(StripDebugInfo)</ExecutionDescription>
```

tvandijck:
Then the detoken changes, pretty cool too... some tokens need to be 'detokened' as paths, rather then just variables... so for that we already have the 'action.pathVars'. This change introduces a 'environ.pathVars' as well, which actually has a higher priority then the one in the action... with this change you can thus write:

```lua
if #r.buildoutputs then
			-- copy visual studio pathVars.
			local pathVars = table.shallowcopy(p.vstudio.pathVars)

			-- add this rules pathVars.
			for _, def in ipairs(r.propertydefinition) do
				pathVars[def.name] = { absolute = true,  token = '%(' .. def.name .. ')' }
			end

			-- create shadow context.
			local ctx = p.context.uncached(r, { pathVars = pathVars })

			-- now use the uncached property to detoken using the new pathVars.
			local outputs = table.concat(ctx.buildoutputs, ";")

			-- write out the result.
			p.x('<Outputs>%s</Outputs>', path.translate(outputs))
		end
```

Which will also correctly resolve the tokens within the scope of this method, but also treats the token as an absolute path, removing everything that comes before it...

tvandijck:
And then the last change updating the Visual STudio backend to use these new features.. which now allows you to write rules like this:

```lua
rule 'example'
    display 'Example compiler'
    fileExtension '.example'

    propertydefinition {
        name = "output_path",
        kind = "string",
        display = "Output Path",
        description = "",
    }

    buildmessage 'Compiling %{file.basename} with example-compiler...'
    buildcommands {
        'package-example-compiler.exe %{output_path} "%{file.relpath}"'
    }
    buildoutputs {
        '%{output_path}%{file.basename}.example.cc',
        '%{output_path}%{file.basename}.example.h'
    }
```

no more Visual Studio native tokens anywhere.....

src/actions/vstudio/vs2010_rules_props.lua
```diff
@@ -145,8 +145,22 @@
...
+
+			-- create shadow context.
+			local ctx = p.context.uncached(r, env)
+
+			-- now use the uncached property to detoken using the new pathVars.
+			local cmds = os.translateCommands(ctx.buildcommands, p.WINDOWS)
+
+			-- write out the result.
 			cmds = table.concat(cmds, p.eol())
+			print(cmds)
```
dcourtois: Is this `print` a debug thing ?

dcourtois:
Just noticed the new Custom Rules thing, so not really familiar with all this, but the patch seems fine. Just found a "print" that looks like a debug thing (I commented the commit where it is)

src/actions/vstudio/vs2010_rules_props.lua
```diff
@@ -163,15 +177,38 @@
...
+			for _, def in ipairs(r.propertydefinition) do
+				pathVars[def.name] = { absolute = true,  token = '%(' .. def.name .. ')' }
+			end
+
+			-- create shadow context.
+			local ctx = p.context.uncached(r, { pathVars = pathVars })
+
+			-- now use the uncached property to detoken using the new pathVars.
+			local outputs = table.concat(ctx.buildoutputs, ";")
+			print(outputs)
```
dcourtois: same question here ?

starkos:
Looking at this now. First impressions are that an `uncached()` function seems useful. There might be a better/more general way to do what you're trying to do with the rules. Thinking…

starkos:
It seems like what you're really doing here is not turning off caching so much as creating a new context that extends an existing context with an extended token expansion environment—that is *definitely* useful. We could then use the same pattern to implement the `pathVars` feature too.

What if instead of `context.uncached()` it were called something more like `context.extend(baseContext, newEnvVars)`? The context it returns would still cache results, but would start out empty so any tokens would be re-expanded into the new environment?

This would allow any exporter to provide their own token replacements, in any context, in a pretty straightforward way.

Separately, maybe it would be useful to have a function under `rule` that can do the property iteration into an environment for you? Maybe something like `rule.prepareEnvironment()` that takes an `env` table and either a list of replacement patterns or function telling it expand vars to `"[var]"` or whatever? Then a Makefile rule exporter could maybe use it to expand to `"$(var)` instead, possibly maybe.

tvandijck:
> It seems like what you're really doing here is not turning off caching so much as creating a new context.

Yes, this change evolved a little since I first wrote the initial comments...
That said however, due to the different metatable assignment, it is also very much 'uncached'. As in it will not store the result of the configset.fetch inside of this 'extended' context..  I very much like the different name you are giving it... so I'll make that change...

> This would allow any exporter to provide their own token replacements, in any context, in a pretty straightforward way.

Yes, that is exactly how I'm using it currently for the rules in the Visual Studio action.
I was planning on doing the same in our xcode and gmake backend, so that I no longer have to write individual rules for each platform. It bothered me, and so I tried to come up with a solution, and this is the best I could come up with, and it looks really neat...

> Maybe something like rule.prepareEnvironment() that takes an env table and either a list of replacement patterns or function telling it expand vars to "[var]" or whatever?

Yeah, any helpers are going to be useful for this for sure... something like:
```lua
rule.prepareEnvironment(env, "[%s]") 
```
would probably be perfect, I'll add something like that..
The issue is that not in all cases you want the rules to expand as 'environment' values, but rather as pathvars. pathVars are special in that they are 'post' processed with a 'isAbsolute' pass. environment variables are not... So sometime you want to do:

```
rule.preparePathVars(pathVars, "%%(%s)")
```

or maybe just

```lua
rule.prepareEnvironment(env, "%%(%s)", true) 
```

src/actions/vstudio/vs2010_rules_props.lua
tvandijck: oops

src/actions/vstudio/vs2010_rules_props.lua
tvandijck: oops...

tvandijck:
> We could then use the same pattern to implement the pathVars feature too.

I'm actually using that in m.outputs in the vs2010_rules_props.lua
And in all honesty, I actually find this more expressive then the action.pathVars.
In the new gmake backend I've been working on I completely removed the global action pathVars in favor of this context sensitive version.

tvandijck:
All updated and squashed.

tvandijck:
There is some minor issues with the visual studio rule stuff, I got fixes here, but I have to apply them to this change list.... so please don't merge this yet...

while this commit works fine in most cases, there is a few cases I hit that make it go boom boom boom.

tvandijck:
OK, so these latest changes, fixes a few issues with how the custom rules was outputting the "Outputs" property. if it is in the props file, it is only computed when that file is loaded. Since this is a property that generally contains other tokens (since it is based on the "buildoutputs") it ends up not beeing recomputed if for example you have a "%(outputdir)" token in it that is configurable in your rule.
By moving it into the targets file, it gets recomputed when the target is executed.

starkos:
I haven't had an opportunity to try this, but I think it looks pretty good. It would be nice to have some unit tests against the issues you were seeing with the old implementation, to make sure we don't regress any of those cases with future changes.

tvandijck:
Added tests... when those pass I'll merge this.

PR #561, opened 2016-08-07
Add vectorextensions "IA32" support for VS>=2012 on x86. (#560)

### LINKED ISSUE #560: vectorextensions for VS2012, VS2013, VS2015 should support "IA32" value
VS >= 2012 defaults to SSE2 instruction set for x86-32 targets. There is an option to set the compiler to no SSE instructions again by specifying `/arch:IA32` on the command line (and similar in project files). However, even when not specifying `vectorextensions` at all in premake, the project files will use the default (i.e. require SSE2 support).
Adding another value `"IA32"` to the option, which would explicitly force this setting, should solve the issue, as should changing the default behaviour if no `vectorextensions` is specified at all.
The `.vcxproj` entry is called `<EnableEnhancedInstructionSet>NoExtensions</EnableEnhancedInstructionSet>`.
 Only x86 32bit is affected, because 64bit has always SSE2 available. Earlier MSVC versions like 2010 or 2008 default to no SSE instructions, and are thus also not affected.

manxorist:
Fixes #560 .

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1098,6 +1098,7 @@
 			"Default",
 			"AVX",
 			"AVX2",
+			"IA32",
 			"SSE",
 			"SSE2",
 			"SSE3",
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1213,6 +1213,8 @@
 				v = "StreamingSIMDExtensions2"
 			elseif x == "SSE" then
 				v = "StreamingSIMDExtensions"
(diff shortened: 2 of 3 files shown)


samsinsane:
This needs tests so it isn't broken in the future, and I believe you'll also need to update _premake_init.lua to allow for the `IA32` value for `vectorextensions`.

samsinsane:
Sorry to be a pain, can you add one more test, VS2010 and IA32, ensuring that nothing is emit.

Other than that, LGTM.

samsinsane:
Wonderful, thanks for that!

Issue #565, opened 2016-08-11
CodeLite / GCC problem with generating projects for SharedLibs


GuildMasterInfinite:
I have a problem trying to generated files for CodeLite, using the GCC toolset (MinGW64, to be specific). 

* I have a ConsoleApp in my workspace, it works well, loads, etc
* I also have a SharedLib project and the problem is: it is not loaded in the workspace. When I try to load it, it says **"Corrupted project file '... MyProject.project'"**

I opened the .project file and found the problem, I found this snippet twice:
```xml
<Linker Required="yes"
 Options="-m64;-s;-shared;-Wl,--out-implib="... MyProject.lib"">
</Linker>
```
As you can see, it doesn't properly escapes the quotes: if I replace **"... MyProject.lib"** with **'... MyProject.lib'** it works.

~~I tried searching the source code, and I think the problem might be [at this line](https://github.com/premake/premake-core/blob/master/src/tools/gcc.lua#L216), but I'm not familiar enough with Lua to find any workaround.~~

GuildMasterInfinite:
I've added a pull request #566, but I've only tested it on Windows (it does work now).

samsinsane:
I believe this was fixed in the CodeLite repository but looks like the submodule reference wasn't updated in core. @tvandijck it looks like you fixed this, can you confirm? Also, can you fix up the submodule reference?

tvandijck:
I pushed the submodule update to master...

PR #566, opened 2016-08-11
Fix CodeLite / GCC project generation

GuildMasterInfinite:
This is a fix for #565. It replaces the use of double quotes in "-Wl,--implib-" with single quotes. Also edited the test to reflect this.


GuildMasterInfinite:
Already fixed in the CodeLite module, but not updated.

PR #568, opened 2016-08-17
Use rpaths & soname/install_name to link sibling shared libraries on OS X and Linux

tritao:
Original patch by Renaud Guillard (@noresources) updated to recent Premake.

### DIFF
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -178,9 +178,21 @@
 -- Format a list of values to be safely written as part of a variable assignment.
 --
 
-	function make.list(value)
+	function make.list(value, quoted)
+		quoted = false
 		if #value > 0 then
-			return " " .. table.concat(value, " ")
+			if quoted then
+				local result = ""
+				for _, v in ipairs (value) do
+					if #result then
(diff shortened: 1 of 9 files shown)


tvandijck:
It looks good to me.... should I close the other PR and merge this one then?

tritao:
Sounds good to me.

src/actions/make/_make.lua
```diff
@@ -178,9 +178,21 @@
 -- Format a list of values to be safely written as part of a variable assignment.
 --
 
-	function make.list(value)
+	function make.list(value, quoted)
+		quoted = false
```
noresources: Is that the line was intentionally kept or is it a test relict ? If your intention was to suppress the new behavior, you should just have to revert make.list as it was before. No ?

tritao:
@tvandijck mind removing that line from master?

src/actions/make/_make.lua
tritao: My mistake, I was testing your PR locally and checking if this change made any difference in the test suite and forgot to remove it the debug line.

@tvandijck mind removing that line from master?

PR #570, opened 2016-08-25
Fix path normalization within os.execute

nmoehrle:
This is a merge of #320 and #538 which resolves the issues  #316 and #537 by normalizing only arguments that should be file system paths.


tvandijck:
lgtm

tritao:
👍

PR #571, opened 2016-08-30
Add a way to set the current working directory

Spacebrain:
I found myself needing a way to change the current working directory of Premake due to my particularly setup.
It could be relevant in some particular cases so I see no reason not including into Premake.


tvandijck:
I'm guessing you found it already, which is why you closed this.... but just in case there is "os.chdir" exposed premake.

Issue #572, opened 2016-08-31
Missing edge case in deprecated value handling


starkos:
Here is a unit test to demonstrate the problem:

```lua

local suite = test.declare("api_deprecations")
local api = premake.api

local sln, prj, cfg

function suite.setup()
	sln, prj = test.createWorkspace()
end

local function prepare(platform)
	cfg = test.getconfig(prj, "Debug", platform)
end


function suite.setsNewValue_whenOldValueIsRemovedViaWildcard_inSubConfig()
	filter { "configurations:Debug" }
		flags { "Symbols" }

	filter { "*" }
		removeflags { "*" }

	prepare()

	test.isequal("Off", cfg.Symbols)
end
```

The `Symbols` flag is set for the "Debug" configuration. Then, all flags are removed from the project in the next block.

The `Symbols` flag has been deprecated in favor of the new `symbols("On")` and `symbols("Off")` syntax. When the flag is set in the first block, the deprecation handler runs, showing a warning message and calling `symbols("On")` to translate the flag to the new approach.

When the flag is removed via the wildcard in the next block, some logic is run to try and match any existing flags and run the appropriate deprecation logic for them. But because the flag was originally set with a different filter, it isn't picked up. As a result, `symbols("Off")` never gets called.

I've spent several hours trying to come up with a fix for this, but I don't think I can do it without a major rewrite or a huge performance hit. If anyone has any ideas I would love to hear them!

tvandijck:
very interesting... I don't really have a good idea for this one... never used this pattern.. but I have a slew of bugs for premake to look at tomorrow here, so I might as well take a stab at this one too.

tvandijck:
OK... if I change line 565 of api.lua from 
```lua
local current = configset.fetch(target, field)
```
to 
```lua
local current = field.allowed
```

I actually get it to return "Symbols == Default"... rather then "Off"...
But Default is the initial value we start with... so maybe that is actually what we're expecting..

Basically what the change does, is instead of fetching the flags, I just explicitly remove all of them.

The alternative is we're looking at flags that have been previously set in any block that matches the current blocks filter...

starkos:
I posted this on the PR too, but for completeness: My concern with this approach is that there are a number of deprecated flags out there now (and I have even more over here). If a project has switched over to the new APIs (which most will have done), this will step on those values:

```lua
floatingpoint "Fast"
removeflags "*"
-- floatingpoint is now "Default"
```

PR #573, opened 2016-09-01
Workaround for #572

tvandijck:
https://github.com/premake/premake-core/issues/572

### DIFF
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -555,15 +555,20 @@
 				table.foreachi(value, recurse)
 
 			elseif hasDeprecatedValues and value:contains("*") then
-				local current = configset.fetch(target, field)
+				local current = configset.fetch(target, field, {
+					matcher = function(cset, block, filter)
+						local current = cset.current
+						return criteria.matches(current._criteria, block._criteria.terms or {}) or
+							   criteria.matches(block._criteria, current._criteria.terms or {})
+					end
+				})
+
 				local mask = path.wildcards(value)
 				for _, item in ipairs(current) do
 					if item:match(mask) == item then
 						recurse(item)
 					end
 				end
-				table.insert(removes, value)
-
 			else
 				local value, err, additional = api.checkValue(field, value)
 				if err then
@@ -1100,9 +1105,7 @@
 
 	function configuration(terms)
 		if terms then
-			if (type(terms) == "table" and #terms == 1 and terms[1] == "*") or
-			   (terms == "*")
-			then
+			if (type(terms) == "table" and #terms == 1 and terms[1] == "*") or (terms == "*") then
 				terms = nil
 			end
 			configset.addblock(api.scope.current, {terms}, os.getcwd())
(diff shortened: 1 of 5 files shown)


starkos:
My concern with this approach is that there are a number of deprecated flags out there now (and I have even more over here). If a project has switched over to the new APIs (which most will have done), this will step on those values:

```lua
floatingpoint "Fast"
removeflags "*"
-- floatingpoint is now "Default"
```

tvandijck:
I see your point.... 

But that means the only way we can figure this out then is to search all blocks for the flags that we set upstream... I'll see if I can figure something else out... All the information is in blocks somewhere, we just need to figure out which ones are "compatible" with "current".

starkos:
Right. In theory, I was thinking it might be possible to do this during baking at the point that the configuration set is compiled, since we're stepping through the blocks there anyway. The catch is the deprecation logic is written using the project APIs (i.e. it calls `symbols("On")`) which means you need to set the active container and point the API code to the right block. Which means that code probably needs to get abstracted out a bit more than it is. 

I'm still looking at this as I can, but it is a rabbit hole.

tvandijck:
OK... so this one works in your case... but the test I added breaks... looking into that.

tvandijck:
Basically what I'm doing now, is going through each block and see if the terms used apply to the block we're currently doing the remove in... That should in theory give me a list of all the flags that were applied for the config we're in... It's not conclusive clearly, given the second test I created.

tvandijck:
Turns out... we need to test both ways.... not really fully understanding the theory behind that, but maybe you have an idea??

Anyway, this way it passes both tests... But I think we need to write a couple more tests for a bunch of cases, because I don't think this covers it...

starkos:
Reversing the criteria tests is a clever approach. This will fall down if the value is set in a parent container (i.e. `flags { "Symbols" }` is in the workspace, and the remove is in the project). It will also fail on any path fields, since `getvalues()` doesn't consider the path handling logic.

This looks like a promising approach, but it will probably have to go through the `configset._fetchMerged` code path. I haven't figured out a clean way to make that happen…

tvandijck:
> but it will probably have to go through the configset._fetchMerged code path.

Yeah, this is obviously more of a proof of concept approach... I did consider the parent container but didn't test for that yet, so that is basically just the next test we need to add to make sure we cover the case, and fix things accordingly...

I think what we can do with the _fetchMerged, is give it a callback argument which does the matching.. instead of hardcoding it to match against the criteria and cset.compiled... That way the normal fetch can default to that... but this particular fetch can provide a different callback that reverses the criteria test.

tvandijck:
OK... so one more try.... ;) it now goes through the configset.fetch method, but I changed the way it handles the "filter" argument... it now has an optional "matcher" callback function, and so in the api.remove we call the configset.fetch with this special callback..

```lua
				local current = configset.fetch(target, field, {
					matcher = function(cset, block, filter)
						local current = cset.current
						return criteria.matches(current._criteria, block._criteria.terms or {}) or
							   criteria.matches(block._criteria, current._criteria.terms or {})
					end
				})
```

starkos:
Thanks for that! I think this looks pretty good now, and ought to cover things well. Nice!

I'm going to be out of commission for the next week or so, starting right about now, so I won't be able to verify this locally, but I think it looks good. I'd like to include it on the alpha10 release branch, as this was the last thing holding up the release. Would it be possible to rebase it there? If not, I can apply it easily enough when I get back.

Thanks again for looking into this!

tvandijck:
I'll merge it, and cherry pick the commits over to alpha10, should be easy.

PR #575, opened 2016-09-01
Add raw module as a standard module.

tvandijck:
This is just a really handy module to inspect the raw lua tables...
Not sure if making it a standard module is really needed, but I figured I could at least propose it with the PR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "modules/d"]
 	path = modules/d
 	url = https://github.com/premake/premake-dlang.git
+[submodule "modules/raw"]
+	path = modules/raw
+	url = https://github.com/premake/premake-raw.git
--- a/src/_modules.lua
+++ b/src/_modules.lua
@@ -9,4 +9,5 @@
 		"monodevelop",
 		"codelite",
 		"d",
+		"raw",
 	}
--- a/src/_modules.lua
+++ b/src/_modules.lua
@@ -9,5 +9,4 @@
 		"monodevelop",
 		"codelite",
 		"d",
-		"raw",
 	}


samsinsane:
I've been thinking about this quite a bit, I think we should probably merge this in. I think there's value in having a diagnostics mode that allows a developer/user to view the state as a whole.

starkos:
If we merge, can we leave it disabled by default, like the test module?

samsinsane:
So to use it, the premake5.lua would need to have `require "raw"`? I don't really see a problem with that, out of curiosity why would you prefer it to be disabled by default?

tvandijck:
> If we merge, can we leave it disabled by default, like the test module?

I'm all in favor, but how is that done?? just not mentioning it in the _modules.lua ?? would the action still work doing that? or is there other changes involved?

starkos:
Sorry…coming back slowly…yes, just don't mention it in `_modules.lua`. Developers will need to trigger the module to load, either by `require("raw")` in their project script or premake-system.lua. This isn't a huge deal though, this is less likely to cause confusion than the self-test action would have.

tvandijck:
OK, well I can certainly do that... not entirely sure what the benefits are? but I have no real objections either.

starkos:
It prevents that action from showing up in the help text unless it has been explicitly enabled (via `require`). I'm trying to do this for features that are primarily targeted at advanced users or Premake developers, to avoid cluttering things up and making it more complicated for new users who are just trying to figure out how to use Premake to generate some project files for their IDE.

That's the motivation anyway?

tvandijck:
that makes perfect sense ;)... I made the change...

samsinsane:
Yeah, that does make sense. LGTM :)

PR #580, opened 2016-09-20
fix windows platform compile error

lailongwei:


samsinsane:
LGTM. Out of curiosity, what compiler (and version) were you using?

lailongwei:
I'm using visual studio 2012:), visual studio 2008 could not be compile too:)

samsinsane:
Interesting, thanks for the information!

@tvandijck @starkos should we add some of these older versions of VS to the AppVeyor builds? Additionally, does AppVeyor support the older versions of VS? It should increase the likelihood of contributors and users if our stuff just works regardless of what they have access to. Thoughts?

PR #597, opened 2016-10-17
Enable Edit and Continue for x64 builds on VS2015

akaStiX:
VS 2015 supports Edit and Continue for x64 builds! Let's enable it


starkos:
Thanks!

Issue #598, opened 2016-10-19
5.0.0-alpha10 symbolspath() produces .pdb without debugging information (VS2015)


ccw808:
`symbolspath()` currently sets the compiler /Fd option, which is the location of an intermediate .pdb file. Possible solution is to set the linker /PDB option.

(Visual Studio 2015 C++)

samsinsane:
#600 should resolve this issue.

PR #600, opened 2016-10-20
Fix for #598

sbx320:
ProgramDataBaseFileName (Compiler Option) describes the location of the debug information for the intermediate object files (.obj). 
ProgramDatabaseFile (Linker Option) instead sets the location of the debug information for the final binary (shared library or executable) which is the documented (and probably also intended) behavior.

Related MSDN Articles:
https://msdn.microsoft.com/en-us/library/9wst99a9.aspx (/Fd (ProgramDataBaseFileName))
https://msdn.microsoft.com/en-us/library/kwx19e36.aspx (/PDB (ProgramDatabaseFile))

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -308,7 +308,6 @@
 			m.clCompileAdditionalUsingDirectories,
 			m.forceIncludes,
 			m.debugInformationFormat,
-			m.programDataBaseFileName,
 			m.optimization,
 			m.functionLevelLinking,
 			m.intrinsicFunctions,
@@ -409,6 +408,7 @@
 				m.largeAddressAware,
 				m.targetMachine,
 				m.additionalLinkOptions,
+				m.programDatabaseFile,
 			}
 		end
 	end
@@ -1793,9 +1793,9 @@
 	end
 
 
-	function m.programDataBaseFileName(cfg)
+	function m.programDatabaseFile(cfg)
 		if cfg.symbolspath and cfg.symbols == p.ON and cfg.debugformat ~= "c7" then
-			m.element("ProgramDataBaseFileName", nil, p.project.getrelative(cfg.project, cfg.symbolspath))
+			m.element("ProgramDatabaseFile", nil, p.project.getrelative(cfg.project, cfg.symbolspath))
 		end
 	end
 


samsinsane:
Very interesting, everyone in #524 seems to indicate it works as expected? @tvandijck is it supposed to be the compiler or linker flag?

tvandijck:
Yeah, our game ships with this stuff.... I don't think we have encountered an issue so far... but the linked articles seem to indicate that we're doing something wrong maybe?? I don't know...

I think what we want is the 'resulting' PDB to be in the right spot, I don't really care much for the intermediate one, but I was unaware that there were two separate options for it..

sbx320:
Are you sure the generated PDBs are non-intermediate PDBs? Also if you're using symbolspath "$(OutDir)$(TargetName).pdb" you'll likely not see any impact, due to that being the default if no ProgramDatabaseFile is set. You'd just end up overwriting the debug information all the time.

When I compile Lua into a shared library with the current behavior and symbolspath "Build/Symbols/$(ProjectName).pdb" I get two pdbs:

- The first one in the intended Build/Symbols/ folder with a name of $(ProjectName).pdb with a size of about 156kb
- The second one located alongside the Lua binary with the same name as the binary just with a .pdb file ending and 1164kb in size. 

The former is unusable for debugging. The second one is generated as part of the VS defaults.

(This is all on VS2015 Update 3)

tvandijck:
> Are you sure the generated PDBs are non-intermediate PDBs?

I'm not... all I know is that when we deploy these pdbs to our symbol server we can debug our live versions of Heroes of the Storm and Starcraft 2... 

The settings for heroes are like this:

linker settings:
image

compiler settings:
image

sbx320:
It works in that case due to ProgramDatabaseName being the same as ProgramDataBaseFileName. 
This results in the intermediate (compiler) pdb being overwritten by the final (linker) pdb.

If you attempt to adjust symbolspath to something else (e.g. MySymbolsDir/$(TargetName).pdb) you end up with the two pdbs I described above.

samsinsane:
@tvandijck it looks like these changes won't break your projects, but just in case you want to verify that everything continues to work on your end, I'll leave this for you to merge in. 😄

tvandijck:
Yeah, I haven't had the time unfortunately... maybe our next "sprint" will allow me some "premake time".
In the meantime, if we could rebase this change to latest master.. github says it's out of date.

samsinsane:
GitHub seems to say "This branch is out-of-date with the base branch" if it's not on the current commit for that branch. I'm not sure if this is a good feature or not, any idea what the "Update branch" button does?

samsinsane:
Oh I had another PR open, I clicked it on that just incase it breaks anything. Looks like it just rebases the branch for you, cool.

PR #605, opened 2016-10-27
Fixed bug with objdirs prefix code '!'

samsinsane:
- Fixed flawed tests that manipulated the actual result before testing it
- Added test to ensure prefix code '!' works


tvandijck:
LGTM.

PR #608, opened 2016-10-30
VStudio: Add additional deps & additional libdirs for static lib

mikisch81:
Currently if static library links with another non-sibling library build fails.

For example: 
- If using the socket() API on windows in a static library, need to link with ws2_32.lib.
  In current premake, linking fails due to unresolved external symbol of the socket() symbol. @mindw

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -431,6 +431,8 @@
 	m.elements.lib = function(cfg, explicit)
 		if cfg.kind == p.STATICLIB then
 			return {
+				m.additionalDependencies,
+				m.additionalLibraryDirectories,
 				m.treatLinkerWarningAsErrors,
 				m.targetMachine,
 				m.additionalLinkOptions,
--- a/tests/actions/vstudio/vc2010/test_link.lua
+++ b/tests/actions/vstudio/vc2010/test_link.lua
@@ -155,6 +155,20 @@
 		]]
 	end
 
+	function suite.additionalDependencies_onSystemLinksStatic()
+		kind "StaticLib"
+		links { "lua", "zlib" }
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+</Link>
+<Lib>
+	<AdditionalDependencies>lua.lib;zlib.lib;%(AdditionalDependencies)</AdditionalDependencies>
+</Lib>
+		]]
+	end
+
 
 --
 -- Any system libraries specified in links() with valid extensions should
@@ -171,6 +185,20 @@
 		]]
 	end
 
+	function suite.additionalDependencies_onSystemLinksExtensionsStatic()
+		kind "StaticLib"
+		links { "lua.obj", "zlib.lib" }
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+</Link>
+<Lib>
+	<AdditionalDependencies>lua.obj;zlib.lib;%(AdditionalDependencies)</AdditionalDependencies>
+</Lib>
+		]]
+	end
+
 
 --
 -- Any system libraries specified in links() with multiple dots should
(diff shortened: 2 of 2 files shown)


starkos:
Sorry for the late response on this one, but want to point out that this behavior is Visual Studio specific and *not* portable to other toolsets. In previous discussions (which I seem completely unable to find) it was considered better to fail everywhere, so it could be fixed, then to silently succeeded on one toolset and fail on another.

jspohr:
@starkos Did you mean this discussion? #627

@mikisch81 This change appears to cause major problems for my static builds. I have a project with a lot of libraries, and I use links between them for dependencies, because I use dynamic libs for development, and static libs for deployment. When building an app, I previously had to put all libraries and their dependencies as links. This is not ideal, however, the new behavior doesn't work for this case, either. Visual Studio's Librarian will now combine all dependent libraries into the output library, which leads to very large file sizes. For example, assume every library on its own is 10mb. For every library it depends on, this adds another 10mb. So I may get to 110mb if my library has 10 dependencies. Now, another library that depends on this 110mb library will get bigger by 110mb. This escalates quickly, and my build fails because I have libraries with more than 2gb in size.

I could create a workaround by implementing my own links() function, or overwrite the VS generator function to get back the old behavior, but I wanted to discuss this first, because this might be a problem for other projects, too.

Appreciate your thoughts on this! Thanks

mikisch81:
@jspohr @starkos Then there should be a way to implicitly add all the dependencies to the application linking step. I opened this pull requests because we supply a library and not an application and we couldn't just tell the user to add all the nested dependencies by himself when building his application.
Thanks.

jspohr:
True, this was the general sentiment voiced in the issue I referred to.
I workarounded this using premake.override for my use case, so this has low priority for me right now, but I think that in the long run, premake needs a solution for this scenario that "just works".

starkos:
I'm okay with adding this as a feature, but I think it should be something that is switched on explicitly by the project author (e.g. `linkstaticdependencies "true"` maybe), with the existing, more portable behavior left as the default.

samsinsane:
Prior to this PR, didn't `links` just act like `dependson` for a static library? I don't really see any benefit in maintaining this functionality, why not just use `dependson` instead?

jspohr:
For my specific use case, I need `links` to work for both static and dynamic libs, because my projects can be one or the other, depending on the configuration. However, I can just wrap that in a filter and use `links` for the dynamic, and `dependson` for the static case. This still means that I have to manually add all dependencies to the executable for the static configuration, which this PR tried to solve.

tvandijck:
To me.... 
- if you put a ```links```, I expect it to link... 
- if you put a ```dependson```, I expect it to add a build dependency. Nothing more, nothing less.

> @jspohr : I have a project with a lot of libraries, and I use links between them for dependencies

To me that feels a little like it is exploiting a premake behavior that isn't intended, but clearly causes a lot of discussion... I think we touched on this particular set of API's a few times now. I understand the implication, and your use-case.. but static libs, if you specify links, I expect it to link, since that is what you say you want.. If you expect a dependency you should be using the dependson API... I also understand you have both dynamic and static uses... then:

```
filter "kind:staticlib"
   dependson "a"
filter "kind:dynamiclib"
   link "a"
```

will do the right thing, and you can easily wrap that in a function and do

```
dynamiclink "a"
```

In addition to this entire discussion, I would argue that dependencies between static libs are almost always unnessesairy... unless you do any kind of codegen, there is no need for a lib to link/dependson any other lib. And adding dependencies actually leads to increased build times due to a decrease in the ability of the compiler to parallelize the build, since now every lib needs to wait for the next lib to finish, and so you get all these sync points in your build.. it's just less efficient for often no reason at all.

The only reason for wanting to link one lib to another is for inclusion, so that once an "app" links against lib A, it automatically gets opengl, winsock, etc as well, since that is what A is using, rather then having to explicitly call this out in the "app" lua script... We've discussed this many times too, and talked about introduction an API that allows you to create links that will propagate down to the app, while keeping the links&dependson in libs clean...

Anyway, no real satisfying answer from me I guess... I'm not against this PR at all, I just think it is clearly a change that breaks people that relying on links to do the "wrong" thing (in my opinion)..

jspohr:
Some more thoughts:

Static libs are obviously not the same thing as dynamic libs. There is no linking involved, just copying object files into large blobs. However, static libs can have dependencies. Resolving those dependencies by just lobbing their contents together is a brute force solution, which wastes disk space and build time, and can even lead to broken builds in case size limits are straddled.
So what do we actually want? When linking a dynamic library or an executable, we want the linker to know all the dependencies (and the dependencies of those dependencies). Premake has that information, the question is how the information is transported to the linker.
@starkos Would it be possible to resolve all dependencies for a project recursively during a Premake run, and putting all of those as inputs to the linker? That's what I'm doing manually now, anyway. If Premake could do this automatically, I hope it would solve the issues at hand.

jspohr:
@tvandijck Sorry I wrote the above without seeing your post. I agree, `links` in the context of static libs is not what I should be using. I probably will roll a custom solution for my issue, anyway.
But on the topic of propagating dependencies, it would still be a great feature, although I'm unsure about the implications and complexity of the implementation. I'm usually not a fan of tools being too smart, and having Premake generate dependencies implicitly could maybe lead to surprises that are hard to track down.
However, IMO build tools should prioritize user's productivity, so maybe a semi-smart solution can be found :)

PR #609, opened 2016-10-30
Fix bug in premake.filename(obj, ext)

mikisch81:
- If "ext" is given and project was configured with filename which ends with "ext" than the returned filename will be "filename.ext.ext".
- It can be relevant in external projects where you want to give the filename of the project file explicitly.
  @mindw

### DIFF
--- a/src/base/premake.lua
+++ b/src/base/premake.lua
@@ -192,7 +192,7 @@ function premake.filename(obj, ext)
 		fname = path.join(fname, ext)
 	else
 		fname = path.join(fname, obj.filename)
-		if ext then
+		if ext and not fname:endswith(ext) then
 			fname = fname .. ext
 		end
 	end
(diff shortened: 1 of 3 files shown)


samsinsane:
I can't really see any use for this, can you please provide an example of why this is required? It seems to me that this would only encourage users to provide the extension which would make it harder to swap between Visual Studio, CodeLite, XCode or Makefiles.

Alex---:
Hi @samsinsane,
I think documentation is misleading regarding this `API.`
[Documentation ](https://github.com/premake/premake-core/wiki/filename#parameters) states: `name` is the desired file name for the generated workspace or project file.
Desired project file is my_proj.vcxproj.
Please consider either modifying documentation to explicitly notify that file name should be provided without extension, or sanitizing input value.
thx

samsinsane:
@Alex--- I've updated the documentation to explicitly notify users that the filename shouldn't include the extension.

mikisch81:
@samsinsane @Alex--- Now the description is more precise, the pull request can be closed.

Alex---:
Tnx. I saw the change. It is much clearer now

PR #617, opened 2016-11-07
fix switch/separator in rules_xml

tvandijck:
This is a workaround for a bug in msbuild, kind of hard to explain without the source for msbuild next to it.


samsinsane:
Are you able to add tests for this?

tvandijck:
not today, but I'll have a look at it, I have to figure out why we needed this again ;)

tvandijck:
Added tests.. will merge once appveyor & travis pass.

PR #619, opened 2016-11-07
Add support in csproj backend to add files to certain configurations.

tvandijck:
and mark files not in that config as 'exclude from build'.


samsinsane:
This should probably have some tests to ensure it isn't broken in the future.

tvandijck:
Test added...

PR #620, opened 2016-11-07
Allow require("modulename") from a http server.

tvandijck:
Work in progress, (I didn't finish the 'server' part yet)...

### DIFF
--- a/src/base/moduledownloader.lua
+++ b/src/base/moduledownloader.lua
@@ -1,0 +1,125 @@
+--
+-- moduledownloader.lua
+-- Downloads a module from a package server
+-- Copyright (c) 2002-2016 Jason Perkins and the Premake project
+--
+
+
+---
+-- progress bar.
+--
+	local function _http_progress(total, current)
+		local width = 78
+		local progress = math.floor(current * width / total)
+
+		if progress == width then
+			io.write(string.rep(' ', width + 2) .. '\r')
+		else
+			io.write('[' .. string.rep('=', progress) .. string.rep(' ', width - progress) .. ']\r')
+		end
+	end
+
+---
+-- escape parameters in a url.
+--
+	local function escape_url_param(param)
+		local url_encodings = {
+			[' '] = '%%20',
+			['!'] = '%%21',
+			['"'] = '%%22',
+			['#'] = '%%23',
+			['$'] = '%%24',
+			['&'] = '%%26',
+			['\''] = '%%27',
+			['('] = '%%28',
+			[')'] = '%%29',
+			['*'] = '%%2A',
+			['+'] = '%%2B',
+			['-'] = '%%2D',
+			['.'] = '%%2E',
+			['/'] = '%%2F',
+			[':'] = '%%3A',
+			[';'] = '%%3B',
+			['<'] = '%%3C',
+			['='] = '%%3D',
+			['>'] = '%%3E',
+			['?'] = '%%3F',
+			['@'] = '%%40',
+			['['] = '%%5B',
+			['\\'] = '%%5C',
+			[']'] = '%%5D',
+			['^'] = '%%5E',
+			['_'] = '%%5F',
+			['`'] = '%%60'
+		}
+
+		param = param:gsub('%%', '%%25')
+		for k,v in pairs(url_encodings) do
+			param = param:gsub('%' .. k, v)
+		end
+
+		return param
+	end
+
+
+---
+-- Downloads a module from a package server
+--
+-- @param modname
+--    The name of the module to download.
+-- @param versions
+--    An optional version criteria string; see premake.checkVersion()
+--    for more information on the format.
+-- @return
+--    If successful, the module was downloaded into the .modules folder.
+---
+
+	function premake.downloadModule(modname, versions)
+		if http == nil then
+			return false
+		end
+
+		-- get current user.
+		local user = 'UNKNOWN'
+		if os.get() == 'windows' then
+			user = os.getenv('USERNAME') or user
+		else
+			user = os.getenv('LOGNAME') or user
+		end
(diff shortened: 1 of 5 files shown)


tvandijck:
Make a super simple server here:
https://github.com/tvandijck/packagesrv-web

src/base/moduledownloader.lua
```diff
@@ -0,0 +1,125 @@
...
+-- moduledownloader.lua
+-- Downloads a module from a package server
+-- Copyright (c) 2002-2016 Jason Perkins and the Premake project
+--
+
+
+---
+-- progress bar.
+--
+	local function _http_progress(total, current)
```
tritao: I wonder if this could be added as a public HTTP function in Premake.

We also have something like this in our build scripts, but your version seems more solid, I'd rather just drop our code and use this one.

src/base/moduledownloader.lua
```diff
@@ -0,0 +1,125 @@
...
+			io.write(string.rep(' ', width + 2) .. '\r')
+		else
+			io.write('[' .. string.rep('=', progress) .. string.rep(' ', width - progress) .. ']\r')
+		end
+	end
+
+---
+-- escape parameters in a url.
+--
+	local function escape_url_param(param)
```
samsinsane: Doesn't curl already do this?

src/base/moduledownloader.lua
tvandijck: works for me, the only thing I would add is that on build servers this is annoying as hell, since the \r doesn't work in captured console output, and so you just get hundreds of lines of this in your log... so a way to disable it would be nice, or have it disabled by default, and only enable with --verbose... either way...

src/base/moduledownloader.lua
tvandijck: It might... I don't know ;) this has been in our package system for a long time now, I've just been copying small bits and pieces to get this module system off the ground, an idea I've had for a few months now. (to do it through the require).

tvandijck:
> A potential problem I can think of is versioning. It looks like the existing code does not deal with that at all.

Actually, the require overload that Jason added deals with versioning... And my commit takes that version to download the version specified (if specified)... The thing that this commit doesn't deal with yet is "what happens when a new version is specified and the module already exists in the .module folder".

It's why it's a work in progress... I figured I might as well just post the PR, so people can test and provide feedback as I hobby my way through this.

tvandijck:
> Also any thoughts about packages? I think you guys had a pretty well developed package system at one point, I remember checking the workings. Is this related to that?

We still do, and it's still very much in use... more and more teams are leveraging it, not just Heroes and SC2. Is this related? not really... But we've been dancing around the module issue for so long now, I really wanted to contribute to that and figure that one out first.

src/base/moduledownloader.lua
tritao: yeah I noticed the same problem on our Travis builds. A simple solution I can think of is maybe disable it if a `CI` environment variable exists.

tritao:
My bad, I can see now that it can already deal with versioning. I didn't want to be negative, I think whatever edge cases there are can be solved after we get the basics of it working.

tritao:
Any reason why this has not been merged yet?

tvandijck:
No reason other then that I don't want to merge my own PR's unless they are trivial or actual bugfixes... this is a feature... and I'm going to leave it up to someone else to make a judgement.

On a side note however.... the "server infrastructure" is far from finished, and any help in that area would be greatly appreciated...
https://github.com/tvandijck/packagesrv-web

It's live at http://packagesrv.com

Also note, this work is unrelated to any blizzard stuff, so I've been mostly working on this very sporadically at home.

tvandijck:
I was mostly requesting comments, and people gave those, I made some cleanups and modifcations, but this PR as is, I consider "done".. I'm sure it'll evolve over time...

The server... that is a whole different story... I'm fully willing to support it, and keep a public server going on one of my dedicated machines (hosted by delimited.com). But in terms of development, I cannot spend any Blizzard time on that, and so it comes down to the 3-4 hours a month I have at home. Any community involvement would greatly accelerate that.

tritao:
What's needed on the server front?

As an alternative to a stateful server, what do you think about using a Github repository as a stateless index? Libraries could be managed via PRs and a Travis CI job could be setup that automatically recomputes a commits a json/Lua table index file cache on the repository.
Package files could be fetched/uploaded to any S3-compatible file store.

CC'ing @realvictorprm since he mentioned to me some interest on helping on this area before.

tvandijck:
> What's needed on the server front?
it's a single static webpage right now ;) with one "module" as a test module...

> As an alternative to a stateful server, what do you think about using a Github repository as a stateless index? 

Yeah, whatever we want really... That said, having an intermediate 'server' that serves the content has proven to be the right thing to do... rather then directly talking to the github API, which often is rate limitted and requires authentication.. which I really don't want to add to my code.

One thing I did do, setup oauth authentication with github, with the intent that we could make it setup webhooks etc.... If we're going with a fixed index in a single repository, then I don't think we need any of that, since we could just setup a single API Token to fetch that index when it changes, and process it to serve the new content... But again, the server code is all extremely simple right now, and anyone working on it, can rip it apart as much as they want...

As for storing files on an S3 file store... again, I really don't care much either way... AWS costs money, I'm already paying 200,= a year per dedicated box I own, which I'm offering to the community to keep up, I think they have 500GB of storage each, but I don't have any backup mechanism setup, but that is easily fixed.

tvandijck:
Anyway, I'm merging this PR then... @starkos & @samsinsane approved it...

Lets take the server discussion to the group.
https://groups.google.com/forum/#!topic/premake-development/U9fLftNY4Vg

PR #623, opened 2016-11-08
Added C90 and C99 flags to gcc

mikisch81:
We needed to build our project for c99.
I saw that C++11 & C++14 was already addded as flags, so why not adding C90 & C99 as well..

Explanation about the other changes:
1. In make_cpp.lua: make.cxxFlags(cfg, toolset) which generates the ALL_CXXFLAGS makefile variable, $(ALL_CFLAGS) should not be there as these are specific flags for C files only. On the other hand $(ALL_CPPFLAGS) should be there and wasn't. In our case there could be a project with .c and .cpp files and -std=c99 can't co-exist with -std=c++14.
2. To support this split, in gcc.lua I split gcc.cflags to 2 tables: gcc.shared and gcc.cflags.
gcc.getcflags() will return shared flags, c flags and warnings.
gcc.getcxxflags() will return shared flags, c++ flags and warnings.

Currently only 1 test in the codelite module tests fails, would appreciate if anyone can help with that . :-)
@mindw
@alzix

### DIFF
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -196,7 +196,7 @@
 		end
 
 		local toolset = m.getcompiler(cfg)
-		local cxxflags = table.concat(table.join(toolset.getcflags(cfg), toolset.getcxxflags(cfg), cfg.buildoptions), ";")
+		local cxxflags = table.concat(table.join(toolset.getcxxflags(cfg), cfg.buildoptions), ";")
 		local cflags   = table.concat(table.join(toolset.getcflags(cfg), cfg.buildoptions), ";")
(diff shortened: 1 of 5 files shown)


tvandijck:
this would be easier to fix if we merged the submoly PR first...
https://github.com/premake/premake-core/pull/625

other then that, the commit looks fine with me..

mikisch81:
Fixed codelite module failure (after rebasing and got rid of the git submodules).
clang tool was needed some handling..

tvandijck:
Can you please document these new flags here:
https://github.com/premake/premake-core/wiki/flags

PR #624, opened 2016-11-09
Add `softrequire()`

starkos:
Behaves just like the existing `require()` call, but returns `nil` if the module does not exist instead of raising an error.

### DIFF
--- a/src/base/globals.lua
+++ b/src/base/globals.lua
@@ -95,10 +95,14 @@
 ---
 
 	function softrequire(modname, versions)
+		local loaded = table.shallowcopy(package.loaded)
+
 		local ok, mod = pcall(require, modname, versions)
+
 		if ok then
 			return mod
 		else
+			package.loaded = loaded
 			return nil
 		end
 	end
--- a/src/base/globals.lua
+++ b/src/base/globals.lua
@@ -4,6 +4,8 @@
 -- Copyright (c) 2002-2014 Jason Perkins and the Premake project
 --
 
+	local p = premake
+
 
 --
 -- Find and execute a Lua source file present on the filesystem, but
@@ -48,7 +50,7 @@
 	io._includedFiles = {}
 
 	function include(fname)
-		local fullPath = premake.findProjectScript(fname)
+		local fullPath = p.findProjectScript(fname)
 		fname = fullPath or fname
 		if not io._includedFiles[fname] then
 			io._includedFiles[fname] = true
@@ -73,14 +75,30 @@
 --    global package.loaded table.
 ---
 
-	premake.override(_G, "require", function(base, modname, versions)
-		local result, mod = pcall(base,modname)
-		if not result then
-			error( mod, 3 )
+	p.override(_G, "require", function(base, modname, versions)
+		local ok, mod = pcall(base, modname)
+		if not ok then
+			error(mod, 3)
 		end
-		if mod and versions and not premake.checkVersion(mod._VERSION, versions) then
+
+		if mod and versions and not p.checkVersion(mod._VERSION, versions) then
 			error(string.format("module %s %s does not meet version criteria %s",
 				modname, mod._VERSION or "(none)", versions), 3)
 		end
+
 		return mod
 	end)
+
+
+---
+-- Returns the specified module if it exists, or `nil` otherwise.
+---
+
+	function softrequire(modname, versions)
+		local ok, mod = pcall(require, modname, versions)
+		if ok then
+			return mod
+		else
+			return nil
+		end
+	end
--- a/tests/_tests.lua
+++ b/tests/_tests.lua
@@ -13,6 +13,7 @@ return {
 	"base/test_override.lua",
 	"base/test_path.lua",
 	"base/test_premake_command.lua",
+	"base/test_softrequire.lua",
 	"base/test_table.lua",
 	"base/test_tree.lua",
 	"base/test_uuid.lua",
--- a/tests/base/test_softrequire.lua
+++ b/tests/base/test_softrequire.lua
@@ -6,13 +6,29 @@
 
 	local suite = test.declare("base_softrequire")
 
+	local TEST_MODULE = {
+		_VERSION = "5.0"
+	}
+
+
+	function suite.setup()
+		package.preload.TEST_MODULE = function()
+			return TEST_MODULE
+		end
+	end
+
+	function suite.teardown()
+		package.loaded.TEST_MODULE = nil
+		package.preload.TEST_MODULE = nil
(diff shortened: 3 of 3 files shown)


src/base/globals.lua
```diff
@@ -73,14 +75,30 @@
...
+
 		return mod
 	end)
+
+
+---
+-- Returns the specified module if it exists, or `nil` otherwise.
+---
+
+	function softrequire(modname, versions)
```
tvandijck: This will also return 'nil' if the version is incorrect... it will however still have loaded the module.

starkos: Ah, good point. I'll fix it up.

starkos:
Patched…if the soft require fails due to a version mismatch, the module's load state is left as it was before the call.

tvandijck:
I think this OK... my only comment would be that _G is still modified, so if a module did create methods or other tables into the global namespace, those would all be there either way.

That said, fixing that would just be incredibly hard, unless we loaded a specific script in the module first that had just it's version number in it or something.. But then we're making the whole module system so much more complicated..

I think we can always revisit this if this turns out to become an issue.

starkos:
Give me a chance to think about it a little more before you merge it. Maybe I can come up with something.

starkos:
Okay, I don't have any clever way around the "must load the module to get the version" issue. If you're okay with this PR as-is, I will plan to mention it as a caveat in the documentation. Seem reasonable?

tvandijck:
Yeah, no object from my end...

starkos:
How do people feel about adding a `loaded()` call to the Premake module spec? If it exists, it would be called once after the module is first loaded. This would provide a place to put your initialization logic that would only be triggered *after* the module's version check.

tvandijck:
I'd rather see a _metadata.lua spec that MUST return a table...

for example:
```lua
return {
  version = '1.2.0',
  author = 'tom van dijck',
  description = 'this is an example module',
}
```

then the softrequire (and require) can load that first, before trying anything else.
I think we can also run this file in a completely separate context with no access to any other lua functionality, state or anything.

samsinsane:
> I'd rather see a _metadata.lua spec 

Like a [table of contents](http://wowwiki.wikia.com/wiki/TOC_format) or something? 😜 

On a more serious note, I think that's a pretty good idea and definitely something we should look into. Should we just extend the _manifest.lua?
```lua
return {
  version = '1.2.0',
  author = 'tom van dijck',
  description = 'this is an example module',
  files = {
    'blizzard.lua',
    'util.lua',
    'package.lua',
    'packageman.lua',
    'cache.lua',
    'consoles.lua',
    'context.lua',
    'prebake.lua',
    'unity.lua',
    'vpaths.lua',
    'json.lua',
    'visualsvn.lua',
    'export.lua',
  },
  -- future things
}
```

While I'm not a fan of breaking the current functionality (which we could fall back to), I think it's better to keep the module system simple. We don't want to end up having a half dozen mandatory files to create a module. Thoughts?

starkos:
I agree with this. I'll look into it.

starkos:
Is there any value to merging this as is? Or is it too dangerous with the broken version check?

samsinsane:
> Is there any value to merging this as is?

I would say there is value, you've added `softrequire` which alone is enough.

> Or is it too dangerous with the broken version check?

It's probably better to keep the new manifest changes separate to this. It introduces people to the idea of `softrequire` and then in a few weeks we can release a bit of an "overhaul" of the module manifest.

I can't really see a reason why the broken version check would be dangerous though. So I guess if there's an argument against it, then my opinion might change?

starkos:
The problem is that by the time we get the version the module script has already been evaluated. If the module has made changes to the global environment, or installed overrides to other code, those will still be in place even if the version check fails.

This isn't a problem for `require()` because a version mismatch will hard fail and end everything then and there.

samsinsane:
> those will still be in place even if the version check fails.

That's not ideal, but the modules I've dealt with are dormant until some additional conditions are satisfied.

While not a great solution, maybe we should emit a warning that `softrequire` has failed to load the specified version of the module and currently this feature can result in unintended side effects? That's too long, but we emit warnings for deprecated APIs, why not emit warnings for "beta" features with known issues? I'm not overly attached to this idea, it seems like a pretty slippery slope. Thoughts?

tvandijck:
I'd merge it as is, and indeed just add a warning in both the documentation and the code on a version check failure.... We know this is a potential issue, we have an idea on how to fix it in the future, lets merge this to get the API down, and then work on the fix..

TurkeyMan:
So, I just had an idea... and then skipped half the coversation ;)
The issue is that loading the module may pollute `_G`?
It might be possible to create a temporary table, assign it a metatable, and assign `_G` to the metatable's `__index`... then `_G` will be available while the module is considering if it should be loaded, but any writes to `_G` will go to the temp table.
After it's loaded, you can restore `_G` and if load was successful, copy all contents of the temp table into `_G`.
If load failed, then would-be pollution to `_G` will just disappear.

...?

starkos:
> It might be possible to create a temporary table, assign it a metatable, and assign _G

I did consider it (Lua is the dark side, I think). But since our module system is still relatively new, Tom's suggestion of adding a ToC, even if it makes backwards compatibility a little weird, seems like the better long term solution.

tvandijck:
> It might be possible to create a temporary table, assign it a metatable, and assign _G to the metatable's __index, then assign the temp table to _G. 

tell me more... I actually want to do this for something else as well...
there is this page on sandboxing lua: 
http://lua-users.org/wiki/SandBoxes

but in premake that doesn't seem to actually work, _G is still completely populated with all the public global methods premake exposes. Basically a pure sandbox would have an empty _G, but I think for that we would need an actual C function to create an entirely new context..

I didn't go that route yet..  but if that is useful for other things, like the module loading, then I'm totally in favor of adding such functionality...

starkos:
So, the problem with this one is that the return value mixes together "module does not exist" with "module does exist but has issues". We really it to:

1. Return `nil` if the module doesn't exist
2. Return `nil` if the version requirement isn't met, *without actually loading the module*
3. If it does exist and the version checks out, load the module and pass any errors through to the caller, matching the behavior of `require()`.

At the moment we don't have calls to handle (1) or (2). We should probably have a call like `locateModule()` which returns the path to the module if found (and maybe a `moduleExists()` too). Then we'd need to break `_VERSION` out from the rest of the module so we could check it without loading (put it in the manifest maybe?).

Rather than continuing to leave this PR to rot, I opened issue #1344, which link here and to the original issue. If someone wants to tackle this, feel free to revive this PR or create a new one.

PR #625, opened 2016-11-09
Move core modules into main repository; drop submodules

starkos:
As discussed numerous times, get rid of the submodule approach for core modules and just included the sources directly in the main repository.

I have no idea how this will play with existing working copies. It might be necessary to [manually clear out the existing submodules](http://stackoverflow.com/questions/1260748) first.


tvandijck:
We should probably also delete the repositories for these submodules once this is merged...

starkos:
Thanks! I'll take care of other repositories.

starkos:
Fixed with PR #648.

PR #628, opened 2016-11-18
Add NoLinkObjects flag

aleksijuvani:
The default behaviour for the Visual Studio and Makefile exporters is to link .obj files if a custom build command outputs them. There are some cases where we don't want this to happen, such as when dealing with Wavefront .obj model files.

I ran into this problem when trying to copy model files around with custom build commands:

```lua
filter "**/models/**.obj"

	-- Copy these files into the target directory while preserving the folder
	-- structure.

	buildcommands
	{
		os.translateCommands "{mkdir} \"%{ path.join(cfg.buildtarget.directory, path.getdirectory(file.relpath)) }\"",
		os.translateCommands "{copy} \"%{ file.relpath }\" \"%{ path.join(cfg.buildtarget.directory, path.getdirectory(file.relpath)) }\""
	}

	buildoutputs "%{ path.join(cfg.buildtarget.directory, file.relpath) }"
```

The linker would try to link these files, resulting in a build error. This can now be solved by specifying the new NoLinkObjects flag for these files:

```lua
filter "**/models/**.obj"

	flags "NoLinkObjects"
````

Please let me know if there's anything that I missed or anything that needs changing.

### DIFF
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -257,12 +257,8 @@
 					for cfg in project.eachconfig(prj) do
 						local filecfg = incfg[cfg]
 						if filecfg then
-							-- If the custom build outputs an object file, we
-							-- add it to the link step automatically to match
-							-- Visual Studio. This behaviour can be turned off
-							-- by specifying the NoLinkObjects flag.
 							local output = project.getrelative(prj, filecfg.buildoutputs[1])
-							if path.isobjectfile(output) and not filecfg.flags.NoLinkObjects then
+							if path.isobjectfile(output) and (filecfg.linkbuildoutputs == true or filecfg.linkbuildoutputs == nil) then
 								table.insert(configs[cfg].objects, output)
 							else
 								table.insert(configs[cfg].customfiles, output)
--- a/src/actions/make/make_cpp.lua
+++ b/src/actions/make/make_cpp.lua
@@ -257,10 +257,12 @@
 					for cfg in project.eachconfig(prj) do
 						local filecfg = incfg[cfg]
 						if filecfg then
-							-- if the custom build outputs an object file, add it to
-							-- the link step automatically to match Visual Studio
+							-- If the custom build outputs an object file, we
+							-- add it to the link step automatically to match
+							-- Visual Studio. This behaviour can be turned off
+							-- by specifying the NoLinkObjects flag.
 							local output = project.getrelative(prj, filecfg.buildoutputs[1])
-							if path.isobjectfile(output) then
+							if path.isobjectfile(output) and not filecfg.flags.NoLinkObjects then
 								table.insert(configs[cfg].objects, output)
 							else
 								table.insert(configs[cfg].customfiles, output)
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1106,8 +1106,8 @@
 
 
 	function m.linkObjects(fcfg, condition)
-		if fcfg.linkbuildoutputs == false then
-			m.element("LinkObjects", condition, "false")
+		if fcfg.linkbuildoutputs ~= nil then
+			m.element("LinkObjects", condition, tostring(fcfg.linkbuildoutputs))
 		end
 	end
 
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -719,6 +719,7 @@
 				m.excludedFromBuild,
 				m.buildCommands,
 				m.buildOutputs,
+				m.linkObjects,
 				m.buildMessage,
 				m.buildAdditionalInputs
 			}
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by samsinsane** [comment]
There's an API to stop files generated by the build command system from building, `compilebuildoutputs`. Does this fix the issue for you?
```lua
filter "**/models/**.obj"

	-- Copy these files into the target directory while preserving the folder
	-- structure.

	buildcommands
	{
		os.translateCommands "{mkdir} \"%{ path.join(cfg.buildtarget.directory, path.getdirectory(file.relpath)) }\"",
		os.translateCommands "{copy} \"%{ file.relpath }\" \"%{ path.join(cfg.buildtarget.directory, path.getdirectory(file.relpath)) }\""
	}

	buildoutputs "%{ path.join(cfg.buildtarget.directory, file.relpath) }"
	compilebuildoutputs "Off"
```

aleksijuvani:
It does not. It doesn't even seem to affect the generated project files in any way.

samsinsane:
Right, the gmake exporter kind of ignores the new `compilebuildoutputs` API. I guess technically it's not "compiling" the build output, but I see no reason why the API should be limited to compiling. I could be wrong, but I think the best solution would be to change the gmake exporter, around where you've edited, to handle the `compilebuildoutputs` API.

There's a few things that would need to change, [this](https://github.com/aleksijuvani/premake-core/blob/1a37f3de07865a901425a7bfc68118fa312141d1/src/actions/make/make_cpp.lua#L216) and [this](https://github.com/aleksijuvani/premake-core/blob/1a37f3de07865a901425a7bfc68118fa312141d1/src/actions/make/make_cpp.lua#L221) should be removed. Then [this](https://github.com/aleksijuvani/premake-core/blob/1a37f3de07865a901425a7bfc68118fa312141d1/src/actions/make/make_cpp.lua#L227-L272) should become something like this:
```lua
-- identify the file type
local kind
local ext
if path.iscppfile(node.abspath) then
  kind = "objects"
  ext = ".o"
elseif path.isresourcefile(node.abspath) then
  kind = "resources"
  ext = ".res"
elseif node.generated and path.isobjectfile(node.abspath) then
  kind = "objects"
elseif node.generated then
  kind = "customfiles"
end

-- skip files that aren't compiled
if not kind then
  return
end

local objectname
if ext then
  objectname = "$(OBJDIR)/" .. node.objname .. ext
else
  objectname = node.relpath
end

-- if this file exists in all configurations, write it to
-- the project's list of files, else add to specific cfgs
if inall then
  table.insert(root[kind], objectname)
else
  for cfg in project.eachconfig(prj) do
    if incfg[cfg] then
      table.insert(configs[cfg][kind], objectname)
    end
  end
end
```

I'm not 100% sure if this will work, but something like this would be compatible with the `compilebuildoutputs` API. Using that API makes sense to me, but I could be wrong about this?

aleksijuvani:
If we did do this, it would no longer match current behaviour for .obj files, since this API defaults to off. This change could break existing projects, and it would no longer match default behaviour for Visual Studio.

aleksijuvani:
If we bundled this behaviour into `compilebuildoutputs`, we could do something weird like adding a default rule that turns this on for .obj files, but I don't think that's a good way to go about doing this.

I would rather keep this as a separate option.

Also, while a purely cosmetic change, it might make sense to change this from a flag to a `linkbuildoutputs` API, so that it matches how `compilebuildoutputs` works.

@tvandijck Any thoughts on all this? Looks like you wrote the `compilebuildoutputs` API in the first place.

tvandijck:
the compilebuildoutputs api was added to compile c/cpp outputs from for example codegen solutions like google protobufs... I think it would be confusing to mix 'object' files into this...

I'm not against a "linkbuildoutputs" API. I certainly would prefer it over a flag, as we're mostly shying away from adding more flags...

src/actions/make/make_cpp.lua
```diff
@@ -257,10 +257,8 @@
 					for cfg in project.eachconfig(prj) do
 						local filecfg = incfg[cfg]
 						if filecfg then
-							-- if the custom build outputs an object file, add it to
-							-- the link step automatically to match Visual Studio
 							local output = project.getrelative(prj, filecfg.buildoutputs[1])
-							if path.isobjectfile(output) then
+							if path.isobjectfile(output) and (filecfg.linkbuildoutputs == true or filecfg.linkbuildoutputs == nil) then
```
samsinsane: This could just be `and filecfg.linkbuildoutputs ~= false`, right?

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1104,6 +1105,13 @@
 	end
 
 
+	function m.linkObjects(fcfg, condition)
+		if fcfg.linkbuildoutputs == false then
+			m.element("LinkObjects", condition, "false")
```
samsinsane: Should this be:
```lua
if fcfg.linkbuildoutputs then
  m.element("LinkObjects", condition, fcfg.linkbuildoutputs)
```

Generally, an element is emit when the user specifies a value for it, regardless of whether it's the default or not.

src/actions/make/make_cpp.lua
samsinsane: I thought it occurred in the codebase all over the place, but it doesn't, ignore me. Hopefully, this is the last bad suggestion I make on this PR, haha.

src/actions/vstudio/vs2010_vcxproj.lua
aleksijuvani: You meant `fcfg.linkbuildoutputs ~= nil`, right? Done. I added some tests while I was at it.

tvandijck:
Can you make sure you document this on the wiki here as well?
https://github.com/premake/premake-core/wiki/Project-API

aleksijuvani:
Done.

PR #635, opened 2016-11-22
Makefile generation - Don't force serial compilation for all projects

robertop:
Only restrict parallel compilation when using precompiled
headers. Currently, the generated makefile has a `.NOTPARALLEL`
target, which means that make will ignore parallel builds, even
for projects which can build in parallel just fine.

### DIFF
--- a/src/actions/make/_make.lua
+++ b/src/actions/make/_make.lua
@@ -144,8 +144,19 @@
 		_p('')
 
 		if kind == "workspace" then
-			_p('.NOTPARALLEL:')
-			_p('')
+			local haspch = false
+			for _, prj in ipairs(target.projects) do
+				for cfg in project.eachconfig(prj) do
+					if cfg.pchheader then
+						haspch = true
+					end
+				end
+			end
+
+			if haspch then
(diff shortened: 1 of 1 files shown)


src/actions/make/_make.lua
```diff
@@ -144,8 +144,17 @@
 		_p('')
 
 		if kind == "workspace" then
-			_p('.NOTPARALLEL:')
-			_p('')
+			local haspch = false
+			for _, prj in ipairs(target.projects) do
+				if prj.pchheader then
```
tvandijck: unfortunately the scope of pchheader == config... 
```lua
	api.register {
		name = "pchheader",
		scope = "config",
		kind = "string",
		tokens = true,
	}
```

meaning that looking at this value at the project level is insufficient. 

```lua
project 'a'
    filter { 'configuration:debug' }
        pchheader 'b.h'
```

haspch would be false in this case.

robertop: Thanks. Updated the PR to account for pchheaders at the config level. 

Rebased from master.

PR #637, opened 2016-11-23
Fix VS'15 <GenerateDebugInformation> to match Update 3

starkos:
As of VS'15 Update 3, the correct values for <GenerateDebugInformation> are "true", "false", and "DebugFastLink". Previous of versions defaulted this element to "true", but would change it to "Debug" is set via the property page; that's been fixed in Update 3. The use of "true" and "false" is compatible with the earlier versions of VS'15.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -391,12 +391,14 @@
 		if cfg.kind == p.STATICLIB then
 			return {
 				m.subSystem,
+				m.fullProgramDatabaseFile,
 				m.generateDebugInformation,
 				m.optimizeReferences,
 			}
 		else
 			return {
 				m.subSystem,
+				m.fullProgramDatabaseFile,
 				m.generateDebugInformation,
 				m.optimizeReferences,
 				m.additionalDependencies,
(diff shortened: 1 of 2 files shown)


tvandijck:
We need a way to test for compile version.... SC2/Heroes still use Update 1 and for various reasons upgrading is simply not an option... this fix, while correct, breaks our build...

It would be awesome if there is some way we can simply run "msdev.exe --version" or whatever and actually know which one we're targeting.. I guess in the meantime I can simply introduce a "vs2015.1" action... but stuff like this is annoying...

starkos:
Hmm…I did test with Update 1 before I pushed the PR and it *seemed* fine. That's unfortunate (and annoying, yes). Can you provide more information on what breaks, if only for future support questions?

> It would be awesome if there is some way we can simply run "msdev.exe --version" or whatever and actually know which one we're targeting

That might be tricky to do in the public version, since we have so little control over the runtime environment. But if you can count on Visual Studio being on the PATH, you could run `devenv /?` and parse out the version string from the first line of output, and then override the `generateDebugInformation()` call if it hits your version range?

Another option might be a compatibility flag, maybe `premake --update1 vs2015` or something. That is something we could put in the public version easily enough, though I don't know if it is any better than an entirely new action.

Issue #578, opened 2016-09-15
Premake bootstrap does not work w/ just 'Visual C++ Build Tools'


nta:
```
        devenv .\build\bootstrap\Premake5.sln /Upgrade
'devenv' is not recognized as an internal or external command,
operable program or batch file.
NMAKE : fatal error U1077: 'devenv' : return code '0x1'
Stop.
```

The [standalone build tools](https://blogs.msdn.microsoft.com/vcblog/2016/03/31/announcing-the-official-release-of-the-visual-c-build-tools-2015/) do not contain the IDE, including the `devenv.com` command-line binary for the IDE, and the shipped bootstrap makefile expects this to exist for /Update and /Build.

tvandijck:
The question is... how do you build a .sln without devenv? should we use msbuild instead? the bootstrap for windows indeed does assume you have visual studio installed, since that is what the bootstrap targets...

since this fails in the step where it tries to build the .sln... it means it already generated one through the bootstrap binary... so if you know what command line to use to build that particular SLN, you should be able to get a full premake binary, and also tell us what command line you used, so we could use that instead of the devenv command line.

ntauthority:
>The question is... how do you build a .sln without devenv? should we use msbuild instead? the bootstrap for windows indeed does assume you have visual studio installed, since that is what the bootstrap targets...

Yes, msbuild will convert a .sln to an internal project representation upon loading for compatibility behavior. Of course, /Upgrade doesn't exist here and it will use whatever toolset is set in the solution, so it should presumably be an override like the MSDEV variable.

tvandijck:
I think you could easily try this out in the bootstrap.mak, change the windows target to use msbuild, and don't specify /upgrade.... I'd potentially just make a different target for it... so you could do:

```nmake -f Bootstrap.mak MSDEV=vs2015 windows-standalone```

or something.

samsinsane:
I don't see any reason why the `devenv /Upgrade` command should be called if you specify `MSDEV`. Furthermore, I don't see any reason why we'd continue to use `devenv /Upgrade` when the developer can just specify their version of VS via `MSDEV`.

PR #639, opened 2016-11-28
Add MSBuild target to Bootstrap.mak

nta:
As I [mentioned](https://github.com/premake/premake-core/issues/578) a while ago, Bootstrap.mak will not directly work when no full VS install is present.

This attempts to resolve said issue by adding a separate target (`windows-msbuild`) to the Makefile.

Another solution would be to handle upgrade behavior based on the MSDEV variable and use `msbuild` to build in all cases, but this would break (or require conditions, and I'm not too sure how well cross-platform Makefiles like those) for VS2008< which use VCBuild project files rather than the VS2010+ MSBuild .vcxproj w/ toolset configuration, and might potentially cause different behavior for build processes relying on the current `windows` target.

For VS2010+, there is no functional difference between binaries built by either method (except not specifying an explicit MSDEV argument will default to `v110` due to the vs2012 toolchain setting - and these will not be upgraded) as `devenv /Build` will dispatch to MSBuild for these projects anyway (though handling solution parsing internally, rather than using MSBuild compatibility behavior).

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -65,10 +65,15 @@ linux: $(SRC)
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake
 	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN`
 
-windows: $(SRC)
+windows-base: $(SRC)
 	if not exist build\bootstrap (mkdir build\bootstrap)
 	cl /Fo.\build\bootstrap\ /Fe.\build\bootstrap\premake_bootstrap.exe /DPREMAKE_NO_BUILTIN_SCRIPTS /I"$(LUA_DIR)" user32.lib ole32.lib $**
 	.\build\bootstrap\premake_bootstrap.exe embed
 	.\build\bootstrap\premake_bootstrap --to=build/bootstrap $(MSDEV)
+
+windows: windows-base
 	devenv .\build\bootstrap\Premake5.sln /Upgrade
 	devenv .\build\bootstrap\Premake5.sln /Build Release
+
+windows-msbuild: windows-base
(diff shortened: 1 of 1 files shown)


tvandijck:
Also, I'm not sure how important it really is to keep VS2008 support... I don't even thing premake would build with VS2008 right now. I can test if anyone is interested..

nta:
Testing this would indeed be appreciated - if so this could be simplified to _always_ build using MSBuild and use a batch-style `if` condition (as done for file existence already, so the `windows` target depended on nmake anyway) to not upgrade if a specific `MSDEV` variable is specified.

A more decent solution would be checking environment variables to default to the toolset version that has currently had the `vcvars*,bat` file run, but within the makefile context that might be a problem.

tvandijck:
I'll see if I can spin up a VM with VS2008 and give this a shot...

tvandijck:
OK, finally had the time to check this in a VM with VS2008...

The current codebase does not build with VS2008.. 
Neither zlib or mbedtls that is embedded in premake compile at all, and the the bootstrap part itself, does not compile due to string_sha1.c's use of <stdint.h>, zlib and mbedtls suffer the same issues. Disabling that however gives me a successful version of premake without zlib or http support.

Quite honestly, I think it is going to be increasingly difficult to keep supporting an 8 year old version of Visual Studio. Certainly now that Visual Studio for non-commercial use is effectively free through the community editions. 

We're going to have to decide at some point.... "we gotta move on"..

starkos:
FWIW, I'm okay with "gotta move on".

tvandijck:
@nta @starkos OK, in that case I guess we can just always use MSBuild to build through the bootstrap.

PR #641, opened 2016-11-29
Fix bugs introduced by shared compiler flags

starkos:
Fix a few spots where the shared compiler flag changes (PR #623) caused things to misbehave.

- Fixed duplicate flags in VC 200x external compiler options
- Added missing C and CXX flag maps to Clang adapter
- Reworked SNC adapter to follow same shared flags approach

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -60,22 +60,21 @@
 		symbols = gcc.shared.symbols
 	}
 
+	clang.cflags = table.merge(gcc.cflags, {
+	})
+
 	function clang.getcflags(cfg)
-		local shared_flags = config.mapFlags(cfg, clang.shared)
+		local shared = config.mapFlags(cfg, clang.shared)
+		local cflags = config.mapFlags(cfg, clang.cflags)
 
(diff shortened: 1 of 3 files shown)


src/tools/clang.lua
```diff
@@ -60,22 +60,21 @@
 		symbols = gcc.shared.symbols
 	}
 
-	function clang.getcflags(cfg)
-		local shared_flags = config.mapFlags(cfg, clang.shared)
+	clang.cflags = table.merge(gcc.cflags, {
```
tvandijck: can't this just be ```clang.cflags = gcc.cflags```

src/tools/clang.lua
```diff
@@ -90,13 +89,13 @@
 --    An array of C++ compiler flags.
 --
 
-	function clang.getcxxflags(cfg)
-		local shared_flags = config.mapFlags(cfg, clang.shared)
+	clang.cxxflags = table.merge(gcc.cxxflags, {
```
tvandijck: can't this just be clang.cxxflags = gcc.cxxflags

PR #643, opened 2016-11-30
Fix VS mapping for `file.path` tokens

starkos:
The `%{file.path}` token is an alias of `%{file.relpath}`, and should map to the same VS symbol. Specifically, it should be the relative path including the file name, and not just the path.


starkos:
I'm not sure why this one failed, the message is:

     warning: Could not find remote branch fix-vs-file-path-map to clone. 
     fatal: Remote branch fix-vs-file-path-map not found in upstream origin

Safe to ignore?

starkos:
Merging this despite the one CI failure; not sure why it couldn't see the branch when the other tests were passing.

PR #644, opened 2016-12-01
Don't treat "dependson" libraries as "links" libraries in Xcode

### LINKED ISSUE #631: Xcode: App that "dependson" a SharedLib creates "links" relationship instead
We have a repository that has several executables as well as a bunch of dynamic libraries which can be dynamically loaded by one of the executables at runtime using dlopen(). I want to set it up so that building the main executable builds the other executables and all the dynamic libraries as well. However, in the exported Xcode project, when I specify that an executable `dependson` the shared libraries, it `links` the executable to them as well, which is not what we want (it causes the executable to fail to load if the libraries are unavailable).

Example:

    workspace "test"
    configurations "Debug"
    language "C"

    project "test"
    kind "ConsoleApp"
    files "test.c"
    dependson "testlib"

    project "testlib"
    kind "SharedLib"
    files "testlib.c"

This exports the following:

screen shot 2016-11-19 at 10 00 44 pm

The "Target Dependencies" section is correct, but I don't want to actually link to that library. When I check the executable, I do see a link dependency:

    $ otool -L test
    test:
        /usr/local/lib/libtestlib.dylib (compatibility version 1.0.0, current version 1.0.0)

If I correctly understand the difference between `links` and `dependson`, a built library should not be linked if the relationship is `dependson`. A workaround for us might be to create a dummy executable that `dependson` the dynamic library, then the executable we build would not be affected. However, if this is in fact unexpected behavior, then perhaps others would benefit from a fix as well.

macsforme:
This patch corrects an issue with the xcode module where if you list a library you're building under "dependson," Xcode would add it to the target dependencies but also to the link dependencies, which is undesired. With this fix, you can now use "links" and "dependson" as desired and they both work as intended.

Fixes #631.

### DIFF
--- a/modules/xcode/xcode_project.lua
+++ b/modules/xcode/xcode_project.lua
@@ -88,7 +88,7 @@
 
 		-- the special folder "Projects" lists sibling project dependencies
 		 tr.projects = tree.new("Projects")
-		 for _, dep in ipairs(project.getdependencies(prj, "sibling", "object")) do
+		 for _, dep in ipairs(project.getdependencies(prj)) do
 		 	-- create a child node for the dependency's xcodeproj
 		 	local xcpath = xcode.getxcodeprojname(dep)
 		 	local xcnode = tree.insert(tr.projects, tree.new(path.getname(xcpath)))
@@ -105,6 +105,14 @@
 		 	node = tree.insert(xcnode, tree.new(cfg.linktarget.name))
 		 	node.path = cfg.linktarget.fullpath
 		 	node.cfg = cfg
+
+			-- don't link the dependency if it's a dependency only
+			for _, deponly in ipairs(project.getdependencies(prj, "dependOnly")) do
+				if dep == deponly then
+					node.nobuild = true
+					break
+				end
+			end
 		end
 
 		 if #tr.projects.children > 0 then
@@ -120,7 +128,7 @@
 		 		node.isResource = xcode.isItemResource(prj, node)
 
 		 		-- assign build IDs to buildable files
-		 		if xcode.getbuildcategory(node) then
+		 		if xcode.getbuildcategory(node) and not node.nobuild then
 		 			node.buildid = xcode.newid(node.name, "build", node.path)
 		 		end
 
--- a/modules/xcode/xcode_project.lua
+++ b/modules/xcode/xcode_project.lua
@@ -108,7 +108,7 @@
 		 		node.isResource = xcode.isItemResource(prj, node)
 
 		 		-- assign build IDs to buildable files
-		 		if xcode.getbuildcategory(node) and not node.nobuild then
+				if xcode.getbuildcategory(node) and not node.excludefrombuild then
 		 			node.buildid = xcode.newid(node.name, "build", node.path)
 		 		end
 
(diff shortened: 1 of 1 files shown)


macsforme:
Any thoughts? I'm trying to port the [BZFlag](https://github.com/macsforme/bzflag/tree/premake) build system to premake5, but it will be easier for me to sell our other developers on it if our premake scripts don't require workarounds, such as for this bug.

macsforme:
I can rework it to get rid of the double negative wording.

> Consider making the internals of the loop a function instead with a simple build/link parameter.

Can you explain more what you mean by this? I'm new to lua, but from what I've read the only way to test whether a value is in a table is by iterating through it. So, as long as we depend on the existing `project.getdependencies` to distinguish between `links` and `dependson`, I don't see any other way except iterating through the list.

I could, however, probably move my call to `project.getdependencies("dependOnly")` outside of the parent loop, since that result doesn't change between dependencies (I just realized that). That should increase efficiency somewhat.

samsinsane:
Basically, something like this:
```lua
function xcode.addDependency(tr, dep, build)
  -- create a child node for the dependency's xcodeproj
  local xcpath = xcode.getxcodeprojname(dep)
  local xcnode = tree.insert(tr.projects, tree.new(path.getname(xcpath)))
  xcnode.path = xcpath
  xcnode.project = dep
  xcnode.productgroupid = xcode.newid(xcnode.name, "prodgrp")
  xcnode.productproxyid = xcode.newid(xcnode.name, "prodprox")
  xcnode.targetproxyid  = xcode.newid(xcnode.name, "targprox")
  xcnode.targetdependid = xcode.newid(xcnode.name, "targdep")

  -- create a grandchild node for the dependency's link target
  local lprj = premake.workspace.findproject(prj.workspace, dep.name)
  local cfg = project.findClosestMatch(lprj, prj.configurations[1])
  node = tree.insert(xcnode, tree.new(cfg.linktarget.name))
  node.path = cfg.linktarget.fullpath
  node.cfg = cfg

  if build == false then
    node.nobuild = true
  end
end

-- Back inside xcode.buildprjtree
for _, dep in ipairs(project.getdependencies(prj, "linkOnly")) do
  addDependency(tr, dep, true)
end

for _, dep in ipairs(project.getdependencies(prj, "dependOnly")) do
  addDependency(tr, dep, false)
end
```

macsforme:
I fixed the inefficiency per your suggestion.

I looked at alternatives for the `not nobuild` line, but because some nodes are created outside of that loop, if there was a `build` variable which could be `true`, `false`, or `nil`, we would have to include items where `build` is either `true` or `nil`. Otherwise, we could test for `build != false` as you suggested, and exclude those files, but it seems to me that having a significant distinction between `false` and `nil` would be more confusing than the original `not nobuild` test (and kind of backwards from the conventional `if` test). If that's what you want I'll change it, but to me it seems less confusing the way it is now.

samsinsane:
Maybe renaming `nobuild` to `exclude` would be the better solution? It removes the issue of the variable name, and avoids the issue you described above. At this point I feel like I'm being a bit pedantic about this, but I've seen this trip people up before, plus it reads a bit silly "if not no build then" as opposed to "if build then" or "if not exclude then".

macsforme:
Changed `nobuild` to `excludefrombuild` (so it's clear what the item is being excluded from).

PR #649, opened 2016-12-14
Fix VS2015 sln file icon

jstewart-amd:
This is a minor fix to make the correct sln file icon appear in Windows.
The Icon Handler for sln files is apparently quite picky.

VS2015 sln files contain this:
`# Visual Studio 14`

Not this:
`# Visual Studio 2015`

Having 2015 instead of 14 does not appear to cause any major issues. But
you get the generic sln file icon in Explorer instead of the
VS2015-specific one.

Test results:
.\bin\release\premake5.exe test
Running action 'test'...
1435 tests passed, 0 failed in 1.76 seconds

### DIFF
--- a/src/actions/vstudio/vs2015.lua
+++ b/src/actions/vstudio/vs2015.lua
@@ -65,7 +65,7 @@
 
 		vstudio = {
 			solutionVersion = "12",
-			versionName     = "2015",
+			versionName     = "14",
 			targetFramework = "4.5",
 			toolsVersion    = "14.0",
 			filterToolsVersion = "4.0",


jstewart-amd:
Before the change, Explorer shows the generic sln file icon:
premake_vs2015_sln_icon_before

After the change, Explorer shows the version-specific file icon:
premake_vs2015_sln_icon_after

The version-specific file icons already work for vs2010, vs2012, and vs2013. This change is only for vs2015. (I don't run versions prior to vs2010, and I haven't tried vs2017 RC yet.)

This change was verified on Windows 7 SP1 with the [Platform Update](https://support.microsoft.com/en-us/kb/2670838) and Windows 10 Version 1607. And I've been running this change locally for ages (since alpha6) with no issues, including on previous Windows 10 versions.

samsinsane:
Thanks!

PR #650, opened 2016-12-15
Remove defines and include directories "None" projects

starkos:
Projects with a type of "None" do not build anything (are excluded from the build entirely, actually) and should not include these values.

I took the opportunity to also break the NMake elements out into a call array so they can be extended by modules a little more easily.


starkos:
Rejecting—let me add a unit test for this.

PR #660, opened 2016-12-24
Prefixed Tokens

CodeAnxiety:
With this change I was looking to add some sugar, making it easier to use environmental variables.  I ended with something a little more extensible.

This is my first pull request.  😃  I would greatly appreciate any feedback.

### DIFF
--- a/src/base/detoken.lua
+++ b/src/base/detoken.lua
@@ -50,37 +50,19 @@
 		-- enable access to the global environment
 		setmetatable(environ, {__index = _G})
 
-		function esc(text)
-			text = text:gsub("\a", "\\a")
-			text = text:gsub("\b", "\\b")
-			text = text:gsub("\f", "\\f")
-			text = text:gsub("\t", "\\t")
-			text = text:gsub("\v", "\\v")
-			text = text:gsub("\r", "\\r")
-			text = text:gsub("\n", "\\n")
-			text = text:gsub("\'", "\\'")
-			text = text:gsub('\"', '\\"')
-			text = text:gsub("\\", "\\\\")
-			return text
-		end
 
-		function resolvePrefix(token)
+		function evalprefix(token)
 			local prefix, argument = token:match("(.*):(.*)")
 			if prefix ~= nil then
 				local resolver = prefixMap[prefix]
 				if type(resolver) == "function" then
-					return '"' .. esc(resolver(argument) or "") .. '"'
-				elseif type(resolver) == "string" then
-					return resolver .. '("' .. esc(argument) .. '")'
+					return resolver(argument)
 				end
 			end
-			return token
+			return nil
 		end
 
-		function expandtoken(token, e)
-			-- attempt to resolve prefixed tokens
-			token = resolvePrefix(token)
-
+		function evalobject(token, e)
 			-- convert the token into a function to execute
 			local func, err = loadstring("return " .. token)
 			if not func then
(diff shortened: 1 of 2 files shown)


samsinsane:
This is a pretty good attempt, but unfortunately, it kind of breaks things and I think there's a few inefficiencies in there too. It's a bit hard to sum this up in a nice concise way, so I apologise if this gets a bit chaotic.

Firstly, the detoken system is quite complex, I'm not sure if you know this already or not, so I'll just point out that it already allows you to call arbitrary functions. `%{os.getenv('path')}` is entirely valid and will do the same as your `%{env:path}`, it will actually be slightly faster too as it doesn't need to escape `path` like your solution does. This seems insignificant, but if you provide a function instead of a string to `detoken.addPrefix`, it can be incredibly slow as it would need to escape the entire return value. For something like the path environment variable this is going to be slow, mine is 690 characters and has a bunch of backslashes that would need to be escaped.

Knowing that the detoken system allows for function calls, it allows for things such as `%{wks.path:sub('foo', 'bar')}`. Which obviously conflicts with your solution, and I think I actually use this somewhere. I think it's very important that we don't break the existing functionality as it's a lot more versatile than your solution. I think having more shortcuts is important though, so I don't think we should abandon what you're attempting here.

Finally, there's some issues with the tests. They don't seem to work in Linux and OSX, but this shouldn't be too hard to fix. In the `resolvesEnvVars` test, you have `foo` and `bar` around the wrong way in `test.isequal`. Finally, you need to add tests that run through more paths of that `esc` function, should `\a` come out as `\\a` or `\\\\a`? It currently does the latter.

Just to reiterate, I think the idea of a shortcut for environment variables is a great idea. I'm just not entirely sure how we could achieve that without breaking existing functionality.

CodeAnxiety:
Thanks for the review.  😅 

All tests were passing, so I assumed it had no side effect.  That's probably naive of me.

I'm going to take a look from Linux and see where its going wrong.  I'll also see if I can make a more efficient version.  A lot of it comes from the desire to not touch `detoken.expandvalue` as much as possible, which meant it going through the loadstring evaluation.

samsinsane:
> Thanks for the review.

No worries! Thanks for contributing this cool feature.

> All tests were passing, so I assumed it had no side effect. That's probably naive of me.

That's a fair assumption, we don't test everything though, so that's our bad. Additionally, the point I made about the side effects is actually wrong, sorry about that. There's only problems when the `prefixMap` contains a matching string, which will probably quite rare, so you can probably ignore that point.

> I'll also see if I can make a more efficient version. A lot of it comes from the desire to not touch `detoken.expandvalue` as much as possible, which meant it going through the loadstring evaluation.

That makes sense, I think you can use that `loadstring` a different way though. Have your prefix function return `token, argument` if `token` is a function skip the `loadstring` and do `token(argument)`. You'll probably still want the `setfenv` call though. Otherwise, you pass `token` into `loadstring` like normal, but after the `pcall` you can check if `result` is a function, if it is and `argument` isn't nil, you call `result(argument)`. This should eliminate the need for `esc` which I find to be pretty expensive, given it will iterate over the input 10 times.

CodeAnxiety:
I can't reproduce the travis-ci error on Linux.

As best I can tell it is failing to match the prefix so the "env:path" continues down to the evalobject logic and fails.

Can anyone else reproduce this problem?

PR #662, opened 2016-12-26
New Flag: WholeArchive

CodeAnxiety:
- Adds support for the GCC whole archive linker flags, which forces all symbols in libraries to be exported


samsinsane:
We're trying not to add anymore flags, instead you should make this an API like `largeaddressaware`, of kind `boolean`.

CodeAnxiety:
Thanks, will do.

PR #663, opened 2016-12-28
Windows Registry

CodeAnxiety:
- Added a new os command `getreg` which takes a registry path and returns the value or nil


tvandijck:
I like this, even though it's pretty windows specific....

that said, if I could make any change, it's the name "getreg", while I appreciate concise and short names as anyone else... This is such an OS specific thing, I'd rather be more explicit and call it "getWin32RegistryValue" or something like that...

neico:
I'd even prefer it more if there were some OS specific namespaces for very specific features.

ex. `windows.getregistry( <namespace, ex. "HKLM">, [<filter string, supporting regex>] )`
and maybe let that table have a metatable with a getvalue() function, this might look verbose and is a bit more work, but might abstract away the clutter that the registry is to a point that it can be used cross-platform.

That being said, some sort of registry could be implemented on other systems by reading in the files in /etc/ (/etc/default/ is a good example) so it might be possible to make a cross-platform solution sometime in the future

starkos:
We've been trying to move to camel-case for API functions, so I would probably use one of:

    os.getWindowsRegistry()
    os.setWindowsRegistry()

    os.windows.getRegistry()
    os.windows.setRegistry()

I wouldn't use "Win32" personally, even though it is sadly true. :)

Otherwise, this is definitely a nice feature to have.

tvandijck:
see: https://github.com/premake/premake-core/pull/687

PR #664, opened 2016-12-28
bring os.getversion() on par with the present

neico:
sadly this doesn't include distribution names for linux, as that seems to be a major headache...

the best way would be probably `cat /etc/*release` and fetch the `ID=*` line, but I'm not up for writing that file parser...

### DIFF
--- a/src/host/os_getversion.c
+++ b/src/host/os_getversion.c
@@ -75,10 +75,14 @@ SYSTEM_INFO getsysteminfo()
 
 void getversion(struct OsVersionInfo* info)
 {
-	OSVERSIONINFOEX versionInfo = {0};
+	RTL_OSVERSIONINFOEXW versionInfo = {0};
 
-	versionInfo.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
-	GetVersionEx((OSVERSIONINFO*)&versionInfo);
+	versionInfo.dwOSVersionInfoSize = sizeof(RTL_OSVERSIONINFOEXW);
+
+	typedef LONG (WINAPI* RtlGetVersionSig)(PRTL_OSVERSIONINFOEXW);
+	RtlGetVersionSig getVersion = (RtlGetVersionSig)
+	GetProcAddress(GetModuleHandle(TEXT("ntdll.dll")), "RtlGetVersion");
+	if ( getVersion ) getVersion(&versionInfo);
 
 	info->majorversion = versionInfo.dwMajorVersion;
 	info->minorversion = versionInfo.dwMinorVersion;
@@ -124,7 +128,7 @@ void getversion(struct OsVersionInfo* info)
 			info->description = "Windows Server 2008";
 		}
 	}
-	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 1 )
+	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 1)
 	{
 		if (versionInfo.wProductType != VER_NT_WORKSTATION)
 		{
@@ -135,7 +139,7 @@ void getversion(struct OsVersionInfo* info)
 			info->description = "Windows 7";
 		}
 	}
-	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 2 )
+	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 2)
 	{
 		if (versionInfo.wProductType != VER_NT_WORKSTATION)
 		{
@@ -146,7 +150,7 @@ void getversion(struct OsVersionInfo* info)
 			info->description = "Windows 8";
 		}
 	}
-	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 3 )
+	else if (versionInfo.dwMajorVersion == 6 && versionInfo.dwMinorVersion == 3)
 	{
 		if (versionInfo.wProductType != VER_NT_WORKSTATION)
 		{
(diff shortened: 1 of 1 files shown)


src/host/os_getversion.c
```diff
@@ -75,10 +75,14 @@ SYSTEM_INFO getsysteminfo()
...
+	RTL_OSVERSIONINFOEXW versionInfo = {0};
 
-	versionInfo.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
-	GetVersionEx((OSVERSIONINFO*)&versionInfo);
+	versionInfo.dwOSVersionInfoSize = sizeof(RTL_OSVERSIONINFOEXW);
+
+	typedef LONG (WINAPI* RtlGetVersionSig)(PRTL_OSVERSIONINFOEXW);
+	RtlGetVersionSig getVersion = (RtlGetVersionSig)
+	GetProcAddress(GetModuleHandle(TEXT("ntdll.dll")), "RtlGetVersion");
+	if ( getVersion ) getVersion(&versionInfo);
```
samsinsane: I'm not quite sure why we would want these four lines over just `RtlGetVersion(&versionInfo);`? Is there a benefit to doing it this way over just calling the function directly?

neico: I wasn't sure if the ntddk.h is available on all build machines.

Plus you'll need to link against NtosKrnl.lib if you use it.
So imo this has the lowest overhead.

samsinsane: According to [this](https://msdn.microsoft.com/en-us/library/mt723418(v=vs.85).aspx), you only need to include `ntddk.h` and link `ntdll.lib`. I personally don't see an issue with linking Windows libraries in, I prefer the code to be simple.

neico: yikes, need to install the windows driver sdk for those. (they aren't included in the Win10 SDK at least)
And I highly doubt that's installed in a default environment...

imo the code is simple enough, but if you really want to force people to install that extra sdk just for simple version info's, be my guest...

samsinsane: > yikes, need to install the windows driver sdk for those. (they aren't included in the Win10 SDK at least)

Oh ok, I didn't really have the time to look into this much further than that one page, I just assumed the function would have a similar level of accessibility as `GetVersionEx`. My bad.

tvandijck: None of this should be used anymore...
"on par with the present "
would be using the recommended microsoft API's for user applications... "Rtl" methods are typically reserved for driver developement.

Instead on par would be using these new API's:
https://msdn.microsoft.com/en-us/library/windows/desktop/dn905474(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/dn424964(v=vs.85).aspx

granted they feel somewhat weird... but that's what MS is recommending.

neico:
I've had those on my radar at first but didn't feel like rewriting half of that code just for them, while there's an easier replacement for the more defunct method.

I guess I could rewrite the code to actually do so if there's a consensus about relying on them (which might or might not be discouraged in the future, getting windows versions / features never was reliable thanks to microsoft)

samsinsane:
@tvandijck I tend to agree that using those APIs should be the way forward. We will need to do [this](https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx) though.

tvandijck:
Yeah, the weird thing is that the GetVErsionEx is actually broken, unless you provide a manifest resource in your binary... which is just odd.... there is also the GetProductVersion API, which actually returns what GetVersionEx previously returned... but it's a Vista+ API only, so it would crash the binary on XP (which as much as I hate to admit it... we still target for Heroes and SC2).

samsinsane:
> so it would crash the binary on XP (which as much as I hate to admit it... we still target for Heroes and SC2).

That's unfortunate, my thoughts are with you and your colleagues!

Technically, we could still try to use `GetProductInfo` if it exists? There's other code in this file, I think, that checks if an API exists, and if not, it falls back to something else. We could do the same thing here?

starkos:
The code does currently use `GetModuleHandle()` and `GetProcAddress()` to check for the existence of the library and symbol before using them. That would be the approach to take to check for XP support (and to future-proof against other changes).

neico:
Well what action should I take now?

- Keep the code as is (in that case merge it in)
- Change to `IsWindows*OrGreater` (where `IsWindows10OrGreater` is only available with Windows 10, meaning it'll probably error on older versions)
- Seems you found other functions that could be used, but also seem to have caveats

But my initial reasoning for using such a low level function was because it's what's windows has been using since the kernel exists, and is unlikely to change, while everything microsoft tried and **build upon it** has some issues depending on your windows version, which is not a good thing... (therefore I still consider my solution the best overall, even I myself don't like it that much, it's just how windows development is...)

tvandijck:
My personal opinion? 

Use GetProductInfo if available (through a GetProcAddress on kernel32.dll). This should cover anything Vista+. If not, use GetVersionEx as it is right now, which should cover anything prior to Vista.

tvandijck:
So I've actually been reading up on this, and this is even more complicated then I though..
Turns out, all methods we've mentioned so far are actually all deprecated, and according to this page:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724429(v=vs.85).aspx

the only reliable way forward to get the version number of the OS, is by looking at the version of a system dll...

I quote: "To obtain the full version number for the operating system, call the GetFileVersionInfo function on one of the system DLLs, such as Kernel32.dll, then call VerQueryValue to obtain the \\StringFileInfo\\<lang><codepage>\\ProductVersion subblock of the file version information."

pretty crazy... I'll see if I can get that to work...

tvandijck:
https://github.com/premake/premake-core/pull/688 merged instead.

PR #671, opened 2017-01-15
Add support for MSVC object files

ifarbod:
MSVC object files previously were detected wrong by premake's parser and the linker wouldn't link them, with this patch .obj files get placed into the right section in .vcxproj projects.

Given:
```lua
files
{
    "file.obj"
}
```

It would end up like:

```xml
<None Include="file.obj" />
```

With this patch it will have the correct behavior:

```xml
<Object Include="file.obj" />
```

And ofcourse MSBuild would link them.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -675,13 +675,35 @@
 		end
 	}
 
+---
+-- Object file group
+---
+	m.categories.Object = {
+		name       = "Object",
+		extensions = ".obj",
+		priority   = 3,
+
+		emitFiles = function(prj, group)
+			local fileCfgFunc = {
+				m.excludedFromBuild
+			}
+
+			m.emitFiles(prj, group, "Object", nil, fileCfgFunc, function(cfg)
+				return cfg.system == p.WINDOWS
+			end)
+		end,
+
+		emitFilter = function(prj, group)
+			m.filterGroup(prj, group, "Object")
+		end
+	}
 
 ---
 -- None group
 ---
 	m.categories.None = {
 		name = "None",
-		priority = 3,
+		priority = 4,
 
 		emitFiles = function(prj, group)
 			m.emitFiles(prj, group, "None", {m.generatedFile})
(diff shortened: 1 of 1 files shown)


samsinsane:
This doesn't allow users to do pre-processing on `.obj` files, which is a pretty important feature to lose. If you're not already aware, `.obj` files are used for [other things](https://en.wikipedia.org/wiki/Wavefront_.obj_file).

tvandijck:
Unfortunately this is one of those things where premake just cannot do the right thing no matter what...
for YOU you need it to link as object files..., for OTHERS it needs to include Alias Wavefront files for some reason...

that said... if you do:
```
links { 'file.obj' } 
```
premake will add the object file as a link input... So you can work around the issue of premake not automatically interpreting the object file as a link input... 

I think this comes down to a fundamental design decision in premake, where we treat things by extension, rather then having independent "lists"... For example if the API was something like this:
```
files {
   ['clang']   = { '*.cpp', '*.c' }
   ['protoc'] = {'*.proto' }
   ['none']   = { '*.txt' }
}
```
it would be a possible to configure custom toolsets for specific sets of files, and in that case you could also do something like
```
file {
   ['linkable'] = { 'myobj.obj' }
   ['none'] = {'aliaswavefront.obj' }
}
```

starkos:
I thought we were linking .obj files by default, with [an API to turn it off](https://github.com/premake/premake-core/pull/628)?

samsinsane:
That [API](https://github.com/premake/premake-core/wiki/linkbuildoutputs) only applies to generated files rather than the existing file. I think what @tvandijck has suggested is a decent idea, but I think something like this would be "better":

```lua
files {
  '**.cpp',
  '**.c',
  '**.h',
  'libs/**.obj',
  { ['none'] = { 'models/**.obj' } },
}

files { '**.cpp', '**.c', '**.h', 'libs/**.obj' }
files { ['none'] = { 'models/**.obj' } }
```

`files` would work exactly the same as it currently does, it would just have the extra ability to handle objects. I imagine it would need to handle "overrides" though, such as:
```lua
files { `**.obj` }
files { ['none'] = { 'models/**.obj' } }
```

ifarbod:
@samsinsane That's a nice idea 👍
However, I'm not sure how could I implement that.

starkos:
Could we leverage [buildaction("None")](https://github.com/premake/premake-core/wiki/buildaction) for this?

samsinsane:
@starkos That would probably make more sense in the short term. Long term, maybe we want to have it all in the declaration of files?

PR #673, opened 2017-01-19
Add additional tokens to fileconfig and vs2010+.

tvandijck:
This provides us with a workaround for a fix that was made that broke our builds.

https://github.com/premake/premake-core/pull/643

we relied on "path" being the "%(RelativeDir)" token in visual studio... obviously wrong, but we relied on it, so we need another way to get there..

### DIFF
--- a/src/actions/vstudio/vs2010.lua
+++ b/src/actions/vstudio/vs2010.lua
@@ -31,7 +31,10 @@
 		["file.basename"]               = { absolute = false, token = "%(Filename)" },
 		["file.abspath"]                = { absolute = true,  token = "%(FullPath)" },
 		["file.relpath"]                = { absolute = false, token = "%(Identity)" },
-		["file.path"]                   = { absolute = true,  token = "%(Identity)" },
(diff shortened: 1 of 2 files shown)


src/actions/vstudio/vs2010.lua
```diff
@@ -31,7 +31,10 @@
 		["file.basename"]               = { absolute = false, token = "%(Filename)" },
 		["file.abspath"]                = { absolute = true,  token = "%(FullPath)" },
 		["file.relpath"]                = { absolute = false, token = "%(Identity)" },
-		["file.path"]                   = { absolute = true,  token = "%(Identity)" },
+		["file.path"]                   = { absolute = false, token = "%(Identity)" },
```
samsinsane: Just want to verify that this should be changed to `false`?

tvandijck: Yes, %(Identity) is a relative path in msbuild... and in premake itself, path is an alias for relpath.

PR #674, opened 2017-01-20
Define _HAS_EXCEPTIONS=0 when exceptionhandling is "off".

tvandijck:
When exceptionhandling "Off" is declared, "_HAS_EXCEPTIONS=0" should also be declared to avoid warnings about exception use in the standard library.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1149,7 +1149,11 @@
 
 
 	function m.clCompilePreprocessorDefinitions(cfg, condition)
-		m.preprocessorDefinitions(cfg, cfg.defines, false, condition)
+		local defines = cfg.defines
+		if cfg.exceptionhandling == p.OFF then
+			defines = table.join(defines, "_HAS_EXCEPTIONS=0")
+		end
+		m.preprocessorDefinitions(cfg, defines, false, condition)
 	end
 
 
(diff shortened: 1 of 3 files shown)


samsinsane:
I vaguely recall this having issues unless you were running either Update 2 or 3 of VS2015, I might be remembering this wrong though. This also only applies to VS2015+, while this will actually emit into all VC++ projects.

starkos:
Update 1 definitely has different behavior that 2 or 3, and even from the original release version. Not sure why they did it that way. I just ran a quick test on v3 and wasn't able to get the standard library errors Tom mentioned.

tvandijck:
I only got the report from one of our engineers... not getting the errors here either, although we may have simply disabled the warnings/errors before I made this patch... I can certainly make it VS2015 only... Our game is on Update 1, seems to at least do no harm...

samsinsane:
I must be remembering wrong, there were a few issues around the same time that I was dealing with, so the version of VS2015 could be completely unrelated.

We can merge as-is, I'd prefer if it was VS2015+ only but I'm not overly concerned by it. Up to you guys :)

starkos:
> I only got the report from one of our engineers... not getting the errors here either, although we may have simply disabled the warnings/errors before I made this patch...

Maybe we should hold off on this one?

tvandijck:
This is a repro case:

[_HAS_EXCEPTIONS.zip](https://github.com/premake/premake-core/files/740722/_HAS_EXCEPTIONS.zip)

On VS2015 you get: 
1>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\limits(210): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc

tvandijck:
Updated to be for VS2013 and later....

tvandijck:
And added a test to specifically test the VS2010 vs VS2013 behavior differences.

samsinsane:
> Updated to be for VS2013 and later....

Oh, I thought it was VS2015 and later, does this occur in VS2013 too?

tvandijck:
Yes, at least with Update 5, and it appears to do no harm with earlier versions.

samsinsane:
> Yes

Fair enough, we noticed the issue when we migrated from VS2013 to VS2015. I imagine we just never updated VS2013 beyond the base install. I've kicked off the OSX build, if someone would like to merge this once it's successful (it looks like it could be a while).

PR #680, opened 2017-01-30
Detect duplicate "newoptions" calls.

tvandijck:


tvandijck:
This breaks more the I want to break, and it seems like defining them multiple times doesn't actually harm anything...

PR #682, opened 2017-01-31
stringpooling api.

tvandijck:
adding:
```lua
stringpooling 'On' or 'Off'
```

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -970,6 +970,12 @@
 	}
 
 	api.register {
+		name = "stringpooling",
+		scope = "config",
+		kind = "boolean"
+	}
+
+	api.register {
 		name = "symbols",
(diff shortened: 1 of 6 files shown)


starkos:
> I think new API's should emit options to makefiles at least

Not sure I understand this?

tvandijck:
> Not sure I understand this?

@TurkeyMan wants that all new API's work for all backends... gmake/xcode/vs20xx...

I thought about it... have no way of testing it... so didn't do it because I really don't have the resources available to me to support a gmake or xcode backend we don't use at Blizzard.

I'm happy to accept pull requests on my pull requests if anyone feel passionate about adding support for these flags in xcode and gmake...

tvandijck:
@TurkeyMan there is no equivalent setting for string pooling in gcc/clang btw.

Issue #612, opened 2016-11-03
"c++" file extension


d-sviridov:
Hi,
Wouldn't that be nice to add ".c++" into function path.iscppfile(fname)?

samsinsane:
I don't think I've ever encountered a file with the extension `.c++`, are there any others that we should add while we're at it? `.h++`?

tvandijck:
I don't think it'll do any harm... however, it would be easy for anyone with such files to simply override the path.iscppfile method, and add their own extensions to the list... maybe we can make that easier by getting the list from the toolset and allow people to simply do a table.insert(toolset.cppfiles, ".c++")

d-sviridov:
@samsinsane Don't know about h++, never encountered it.
@tvandijck 
> it would be easy for anyone with such files to simply override the path.iscppfile method
it is much less easy to find what to override, how and where. Especially if you are seeing lua first time.
Trust me on that! ;)
> we can make that easier by getting the list from the toolset and allow people to simply do a table.insert(toolset.cppfiles, ".c++")
Unlikely. Some files, like c++ stl headers simply don't have any extension. So if you **really** want to fix it in a general way, its better to switch from `files` to something like  `headerFiles`, `compileFiles` etc

PR #685, opened 2017-01-31
Detect native source files.

TurkeyMan:
Separated the concept of 'native' languages from language detection.
There is code that assumes native code == c/c++ code.
I also feel it's weird to conflate asm and obj-c... I can imagine myself writing code to detect those languages with the intent to do something language-specific that's probably not applicable to asm or whatever.

### DIFF
--- a/src/base/path.lua
+++ b/src/base/path.lua
@@ -142,23 +142,52 @@
 
 
 --
--- Returns true if the filename represents a C/C++ source code file. This check
--- is used to prevent passing non-code files to the compiler in makefiles. It is
--- not foolproof, but it has held up well. I'm open to better suggestions.
+-- Returns true if the filename represents various source languages.
 --
 
+	function path.isasmfile(fname)
+		return path.hasextension(fname, { ".s" })
+	end
+
 	function path.iscfile(fname)
-		return path.hasextension(fname, { ".c", ".s", ".m" })
+		return path.hasextension(fname, { ".c" })
+			or path.isasmfile(fname)	-- is this really right?
+			or path.isobjcfile(fname)	-- there is code that depends on this behaviour, which would need to change
 	end
 
 	function path.iscppfile(fname)
-		return path.hasextension(fname, { ".cc", ".cpp", ".cxx", ".c", ".s", ".m", ".mm" })
+		return path.hasextension(fname, { ".cc", ".cpp", ".cxx" })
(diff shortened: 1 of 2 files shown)


TurkeyMan:
My gmake fixes depend on this...

src/base/path.lua
```diff
@@ -142,24 +142,53 @@
...
 	function path.iscfile(fname)
-		return path.hasextension(fname, { ".c", ".s", ".m" })
+		return path.hasextension(fname, { ".c" })
+			or path.isasmfile(fname)	-- is this really right?
+			or path.isobjcfile(fname)	-- there is code that depends on this behaviour, which would need to change
 	end
 
 	function path.iscppfile(fname)
-		return path.hasextension(fname, { ".cc", ".cpp", ".cxx", ".c", ".s", ".m", ".mm" })
+		return path.hasextension(fname, { ".cc", ".cpp", ".cxx" })
```
samsinsane: While you're changing this, can you throw `.c++` in as per #612?

TurkeyMan: i gueeeeess... although that's a really weird file extension :/

starkos:
A good idea. A couple of thoughts, if you don't mind considering them while you're in the code:

- If we move the file extension lists out of the functions themselves, people and modules would be able to add/remove new entries to suit.

- Perhaps this would be better as a toolset function i.e. `isbuildablefile()`? Clang can build Objective C, for instance, but MSC cannot.

src/base/path.lua
samsinsane: I don't disagree, it's completely valid (I tested it on Windows lol) and it seems to be used by things. I think @starkos has a better idea though.

src/base/path.lua
```diff
@@ -142,24 +142,53 @@
...
 
 
+--
+-- Returns true if the filename represents a native language source file.
+-- These checks are used to prevent passing non-code files to the compiler
+-- in makefiles. It is not foolproof, but it has held up well. I'm open to
+-- better suggestions.
+--
+
+	function path.isnativefile(fname)
```
tvandijck: I'm with @starkos on this one... this should move to toolset, and rename to "isbuildablefile".

tvandijck:
I'm going to merge this, it's better then what we currently have, even though I still think some of this behavior should move to toolset.

PR #686, opened 2017-01-31
Refactor language flags to go through the language API.

tvandijck:
So you can use 'language "C++11"' instead of 'flags { "C++11" }'

### DIFF
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -51,9 +51,9 @@
 			error("Invalid toolset '" + (_OPTIONS.cc or cfg.toolset) + "'")
 		end
 
-		if cfg.language == "C" then
+		if p.languages.isc(cfg.language) then
 			return m.ctools[tool]
-		elseif cfg.language == "C++" then
+		elseif p.languages.iscpp(cfg.language) then
 			return m.cxxtools[tool]
 		end
 	end
@@ -354,8 +354,19 @@
 		_p(3, '</AdditionalRules>')
 	end
 
+	function m.isCpp11(cfg)
+		return (cfg.language == 'gnu++11') or (cfg.language == 'C++11') or cfg.flags["C++11"]
+	end
+
+	function m.isCpp14(cfg)
+		return (cfg.language == 'gnu++14') or (cfg.language == 'C++14') or cfg.flags["C++14"]
+	end
+
 	function m.completion(cfg)
-		_p(3, '<Completion EnableCpp11="%s" EnableCpp14="%s">', iif(cfg.flags["C++11"], "yes", "no"), iif(cfg.flags["C++14"], "yes", "no"))
+		_p(3, '<Completion EnableCpp11="%s" EnableCpp14="%s">',
+			iif(m.isCpp11(cfg), "yes", "no"),
+			iif(m.isCpp14(cfg), "yes", "no")
+		)
 		_p(4, '<ClangCmpFlagsC/>')
 		_p(4, '<ClangCmpFlags/>')
 		_p(4, '<ClangPP/>') -- TODO: we might want to set special code completion macros...?
--- a/modules/xcode/xcode_common.lua
+++ b/modules/xcode/xcode_common.lua
@@ -959,6 +959,40 @@
 	end
 
 
+	function xcode.XCBuildConfiguration_CLanguageStandard(settings, cfg)
+		if (p.languages.isc(cfg.language)) then
+			if cfg.language ~= "C" then
+				settings['GCC_C_LANGUAGE_STANDARD'] = cfg.language:lower()
+			else
+				if cfg.flags['C99'] then
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C99'
+				elseif cfg.flags['C11'] then
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C11'
+				else
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C90'
+				end
+			end
+		else
+			settings['GCC_C_LANGUAGE_STANDARD'] = 'gnu99'
+		end
+	end
+
+
+	function xcode.XCBuildConfiguration_CppLanguageStandard(settings, cfg)
+		if (p.languages.iscpp(cfg.language)) then
+			if cfg.language ~= "C++" then
+				settings['CLANG_CXX_LANGUAGE_STANDARD'] = cfg.language:lower()
+			else
+				if cfg.flags['C++11'] then
+					settings['CLANG_CXX_LANGUAGE_STANDARD'] = 'c++11'
+				elseif cfg.flags['C++14'] then
+					settings['CLANG_CXX_LANGUAGE_STANDARD'] = 'c++14'
+				end
+			end
+		end
+	end
+
+
 	function xcode.XCBuildConfiguration_Project(tr, cfg)
 		local settings = {}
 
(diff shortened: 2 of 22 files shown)


modules/codelite/codelite_project.lua
```diff
@@ -366,7 +366,10 @@
 	end
 
 	function m.completion(cfg)
-		_p(3, '<Completion EnableCpp11="%s" EnableCpp14="%s">', iif(cfg.flags["C++11"], "yes", "no"), iif(cfg.flags["C++14"], "yes", "no"))
+		local cpp11 = (cfg.language == 'gnu++11') or (cfg.language == 'C++11') or cfg.flags["C++11"]
```
samsinsane: This should probably be a function, to allow others to add other language extensions like GNU++11. For example, VS2015 Update 3 (probably VS2017 too) allows you to specify "C++Latest", we should have support for this to translate to C++17. When C++20 starts happening, it will need to map to that, and changing it everywhere is going to be time consuming.

modules/xcode/xcode_common.lua
```diff
@@ -987,12 +987,32 @@
 			settings['COPY_PHASE_STRIP'] = 'NO'
 		end
 
-		settings['GCC_C_LANGUAGE_STANDARD'] = 'gnu99'
+		if (p.languages.isc(cfg.language)) then
+			if cfg.language ~= "C" then
+				settings['GCC_C_LANGUAGE_STANDARD'] = cfg.language:lower()
```
samsinsane: Based on previous comment, this will need to handle things like "C++Latest" (but a C equivalent), so it's probably best this entire block for adding `GCC_C_LANGUAGE_STANDARD` be a function instead, same for the C++ block.

src/_premake_init.lua
```diff
@@ -1348,6 +1344,13 @@
 	end)
 
 
+	-- 31 January 2017
+
+	api.deprecateValue("flags", "C++11", 'Use `language "C++11"` instead', function(value) end, function(value) end)
+	api.deprecateValue("flags", "C++14", 'Use `language "C++14"` instead', function(value) end, function(value) end)
+	api.deprecateValue("flags", "C90",   'Use `language "C90"` instead',   function(value) end, function(value) end)
+	api.deprecateValue("flags", "C99",   'Use `language "C99"` instead',   function(value) end, function(value) end)
+	api.deprecateValue("flags", "C11",   'Use `language "C11"` instead',   function(value) end, function(value) end)
```
samsinsane: Shouldn't these all set the language?

tvandijck: @starkos suggested we don't do that in his last comment on this thread: https://github.com/premake/premake-core/pull/659

We could though, I don't think with the changes I made it would break anything.

starkos: Agreed, shouldn't break anything. Just thought it would be easier to handle both the language and the flags at the point of application, which you did.

tvandijck: I think the suggestion was ultimately the correct one:

```lua
	api.deprecateValue("flags", "C++11", 'Use `language "C++11"` instead', 
	function(value) 
		language "C++11"
	end, 
	function(value) 
		<what to do here> ???
	end)
```

I've tried a few things,
```lua
removelanguage "C++11"
or
language "C++"
```
but they all break...

samsinsane: Add a `Default` value to `language`? I'm picturing a situation like this:
```lua
project "NiceProject"
  language "C++14"
  files { "**.h", "**.c" }

  filter { "files:RealCFiles/**.c" }
    language "Default"
```

Would the language value flow down regardless of extension? Assuming it would, the above might be desired functionality?

starkos: Ah that's right, forgot about that part. We need the ability to set the language (or standard, or whatever) for each "dialect".

```
clanguage "C99"
cpplanguage "C++14"
```

We can still have the generic `language`, which is used if a more specific value hasn't been set. And of course this needs to extend to any supported language:

```
dlanguage "SafeD"
```

So yeah…there's that.

tvandijck: You could just do:

```
language "C++14"
filter { "files:*.c" }
    language "C99"
```

but that said, if we went the "dialect" route, I would certainly prefer to actually use that name.
So either:
```
dialect {
    cpp = "C++14"
    c = "C90",
    d = "D3.4",  -- if such a thing was a thing.
}
```
or as suggested: 
```
cdialect 'C90'
cppdialect 'C++14'
ddialect 'D3.4"
```
I'm somewhat indifferent to the approach, I think both have merit..

You guys want me to prototype that instead??

starkos: > You could just do:

```lua
language "C++14"
filter { "files:*.c" }
    language "C99"
```

You're quite right, I was overthinking. It is really only an issue in multi-language (or dialect) projects, and this suggestion is the right way to go.

tvandijck:
Can we merge this? or is there more work left?

samsinsane:
@tvandijck There's a few places where you remove checks for `iscpp`, these should probably stay. In some cases, there's more code in the function and by removing the check you're causing module developers to copy code from core. `vs2010.generateProject` is a good example of this, the D module will end up generating a `.vcxproj` and `.visualdproj` if you select any VS action for a D project.

tvandijck:
> by removing the check you're causing module developers to copy code from core.

not sure I understand this...

> the D module will end up generating a .vcxproj and .visualdproj if you select any VS action for a D project.

I see that one, but the check was wrong... vcxproj is for both C and C++ projects.
I guess I can add a check to check for either.. I'm just wondering where this is needed exactly.

In that same boat, I then also think the "project.isnative" method is confusing, because there really isn't such a thing...

samsinsane:
Well, ideally, we'd change the functions so that generic and specific code don't exist in the same function. Maybe for now, the best move would be to check for C and C++ and then, eventually, we can move the generic code out. Thoughts?

neico:
While you're at it, add the new switches for Visual Studio as well? (`/std:c++latest` for example, see: https://blogs.msdn.microsoft.com/vcblog/2016/06/07/standards-version-switches-in-the-compiler/)

tvandijck:
@samsinsane what changes do we still need here? other then what @neico suggests to add?

tvandijck:
I rebased this, and made some more modifications to solve some of the comments @samsinsane made about me removing some of the required "iscpp" calls... I think I got them all correct now, but please review.

tvandijck:
@neico this commit just modifies the flag/api.... I've not actually implemented usage of it inside of the Visual Studio action at all... Feel free to submit further PR's on top of this change for that once this is merged (or even against this PR).

modules/xcode/xcode_common.lua
```diff
@@ -959,6 +959,40 @@
...
+				if cfg.flags['C99'] then
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C99'
+				elseif cfg.flags['C11'] then
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C11'
+				else
+					settings['GCC_C_LANGUAGE_STANDARD'] = 'C90'
+				end
+			end
+		else
+			settings['GCC_C_LANGUAGE_STANDARD'] = 'gnu99'
```
samsinsane: Should this map to one of the C11 standards when the language is C++17? I believe C++17 is based on C11, rather than C99, or is it the plan for the next C++ standard to be on C11?

src/actions/vstudio/vs2005.lua
```diff
@@ -51,16 +51,14 @@
 			if #user > 0 then
 				p.generate(prj, ".csproj.user", function() p.outln(user) end)
 			end
-
-		elseif premake.project.iscpp(prj) then
+		else
```
samsinsane: Technically, this should probably have the check for C/C++, but I'm happy to ignore this since it's for VS2005 and VS2008, and VS2008 will be EOL in a little under a year.

starkos:
Are we good to merge? I think it looks pretty good, major props to @tvandijck for putting it together!

aleksijuvani:
Looks like the line endings in ` src/actions/vstudio/vs2010_vcxproj.lua` have changed, so I'm getting a huge merge conflict with #734. I feel like we should revert this and fix the line endings here.

samsinsane:
I've noticed that line ending issue, I'm fairly certain it was already in master from somewhere else.

aleksijuvani:
I fixed this in my branch by running `git rebase master -s recursive -X ignore-space-change` and then manually fixing up line endings in every commit that touches `src/actions/vstudio/vs2010_vcxproj.lua`.

I think git provides facilities to automatically normalize line endings on push with a `.gitattributes` file, so we might want to look into that at some point.

neico:
@aleksijuvani `.gitattributes` and `.editorconfig` together can avoid any of those.

if you throw in `.clang-format` you get even actual style guidelines enforced.

So yeah any git project should at least use the first 2 mentioned files.

PR #687, opened 2017-01-31
Windows Registry

tvandijck:
This is an update to https://github.com/premake/premake-core/pull/663
with the requested renames.


tvandijck:
This was previously reviewed, and accepted...

neico:
Shouldn't there be a test for non-windows as well that running the command doesn't negatively impact those? (missing function, corrupting data)  
The current code certainly doesn't do anything harmful, but it might in the future due to a regression...

PR #688, opened 2017-01-31
bring os.getversion() on par with the present

tvandijck:
@neico, this is your change, rebased against head + some additional changes...

see https://github.com/premake/premake-core/pull/664 for further discussion.

### DIFF
--- a/src/host/os_getversion.c
+++ b/src/host/os_getversion.c
@@ -16,13 +16,15 @@ struct OsVersionInfo
 	int isalloc;
 };
 
-static void getversion(struct OsVersionInfo* info);
-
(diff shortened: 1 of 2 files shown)


neico:
Something I just noticed is that the mac part isn't accounting for [MacOS Server](https://en.wikipedia.org/wiki/MacOS_Server).

As I don't have mac or it's server I couldn't even start guessing how to detect it, so I'd probably leave it for a followup PR, unless you're able to integrate it yourself.

tvandijck:
I'll take care of the MacOS names, and merge it after....
travis-ci's mac builds are down, so that's why they are failing.... I tested all these branches locally here on our own buildfarm, so we should be all good.

PR #691, opened 2017-02-09
Fix a bunch of compile warnings.

tvandijck:
This just takes care of some compiler warnings when compiling premake in 64-bit.

I think we should support and test both 32-bit and 64-bit versions of premake.


samsinsane:
Sounds like a good idea to me.

starkos:
Updated to latest master branch, waiting for tests to pass…

PR #694, opened 2017-02-11
Add support Visual Studio 2017 DebugFull option

rorydriscoll:
In Visual Studio 2017, Microsoft changed the behavior of the default debug info flag to use fastlink rather than emitting a full PDB file. There is a new debug info flag /DEBUG:FULL that will work that same way that /DEBUG worked in Visual Studio 2015.

This change adds a new flag for symbols that will produce a full PDB. I think that this is the correct thing to do since symbols on should probably do whatever the current version of Visual Studio thinks is the default.

There's now the following logic for what to place in the GenerateDebugInformation attribute:

|          |        vs2017 |        vs2015 | vs20XX |
|----------|---------------|---------------|--------|
| On       |          true |          true |   true |
| Off      |         false |         false |  false |
| FastLink | DebugFastLink | DebugFastLink |   true |
| Full     |     DebugFull |          true |   true |

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1376,14 +1376,21 @@
 
 	function m.generateDebugInformation(cfg)
 		local lookup = {}
-		if _ACTION >= "vs2015" then
+		if _ACTION >= "vs2017" then
 			lookup[p.ON]       = "true"
 			lookup[p.OFF]      = "false"
 			lookup["FastLink"] = "DebugFastLink"
+			lookup["Full"]     = "DebugFull"
+		elseif _ACTION == "vs2015" then
+			lookup[p.ON]       = "true"
+			lookup[p.OFF]      = "false"
+			lookup["FastLink"] = "DebugFastLink"
+			lookup["Full"]     = "true"
 		else
 			lookup[p.ON]       = "true"
 			lookup[p.OFF]      = "false"
 			lookup["FastLink"] = "true"
+			lookup["Full"]     = "true"
 		end
 
 		local value = lookup[cfg.symbols]
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1002,6 +1002,7 @@
 			"On",
 			"Off",
 			"FastLink",    -- Visual Studio 2015+ only, considered 'On' for all other cases.
+			"Full",        -- Visual Studio 2017+ only, considered 'On' for all other cases.
 		},
 	}
 
--- a/tests/actions/vstudio/vc2010/test_link.lua
+++ b/tests/actions/vstudio/vc2010/test_link.lua
@@ -151,6 +151,17 @@
 		]]
 	end
 
+	function suite.generateDebugInfo_onSymbolsFull_on2010()
+		premake.action.set("vs2010")
+		symbols "Full"
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+	<GenerateDebugInformation>true</GenerateDebugInformation>
+		]]
+	end
+
 	function suite.generateDebugInfo_onSymbolsOn_on2015()
 		premake.action.set("vs2015")
 		symbols "On"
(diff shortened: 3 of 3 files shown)


starkos:
This looks good—thanks for the tests! Giving Tom a chance to look before merging…

src/actions/vstudio/vs2010_vcxproj.lua
```diff
@@ -1376,14 +1376,21 @@
 
 	function m.generateDebugInformation(cfg)
 		local lookup = {}
-		if _ACTION >= "vs2015" then
+		if _ACTION >= "vs2017" then
```
tvandijck: I think this chance is good, and we can merge this as is....

I do however like to see someone create a PR to move this table into the action definition (vs2010/vs2012/vs2013/vs2015/vs2017.lua)

rorydriscoll: I can do that since I touched it last. I'll create a separate PR for that.

jspohr:
Hello, I apologize for digging this up, but I'm confused regarding the "FastLink" behavior on VS2015. 
When I use `symbols "FastLink"`, Premake emits an element `FullProgramDatabaseFile` into the vcxproj, which causes a full PDB to be created. Might this be a workaround because there was no `/DEBUG:FULL` before VS2017?
Maybe this line:
https://github.com/premake/premake-core/blob/cda4f94e288393d6672c4e0d7fdbebee2d2439a3/modules/vstudio/vs2010_vcxproj.lua#L1772
just needs to be changed to
`		if _ACTION == "vs2015" and cfg.symbols == "FastLink" then`
instead?

Thanks!

samsinsane:
I'm not sure I'm following, are you using VS2015 or VS2017?

jspohr:
I'm using VS2015 mainly, but the issue is present on VS2017 as well. I don't really understand the implementation myself, that's why I'm failing to explain my issue with it properly. I assumed that I would benefit from FastLink even on VS2015, which I think I could if it wasn't for the `FullProgramDatabaseFile` element. I omitted the element using `premake.override`, and it seems to work fine, I get partial PDBs and faster link times with VS2015. But perhaps I'm missing something?
Additionally, I found it curious that even on VS2017 it appears FastLink's benefit is limited by emitting `FullProgramDatabaseFile`.

WorldofBay:
from blogs.msdn.com:

> Developers can opt-out of this behavior by adding the ‘/debug:full’ linker switch to linker command line or by enabling the ‘Generate Full Program Database File’
so premake does an opt-out of the fastlink option. that looks weird, especially weird sind it's already the case in #525 

looks wrong to me, did you look if any test fails on your override? if not, file a PR and it should be safe to accept.

jspohr:
I added the override inside my project, I have little experience running tests or modifying premake itself. I was hoping that someone could provide some background on why the opt-out was added. AFAICT, the linker option was there since the feature was initially added to the vcxproj generator, maybe there was a reason to use it back then?
If this is really unintended behavior, creating a pull request to fix it would be easy, just remove the function `vstudio.vc2010.fullProgramDatabaseFile` altogether.

samsinsane:
I would say it was unintended from @tvandijck, it's likely someone at Blizzard chose to opt out of FastLink and it was just brought across to here. If a PR is opened for this I'd be more than happy to review and accept it :)

WorldofBay:
the PR should be bigger than that an cover more of the vs2015 symbols handling. i've found some old test outputs:
https://github.com/premake/premake-core/blob/cda4f94e288393d6672c4e0d7fdbebee2d2439a3/modules/vstudio/tests/vc2010/test_link.lua#L187
since vs2015 this value is an enum instead of a boolean, so all of the tests refering to boolean output are likely wrong

i might look at it tomorrow if i find time

rorydriscoll:
To be clear, this PR didn't add the fast-link support. In VS2015, 'on' would produce a full PDB so fast-link was opt-in. In VS2017, 'on' would use fast-link by default so there was no way to get a full PDB anymore. This change just added back in a way to get a full PDB in VS2017.

WorldofBay:
@rorydriscoll your PR is not responsible for the strange behaviour, that has been in the code ever since tvandijk added support for the fastlink option for vs2015. the discussion just happens to be here.

WorldofBay:
digging into it i saw that vs2017 still puts a true and false in the vcxproj files even though it's an enum, so it's better to keep it. i'll keep the change simple then.

WorldofBay:
bumping this again:
i tested it with an internal project with lots of dependencies and strangely it built faster with the fullprogramdatabase enabled (VS2017). the difference wasn't even that small, about 10%.

so i need someone to do a "mini-benchmark" on VS2015 and maybe a confirmation on VS2017.

PR #695, opened 2017-02-13
Move the Visual Studio symbols lookup table into the respective action files

rorydriscoll:
I extracted the table information for the mapping between premake symbol flags and the various Visual Studio version flags. 

I duplicated the tables for vs2005-vs2013 since I wasn't sure if I should rather define them somewhere in the vs2005 file and then have each action reference them or not. It looks like some of the functions do this already, but I didn't find an example of a table doing this.

Also, I wasn't sure whether to put the new symbol lookup table inside the vstudio table that's part of the action. I decided not to in the end since that table looks more like it's for the general visual studio version configuration.

I added a few more tests just to verify that every current version of visual studio generates the correct value for symbols on.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1002,6 +1002,7 @@
 			"On",
 			"Off",
 			"FastLink",    -- Visual Studio 2015+ only, considered 'On' for all other cases.
+			"Full",        -- Visual Studio 2017+ only, considered 'On' for all other cases.
 		},
 	}
 
--- a/src/actions/vstudio/vs2005.lua
+++ b/src/actions/vstudio/vs2005.lua
@@ -124,5 +124,12 @@
 			productVersion      = "8.0.50727",
 			solutionVersion     = "9",
 			versionName         = "2005",
+		},
+
+		symbols = {
+			[p.ON]       = "true",
+			[p.OFF]      = "false",
+			["FastLink"] = "true",
+			["Full"]     = "true"
 		}
 	}
--- a/src/actions/vstudio/vs2008.lua
+++ b/src/actions/vstudio/vs2008.lua
@@ -5,6 +5,8 @@
 --
 
 	premake.vstudio.vs2008 = {}
+
+	local p = premake
 	local vs2008 = premake.vstudio.vs2008
 	local vstudio = premake.vstudio
 
@@ -50,5 +52,12 @@
 			solutionVersion     = "10",
 			versionName         = "2008",
 			toolsVersion        = "3.5",
+		},
+
+		symbols = {
+			[p.ON]       = "true",
+			[p.OFF]      = "false",
+			["FastLink"] = "true",
+			["Full"]     = "true"
 		}
 	}
--- a/src/actions/vstudio/vs2010.lua
+++ b/src/actions/vstudio/vs2010.lua
@@ -163,5 +163,12 @@
 			versionName         = "2010",
 			targetFramework     = "4.0",
 			toolsVersion        = "4.0",
+		},
+
+		symbols = {
+			[p.ON]       = "true",
+			[p.OFF]      = "false",
+			["FastLink"] = "true",
+			["Full"]     = "true"
 		}
 	}
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1375,25 +1375,8 @@
 
 
 	function m.generateDebugInformation(cfg)
-		local lookup = {}
-		if _ACTION >= "vs2017" then
-			lookup[p.ON]       = "true"
-			lookup[p.OFF]      = "false"
-			lookup["FastLink"] = "DebugFastLink"
-			lookup["Full"]     = "DebugFull"
-		elseif _ACTION == "vs2015" then
-			lookup[p.ON]       = "true"
-			lookup[p.OFF]      = "false"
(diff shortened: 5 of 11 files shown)


samsinsane:
VS2005 and VS2008 aren't required, they don't use the vcxproj code path.

@tvandijck is this correct move forward? Any additional symbol values will need to be added in five different locations, maybe more if the next release of VS adds the new value. This could have been done in the function with four lines instead:
```lua
lookup[p.ON] = "true"
lookup[p.OFF] = "false"
lookup["FastLink"] = iif(_ACTION >= "vs2015", "DebugFastLink", "true")
lookup["Full"] = iif(_ACTION >= "vs2017", "DebugFull", "true")
```

tvandijck:
@samsinsane I asked for this change... so in my opinion it is the right move yes....
We already merged the change that made it do things locally in the function as you suggested.

I think that the VS2017 backend is going to diverge much more then just this simple change... for example VS2017 supports: <ClInclude Include="*.cpp"> rather then specifying each file individually like what all previous versions of the IDE required... 

In addition, I really don't like all those "if _ACTION >= "vs2015"" etc if statements... for one it's a string compare that really only works because MS keeps giving it a name that coincides with the year of release, but if they ever changed it, that may no longer holds up.

So yes, in general I think this is a cleaner approach, despite it requiring slightly more code, and in my opinion we should rely on the action a lot more for storing "action" specific code or tables. Although the Visual Studio "action" is somewhat unique in that it shares one action's code to many actions.

samsinsane:
> I think that the VS2017 backend is going to diverge much more then just this simple change... for example VS2017 supports: rather then specifying each file individually like what all previous versions of the IDE required...

I don't have VS2017 but I've seen VS pick up new files in folders, I think this is part of VS2015 too. Anyway, that doesn't really matter. I would have expected VS2015 to diverge too given how much cross-platform support it gained, but core still assumes it's Windows-only and the above change kind of reiterates that assumption too.

> In addition, I really don't like all those "if _ACTION >= "vs2015"" etc if statements... for one it's a string compare that really only works because MS keeps giving it a name that coincides with the year of release, but if they ever changed it, that may no longer holds up.

We should definitely make helper functions for this that take the action name and then do a comparison internally against the version numbers defined in the action table. I'm not overly fond of the existing comparisons either, they're quite prone to errors without any warnings at all. Which helper functions could help with, e.g. "vs2014 is not a valid action.".

> So yes, in general I think this is a cleaner approach, despite it requiring slightly more code, and in my opinion we should rely on the action a lot more for storing "action" specific code or tables. Although the Visual Studio "action" is somewhat unique in that it shares one action's code to many actions.

I agree, for core, this is a cleaner approach. I disagree with this being cleaner for module developers though. How do you override a table under certain conditions, that may include the version of VS? Furthermore, any functions stored in the VS actions, would mean modules would require updates to add support for new versions of VS. Maybe this is a good thing though, I'm not sure.

I think this might require more discussion, or fleshing out, before we head down this path. I'm all for progress, but we seem to frequently ignore module development and that isn't good for the users that write or use modules. I am biased though, I've written two modules and I've extended a third, and I'd really like the modules to work for everyone, not just me.

starkos:
>  I'm all for progress, but we seem to frequently ignore module development and that isn't good for the users that write or use modules. I am biased though, I've written two modules and I've extended a third, and I'd really like the modules to work for everyone, not just me.

I'm guilty of this for sure. Please do raise a ruckus if you see something that can be improved for module development, your bias is appreciated!

tvandijck:
As it turns out however... as much as I love this change and thing this is better then the hardcoded approach... it's apparently not even correct...

I just opened up https://github.com/premake/premake-core/issues/699 which is a clear indication that this table shouldn't live in the action, but in the toolset.. Along with a bunch of other tables.

samsinsane:
@tvandijck the more I think about it, the less I think tables will work. Clang is a pretty good example, there will be a VS table for Linux, Android, iOS, Emscripten, Native Client and others. Within Android, there will be a table for the Microsoft implementation, the vs-android implementation and the NVidia implementation. Furthermore, there will be tables for the different versions of Clang supported by these systems and implementations. Also, there's currently two versions of the Microsoft implementation, which could also have incompatibilities.

Setting the language standard wasn't supported until VS2015 Update 3, a number of these VS extensions supported setting the language standard *before* that update was released. I'm not sure if any of the them changed to support the "official" element name and values, but this creates another discrepancy that can't be represented with a simple table.

I think lookup tables are definitely needed, `floatingpoint "default"` doesn't work in VS2010+ (#701). But I think using a global lookup table is a bad idea, `floatingpoint "strict"` causes Clang builds to fail with unknown flag errors (#700).

Thoughts?

tvandijck:
@samsinsane Yeah, that is kind of what I mentioned in my previous comment.. I don't think putting this table in the action is going to be sufficient, and rather putting things in the toolset is going to be a better approach...

But even there, we currently have a single toolset for "clang" rather then clang3.7, clang3.8, etc... and the same for VS2015, we don't have a VS2015.1, VS2015.3, etc. And you mentioned a bunch of other combinations as well.... clang3.7 on android vs clang3.7 on windows apparently is also different... It's kind of messy to deal with, but we should probably think of a way to do that...

tvandijck:
Closing this... we need to come up with a more general way of dealing with things as per the above discussion. Until that time what we currently do seems to be sufficient... I'm sorry I wasted your time with my request to do this work. It seemed a good idea to me, but is clearly contentious, and created some good questions and discussion.

PR #697, opened 2017-02-20
UTs fix: Use Windows registry locations that exist for built-in users

nsprljan:
I have some Jenkins jobs that run premake UTs and fail because the registry keys used to test os.getWindowsRegistry do not exist for the SYSTEM user (which Jenkins runs under). So I've tried to find the keys that are always present on the system. Tested only on Win7.


tvandijck:
Our jenkins seems to agree....

Issue #706, opened 2017-03-08
http.* with SSL doesn't work on Windows


Jusonex:
Steps to reproduce:
1) Download the _alpha11_ release.
2) Create a `premake5.lua` file with the following content:
   ```lua
   local resource, result_string = http.get("https://google.com")
   print(tostring(result_string))
   ```
3) The output is:
   ```
   PS> .\premake5.exe vs2015
   SSL peer certificate or SSH remote key was not OK
   Cert verify failed: BADCERT_NOT_TRUSTED
   ```
_alpha10_ worked so I suspect it has been introduced with the _mbedtls_ transition. The exact same code snippet does also work on Linux.

PR #707, opened 2017-03-08
Fix missing CA certs on Windows

Jusonex:
This pull request fixes https://github.com/premake/premake-core/issues/706 (_http.* with SSL doesn't work on Windows_) and partially reverts the _mbedtls_ transition.

The main reason for using Microsoft's own TLS implementation (Schannel) is that it doesn't explicitly require us to bundle the CA certs, but uses the the OS' built-in certs instead (see the last section for details: https://curl.haxx.se/docs/sslcerts.html)

### DIFF
--- a/contrib/mbedtls/premake5.lua
+++ b/contrib/mbedtls/premake5.lua
@@ -16,4 +16,5 @@ project "mbedtls-lib"
 		"library/*.c"
 	}
 
-
+	filter { "system:windows or macosx" }
+		flags { "ExcludeFromBuild" }
--- a/contrib/mbedtls/premake5.lua
+++ b/contrib/mbedtls/premake5.lua
@@ -16,5 +16,4 @@ project "mbedtls-lib"
 		"library/*.c"
 	}
 
-	filter { "system:windows or macosx" }
-		flags { "ExcludeFromBuild" }
+
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -53,7 +53,7 @@ mingw: $(SRC)
 
 osx: $(SRC)
 	mkdir -p build/bootstrap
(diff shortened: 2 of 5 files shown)


tvandijck:
I'm OK with this change, but then I think we also want to only build/include/link against mbedtls on linux only... so more changes to the premake5.lua script would be needed.

tvandijck:
FWIW, the only reasons I initially made the change from using OpenSSL to mbedtls, was that I could link against it statically, and don't get a premake binary on linux that only works on one specific distribution of linux.

On Mac & Windows that isn't really an issue, so using their internal 'tls' libraries is perfectly fine with me.

contrib/mbedtls/premake5.lua
```diff
@@ -16,4 +16,5 @@ project "mbedtls-lib"
 		"library/*.c"
 	}
 
-
+	filter { "system:windows or macosx" }
```
tvandijck: Let's revert this file, since it would still create a .lib file that is empty...

I think just not linking against it is enough, since the dependency tree will not include it (at least in visual studio that will work)... Thank you for those modifications...

Jusonex:
It apparently doesn't skip the project anymore now. I guess it's caused by building the entire solution instead of only the Premake5 project.
Shall I revert the revert or do we keep it like this?

tvandijck:
No that is fine.... 

the alternative would be to not include the library script at all on mac and windows... which is on this line: https://github.com/premake/premake-core/blob/master/premake5.lua#L195

an ```if not (os.is("windows") or os.is("macosx")) then``` around it would do it I think.

ddobrev:
Hello, is there anything to fix in this pull before merging it? This bug prevents https://github.com/mono/CppSharp from gaining support for VS 2017 because we're forced to update premake to get its support of VS 2017 but at the same time premake now has this bug which prevents us from downloading some dependencies.

starkos:
Thanks everyone!

ddobrev:
Thank you for merging this. When will there be a new release of premake 5 with this fix?

starkos:
Yes, it will be included in the next alpha release.

PR #710, opened 2017-03-11
Treat *.inl files as headers

Zeenobit:
This addresses issues with Visual Studio and Intellisense when using *.inl files. If they're not treated as headers or source files, Intellisense fails to parse them correctly.

I can't see this having undesirable side effects for existing projects. Ideally, the user should be able to specify how different files should be treated.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -624,7 +624,7 @@
 ---
 	m.categories.ClInclude = {
 		name       = "ClInclude",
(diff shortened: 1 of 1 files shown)


tvandijck:
FWIW: this behavior is already entirely 'overridable'... 
```
premake.vstudio.vc2010.categories.ClInclude.extensions
```
is just a table that can be accessed by anyone at all times...
sure, it's a little bit of 'internal' knowledge, and maybe some sort of API can be invented to do this in a more elegant way... but... you could technically do:

```
table.insert(premake.vstudio.vc2010.categories.ClInclude.extensions, ".inl")
```

PR #716, opened 2017-03-24
Refactor curl code a bit, to be more in line with all the other code.

tvandijck:
It splits the code up a little, and allows other files to reuse the curl wrappers.


tritao:
lgtm

PR #718, opened 2017-03-24
Adding support for NotSet characterset in VS2010+ (#79)

tvandijck:


samsinsane:
I don't fully understand this API. `Default` is actually Unicode? I thought `Default` was always nothing, thus defaulting the value to whatever the action/toolset dictates? Also, just out of curiosity what does `NotSet` mean in VS? I thought it was just swapping between the `A` and `W` functions?

tvandijck:
Apparently there is more to is then just A vs W.... @jrepp might be able to explain

starkos:
> `Default` is actually Unicode? I thought `Default` was always nothing

This one has a bit of history, see [this post to the developer forums](https://groups.google.com/d/msg/premake-development/MvfDl6DzS68/ai7zsDHSBAAJ). I have no objection to revisiting it though.

tvandijck:
FYI, Not Set is an actual setting, other then not specified, empty, which is what our "Default" does.

image

starkos:
If you create a new C++ project via the VS project wizard (do we still call them "wizards"? why did we ever call them "wizards"?) in the IDE, it sets CharacterSet to Unicode. So technically our default value is right, from that perspective.

Best I can tell "NotSet" equates to an ASCII or Single-Byte Character Set. Perhaps we could one of those, more informative names instead? That would make it easier to translate them to the other toolsets as well. What do you think?

tvandijck:
@starkos That sounds like an acceptable alternative yes, I'll adjust the PR, using "ASCII"

tvandijck:
changed, and added some tests.

PR #721, opened 2017-03-27
Fixes to codelite module.

tvandijck:
We ran into a problem where libs were not beeing linked in, if using the Library tag...
not sure if we're doing something wrong or not, but we've had this change in our repository for a while now, which resolved a lot of our issues.

### DIFF
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -4,8 +4,9 @@
 -- Author:      Ryan Pusztai
 -- Modified by: Andrea Zanellato
 --              Manu Evans
+--              Tom van Dijck
 -- Created:     2013/05/06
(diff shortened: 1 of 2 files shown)


modules/codelite/codelite_project.lua
```diff
@@ -4,8 +4,9 @@
 -- Author:      Ryan Pusztai
 -- Modified by: Andrea Zanellato
 --              Manu Evans
+--              Tom van Dijck
 -- Created:     2013/05/06
--- Copyright:   (c) 2008-2015 Jason Perkins and the Premake project
+-- Copyright:   (c) 2008-2016 Jason Perkins and the Premake project
```
samsinsane: I'm going to assume you made these changes last year?

Issue #629, opened 2016-11-19
 Do not support chinese path of source file


kinsung:
Write a file named premake5.lua, and put some .cpp file in the same directory.   
If the directory contains  .cpp files with chinese character, such as "test测试中文路径.cpp", the generated .vcxproj file cannot be opened by Visual Studio. The reason is .vcxproj file contains a few characters appear as gibberish.
    <ClCompile Include="..\..\test��������·��.cpp" />

Sample premake5.lua  as beblow:

 ```
 solution "LibTest"
      configurations { "Debug" }
      location ("build")

  project "LibTest"
    targetname "LibTest"
    language "C++"
    kind "ConsoleApp"
    location ("build")

    files
    {
        "*.cpp", "*.h"
    }
    
    configuration "Debug"
        targetdir "bin"
        defines { "_DEBUG", "WIN32" }




```

fun4jimmy:
This is because `FindFirstFileA` and `FindNextFileA` are used in `os_matchstart` and `os_matchnext` respectively, the ANSI versions do not return anything for files containing unicode character.

I started on a fix and swapped them out for the Unicode versions which correctly listed unicode files but this caused errors later on in `path_normalize` when calling `isspace`. That specific error only occurs in debug builds as there is an debug only check (`_ASSERTE(c >= -1 && c <= 255);`) but I think there would be other knock on effects in other functions.

The changes also require the use of `WideCharToMultiByte` and `MultiByteToWideChar` and calls to malloc/free for temporary strings with every call to `os_matchstart` and `os_matchname` which would make the functions slower to run.

A quick look at some other path related functions would required similar changes to fully support Windows Unicode. For example I think `os_getcwd` will currently not work if you are within a Unicode directory.

I think it would make sense to fully support unicode rather than partially support it from some functions but wanted to check that this is something people want before going any further.

I also wasn't sure what to do in the case of a failure to convert from/to wchar_t:
```c
int os_matchname(lua_State* L)
{
	MatchInfo* m = (MatchInfo*)lua_touserdata(L, 1);
	int widelength = wcslen(m->entry.cFileName);
	int multibytelength = WideCharToMultiByte( CP_UTF8, WC_ERR_INVALID_CHARS, m->entry.cFileName, widelength, NULL, 0, NULL, NULL);
	if (multibytelength != 0)
	{
		char* filename = (char*)malloc(sizeof(char)*(multibytelength+1));
[...]

PR #728, opened 2017-04-02
Add support of UTF-8 for Windows

DrLynix:
Hi there,

We're using Premake at my job and we had some troubles when trying to copy files from/to locations containing unicode characters, the reason for this is that Premake uses the ASCII version of the Windows API.

I made a fix by adding two helper functions (`utf8_fromwide` and `utf8_towide`) and using them with the Unicode Windows API everywhere I could (I just didn't do it for the os.getpass function, making the Windows console working with unicode is not an easy thing to do).

Anyway, tell me if I need to fix some stuff, as I'm not sure I did everything in the "right" (premake-team) way 😄


src/host/os_isdir.c
```diff
@@ -42,6 +44,10 @@ int os_isdir(lua_State* L)
 	{
 		lua_pushboolean(L, 0);
 	}
+	
+#ifdef _WIN32
+	lua_pop(L, -2); /* pop wide string */
```
DrLynix: Oops, should be `lua_remove`, can't fix it right now

src/host/utf8handling.c
```diff
@@ -0,0 +1,33 @@
...
+{
+	int size_required = WideCharToMultiByte(CP_UTF8, 0, wstr, -1, NULL, 0, NULL, NULL);
+
+	char* unicode_str = (char*) malloc(size_required * sizeof(char));
+	WideCharToMultiByte(CP_UTF8, 0, wstr, -1, unicode_str, size_required, NULL, NULL);
+
+	lua_pushstring(L, unicode_str);
+	free(unicode_str);
+	
+	return unicode_str;
```
tvandijck: you just free'd that pointer.... should really not return that, as that will be a "use after free" access violation.

src/host/utf8handling.c
```diff
@@ -0,0 +1,33 @@
...
+ * \file   utf8handking.c
+ * \brief  Handles conversions between Unicode (UTF-8) and system native encoding (wide chars on Windows)
+ * \author Copyright (c) 2017 Jérôme Leclercq and the Premake project
+ */
+
+#include "premake.h"
+#include "stdlib.h"
+
+#ifdef PLATFORM_WINDOWS
+const char* utf8_fromwide(lua_State* L, const wchar_t* wstr)
```
tvandijck: I would prefer a ```size_t utf8_fromwide(char* destination, size_t destSize, const wchar_t* wstr);```

which would not need to allocate memory, would allow for the call to WideCharToMultiByte to simply be called once. And wouldn't need the LuaState... pushing the string to the state would be up to the caller.

tvandijck:
Other then the use after free, I have no real objections... I think API wise relying less on lua and more on just C would be nice, but it's no breaking point for me.. we can always refactor that further if we wish for that..

src/host/utf8handling.c
DrLynix: I though using Lua GC to manage memory was a good idea, but I will make the changes.

src/host/utf8handling.c
DrLynix: Indeed, in my original design I wasn't returning anything. Will change that.

src/host/utf8handling.c
DrLynix: Mmh, actually a wrapper function is useless if all it does is calling `WideCharToMultiByte`, since we can do that in places where we are calling this function (since we're already checking for the Windows part).

What if I make something like that:
```c
const char* utf8_fromwide(const wchar_t* wstr)
 {
 	int size_required = WideCharToMultiByte(CP_UTF8, 0, wstr, -1, NULL, 0, NULL, NULL);
 
 	char* unicode_str = (char*) malloc(size_required * sizeof(char));
 	WideCharToMultiByte(CP_UTF8, 0, wstr, -1, unicode_str, size_required, NULL, NULL);
 
 	return unicode_str;
 }
```

be a better solution? Of course the caller has to free the buffer then (which I dislike).

tvandijck: I agree that that is not preferable... In general I was just hoping that a malloc for this isn't something we need... mostly because these match functions are used ALOT, and so have quite a bit of performance impact... on our project we're already hitting 2GB of memory use.

if that removes the need for this method entirely, then I'm OK with just calling it in place to be entirely honest, and simply use stack for this...

DrLynix:
I'll try to think of something that will not impact performances much then,
especially there.

I'll use the stack wherever I can and remove those two functions then, I
hope I can make it in a not-so-dirty way, despite the Win32 calls.

fun4jimmy:
Some of this stuff was discussed as part of #629. I never got round to doing anything that was proposed but I think this might fix that issue as well.

DrLynix:
I remade this using only stack memory (and I'm wondering why I didn't do that in the first place..).
In case any conversion fails, it will trigger a Lua error, as discussed in #629.

I'm wondering, should I apply the same changes to the integrated Lua library to support unicode operations? (Talking about `io` library)

tvandijck:
Thank you very much, that looks pretty clean to me... ready to merge as far as I'm concerned.

> I'm wondering, should I apply the same changes to the integrated Lua library to support unicode operations? (Talking about io library)

We could, but lets make that a separate pull request.. I'm not too fond of actually changing lua itself to be honest, since that makes moving between versions of Lua really cumbersome.. So I'd rather see extensions to the io library to support this... like an io.open_utf8(...) or something of that kind, I don't know how that would look really.. so I don't really have a very well thought out opinion on this, but it should be fairly straightforward to 'override' the standard io library with methods that support utf8 instead.

I would imagine that really all you need to override is the io.open method? all the other methods work on stream handles (FILE*, etc), so those should not be affected.

DrLynix:
No problem with making another pull request.
I can override the files functions of `io` without touching the Lua library, but I think I'll have to override more than one function (since Windows Unicode stuff doesn't work with FILE* handles AFAIK, I'll check that).

I don't think making a separate Unicode open function will work in this case because of the io metatable (for calls like `handle:read()`), so I'll probably override, except if I find a way to make Unicode works with FILE* handles.

Issue #569, opened 2016-08-25
Nuget keyword not respecting nuget.config, project location, actual lib names


SoleilLapierre:
Hi. I'm currently evaluating premake 5 as a possible new build system at work. So far I'm liking it lots.

I'm having some trouble getting nuget dependencies working though. The first thing I tried to do with it was add a dependency on NUnit 3.4.1 in one project of a vs2015 multi-project solution. I realize it's bleeding edge days for premake 5 and I may have to invest some effort into making it work, but the user guide says if something's not working as you expect, submit a ticket. Also from reading some of your dev discussions and pull requests I think this probably touches on larger questions. 

So here goes. :)

1) There appear to be assumptions that a NuGet package contains only one relevant lib and that it has the same name as the package. These are not always the case, but as far as I can tell the .nupkg file doesn't actually tell you which files you want; it might actually be correct to just find and use all the DLLs in the appropriate subdirectory of the package.

2) The content of the generated packages.config file is correct, but it is placed next to the .sln file rather than the .csproj that actually has the dependency. 

3) The properties added to the .csproj are not correct; in addition to #1 above it adds dependencies for every .net framework version, and fails to add the file include for packages.config. Because of this the package will never be downloaded and the build will fail.

4) If you have a nuget.config file set up so that all your packages download to one nice clean place, that location is not respected. See http://docs.nuget.org/release-notes/nuget-2.1#Specify_%e2%80%98packages%e2%80%99_Folder_Location - there are several places this config can go. (Actually, I'd be totally cool with premake generating the nunit.config file next to the .sln as long as it allowed specifying extra package sources as well as a custom repository path.)

Here are the diffs to the .csproj file if I add the NuGet package manually, using the solution generated from the attached repro case with the nuget line commented out. This is what should happen. Also note that it's using the .net 4.5 version even though I use 4.6, because there is no 4.6 version in the NuGet package.

```
42a43,47
>     <Reference Include="nunit.framework, Version=3.4.1.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
>       <SpecificVersion>False</SpecificVersion>
>       <HintPath>..\..\..\nuget_packages\downloaded\NUnit.3.4.1\lib\net45\nunit.framework.dll</HintPath>
>       <Private>True</Private>
>     </Reference>
57a63,65
>   </ItemGroup>
>   <ItemGroup>
>     <None Include="packages.config" />

```

And here is what actually happens:

```
43a44,54
[...]

PR #729, opened 2017-04-02
Add JSON API

aleksijuvani:
In order to fix some issues with the NuGet support, we need to be able to parse JSON responses from the NuGet API. See https://github.com/premake/premake-core/issues/569#issuecomment-289583588.

So, to that end, this pull request adds two new functions, `json.encode` and `json.decode`. I believe these functions will also be useful to other people in cases where they need to interact with web APIs.

Internally, these functions use [Jeffrey Friedl's JSON library](http://regex.info/blog/lua/json), so they *should be* relatively field-tested.

I ran some of the NuGet API responses through this, and from what I can tell, they're indeed getting decoded correctly.

### DIFF
--- a/src/base/jsonwrapper.lua
+++ b/src/base/jsonwrapper.lua
@@ -4,14 +4,32 @@
 -- Copyright (c) 2017 Jason Perkins and the Premake project
 --
 
-	local implementation = dofile('json.lua')
-
 	json = {}
 
+	local implementation = dofile('json.lua')
+	local encode_implementation = implementation
+	local decode_implementation = implementation:new()
+
+	local err
+
+	function decode_implementation.assert(condition, message)
+		if not condition then
+			err = message
+		end
+	end
+
 	function json.encode(value)
-		return implementation:encode(value)
+		return encode_implementation:encode(value)
 	end
 
 	function json.decode(value)
-		return implementation:decode(value)
+		err = nil
+
+		local result = decode_implementation:decode(value)
+
+		if err then
+			return nil, err
+		end
+
+		return result
 	end
--- a/src/base/jsonwrapper.lua
+++ b/src/base/jsonwrapper.lua
@@ -1,0 +1,17 @@
+--
+-- jsonwrapper.lua
+-- Provides JSON encoding and decoding API by wrapping a third-party JSON library
+-- Copyright (c) 2017 Jason Perkins and the Premake project
+--
+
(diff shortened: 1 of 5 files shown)


aleksijuvani:
As a side note, the `json.decode` function throws an error if parsing fails. Do we want to go with that, or maybe return `nil` and an error message like `http.get` does?

src/base/jsonwrapper.lua
```diff
@@ -0,0 +1,17 @@
+--
+-- jsonwrapper.lua
+-- Provides JSON encoding and decoding API by wrapping a third-party JSON library
+-- Copyright (c) 2017 Jason Perkins and the Premake project
+--
+
+	local implementation = dofile('json.lua')
+
+	json = {}
```
tvandijck: I like this wrapper idea... just think it should be in the "premake.json" table, rather then a global json table...

tvandijck:
> Do we want to go with that, or maybe return nil and an error message like http.get does?

good question... I don't really have an opinion one way or the other... the nil+error might be more convenient. At the same time going with the default hasn't really bothered us here with that same json plugin... We've just not wrapped it in our code right now, but I'm somewhat liking that idea.

src/base/jsonwrapper.lua
aleksijuvani: I thought about that (and it actually was that way at first), but I felt like it would be better if it was global. The `http` API that we have is global, and if you're going to be making any kind of web API calls with that, you're probably going to need a JSON parser as well.

aleksijuvani: I'm open to changing this though. It doesn't make any difference for NuGet.

tvandijck: hmm yeah, you got a point there ;)

aleksijuvani:
I did the `nil` and error message thing for `json.decode`. If we don't want that, just revert 52a66a67f1385c637d9912f84f2e1d5839d8e2e2.

I tried doing the same thing for `json.encode`, but turns out that the library is wildly inconsistent with encoding errors. Sometimes it would raise a Lua error and sometimes it would set the error message. I guess we could get around this with `pcall`, but I decided to leave it as it is – raise errors if encoding fails.

aleksijuvani:
Apparently the library expects us to abort in the encode error handler (but unintuitively, it's not required in the decode error handler). If we abort in the error handler, we get proper error messages.

PR #732, opened 2017-04-06
Fix VS2013 and older always rebuilding when debug symbols explicitly …

jstewart-amd:
…disabled

Visual Studio versions 2013 and older have an issue with
DebugInformationFormat set to None. The project will always be out of
date and thus always rebuild. The workaround is to leave the
DebugInformationFormat field blank.

https://connect.microsoft.com/VisualStudio/feedback/details/833494/project-with-debug-information-disabled-always-rebuilds

VS2015 and newer do not have this issue.

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1262,7 +1262,14 @@
 
 			m.element("DebugInformationFormat", nil, value)
 		elseif cfg.symbols == p.OFF then
-			m.element("DebugInformationFormat", nil, "None")
+			-- leave field blank for vs2013 and older to workaround bug
+			if _ACTION < "vs2015" then
+				value = ""
+			else
+				value = "None"
+			end
+
+			m.element("DebugInformationFormat", nil, value)
 		end
 	end
 
--- a/tests/actions/vstudio/vc2010/test_compile_settings.lua
+++ b/tests/actions/vstudio/vc2010/test_compile_settings.lua
@@ -790,9 +790,29 @@
 
 --
 -- Check handling of the explicitly disabling symbols.
+-- Note: VS2013 and older have a bug with setting
+-- DebugInformationFormat to None. The workaround
+-- is to leave the field blank.
 --
 	function suite.onNoSymbols()
 		symbols 'Off'
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<DebugInformationFormat></DebugInformationFormat>
+	<Optimization>Disabled</Optimization>
+		]]
+	end
+
+
+--
+-- VS2015 and newer can use DebugInformationFormat None.
+--
+	function suite.onNoSymbolsVS2015()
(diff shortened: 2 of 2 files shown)


jstewart-amd:
I added a new test, so now there are separate tests for pre-VS2015 (leave the field blank) and VS2015 and later (using None works as expected).

jstewart-amd:
More information about the underlying issue in Visual Studio can be found here:

Project with debug information disabled always rebuilds
https://connect.microsoft.com/VisualStudio/feedback/details/833494/project-with-debug-information-disabled-always-rebuilds

"The issue you are seeing is because the specific code incorrectly doesn't check for None. It looks for OldStyle and Empty."

"You can workaround it by blanking out the property field. It will be passed to the build as an empty string and correctly by pass the 'Forcing recompile'."

jstewart-amd:
.\bin\release\premake5.exe test
Running action 'test'...
1475 tests passed, 0 failed in 1.97 seconds

samsinsane:
I kind of feel like this was recently changed to be the way it is. Are all the files that are generated by the various versions of visual studio, *exactly* the same with and without this PR? I'm a little hesitant to approve a PR that could just be reverted in a few weeks/months because *something* is different.

> .\bin\release\premake5.exe test
> Running action 'test'...
> 1475 tests passed, 0 failed in 1.97 seconds

Thanks for this, but it's not necessary as the build scripts perform these checks when you create your PR. 😃

tvandijck:
LGTM too... won't affect us...

jstewart-amd:
Thanks, everybody. I wasn't ignoring @samsinsane. Rather, I was doing more testing.

To clarify, Premake without this fix is not producing incorrect output for VS2012 and VS2013. But it is running into a Visual Studio bug. The bug manifests when explicitly setting debug symbols to "Off". The symptom is that Visual Studio always thinks the project is out of date and so always does a full recompile (for the build configs where you have explicitly disabled symbols).

The issue showed up in the switch from `flags { "Symbols" }` to the new `symbols` function. Prior to the
`symbols` function, there was no notion of explicitly disabling symbols. You either had `flags { "Symbols" }` or you didn't. Absence of `flags { "Symbols" }` was an implicit disable.

This PR is a workaround for the Visual Studio bug, following the suggestion from Microsoft in the connect.microsoft.com link mentioned in a previous comment.

For VS2010, Premake actually **is** producing incorrect output when symbols are explicitly disabled. VS2010's DebugInformationFormat does not have a None setting. The result is this error:
`error : Element <DebugInformationFormat> has an invalid value of "None".`

This PR also fixes the VS2010 issue.

Summary:
* VS2010 - Does not support DebugInformationFormat None. Leave field blank to get the behavior of None.
* VS2012/VS2013 - Supposedly supports DebugInformationFormat None, but it doesn't work. Leave field blank to get the behavior of None.
* VS2015 and newer - DebugInformationFormat None actually works.

I have tested on VS2010, VS2012, VS2013, and VS2015. I still don't have VS2017 installed anywhere, but output produced for VS2015 and newer is unchanged by this PR.

tvandijck:
I verified correct behavior in VS2017....

samsinsane:
Sounds good to me, thanks for the fixes!

PR #733, opened 2017-04-06
Generated project files are now correctly located beside script

### LINKED ISSUE #371: Projects not being built in the correct directories.
I have a premake5 script that's doing this:

```
local libPremakes = 
{
    "glload/glload.lua",
    "glimg/glimg.lua",
    "glfw/glfw.lua",
    "freeglut/freeglut.lua",
    "glutil/glutil.lua",
    "glmesh/glmesh.lua",
    "glscene/glscene.lua",
}

for i, luaPremake in ipairs(libPremakes) do
    if(#os.matchfiles(luaPremake) ~= 0) then
        dofile(luaPremake)
    end
end
```

Each of these files creates a project. The problem is that it doesn't create the project in those directories. The projects are being created in the root directory

According to the [workspaces and projects](https://github.com/premake/premake-core/wiki/Workspaces-and-Projects) documentation page, "By default, Premake will place generated workspace and project files in the same directory as the script which defined them." If that's supposed to be the case, then I shouldn't have to add `location "."` to all of those Lua scripts, right? So either Premake5 is doing the wrong thing or the docs are wrong.

I'm using Premake5.0.0 alpha-7, and I'm building for vs2015.

samsinsane:
Fixes #371

### DIFF
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -235,7 +235,7 @@
 		-- location. Any path tokens which are expanded in non-path fields
 		-- are made relative to this, ensuring a portable generated project.
 
-		self.location = self.location or wks.location or self.basedir
+		self.location = self.location or self.basedir
 		context.basedir(self, self.location)
 
 		-- This bit could use some work: create a canonical set of configurations


starkos:
I'm trying to remember the conversations that we had around this. I *think* the behavior we were trying to implement is that if `location` were set on the workspace, this value would be inherited by the projects, like any other configuration value. I'm not 100% sure if we have a test for that or not?

```lua
workspace "MyWorkspace"
  location "build"

project "MyProject"
  -- implicit location is "build", inherited from MyWorkspace
```

So if a location is set on the workspace, it should be inherited by the project. But if no location is set, then project should use its `basedir`.

samsinsane:
This PR appears to do that.

Using Premake as the test.
Without this PR:
```
Premake5.sln
Premake5.vcxproj
lua-lib.vcxproj
```

With this PR:
```
Premake5.sln
Premake5.vcxproj
contrib/lua/lua-lib.vcxproj
```

With this PR and `--to=testing`: (which only applies to the workspace)
```
testing/Premake5.sln
testing/Premake5.vcxproj
testing/lua-lib.vcxproj
```

I'm happy to add tests in for this kind of thing, but I'm not sure exactly how to test it?
```lua
wks = test.createWorkspace()
origin = os.getcwd()
os.setcwd(origin + "/folder")
prj = test.createproject(wks)
os.setcwd(origin)

test.issame(origin, wks.location)
test.issame(origin + "/folder", prj.location)
```
This seems like too much of a hack?

starkos:
> I'm happy to add tests in for this kind of thing, but I'm not sure exactly how to test it?

I'm probably wasn't either, which is probably why there aren't tests for it :) What about…?

```lua
wks = test.createWorkspace()
wks.basedir = os.getcwd()

prj = test.createproject(wks)
prj.basedir = path.join(os.getcwd(), "folder")
```

Going to approve this anyway, since you've confirmed we're getting the behavior we want for both cases.

tritao:
Just had to workaround this issue today... thanks for the fix @samsinsane.

PR #734, opened 2017-04-07
NuGet fixes

aleksijuvani:
This pull request should fix everything mentioned #569, except for `nuget.config` customization. Premake now uses `api.nuget.org` to fetch required information about the packages.

I've tested this against the following NuGet packages:

C++:

- boost
- boost_filesystem-vc140
- boost_system-vc140
- sdl2.v140
- sdl2.v140.redist
- glew.v140
- harfbuzz
- physfs.v140
- freetype2
- assimp.v140
- assimp.v140.redist
- pugixml

C#:

- Newtonsoft.Json
- NUnit
- SSH.NET

There is one problem where older C# packages (published before 2016 or sometime 2016?) might not work because the API response for them doesn't always contain a file listing. There isn't really anything that we can do about this, other than to inform the user and abort.

One thing to also note is that this will increase generation times for projects with NuGet packages. With 12 packages, I was getting generation times of around 20 seconds (`api.nuget.org` is slow).

We might want to consider caching the API requests to a file. If we had an asynchronous HTTP API, that could also help because then we could run requests in parallel. Any thoughts?

Anyway, I'd appreciate it if people with C# projects could try out this branch before it was merged, preferably with more and different packages (@tritao, @SoleilLapierre, @tvandijck?). I don't have any personal C# projects, so I only have a dummy project to test this with.


tritao:
Looks awesome, I'll test it out during the weekend.

src/actions/vstudio/vs2010_nuget.lua
```diff
@@ -54,33 +37,148 @@
...
-		local wks = obj.workspace
+			printf("Examining NuGet package '%s'...", id)
+			io.flush()
 
-		local done = {}
-		local packages = {}
-		for prj in p.workspace.eachproject(wks) do
-			for i = 1, #prj.nuget do
-				local package = prj.nuget[i]
+			local response, err, code = http.get(string.format("https://api.nuget.org/v3/registration1/%s/index.json", id:lower()))
```
tvandijck: I really like this to be configurable, there is other nuget streams out there, and certainly within companies like ours (Blizzard) we host our own internal caching streams, through things like artifactory, and myget...

aleksijuvani: I just tried this with MyGet, and while it looks like while they implement the same API as `api.nuget.org`, they don't return a file listing for the packages (at least if they were uploaded by their automatic `packages.config` tool). There's something weird going on here.

`api.nuget.org` doesn't return a file listing for all packages either, but from my testing that only seemed to happen with older packages.

If the API doesn't provide us with a file listing, the only other way that I can think of is to have Premake download and extract the packages.

aleksijuvani: I did some digging and I think that reporting the file listing is unsupported by [NuGet.Server](https://github.com/NuGet/NuGet.Server). I think they have a different server implementation running at `nuget.org` and they populate the file listings there with a background task.

If I'm right about this, we could support custom package sources, but people wouldn't be able to use any servers with it, because the open source server doesn't return the data that we need...

I suppose people could use it with C++ projects because we don't need the file listing for those, but C# projects wouldn't work with it.

One thing that we could do is file an issue to the [NuGet.Server issue tracker](https://github.com/nuget/NuGetGallery/issues?q=is%3Aissue+label%3A%22Area%3A+NuGet.Server%22+is%3Aopen), but I'm not comfortable doing so until I'm sure that I've understood the issue correctly.

tvandijck: API's and support aside, let's at least make this configurable... a nuget.serverlist = {}, or something...

Then as for the filelist.. for now, we can simple error... and if we hit it, we can look at implementing whatever is needed to get it to work... Internally here at Blizzard we are using NuGet.Server, so yeah, we're going to hit that, since those implementations only implement the V2 API of nuget, and it seems like you are querying the V3 API.. Which is all right, if we need it, we'll either get to it ourselves, or someone else who needs it will...

src/base/validation.lua
```diff
@@ -219,6 +220,13 @@
 	end
 
 
+	function m.NuGetHasHTTP(prj)
```
tvandijck: that is a pretty smart solution ;) I like it...

aleksijuvani:
How's this?

aleksijuvani:
I did run into something else when testing this. Some packages, such as MySQL.Data depend on other packages, but this doesn't currently handle those dependencies. So that's going to need some fixing. Should we fix it before merging, or should we do it in a separate PR?

tvandijck:
I'd say, separate PR... I'm sure there will be many more nuget improvements over time...

aleksijuvani:
Sure.

SoleilLapierre:
I just tried from c5bf4b9 but I get errors when creating a NuGet dependency on NUnit. If I use:

`    nuget { "NUnit:3.6.0" }`

then I get: 

```
Running action 'vs2015'...
Examining NuGet package source 'https://api.nuget.org/v3/index.json'...
Error: ** Error: NuGet API error (0)
SSL connect error
schannel: failed to retrieve ALPN result
```

If  leave out the :3.6.0 version number, the error becomes:

```
Running action 'vs2015'...
Generated Zaber.Plugins.sln...
Generated Zaber.Plugins/Zaber.Plugins.csproj...
Error: [string "src/actions/vstudio/vs2010_nuget.lua"]:9: attempt to perform arithmetic on a nil value
```

I was following the instructions in BUILD.txt to build Premake on Windows using VS2015. "premake5 test" fails with this build.

aleksijuvani:
I think we're hitting a libcurl bug here (https://github.com/curl/curl/issues/840), solution would be to update libcurl.

aleksijuvani:
I've opened pull request #738 to update libcurl.

src/actions/vstudio/vs2005_csproj.lua
```diff
@@ -368,33 +368,89 @@
...
+					-- <Culture=neutral, PublicKeyToken=2638cd05610744eb,
+					-- <processorArchitecture=MSIL">
+					--
+					-- We're just outputting:
+					--
+					-- <Reference Include="nunit.framework">
+
+					_x(2, '<Reference Include="%s">', path.getbasename(file))
+					_x(3, '<HintPath>%s</HintPath>', vstudio.path(prj, p.filename(prj.solution, string.format("packages\\%s.%s\\%s", id, packageAPIInfo.verbatimVersion or packageAPIInfo.version, file))))
+					_p(3, '<Private>True</Private>')
```
ClxS: One thing to note about this which we've had to fix on our internal version - this doesn't respect NoCopyLocal. On our version we replaced this line with:
if premake.config.isCopyLocal(prj, package, cfg) then
_p(3, 'True')
else
_p(3, 'False')
end

aleksijuvani: Thanks for letting me know. I've fixed this in bde50fc.

PR #738, opened 2017-04-11
Update curl to 7.53.1

aleksijuvani:
Should fix an issue where all HTTPS requests on Windows versions older than Windows 8.1 or Windows Server 2012 R2 were failing with the error message "schannel: failed to retrieve ALPN result" (https://github.com/curl/curl/issues/840). Doesn't look like there are any breaking changes as far as I can tell.


aleksijuvani:
Looks like Linux and macOS builds are failing. I'll take a look.

aleksijuvani:
There we go.

tritao:
lgtm

PS: We might want to consider submodule'ing curl in the future.

tvandijck:
@tritao I wish we could, but we hand-crafted our curl_config.h to work for the platforms we currently sort of support.... curl itself when build through ./configure, or cmake generates that file through autoconf like tools... Basically we have 'local' changes to curl that have to be integrated into any updates.

Issue #722, opened 2017-03-27
_OS overwritten by action.


tvandijck:
in action.lua
```
		if act then
			_OS = act.os or _OS
		end
```

both the xcode and vs200x set the os variable, which causes those actions to always run as if they run on those platforms... 'macosx' and 'windows' respectively...

The problem is... we then have code like this:
```
			-- Identify the system
			local arch
			if _OS == "windows" then
				arch = os.getenv("PROCESSOR_ARCHITECTURE")
			elseif _OS == "macosx" then
				arch = os.outputof("echo $HOSTTYPE")
			else
				arch = os.outputof("uname -m")
			end
```
such as in os.is64bit(), which assumes that the _OS variable is correctly identifying the platform we're actually running on, which is not the case... and hence this method just returns incorrect results.

usage of _OS, os.is, os.get()... they are basically returning the OS the action is targetting, not what we are executing on.

starkos:
Oops.

A quick code search makes it seem like the majority of the time we are really interested in the target OS, rather than the one we are currently on, i.e. most of the uses of `_OS` are correct as is. So it probably makes sense to add a new method like `os.current()` or `os.real()` to get the actual system identifier? Does that seem reasonable?

tvandijck:
yeah... I was thinking of a similar approach.. it seems a bit of a bandaid though...

I'd rather switch it around....
```os.target()``` is what the action is targeting... 

basically:
```os.get()``` -> ```os.target()```
```os.is()``` -> ```os.targetting()```
```_OS``` -> ```_TARGET_OS```

and then add:
```os.current()```

Then deprecate the old API entirely. We can keep it around, but basically add a ```p.warnonce()``` in it.
The harder one is the _OS variable, so I don't think you can really get away from that one, but we can at least just not use it anywhere in premake-core code.

I'm not sure though....

starkos:
I agree, this is a better long term approach. I'm not opposed to deprecating the `_OS` global too…we could probably catch it and warn via a metatable on _G?

PR #739, opened 2017-04-11
Refactor _OS, os.get() & os.is()

tvandijck:
This fixes: https://github.com/premake/premake-core/issues/722

replaced:
```
os.get() -> os.target()
os.is() -> os.istarget()
_OS -> _TARGET_OS
action.os -> action.targetos
```

added:
```
os.host()
os.ishost()
```


src/host/os_current.c
```diff
@@ -0,0 +1,13 @@
+/**
+ * \file   os_current.c
+ * \brief  Get the current OS we're executing on.
+ * \author Copyright (c) 2014-2017 Tom van Dijck, Jason Perkins and the Premake project
+ */
+
+#include "premake.h"
+
+int os_current(lua_State* L)
```
tvandijck: I did it this way, because I want it to be absolutely read-only at runtime.

tritao:
This is a much needed clarification! 👍 

I wonder if we instead of "current" we should use "host" OS term instead, but might be just be my bias showing due to dealing a lot with that terminology with cross compilers.

tvandijck:
I like ```host``` it does seem a little more explicit..

src/base/os.lua
```diff
@@ -134,26 +134,65 @@
...
+		__newindex = function(t, k, v)
+			if (k == '_OS') then
+				premake.warnOnce("_OS+set", "_OS is deprecated, use '_TARGET_OS'.")
+				rawset(t, "_TARGET_OS", v)
+			else
+				rawset(t, k, v)
+			end
+		end
+	}
+	setmetatable(_G, _G_metatable)
```
samsinsane: Perhaps this should be a helper for variables that are deprecated in the future?

tvandijck: This method actually has quite the impact on performance... ANY access to a global, now has a string compare first.. so I wanted to keep this as simple as possible, and really only keep it in there for the next 2-3 alpha releases. I really don't want to write a 'generic' deprecated globals table...

PR #741, opened 2017-04-13
Removed deprecated APIs and flags older than two years

samsinsane:
Not sure if this is wanted or not, but I think two years is more than enough time to stop using those deprecated APIs and flags.

### DIFF
--- a/modules/d/tools/dmd.lua
+++ b/modules/d/tools/dmd.lua
@@ -218,7 +218,7 @@
 			CodeCoverage	= "-cov",
 			Deprecated		= "-d",
 			Documentation	= "-D",
-			FatalWarnings	= "-w",
+--			FatalWarnings	= "-w",
 			GenerateHeader	= "-H",
 			GenerateJSON	= "-X",
 			GenerateMap		= "-map",
--- a/modules/xcode/tests/test_xcode_project.lua
+++ b/modules/xcode/tests/test_xcode_project.lua
@@ -1153,7 +1153,7 @@
 
 
 	function suite.XCBuildConfigurationProject_OnOptimizeSpeed()
-		flags { "OptimizeSpeed" }
+		optimize "speed"
 		prepare()
 		xcode.XCBuildConfiguration_Project(tr, tr.configs[1])
 		test.capture [[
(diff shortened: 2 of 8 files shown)


ratzlaff:
Instead of wholesale removal of the deprecated values in src/_premake_init.lua, could I perhaps entice you to create a module that people such as myself could 'require' in my premake5.lua that puts them back into the environment? 

We have scripts that work, even with the deprecation warnings. There has been no real incentive to "move over" to new words because there is no new functionality along with the new words. This will completely break my company's scripts (some which I have no control over).

For someone who is not developing premake, it was not readily apparent that things like
'flags { "EnableSSE" }`
needed to be changed to 
`vectorextensions "SSE2"`

Digging through the docs did not tell me immediately and I had to read the premake code around the symbols I knew were generated to figure out what to set in my script to replicate the deprecated behavior.

starkos:
> We have scripts that work, even with the deprecation warnings. There has been no real incentive to "move over" to new words because there is no new functionality along with the new words.

Actually, there is an incentive: deprecated stuff gets removed eventually. I'm with Sam: two years is plenty of time.

If you are interested in contributing a compatibility module, that would be very cool, but you will still want to put a limit on how long you support things.

> For someone who is not developing premake, it was not readily apparent that things like
`flags { "EnableSSE" }` needed to be changed to `vectorextensions "SSE"`

Hmm…the newer deprecations point you to the new APIs, e.g. `flags "Symbols"` outputs:

```
** Warning: the flags value Symbols has been deprecated.
   Use `symbols "On"` instead
```

I'm not sure why the older deprecations aren't doing that as well. And in fact, I just tried a test with `flags "EnableSSE"` and it doesn't even print a deprecation warning at all!

Ugh.

@samsinsane…what about modifying this PR to remove only the ones that have deprecation messages associated with them (`nil` for the third argument). I will submit a PR to add messages the rest and to change the deprecation date to today.

Is that reasonable to everyone?

starkos:
I created PR #743 which adds the missing messages and removes deprecated APIs that had a warning message that are older than one year. I also strengthened the deprecation warning message from "xxx is deprecated" to "xxx is deprecated and will be removed." 

I suggest we close this one and merge that instead. What do you think?

ratzlaff:
That seems reasonable to me

samsinsane:
Sorry about that, I didn't notice the messages weren't there. I noticed the `nil` but it didn't click that it was the messages, it was pretty late when I started that. I've merged #743 and will close this one off.

starkos:
No worries! I assume that I am the one who left the messages off in the first place, though I've no idea why…

PR #742, opened 2017-04-13
Added BSD support

### LINKED ISSUE #309: Doesn't even come close to working on *BSD.
See this [stackoverflow question](http://stackoverflow.com/questions/31169605/premake5-wont-build-on-freebsd-10-1/).

samsinsane:
Resolves #309 - only tested against FreeBSD 11 (all tests pass)


samsinsane:
The tests were passing, including the HTTP tests, which is why the CA bundle location was added. Travis supports using Docker, so we might be able to add BSD to our list, I don't really understand the Travis documentation, it seems to assume you already know how it works. I'll try some things out and see how it goes.

PR #746, opened 2017-04-17
Fixed erroneous escape usages in VS2010+

### LINKED ISSUE #725: '&&' gets translated to '&amp;amp;&amp;amp;' in VS projects
Hello again.

I needed to define the following expression for one of my projects:

    defines
    {
        "DOCTOPT_USE_BOOST_REGEX=(  _DEBUG && _WIN64 && _ITERATOR_DEBUG_LEVEL != 0 )"
    }

This got translated to the following in the vcxproj file (I see this code is located [here](https://github.com/premake/premake-core/blob/1b776d8f6bf32bbbbb0237dd8e6b27a7c9762321/src/actions/vstudio/vs2010.lua#L100) ):

(newlines added only to separate each definition / for easier reading)

    <PreprocessorDefinitions>
        WIN64;
        DEBUG;
        WINDOWS;
        WIN32;
        DOCTOPT_USE_BOOST_REGEX=( _DEBUG &amp;
        amp;
        &amp;
        amp;
        _WIN64 &amp;
        amp;
        &amp;
        amp;
        _ITERATOR_DEBUG_LEVEL != 0 );
        %(PreprocessorDefinitions)
    </PreprocessorDefinitions>`

As we can see each `&&` got translated into `&amp;amp;&amp;amp;`, when it should be `&amp;&amp;`.

So, ofcourse I rewrote my expression to something that would not be translated:

    defines
    {
        "DOCTOPT_USE_BOOST_REGEX=!( !_DEBUG || !_WIN64 || _ITERATOR_DEBUG_LEVEL == 0 )"
    }

But people might say that this is not readable, to which I agree.
However it works nicely as a temporary workaround for now.

samsinsane:
- Added tests for sensible escape usage locations (defines, not undefines or warnings)

Resolves #725

### DIFF
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1797,7 +1797,7 @@
 	function m.nmakePreprocessorDefinitions(cfg)
 		if cfg.kind ~= p.NONE and #cfg.defines > 0 then
 			local defines = table.concat(cfg.defines, ";")
-			defines = p.esc(defines) .. ";$(NMakePreprocessorDefinitions)"
+			defines = defines .. ";$(NMakePreprocessorDefinitions)"
 			m.element('NMakePreprocessorDefinitions', nil, defines)
 		end
 	end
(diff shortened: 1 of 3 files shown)


tvandijck:
not sure I understand why the p.esc call is removed? why is that wrong? is m.element already escaping everything?

samsinsane:
Yeah, `m.element` calls `p.esc` on the third parameter if there are only three parameters.

tvandijck:
thanks for adding tests too... good to merge.

PR #747, opened 2017-04-19
Bootstrap makefile now cleans up previous builds before building

### LINKED ISSUE #334: Remove intermediate files in bootstrap build
I ran into an issue where I wasn't able to build Premake properly after an update. I tried to use the boostrap makefile to get things back to a working state but got flooded with a bunch of linker errors as it tried to link against the object files from the previous build. I was able to fix it by removing all of the `bin/` and `obj/` folders manually.

It would probably be a good idea to always clean those folders out when doing a bootstrap build, just to make sure we're getting a clean build with the latest code.

samsinsane:
Tested with "windows", "windows-msbuild", "linux" and "bsd", assuming it works for "osx" and "mingw".

Resolves #334

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -45,6 +45,9 @@ none:
 	@echo "   osx linux"
 
 mingw: $(SRC)
+	$(SILENT) rm -rf ./bin
+	$(SILENT) rm -rf ./build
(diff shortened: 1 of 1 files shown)


starkos:
Nice change, thanks

PR #752, opened 2017-04-24
Fixed various issues with compiling with VS2012

### LINKED ISSUE #696: Bootstrap fails with Visual Studio 2012
The bootstrap makefile fails when using VS 2012. The command `nmake -f Bootstrap.mak MSDEV=vs2012 windows` produces:

```
src/host/os_getversion.c(88) : error C2275: 'HKEY' : illegal use of this type as an expression
        C:\Program Files (x86)\Windows Kits\8.0\include\shared\minwindef.h(239) : see declaration of 'HKEY'
src/host/os_getversion.c(88) : error C2146: syntax error : missing ';' before identifier 'key'
src/host/os_getversion.c(88) : error C2065: 'key' : undeclared identifier
src/host/os_getversion.c(89) : error C2065: 'key' : undeclared identifier
src/host/os_getversion.c(89) : warning C4047: 'function' : 'PHKEY' differs in levels of indirection from 'int *'
src/host/os_getversion.c(89) : warning C4024: 'RegOpenKeyExA' : different types for formal and actual parameter 5
src/host/os_getversion.c(94) : error C2065: 'key' : undeclared identifier
src/host/os_getversion.c(94) : warning C4047: 'function' : 'HKEY' differs in levels of indirection from 'int'
src/host/os_getversion.c(94) : warning C4024: 'RegQueryValueExA' : different types for formal and actual parameter 1
src/host/os_getversion.c(95) : error C2065: 'key' : undeclared identifier
src/host/os_getversion.c(95) : warning C4047: 'function' : 'HKEY' differs in levels of indirection from 'int'
src/host/os_getversion.c(95) : warning C4024: 'RegCloseKey' : different types for formal and actual parameter 1
```

The bootstrap *does* work on VS 2013 and 2015.

samsinsane:
I was able to somehow target v110 with VS2015, but it wasn't able to do the full build, so I may have missed things.

Resolves #696

### DIFF
--- a/src/host/os_chdir.c
+++ b/src/host/os_chdir.c
@@ -11,8 +11,6 @@ int do_chdir(lua_State* L, const char* path)
 {
 	int z;
 
-	(void)(L);  /* warning: unused parameter */
-
 #if PLATFORM_WINDOWS
 	wchar_t wide_buffer[PATH_MAX];
 	if (MultiByteToWideChar(CP_UTF8, 0, path, -1, wide_buffer, PATH_MAX) == 0)
@@ -23,6 +21,8 @@ int do_chdir(lua_State* L, const char* path)
 
 	z = SetCurrentDirectoryW(wide_buffer);
 #else
+	(void)(L);  /* warning: unused parameter */
+
 	z = !chdir(path);
 #endif
 
--- a/src/host/os_isfile.c
+++ b/src/host/os_isfile.c
@@ -18,10 +18,9 @@ int os_isfile(lua_State* L)
 
 int do_isfile(lua_State* L, const char* filename)
 {
-	(void)(L);  /* warning: unused parameter */
-
 #if PLATFORM_WINDOWS
 	wchar_t wide_path[PATH_MAX];
+	DWORD attrib;
 
 	if (MultiByteToWideChar(CP_UTF8, 0, filename, -1, wide_path, PATH_MAX) == 0)
 	{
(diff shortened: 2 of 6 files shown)


src/host/os_chdir.c
```diff
@@ -26,6 +24,8 @@ int do_chdir(lua_State* L, const char* path)
 	z = !chdir(path);
 #endif
 
+	(void)(L);  /* warning: unused parameter */
```
tvandijck: if we don't use it... why is it passed in?? I think we should just remove this argument entirely....
bigger change, but probably better long term. I think this is just a remnant for a previous PR.

src/host/os_isfile.c
```diff
@@ -42,5 +41,7 @@ int do_isfile(lua_State* L, const char* filename)
 	}
 #endif
 
+	(void)(L);  /* warning: unused parameter */
```
tvandijck: same thing... if not used, don't pass it in..

tvandijck:
Approved, but I think further changes/cleanups are possible.

src/host/os_chdir.c
samsinsane: For some strange reason, only Windows uses it for an error message. I thought it was very strange, I probably should have scoped it to the non-Windows path, I can make that change if you'd like? :)

tvandijck: Ah yes, I see.... yeah, probably best to put it in an #else then.

PR #757, opened 2017-04-25
Increased consistency of p. usage

samsinsane:
This is a pretty large change, but we went from ~900 usages of `premake.` to ~175. I probably went a bit overboard, and I'm happy to remove the ones where I added `local p = premake`.


modules/d/tools/dmd.lua
```diff
@@ -6,9 +6,10 @@
 
 	local tdmd = {}
 
-	local project = premake.project
-	local config = premake.config
-    local d = premake.modules.d
+	local p = premake
+	local project = p.project
+	local config = p.config
+    local d = p.modules.d
```
tvandijck: tab vs spaces... may want to check other files as well...

samsinsane: The file was already using spaces, I feel like that change might be a bit unrelated to this PR. I'm happy to do up another PR if you'd like?

tvandijck: OK... I think we should certainly make things consistent all around, according to the editorconfig file that is committed to the repository, and anyone committing to this repo should 100% use that tool in all their editors... so yeah, maybe a separate commit for that would be better.

tvandijck:
Change looks harmless to me too... merge it ;)

tvandijck:
BTW... I'm pretty thankful you are willing to go through monkey work like this, so thank you very much.

samsinsane:
Haha, no worries! :)

PR #760, opened 2017-04-26
Adding ARM support for Visual Studio.

tvandijck:
This just adds the ARM architecture to Visual Studio projects using the Microsoft provided ARM compiler.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -26,6 +26,7 @@
 			"universal",
 			p.X86,
 			p.X86_64,
+			p.ARM,
 		},
 		aliases = {
 			i386  = p.X86,
--- a/src/actions/vstudio/_vstudio.lua
+++ b/src/actions/vstudio/_vstudio.lua
@@ -24,6 +24,7 @@
 		x86     = "x86",
 		x86_64  = "x64",
 		xbox360 = "Xbox 360",
+		ARM     = "ARM",
 	}
 
 	vstudio.vs2010_architectures =
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -159,6 +159,7 @@
 				m.platformToolset,
 				m.wholeProgramOptimization,
 				m.nmakeOutDirs,
+				m.windowsSDKDesktopARMSupport,
 			}
 		end
 	end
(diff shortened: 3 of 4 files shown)


samsinsane:
Was thinking about this a bit more, `ARM` isn't all that descriptive and could mean anything really. For the Android extension of VS, `ARM` translates to `ARMv7-a` (I think this would suffice as `ARMv7`), I think we should probably attempt to support the proper names, rather than the simplification that MS has introduced here.

tvandijck:
Yeah, I don't know... I got a request from a collegue to add this, and this is how I got Visual Studio's native ARM compiler to work... so I totally get that this is an over simplification maybe, but I don't know anything about ARM in the realm of android or ios and all the 5 million different architectures...

tvandijck:
ARM64 isn't out yet from Microsoft, so I have no way to test that in VS2017 really... couple more months apparently...

samsinsane:
> I got a request from a collegue to add this

Interesting. Does this require any sort of VS extension or is part of the base VS2017 install?

> I don't know anything about ARM in the realm of android or ios and all the 5 million different architectures

That's totally fair enough, is your colleague able to shed any light on the "proper" architecture? I don't really have an issue with this going in as is, but in the future it would be nice to deprecate this correctly.

> ARM64 isn't out yet from Microsoft, so I have no way to test that in VS2017 really... couple more months apparently...

Ah ok, I know that VS2015 Update 3 has ARM64 for Android, which would require the same changes you've made here (excluding the SDK thing), except "ARM64". But given that this is for something else, and MS have previously created differing elements/values, not adding ARM64 is probably for the best.

tvandijck:
> Does this require any sort of VS extension or is part of the base VS2017 install?

In both VS2015 and VS2017 it's just a checkbox in the installation... not an extension or anything.
In VS2017's installer it's here:
image

tvandijck:
I think it allows you to write native applications for Windows 8 and later that runs on ARM tablets/phones, and has nothing to do with Android/iOS support at all.

samsinsane:
Hmm, maybe we should support the generic ARM and ARM64 (eventually). An end user that's trying to target what you've linked above mightn't know what the actual architecture is. It could also be a multi-architecture binary, similar to UWP (I think?).

Sorry for flip-flopping on this, feel free to merge. I think what you've got is 100% correct, even if we add all the ARM versions that Android and whatever else supports.

PR #761, opened 2017-04-27
Removed echo off in VS rules prop generator

### LINKED ISSUE #698: Output disabled for custom rule command line templates
There’s an extra “@echo off” in custom rule command line templates (.props files), which prevents build output to be displayed. Now, I understand some commands can be quite noisy, but having a total silence is not helpful when debugging builds. Perhaps some option could be added to switch the echo off, but I think by default it should be on. Seems to have been introduced in 8b7b4c1139c7051c3a09123a03ca4f116f20fee

samsinsane:
Resolves #698


tvandijck:
tests fail though...

samsinsane:
All fixed, my bad.

PR #763, opened 2017-04-27
[core] Print warnings in purple

tvandijck:
Jeej, fancy colors on Windows... anyone know how to do this on linux/macosx/posix??

### DIFF
--- a/src/base/_foundation.lua
+++ b/src/base/_foundation.lua
@@ -334,7 +334,10 @@
 		if _OPTIONS.fatal then
 			error(message)
 		else
+			local color = term.getTextColor();
+			term.setTextColor(13);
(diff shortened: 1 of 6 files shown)


src/base/_foundation.lua
```diff
@@ -334,7 +334,10 @@
 		if _OPTIONS.fatal then
 			error(message)
 		else
+			local color = term.getTextColor();
+			term.setTextColor(13);
```
samsinsane: Maybe instead of using the value explicitly, we could create a `term.warningColor` that can be set based on the OS?

tvandijck: sure... I'll update the PR... seems like a good idea.

PR #769, opened 2017-05-01
[core] new cmd path decorations with %[] syntax

tvandijck:
this is for marking cmd options as paths relative to the projects base dir. decorated paths can use tokens / macros decorated paths are wrapped with "".  Final slashes are honored. Slashes are platform specific.

You can use '%[path]' in any kind of buildcommands API, to decorate an option as a path that needs to be resolved...  Currently there is no reliable way to identify parts of a string as a path, which is what this adds, writing buildcommands is quite a bit easier because of that.

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -610,6 +610,47 @@
 
 
 
+---
+-- Translate decorated command paths into their OS equivalents.
+---
+
+	function os.translateCommandsAndPaths(cmds, basedir, location, map)
+		local translatedBaseDir = path.getrelative(location, basedir)
+
+		map = map or os.target()
+
+		local translateFunction = function(value)
+			local result = path.join(translatedBaseDir, value)
+			if value:endswith('/') or value:endswith('\\') or
+			   value:endswith('/"') or value:endswith('\\"') then
+				result = result .. '/'
+			end
+			if map == 'windows' then
+				result = path.translate(result)
+			end
+			return result
+		end
+
+		local processOne = function(cmd)
+			local replaceFunction = function(value)
+				value = value:sub(3, #value - 1)
+				return '"' .. translateFunction(value) .. '"'
+			end
+			return string.gsub(cmd, "%%%[[^%]\r\n]*%]", replaceFunction)
+		end
+
+		if type(cmds) == "table" then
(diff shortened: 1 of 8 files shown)


tests/base/test_os.lua
```diff
@@ -281,3 +281,18 @@
...
+-- os.translateCommandsAndPaths.
+--
+
+	function suite.translateCommandsAndPaths()
+		test.isequal('cmdtool "../foo/path1"', os.translateCommandsAndPaths("cmdtool %[path1]", '../foo', '.', 'osx'))
+	end
+
+	function suite.translateCommandsAndPaths_PreserveSlash()
+		test.isequal('cmdtool "../foo/path1/"', os.translateCommandsAndPaths("cmdtool %[path1/]", '../foo', '.', 'osx'))
+	end
```
samsinsane: These tests are great, but I've noticed a bit of a trend of our Lua patterns not working with the same pattern twice. For example, what happens if you were to have `cmdtool %[path1/] %[path2/]`?

src/base/os.lua
```diff
@@ -610,6 +610,47 @@
 
 
 
+---
+-- Translate decorated command paths into their OS equivalents.
+---
+
+	function os.translateCommandsAndPaths(cmds, basedir, location, map)
+		local translatedBaseDir = path.getrelative(location, basedir)
```
samsinsane: I tend to agree with the code here, just wanted to know your thoughts on having the caller perform this? Would there be any benefit in this function being "flexible"?

tests/base/test_os.lua
tvandijck: We can certainly add a test for that... the output should be ```cmdtool "../foo/path1/" "../foo/path2/" ```

We use this in several places like that... but adding a test for it is probably a good idea.

src/base/os.lua
tvandijck: I don't think it matters much, whatever we prefer... it just moves this single line into 5-6 locations instead...

tests/base/test_os.lua
samsinsane: Awesome, I can never really tell with the Lua patterns. 👍

Issue #766, opened 2017-04-28
CompileAs breaks mixed projects


Jusonex:
In our project, we have third party dependencies that consist of both C and C++ source files (extension `.c` and `.cc`).

Unfortunately, 5b688cac93a2b1ab50ab851e7ccfeb7b5ac9b970 changes the default behavior (based on the extension) to a fixed behavior only based on the `language` property.

In our case VS' default behavior is obviously the correct one. As an alternative to fully reverting 5b688cac93a2b1ab50ab851e7ccfeb7b5ac9b970, I'd suggest a `compileAs` option for premake which accepts the same values as the `language` property.

tvandijck:
This is affecting us too, so I made the following PR:
https://github.com/premake/premake-core/pull/771

Jusonex:
The PR fixed it 👍

Issue #770, opened 2017-05-01
language 'C++' overrides language 'C++11'


tvandijck:
previously I could do this:
```
workspace 'X'
    flags { 'C++11' }

    project 'Y'
        language 'C++'
```

and all my code would be compiled with C++11 standards enabled..

```
workspace 'X'
    language 'C++11'

    project 'Y'
        language 'C++'
```

doesn't work that way... project Y now uses whatever is the default, which is OK, since that is what is specified... but for SC2/Heroes, I now have to basically fix up 98 project lua scripts. I know I did the work, and we had long discussions about this particular refactor, but it's a rather inconvenient API the way we made this work..

samsinsane:
I can appreciate the pain you're going through, I'll be going through this soon enough. I don't think ignoring `C++` will work though, it's a good way to not have anything emit, this is good if you need to emit a standard that Premake doesn't have. An end user shouldn't need to know how to add a new standard into Premake for all of the actions, just how to use `filter` and `buildoptions`.

I do think we need a better solution though, because the above is kind of horrifying.

tvandijck:
> it's a good way to not have anything emit,

Just for what it is worth, premake sets the default language at startup to C++.
I tried removing that line, and then any project that doesn't set it explicitly errors during validation. Which is kind of interesting, as it lead some of the projects to suddenly use CompileAs set to C++, whereas previously they didn't, and hence they started failing...

pdoane:
I would like to find a solution to the problem that keeps something like the current syntax.  I maintain several projects that support multiple C++ language revisions and this change simplifies usage a lot.  Instead of this:

```
    if _OPTIONS["language"] == "C++11" then
        flags { "C++11" }
    elseif _OPTIONS["language"] == "C++14" then
        flags { "C++14" }
    elseif _OPTIONS["language"] == "C++17" then
        flags { "C++17" }   -- future proofing
    end
```

I can now do:

```
    language ( _OPTIONS["language"] )
```

Also, if I have a mixed solution where some projects need different dialects, the script changes from:

```
        removeflags { "C++11" }
        removeflags { "C++17" }
        flags { "C++14" } 
```

to

```
        language "C++14"
```

So I think there needs to be some way to control the '-std=xxx' option with a single API call which this is handling great... the problem is just in the backwards compatibility.

Ideally for me `language 'C++'` would recognize that `language 'C++11'` is more specific and ignore it.

samsinsane:
[...]

PR #771, opened 2017-05-01
[core] Use 'compileAs' API

tvandijck:
Fix for: https://github.com/premake/premake-core/issues/766

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -184,6 +184,17 @@
 	}
 
 	api.register {
+		name = "compileas",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"C",
+			"C++",
+		}
+	}
+
+	api.register {
 		name = "configmap",
 		scope = "project",
 		kind = "list:keyed:array:string",
--- a/src/actions/vstudio/vs2010_vcxproj.lua
+++ b/src/actions/vstudio/vs2010_vcxproj.lua
@@ -1216,9 +1216,9 @@
 
 
 	function m.compileAs(cfg)
-		if p.languages.isc(cfg.language) then
+		if p.languages.isc(cfg.compileas) then
 			m.element("CompileAs", nil, "CompileAsC")
-		elseif p.languages.iscpp(cfg.language) then
+		elseif p.languages.iscpp(cfg.compileas) then
 			m.element("CompileAs", nil, "CompileAsCpp")
 		end
 	end
--- a/tests/actions/vstudio/vc2010/test_compile_settings.lua
+++ b/tests/actions/vstudio/vc2010/test_compile_settings.lua
@@ -38,7 +38,6 @@
 	<PrecompiledHeader>NotUsing</PrecompiledHeader>
 	<WarningLevel>Level3</WarningLevel>
 	<Optimization>Disabled</Optimization>
-	<CompileAs>CompileAsCpp</CompileAs>
 </ClCompile>
 		]]
 	end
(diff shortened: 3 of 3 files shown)


tvandijck:
What else is there to determine project type? The only alternative we have is ```kind```.

But I see what you are kind of hinting at?
"what if we just didn't set language at all, and only set it to be explicit when needed..."

I think that is a good idea, but indeed we'd need something else to determine the project type, and then I think we're falling in the same trap as what we tried to avoid with 'language' vs 'dialect'... say for example we did:
```
project 'foo'
    projecttype 'NativeProject'
    language 'C#'
```
it wouldn't work, because for C# you need a csproj, not a vcxproj...
So just like 
```
project 'foo'
    language 'C++'
    dialect 'C++14'
```

the options are related, and need to be validated for correct combinations..
Basically, I'm coming around on the language determines both "language" and "dialect" like we have implemented right now, and it turns out the side effects of the change are quite substantial, and I don't really know how to correctly fix them without introducing API's like this.

samsinsane:
I hear what you're saying, the inability to link APIs together does create a bit of an issue when things like this occur. But at the same time, there's really not much difference between the above and:
```
project 'foo'
  language 'C#'
  compileas 'C++'
```

Regardless of which direction we go, we're creating an API that is only valid based on another API. We've kind of already got these anyway, `frameworkdirs` isn't really valid outside of macOS, similarly, the `dotnetframework` isn't really valid outside of C# (and perhaps C++/CLI? I really don't know). Well, of the things we support, those two APIs are only valid in the cases I mentioned.

I think the tight coupling we've created between "project type" and "language"/"dialect" is a problem, but perhaps `compileas` is the shortest path forward. I was pondering on the idea that maybe we shift the `<CompileAs>` element to only emit at the file level when `prj.language ~= cfg.language`. This could still allow for the original behaviour (C only) to occur at the project level though, but that creates some strange logic and I'm sure someone will eventually make the same change I did and break everything.

I'm happy for this to go through, I did just notice you've called it `compileAs` though, and it should probably be `compileas` for consistency and whatnot.

tvandijck:
> I think the tight coupling we've created between "project type" and "language"/"dialect" is a problem, but perhaps compileas is the shortest path forward. 

It think the compileas is more of a bandaid to the bigger issue. It solves one particular issue, but not the issue I mentioned here: https://github.com/premake/premake-core/issues/770

We **can** merge this in the intermediate time, but I rather think about solving this the right way for the final 5.0 release that @starkos is working towards.

I rather refactor to a ```language & dialect``` API, basically reverting the language refactor I recently did going back to a simpler approach, and simply add validation of dialect/language pairs..

tvandijck:
made it lowercase as requested.

tvandijck:
> I rather refactor to a language & dialect API...

so, that wouldn't solve a thing either.... for example this would blow up if we added validation:

```
workspace ‘foo’
  language ‘C++’
  dialect ‘C++11’

project ‘bar’
  language ‘C’
```

which is not an unthinkable situation (we have tons of those cases already with the "flags {C++11}".
right or wrong, the flags only applied to the C++ language features...

starkos:
The `complieas()` call was added because…well, because I didn't think things through all the way, clearly. There is no reason for that to exist, you should be able to use `language()` in its place.

tvandijck:
@starkos not sure I understand... using language for this in its place is not working, it emits this <CompileAs> property at all times, because there is no way to not set language.
```language(nil)``` doens't work, since then the project no longer validates, and premake errors out. You are required to set language, which makes this particular vcxproj feature **always** emit something, which is undesirable, we want the ability to not emit anything. Hence this PR, but again, it feels like a bandaid, specifically for the VS* actions.

PR #774, opened 2017-05-02
Assure a default toolset is always set.

tvandijck:
We're running into the issue where:
```
filter { 'toolset:msc-v140' }
```
doesn't work, because the majority of time, the toolset is not actually set. This change assures the filter always has a valid value in it, and not just 'nil' in the majority of cases.

We also normalize the toolset, so when you do
```
toolset 'v140'
```
it normalizes to ```msc-v140```, as well as the other rules implemented in the ```p.tool.canonical``` function.
We should realy deprecate the v140 naming however, and require explicit msc-v140 usage, so I added a p.warnOnce in there too.

### DIFF
--- a/src/base/tools.lua
+++ b/src/base/tools.lua
@@ -29,25 +29,33 @@
 --    returns nil.
 ---
 
-	function p.tools.canonical(identifier)
-		local parts
+	function p.tools.normalize(identifier)
 		if identifier:startswith("v") then -- TODO: this should be deprecated?
-			parts = { "msc", identifier }
-		else
-			parts = identifier:explode("-", true, 1)
-
-			-- couple of little hacks here to fix up version names
-			if parts[2] ~= nil then
-				-- 'msc-100' is accepted, but the code expects 'v100'
-				if parts[1] == "msc" and tonumber(parts[2]:sub(1,3)) ~= nil then
-					parts[2] = "v" .. parts[2]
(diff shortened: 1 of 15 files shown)


src/base/tools.lua
```diff
@@ -29,25 +29,34 @@
...
-				if parts[1] == "msc" and tonumber(parts[2]:sub(1,3)) ~= nil then
-					parts[2] = "v" .. parts[2]
-				end
-
-				-- perform case-correction of the LLVM toolset
-				if parts[2]:startswith("llvm-vs") then
-					parts[2] = "LLVM-" .. parts[2]:sub(6)
-				end
-			end
+			p.warnOnce(identifier, "toolset '%s' is deprecated, toolset 'msc-%s' instead.", identifier, identifier)
```
starkos: That was the original intention behind leaving the "v140" form in place: that matches what VS uses, and what you see in the IDE, so it is a more "expected" format if you're familiar with the VS projects. We did go back and forth on it though.

tvandijck: I just put the deprecation message in there, because there is a comment that says "we should deprecate this"...

tvandijck:
Removed the deprecation warning.

tvandijck:
Removed the deprecation warning a while ago... need more feedback however on what direction we want to go for the normalization?

PR #775, opened 2017-05-02
Languages fixes

tvandijck:
So this is just one possible solution to the language issues explained here:
https://github.com/premake/premake-core/issues/770

Basically it's a set of changes that allows fields to decide how they get stored or merged when accessed. The tests test just a few use-cases that I could think of, more might be needed to fully cover everything, help appreciated.

### DIFF
--- a/modules/d/_preload.lua
+++ b/modules/d/_preload.lua
@@ -20,6 +20,8 @@
 
 	p.D = "D"
 
+
+	p.languages.types[p.D] = { p.D }
 	table.insert(p.languages.all, p.D)
 	api.addAllowed("language", p.D)
 
@@ -42,7 +44,6 @@
 		"UnitTest",
 		"Verbose",
 	})
-
 
 --
 -- Register some D specific properties
--- a/modules/d/d.lua
+++ b/modules/d/d.lua
@@ -19,8 +19,12 @@
 --
 -- Patch the project table to provide knowledge of D projects
 --
+	function p.languages.isd(value)
+		return value == p.D
+	end
+
 	function p.project.isd(prj)
-		return prj.language == p.D
+		return p.languages.isd(prj.language)
 	end
 
 --
@@ -29,7 +33,6 @@
 	function path.isdfile(fname)
 		return path.hasextension(fname, { ".d", ".di" })
 	end
-
 
 --
 -- Patch actions
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -697,8 +697,22 @@
 	api.register {
 		name = "language",
 		scope = "config",
-		kind = "string",
-		allowed = p.languages.all
+		kind = "processed:string",
+		allowed = p.languages.all,
+		process = function(current, value)
+			-- if we try setting a completely different language, allow it.
+			if p.languages.gettype(current) ~= p.languages.gettype(value) then
+				return value
+			end
+
+			-- if we try to set a specific version of the language, allow it.
+			if p.languages.gettype(current) ~= value then
+				return value
+			end
+
+			-- otherwise, stick with the old value.
+			return current
+		end
 	}
 
 	api.register {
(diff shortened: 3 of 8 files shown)


tvandijck:
FYI, this goes the route of ```C++``` does not override ```C++11``` when set, but ```C++11``` would override ```C++14```. Basically, if you are not specific, then it won't set it if something more specific is set. If you set a different language entirely, it will set it regardless of whether you are specific or not.

tvandijck:
> but as I said in #770, I think there should always be a way to tell Premake to not emit anything.

@samsinsane which is currently not possible at all... there is no "removelanguage" API, and premake sets language to "C++" at startup... so unless you set it to something else, it will be "C++".

We can remove the default, for sure... but then this will fail:
https://github.com/premake/premake-core/blob/master/src/base/validation.lua#L206

which we could remove... but then this will no longer work:
https://github.com/premake/premake-core/blob/master/src/actions/vstudio/vs2010.lua#L52

which we could fix, but not using language as the project type, as you suggested, but then you basically just end up with: ```language``` & ```dialect``` api's, maybe with different names ```projecttype``` & ```language```, same difference, with same problems...

So... as it stands, there is currently no way to have premake NOT emit 'language', this PR does not change that... The "compileas" PR fixes the "CompileAs" emit issue separately, which would still be needed with this PR.

If we wanted to remove language default, then we need to figure out a different way to emit language and project types, which is a much bigger change. But... I'm open to suggestions obviously..

samsinsane:
> which is currently not possible at all... there is no "removelanguage" API, and premake sets language to "C++" at startup...

This PR is designed to stop C++ from overriding C++11. What standard was it emitting to the command line? I was under the assumption that it was not emitting any standard? Which is exactly what I think should happen, or at least be an option.

tvandijck:
@samsinsane correct... it assumes that if a higher power sets a standard, then that higher power wins... just like with the flag...
```lua
workspace 'foo'
    language 'C++11'
    project 'x'
        language 'C++'
```
project 'x' gets C++11, because the workspace says so... just like
```lua
workspace 'foo'
    flags { 'C++11'}
    project 'x'
        language 'C++'
```
would have...

Which isn't to say we can't do anything else... but we have a ton of projects that relied on the behavior of the flag that way... The only way to basically not get this behavior without this PR, is to not set language anywhere, and assume someone else higher up (workspace?) sets it, and you only overload it if you want something different....

samsinsane:
Correct, but how would you now do `removeflags { "C++11" }` at the project level?

tvandijck:
That is a valid question indeed... 
You could write:
```
language 'C99'
language 'C++'
```
switch it to a different language, and then back, which would override it... but that is probably the dirtiest workaround I've ever posted here ;)

we could do something like:
```
language 'force:C++'
```

I'm not sure... it seems we're a little bit in a bind though... we want this behavior:
```
workspace 'foo'
    flags { 'C++11'}
    project 'x'
        language 'C++'
```

but we also want:
```
workspace 'foo'
    flags { 'C++11'}
    project 'x'
        language 'C++'
        removeflags { 'C++11' }
```

with this PR, the first is working:
```
workspace 'foo'
    language 'C++11'
    project 'x'
        language 'C++'
```

but the second is no longer available without a dirty.
without this PR, the first isn't working, but the second case is.

samsinsane:
Yeah, that's my only concern really, that we don't support both. I am not against your changes, I'm just against trading one for the other.

Perhaps, the force prefix could be used for your use case? You're forcing the dialect/standard to hold, though, I'm forcing the dialect/standard to drop. Though, your use case would require force to be stored, and then later removed in the oven. Hmm, either way I'm happy with the idea of the prefix, there could be a better solution, I can't think of anything right now though.

Anyone got ideas on how to achieve both cases relatively cleanly?

tvandijck:
closing this, I don't think this is the answer, although I think the "processed" field kind is kinda cool...

PR #785, opened 2017-05-14
Unicode file functions for Windows

DrLynix:
Hello,

As we discussed in my last PR (#728) I'm now proposing unicode override for Lua base functions (io.open, os.remove and os.rename).

I tried to respect the Lua API but this made me duplicate some code unfortunatly.


DrLynix:
I have no idea why AppVeyor is failing

tvandijck:
I gave it a kick to try again... but 0xc0000005 is typically a memory access violation...

tvandijck:
Tried it two times, but it's consistently failing with the same error... I'm assuming some kind of null pointer exception... I unfortunately don't have the time today to try and repro this locally...

DrLynix:
I have troubles to run Premake (output exec) on Windows, I'll try to fix this.

DrLynix:
Finally managed to build and run premake on my computer and it works like a charm.. :/

DrLynix:
I found the problem, it's because of that [weird environnement](https://github.com/premake/premake-core/blob/master/contrib/lua/src/liolib.c#L552) which defines a close function for file handles (causing the null pointer crash when an unicode file tries to close as it has no close function).

I tried many things, but there are a lot of problem involved with the override post-init of io functions (which are fixed in Lua 5.2! Why is Premake still in Lua 5.1?).

I have no choice right now than to override liolib.c in Lua library... Hopefully it's a small change.

tvandijck:
We've talked about upgrading Lua a few times, not sure what the general opinion is by now. At Blizzard we're on 5.3.4, but I think there is a few local changes that premake made to Lua that we didn't integrate. I don't think we're noticing it however... so not sure what the deal is...

That aside, these changes look fine to me. I think it would be awesome if lua itself got these changes committed however. Would help lua itself too, and 5.3 actually has utf8 functionality.

PR #787, opened 2017-05-18
os.comparefiles function

DrLynix:
Hi again,

Here's a function I had to write two times at my job in Lua: os.comparefiles, returns true if two files are identical by content.

My main use of this is to synchronize big files from one directory (build output) to the other (thirdparty repository).

Of course, it is unicode-aware on Windows.

### DIFF
--- a/src/host/os_comparefiles.c
+++ b/src/host/os_comparefiles.c
@@ -46,6 +46,25 @@ int os_comparefiles(lua_State* L)
 	secondFile = fopen(secondPath, "rb");
 	#endif
 
+	if (!firstFile)
+	{
+		if (secondFile)
+			fclose(secondFile);
+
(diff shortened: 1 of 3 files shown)


src/host/os_comparefiles.c
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	firstFile = _wfopen(wide_firstPath, L"rb");
+	secondFile = _wfopen(wide_secondPath, L"rb");
+	#else
+	firstFile = fopen(firstPath, "rb");
+	secondFile = fopen(secondPath, "rb");
+	#endif
+
+
```
DrLynix: Damn, I forgot to check if the file have been opened, hold my beer.

PR #798, opened 2017-06-01
create object directory before generating precompiled header

noresources:
Fix precompiled header generation in gmake action with gcc/clang toolsets
The object directory where the gch file is generated was not created before compiler invocation


noresources:
I added a new commit to the pull request to update the test suite but the appveyor tool failed weirdly.

tvandijck:
kicked of a new appveyor build.

PR #799, opened 2017-06-02
sort file lists

bmwiedemann:
because filesystems return files in random order
and linkers order functions depending on order of .o files

This helps enable reproducible builds in software built with premake
like [0ad](https://play0ad.com/)

See https://reproducible-builds.org/ for why that matters.

### DIFF
--- a/src/host/os_match.c
+++ b/src/host/os_match.c
@@ -104,8 +104,10 @@ int os_matchnext(lua_State* L)
 
 typedef struct struct_MatchInfo
 {
-	DIR* handle;
+	struct dirent** namelist;
 	struct dirent* entry;
+	int n;
+	int i;
 	char* path;
 	char* mask;
 } MatchInfo;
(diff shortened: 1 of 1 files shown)


starkos:
Are you sure about this? I thought Premake already sorted all lists under the hood.

tvandijck:
@starkos it does: https://github.com/premake/premake-core/blob/master/src/base/oven.lua#L671

starkos:
Okay, closing this as unnecessary then. Feel free to reopen if there is something I've not considered!

bmwiedemann:
OK. then maybe 0ad is bundling an older version of premake.

tvandijck:
https://trac.wildfiregames.com/wiki/premake

Note: Currently we bundle a modified version of the Premake 4.3 release, located in build/premake/premake4. The modifications fix a number of bugs and missing features in the 4.3 release.

tvandijck:
Also looks like there is an effort to upgrade to Premake5 here: https://trac.wildfiregames.com/ticket/3729

PR #800, opened 2017-06-03
Fix #794 (PCH creation fails)

### LINKED ISSUE #794: PCH creation fails
After upgrading our premake to HEAD (the previous version was pretty old), gcc fails to create the PCH.

It exactly says:
```../Server/core/StdInc.h:1:0: fatal error: can’t create precompiled header obj/x64/Release/Core/StdInc.h.gch: No such file or directory```.
It turned out that it works fine if `obj/x64/Release/Core` has been created manually before. Also, the compile log is missing a 'Creating ...' line compared to the old one.

That means:
Old premake:
```
==== Building Core (release_x64) ====
Creating ../Bin/server/x64
Creating obj/x64/Release/Core
StdInc.h
```
HEAD premake:
```
==== Building Core (release_x64) ====
StdInc.h
../Server/core/StdInc.h:1:0: fatal error: can’t create precompiled header obj/x64/Release/Core/StdInc.h.gch: No such file or directory
```

Code to reproduce: https://do-not.press/premake-pch.tar.gz

Jusonex:


Jusonex:
Ooops, just noticed my PR is equivalent to #798

PR #801, opened 2017-06-07
Groups in XCode workspace working.

ricka-github:
I noticed the 'group' directive doesn't do anything for XCode, even though XCode workspaces do support groups.

A nice and simple change to get my feet wet in the codebase.

### DIFF
--- a/modules/self-test/test_helpers.lua
+++ b/modules/self-test/test_helpers.lua
@@ -44,7 +44,10 @@
 		return prj
 	end
 
-
+	function m.createGroup(wks)		
+		local prj = group ("MyGroup" .. (#wks.groups + 1))
+		return prj
+	end
 
 
 	function m.getWorkspace(wks)
--- a/modules/xcode/tests/test_xcode4_workspace.lua
+++ b/modules/xcode/tests/test_xcode4_workspace.lua
@@ -75,6 +75,28 @@
 		]]
 	end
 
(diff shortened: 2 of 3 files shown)


starkos:
Looks reasonable to me. Can you add a unit test for it?

ricka-github:
I had to extend the test helper to be able to create a group. It seems to work, but I'm not sure this is the way you want it to be.

modules/self-test/test_helpers.lua
```diff
@@ -44,7 +44,10 @@
 		return prj
 	end
 
-
+	function m.createGroup(wks)		
+		local prj = group ("MyGroup" .. (#wks.groups + 1))
```
tvandijck: just nitpicking I guess... with typically we use 'prj' for projects...
it's a local, so not a big deal, but I would probably rename that local to grp or something.

tvandijck:
Looks good to me... can be merged as is, although I did mark it as 'request changes', it's such a minor thing, that we don't have to hold up the merge for that...

modules/self-test/test_helpers.lua
starkos: Technically, we don't need this function at all…you could just use `group()` directly and have full control over the name. But it doesn't hurt to have it for completeness I suppose?

Either way, the change under test is certainly a good one.

PR #803, opened 2017-06-13
Gmake2 module

tvandijck:
This is the previous gmake-refactor PR "https://github.com/premake/premake-core/pull/516" as a gmake2 module instead.

I'm closing: https://github.com/premake/premake-core/pull/516 in favor of this PR.


samsinsane:
There's quite a bit to read through, but I think we should just merge this and get feedback on it. Perhaps, if not already, clearly mark it as "experimental".

tvandijck:
> Perhaps, if not already, clearly mark it as "experimental".

Well, it's way past the 'experimental' stage at this point... I got about 4 games using this in production ;)
But I'm sure there is bugs and I'm open to any feedback.

And as we discussed prior in the gmake-refactor PR here (#516) there is a lot more ideas floating around to making adding languages easier and more flexible.

PR #805, opened 2017-06-13
Add --insecure option for SSL curl requests.

tvandijck:
We run into a lot of ssl errors when communicating with the outside world due to proxy settings.
This allows us to run some of the nuget tests without having to locally modify the code.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1571,6 +1571,13 @@
 		description = "Display version information"
 	}
 
+	if http ~= nil then
+		newoption {
+			trigger = "insecure",
+			description = "forfit SSH certification checks."
+		}
+	end
+
 
 -----------------------------------------------------------------------------
 --
--- a/modules/self-test/test_runner.lua
+++ b/modules/self-test/test_runner.lua
@@ -120,7 +120,7 @@
 		_.builtin_write = mt.write
 		mt.write = _.stub_stderr_write
 
-		_OPTIONS = {}
+		_OPTIONS = table.shallowcopy(_OPTIONS) or {}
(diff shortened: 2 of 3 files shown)


src/_premake_init.lua
```diff
@@ -1571,6 +1571,13 @@
 		description = "Display version information"
 	}
 
+	if http ~= nil then
+		newoption {
+			trigger = "insecure",
+			description = "forfit SSH certification checks."
```
samsinsane: Minor typo, "forfeit".

tvandijck:
>  I feel like this is something you never want to actually use.

I agree, that in general you do never want to use it... However, there is cases where a company for all kinds of reasons I don't agree with, puts in a proxy server that rewrites all certificates, it's basically an IT endorsed man in the middle attack. For example to monitor employees leaking sensitive data, or for whatever other reason...  The two options I'm disabling here:
```
		curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0);
 		curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0);
```

All nuget unit-tests here fail when executed on my desktop without this option.

samsinsane:
My understanding of proxies for HTTPS is rather limited, but what about options such as [`CURLOPT_PROXY_SSLCERT`](https://curl.haxx.se/libcurl/c/CURLOPT_PROXY_SSLCERT.html)? I'm happy for this to go through, I'm just hoping that curl supports security, even if it's by allowing a specific MitM attack rather than all of them.

samsinsane:
@tvandijck That's fair enough, I noticed those options when I added the proxy URL and just figured that this might be what they're for? Ah well, if it's what you need to get through your proxy server, then I imagine others will need it too.

PR #809, opened 2017-06-14
Added os touchfile and a workaround for vs2010 reloading

tvandijck:
vs2010 (and above) has an issue where if the vcxproj file
isn't modified but the.vcxproj.filters is modified during
premake, then VS doesn't notify the user that the project has
changed.  This can occur if the only change to a project is
with the vpaths.  Before this fix, the user would be need to
manually unload and reload the project to pick up the change.

This VS workaround checks for the case where the vcxproj.filters
file is modified but the .vcxproj file is not. At which point
premake preforms a touch on the .vcxproj file to trigger the IDE.

### DIFF
--- a/src/host/os_touchfile.c
+++ b/src/host/os_touchfile.c
@@ -1,0 +1,94 @@
+/**
+ * \file   os_touchfile.c
+ * \brief markes a file as modified without changing its contents.
+ * \author Blizzard Entertainment (contact tvandijck@blizzard.com)
+ * \author Copyright (c) 2015 Jason Perkins and the Premake project
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "premake.h"
+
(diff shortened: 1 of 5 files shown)


src/host/os_touchfile.c
```diff
@@ -0,0 +1,94 @@
...
+	return TRUE;
+}
+
+int os_touchfile(lua_State* L)
+{
+	FILE* file;
+	const char* dst     = luaL_checkstring(L, 1);
+
+	// if destination exist, mark the file as modified
+	if (do_isfile(dst))
```
samsinsane: This takes a `lua_State*` as the first parameter with the path as the second, I'm guessing you modified this just before pushing?

tvandijck: Yeah, we have a local change I guess that doesn't have that argument...
I'll fix it in the PR.

PR #811, opened 2017-06-14
Add 'preferredtoolarchitecture' API

tvandijck:


samsinsane:
Not overly sure what this is for? Is this a VS2017 thing?

tvandijck:
actually this is a VS2010 and later thing ;)

It makes msbuild use the 64-bit version of cl.exe, link.exe and mspdbsrv.exe.
Which is particularly useful when you are using incredibuild to build. We had instances where the pdb service ran out of memory while linking.. 

it has pretty minimal documentation, but there is something here:
https://msdn.microsoft.com/en-us/library/ee662426.aspx
and here:
https://stackoverflow.com/questions/19820718/how-to-make-visual-studio-use-the-native-amd64-toolchain

samsinsane:
Haha, yeah I've seen that before, I wonder why I didn't think of that.

PR #815, opened 2017-06-14
Errors in red.

tvandijck:
I hard-coded the color to red here, otherwise we have to get it using all kinds of lua functions...
I just didn't think it was worth the effort...

this would work I think though:
```c
lua_getglobal(L, "term");
lua_pushstring(L, "errorColor");
lua_gettable(L, -2);
int color = luaL_checknumber(L, -1);
lua_pop(L, 2);
```

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -98,4 +98,4 @@ windows: windows-base
 	devenv .\build\bootstrap\Premake5.sln /Build Release
 
 windows-msbuild: windows-base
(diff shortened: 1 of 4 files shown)


tvandijck:
well, we exposed the "infoColor" and "warningColor" in lua, so people can make their own picks..  I don't know if anyone ever will, but it's there because someone asked for that... so "errorColor" is there in lua, but since the error handling in C, it's currently not using it... I guess we could query it, just didn't seem worth the effort to me.

PR #816, opened 2017-06-14
Fixed C++17 flag to work with current versions of Clang.

tvandijck:
17 isn't official yet, both gcc and clang still refer to it as '1z'


tvandijck:
I think premake, just like cmake should 'detect' what version is currently 'installed' and being used, and decide based on that... until such time flags like this will always be in flux...

We have similar problems with VS2015 Update 1 vs Update 3.... Update 3 has /std:c++14 support, Update 1 does not and gives a warning that it is an unrecognized option. Premake doesn't know, it just generates what we tell it to generate for "vs2015"... Sure we can make actions like vs2015.1, and vs2015.2, etc... but are we really then also going to make things like:

clang38, clang381, clang39, clang391, and for gcc, oh my.... if you go back to 2010 (4.4.3+) we're talking about 41 releases...

it's a huge matrix of x translates to y for any of these compilers, and premake isn't really that good at doing the right thing for any of them, except those that someone actively uses is targetting. So with blizzard as a really active user, fixes like this essentially just target the compilers we use, but we would really like to see a better more sustainable option.

PR #821, opened 2017-06-19
Attempt at colors on linux

tvandijck:


tvandijck:
test on ubuntu and macOS.

PR #824, opened 2017-06-20
Set default toolsets for vs2005 and vs2008.

tvandijck:


tvandijck:
hmmm, interesting side effect 👎

tvandijck:
Question @starkos & @samsinsane do we wish to maintain the vs2005 and vs2008 backends? I no longer have access to it.. it's going to be increasingly difficult to do the right thing for these.

samsinsane:
I, personally, don't care either way, though I don't think VS2008 is EoL yet, so perhaps the removal should wait until then?

tvandijck:
https://blogs.msdn.microsoft.com/visualstudio/2017/04/10/end-of-support-for-visual-studio-2008-in-one-year/

April 10, 2018 apparently... We can wait...

PR #827, opened 2017-06-20
Added option to create bundle and frameworks on macOS.

tvandijck:


src/tools/gcc.lua
```diff
@@ -264,6 +264,24 @@
 	end
 
 --
+-- get the right output flag.
+--
+	function gcc.getsharedlibarg(cfg)
+		if cfg.system == premake.MACOSX then
+			if cfg.sharedlibtype and cfg.sharedlibtype == "OSXBundle" then
```
samsinsane: The first check isn't required? (Same for elseif below)

tvandijck: uhm.. yeah I guess we don't really have to check if it's set or not to then compare it again... you are right.

PR #829, opened 2017-06-20
Reduced vs2010+ file sizes by removing redundant config settings

tvandijck:


tvandijck:
@samsinsane there is a number of tests in test_files.lua that still exhibit the "per config" behavior, so I feel this is still covered.

PR #840, opened 2017-06-25
Proof of concept for binary modules.

tvandijck:
This change allows you to create .dll (.so on linux/mac?) modules.

I made an example module that shows the setup you have to apply, and a test that calls the example.test function it registers.

I wrapped the entire public Lua API, so in theory any lua code that exists and builds against Lua 5.3.4 can be rebuild against the 'luashim' library, and premake would be able to use it as a shared library. The only requirement is that you call the 'shimInitialize(L);' method as the first thing in your luaopen_<name> function.


tvandijck:
I only tested this on Windows for the moment, so linux/mac are going to fail this build.

starkos:
Interesting idea. I don't have a particular use for this right now, but I would be curious to hear from people that do.

tvandijck:
Just some examples of options:
- This change allows us to integrate LuaRocks if we wanted to, as long as the rock has a .src or a .all spec, we can provide the facilities to build those rocks locally specifically for premake. 

- I currently have a commit in my history that adds a profiler using a lua hook. It's super hardcoded.. could easily be done using a `require("profiler")`

- Embedding lua source inside the module. It would allow for modules to just be distributed as .dll's.

Anyway, I think this is pretty neat, but I don't really have a real use for it either ;) ha...  I just had this idea at 2am saturday night, and had to just implement it, otherwise I'd just keep mulling on it.

ratzlaff:
First thing that comes to my mind is extending premake with commands that utilize third party libraries - like curl/ssl.

tritao:
For what it's worth I've wanted something like this before. Don't really have any use for it right now though, but still think it might be worth to consider adding it to core, especially as it doesn't seem to add too much complexity to the codebase.

tvandijck:
I think indeed like @ratzlaff says... we could basically pull out the zip and http libraries, and just make modules for them... It would get rid of a bunch of command line options to turn these features on/off

curl:
https://github.com/premake/premake-core/blob/master/premake5.lua#L67

zip:
https://github.com/premake/premake-core/blob/master/premake5.lua#L73

just an idea...

tvandijck:
these changes actually make it work on linux, except for that fact that I have to manually rename and move the file...

on windows I get "example.dll" in the same folder as the exe, and windows automatically looks there when loading a DLL... on linux this isn't that case... I get libexample.so, also next to the exe, but that is not included in the searchpath.. so I have to move the file into the root of the project, and rename it to example.so for it to work...

gotta figure out a way to make that more consistent.

tvandijck:
So now only macOS is failng.... linux + windows are good.

tvandijck:
and... it's green ;) whoop whoop...

tritao:
Hey @tvandijck, I'm seeing this on just Debug builds.

```
==== Building example (debug) ====
main.c
Linking example
Undefined symbols for architecture x86_64:
  "_findNode", referenced from:
      _shimInitialize in libluashim-lib.a(luashim.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I've tried cleaning and rebuilding but it persists, Release builds works though...

tvandijck:
I don't currently have access to a mac, but will look at this tomorrow. Although our builds here are all passing ;(

Any particular repro steps? Or just generally debug fails?

PR #858, opened 2017-08-01
Fix Nuget package handling to support Unix-style paths

tritao:
Fixes Nuget package handling to work on OSX.

cc @aleksijuvani

### DIFF
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -12,6 +12,7 @@ return {
 	"cs2005/test_nuget_config.lua",
 	"cs2005/test_nuget_framework_folders.lua",
 	"cs2005/test_nuget_packages_config.lua",
+	"cs2005/test_nuget_references.lua",
 	"cs2005/test_output_props.lua",
 	"cs2005/projectelement.lua",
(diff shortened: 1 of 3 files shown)


aleksijuvani:
LGTM, except that the current NuGet tests are in `test_assembly_refs.lua` so we should either move those tests to this new file or move this new test to that file.

aleksijuvani:
Also, just to note – this is matching the filenames returned by the NuGet API so the host system shouldn't matter.. the packager just used a Unix system so that's why the slashes are different.

tritao:
I thought so too, but actually the issue is some `path.translate` call that is done before this function is called, only realized this after the fix, what do you think? Should we hardcode a separator in the translate call instead?

aleksijuvani:
You're right, it looks like the API always returns paths with forward slashes, so it is a local issue after all. Hardcoding it to use the "\\" separator might be the best course of action. I think this is what Visual Studio expects anyway.

tvandijck:
I'm a little confused with the conclusion of the above discussion? are we good to merge? it looks fine to me.

aleksijuvani:
Should be good to merge, I just got a bit confused.

tritao:
👍

PR #860, opened 2017-08-02
fix for build failure on macOS in debug.

tvandijck:
Fixes issue reported here: https://github.com/premake/premake-core/pull/840


tvandijck:
I also updated the travis-ci script to make a debug build, so this doesn't happen again.

PR #861, opened 2017-08-04
Allow embedding of binary resources.

tvandijck:


scripts/embed.lua
```diff
@@ -67,20 +67,24 @@
...
-				verbosef("Compiling... " .. filename)
-				local output = path.replaceextension(filename, ".luac")
-				local res, err = os.compile(filename, output);
-				if res ~= nil then
-					data = loadScript(output)
-					os.remove(output)
+			if not path.hasextension(filename, ".lua") then
+				data = loadScript(filename)
+			else
+				if _OPTIONS["bytecode"] then
```
samsinsane: An `elseif` seems like it would make more sense here?

PR #863, opened 2017-08-04
fix for duplicate 'allowed'.

tvandijck:


src/base/api.lua
```diff
@@ -349,8 +349,10 @@
 			end
 		else
 			field.allowed = field.allowed or {}
-			table.insert(field.allowed, value)
-			field.allowed[value:lower()] = value
+			if field.allowed[value:lower()] == nil then
+				table.insert(field.allowed, value)
+				field.allowed[value:lower()] = value
```
samsinsane: Should `value:lower()` be stored in a local variable since it's calculated twice here?

tvandijck: yeah, we could... this isn't really a method that gets called very often, so I was lazy.

PR #865, opened 2017-08-07
Remove "test.print" calls.

tvandijck:


tvandijck:
Still have no idea what is going on with the two tests that sporadically fail... but we have output from one of the failures, and it looks like sometimes the returned tables from the fetch are entirely empty rather then containing the values we test for. I added extra asserts to test if the store succeeded, but other then that we have no additional information yet.

PR #872, opened 2017-08-10
Rework binmodule example to silence unit test console output

starkos:
Trying to return from the dead…small fix to remove some spurious console output from Premake's own unit tests. Reworks the binmodule example function to concatenate two strings and return the result, rather than write to the console. Adjusts the corresponding unit test to verify the result.


tvandijck:
welcome back ;) and yes, this was on my bucket list... thank you...

PR #875, opened 2017-08-17
Add lua error handling to curlProgressCallback

adamkay:


adamkay:
With the current implementation, if something went wrong in the lua progress callback, we wouldn't get any notification about it. The download will finish regardless of the callback failing or not, but an error message would be useful for debugging.

tvandijck:
I think this is OK, but I would indeed prefer a pcall to ```premake.error``` as sam suggests.

adamkay:
Pcalling `premake.error` is redundant and it wouldn't print any error message.

```lua
function premake.error(message, ...)
    error(string.format("** Error: " .. message, ...), 0)
end
```

It calls the lua standard `error` function, so a `lua_pcall` would return with `LUA_ERRRUN`, leaving the message printing to the C side.

Instead, forwarding errors to the site where `_premake_main` is called (`premake.c/premake_execute`) would basically have the same behaviour as calling `premake.error` from the script.

```c
/* retrieve the lua progress callback we saved before */
lua_rawgeti(L, LUA_REGISTRYINDEX, state->RefIndex);
lua_pushnumber(L, (lua_Number)dltotal);
lua_pushnumber(L, (lua_Number)dlnow);
int ret = lua_pcall(L, 2, LUA_MULTRET, 0);
if (ret != LUA_OK) {
    // leave error handling to premake_execute
    // unlike in the original commit, this is now terminates script execution
    lua_error(L);
}
```

samsinsane:
I don't really have the time to look at how this all works right now. Just write a test for this, if the error message appears during test execution, the implementation is wrong - it's that simple.

adamkay:
Displaying the error message is the intention. I don't know how exactly should I write a test for this?

tvandijck:
I see what you mean now indeed.... we could display it as a warning instead, since its not fatal... which would be accomplished through ```premake.warn```... other then that, I understand calling premake.error is indeed not solving anything...

I think what this comes down to for me personally, is the hardcoded 12, where a term.errorColor exists.
That said, this pull request: https://github.com/premake/premake-core/pull/876 that @starkos made kind of simplified it a bit, so maybe that PR and this one can be combined, and the ```printLastError``` can be used. (it's static now, but we can easily just make that a public function).

data-man:
General question: why not use the excellent [Lua-cURLv3](https://github.com/Lua-cURL/Lua-cURLv3) library?

tvandijck:
@data-man we could... but it's a LuaRock. Premake doesn't really support luaRocks out of the box, because "Lua" itself is embedded in Premake, and not as a dll. That said, the new binary module system I made would allow for any LuaRock to be compiled against Premake, and should work... nobody tried yet though.

data-man:
@tvandijck 
> but it's a LuaRock

Not necessarily. See [luaopen_lcurl](https://github.com/Lua-cURL/Lua-cURLv3/blob/d1b6b8627f3b4086be53912867d2c76c315b434f/src/lcurl.c#L226) and [luaopen_lcurl_safe](https://github.com/Lua-cURL/Lua-cURLv3/blob/d1b6b8627f3b4086be53912867d2c76c315b434f/src/lcurl.c#L229).

tvandijck:
@data-man yeah, that's what I said... it should be possible to compile it against premake, using the new binary module stuff... it will however require use to copy the code, modify the build settings/scripts/whatever.... So my point is... it won't just work out of the box. it will require some engineering to get it to work.

adamkay:
`premake.warn` wont work either as it also calls the builtin lua error function if the `--fatal` command line flag is used. I've updated the PR, I think it is more acceptable now. Let me know if any modifications needed.

tvandijck:
looks good to me...

tvandijck:
@starkos why closing it?

starkos:
I didn't close it, but I did remove the alpha12 release branch, which this appears to have been forked from? I've restored the branch for now.

starkos:
…but note: the release is done now, so if you merge this request, you should force it over to the master branch, and not leave it on the release branch.

tvandijck:
@starkos, ah I see, I guess github closed it for you then...
@adamkay can you maybe move this to merge against master? or just open a new pull request on master with these changes? I think the changes are good now, so we can merge this as soon as that is done.

adamkay:
For sure,

adamkay:
So yeah, did a rebase against master. Still, github tells me merging is blocked. Is there something I should do to fix it, or you can merge it? Sorry I'm new to this.

samsinsane:
> github tells me merging is blocked

You've done everything correctly, it's blocked due to my requested changes. @tvandijck seems to understand this much better than I do, so I'll dismiss my review and leave it for him to approve and merge. Thanks for the changes!

adamkay:
Thank you!

PR #876, opened 2017-08-17
Fix error reporting for missing core scripts

starkos:
If a core script (something listed in the root `_manifest.lua`), execution fails with the very unhelpful message:

```
PANIC: unprotected error in call to Lua API (bad argument #-1 (number expected, got nil))
Abort trap: 6
```

This PR fixes the return value from `premake_load_embedded_script()` (was `LUA_YIELD`, not `LUA_ERRFILE`). It also adds a nil check to the _premake_main error handling for the case where an error color has not yet been set.

### DIFF
--- a/src/host/lua_auxlib.c
+++ b/src/host/lua_auxlib.c
@@ -1,7 +1,7 @@
 /**
  * \file   lua_auxlib.c
  * \brief  Modifications and extensions to Lua's library functions.
- * \author Copyright (c) 2014-2015 Jason Perkins and the Premake project
+ * \author Copyright (c) 2014-2017 Jason Perkins and the Premake project
  */
 
 #include "premake.h"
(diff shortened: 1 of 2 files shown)


src/host/lua_auxlib.c
```diff
@@ -24,20 +24,21 @@ static int chunk_wrapper(lua_State* L);
...
+LUALIB_API int luaL_loadfilex (lua_State* L, const char* filename, const char* mode)
 {
 	const char* script_dir;
 	const char* test_name;
 
 	int bottom = lua_gettop(L);
 	int z = !OKAY;
 
-	/* If filename is starts with "$/" then we want to load the version that
+	/* If filename istarts with "$/" then we want to load the version that
```
tritao: typo?

tvandijck:
not holding up a merge for a 1 character typo... i'll fix it in place.

PR #878, opened 2017-08-27
Add Support for Headless Actions

SE2Dev:
+ Added eprintf() - a variant of printf() that writes to stderr instead of stdout.
+ Added support for "headless" actions which, when run, prevent Premake from printing various information (such as the "Building configurations..." notification, or the execution time summary). This can be used to allow Premake to function as a standalone Lua engine, as well as allows these "headless" actions to be used to pipe output to other programs.

An action can be made headless by setting the `headless` field of the action to true when calling `newaction`, if this field is not present or is not set to true, the action is not considered headless.

Example of a headless action:
```lua
newaction {
    trigger = "myaction",
    description = "Describe your action!",
    headless = true, -- Make this action headless
    onStart = function()
        print("START")
    end,
    onWorkspace = function(wks)
        printf("WORKSPACE:\t%s", wks.name)
    end,
    onProject = function(prj)
        printf("PROJECT:\t%s", prj.name)
    end,
    onRule = function(rule)
        print("RULE")
    end,
    execute = function()
        print("EXECUTE")
    end,
    onEnd = function()
        print("END")
    end
}
```

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -230,7 +230,7 @@
 		-- script has run to allow for project-specific options
 		ok, err = p.option.validate(_OPTIONS)
 		if not ok then
-			print("Error: " .. err)
+			eprint("Error: " .. err)
 			os.exit(1)
 		end
 
@@ -243,12 +243,12 @@
 
 			local action = p.action.current()
 			if not action then
-				print("Error: no such action '" .. _ACTION .. "'")
+				eprint("Error: no such action '" .. _ACTION .. "'")
 				os.exit(1)
 			end
 
 			if p.action.isConfigurable() and not os.isfile(_MAIN_SCRIPT) then
-				print(string.format("No Premake script (%s) found!", path.getname(_MAIN_SCRIPT)))
+				eprint(string.format("No Premake script (%s) found!", path.getname(_MAIN_SCRIPT)))
 				os.exit(1)
 			end
 		end
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -260,7 +260,7 @@
 ---
 
 	function m.preBake()
-		if p.action.isConfigurable() then
+		if p.action.isConfigurable() and not p.action.isHeadless() then
 			print("Building configurations...")
 		end
 	end
@@ -321,8 +321,10 @@
 ---
 
 	function m.preAction()
-		local action = p.action.current()
-		printf("Running action '%s'...", action.trigger)
+		if p.action.isHeadless() then
+			local action = p.action.current()
+			printf("Running action '%s'...", action.trigger)
+		end
 	end
 
 
@@ -341,7 +343,7 @@
 ---
 
 	function m.postAction()
-		if p.action.isConfigurable() then
+		if p.action.isConfigurable() and not p.action.isHeadless() then
 			local duration = math.floor((os.clock() - startTime) * 1000);
 			printf("Done (%dms).", duration)
 		end
--- a/src/base/action.lua
+++ b/src/base/action.lua
@@ -204,6 +204,21 @@
 	end
 
 
+---
+-- Determines if an action is headless or not.
+-- Headless actions prevent various built-in messages from printing.
+---
+
+	function action.isHeadless(self)
+		if not self then
+			self = action.current() or {}
+		end
+		if self.headless == true then
+			return true
+		end
+		return false
+	end
+
 
 ---
 -- Activates a particular action.
--- a/src/base/moduledownloader.lua
+++ b/src/base/moduledownloader.lua
@@ -60,7 +60,9 @@
 		end
 
 		-- Unzip the module, and delete the temporary zip file.
-		verbosef(' UNZIP   : %s', destination)
+		if not p.action.isHeadless() then
+			verbosef(' UNZIP   : %s', destination)
+		end
 		zip.extract(destination, location)
 		os.remove(destination)
 		return true;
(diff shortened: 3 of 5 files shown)


src/_premake_main.lua
```diff
@@ -230,7 +230,7 @@
 		-- script has run to allow for project-specific options
 		ok, err = p.option.validate(_OPTIONS)
 		if not ok then
-			print("Error: " .. err)
+			eprint("Error: " .. err)
```
tvandijck: can't we just call p.error(...) here... 
it prints the error, and exits, which seems to be what is intended here anyway?

src/_premake_main.lua
```diff
@@ -243,12 +243,12 @@
 
 			local action = p.action.current()
 			if not action then
-				print("Error: no such action '" .. _ACTION .. "'")
+				eprint("Error: no such action '" .. _ACTION .. "'")
```
tvandijck: same

src/_premake_main.lua
```diff
@@ -243,12 +243,12 @@
...
 				os.exit(1)
 			end
 
 			if p.action.isConfigurable() and not os.isfile(_MAIN_SCRIPT) then
-				print(string.format("No Premake script (%s) found!", path.getname(_MAIN_SCRIPT)))
+				eprint(string.format("No Premake script (%s) found!", path.getname(_MAIN_SCRIPT)))
```
tvandijck: same

src/base/action.lua
```diff
@@ -204,6 +204,21 @@
...
+---
+-- Determines if an action is headless or not.
+-- Headless actions prevent various built-in messages from printing.
+---
+
+	function action.isHeadless(self)
+		if not self then
+			self = action.current() or {}
+		end
+		if self.headless == true then
```
tvandijck: wouldn't ```return self.headless``` be the same? or if you really wanted to enforce a boolean return: ```return self.headless == true```

src/base/moduledownloader.lua
```diff
@@ -60,7 +60,9 @@
 		end
 
 		-- Unzip the module, and delete the temporary zip file.
-		verbosef(' UNZIP   : %s', destination)
+		if not p.action.isHeadless() then
```
tvandijck: verbosef only prints when --verbose is specified... I think this isn't really needed, or the ```if not headless``` could be moved to here (https://github.com/premake/premake-core/blob/master/src/base/_foundation.lua#L402) maybe?

src/_premake_main.lua
SE2Dev: Yes, I only added eprint() because I wasn't entirely sure if there was some reason I wasn't aware of that you were using print() instead of p.error().

src/base/action.lua
SE2Dev: Correct, I suppose `if self.headless then ...` would also work if one wanted to allow values like `1` to be used to enable headless mode as well; I've only been messing with Lua for a few days, so I'm not extremely familiar with any language specific tricks yet.

src/base/moduledownloader.lua
SE2Dev: It *could* be added there, but if the user uses `verbosef()` to print output in their headless action, it wouldn't function correctly.

The reason I added a check for that specific verbosef() call was primarily a safety measure, as I intended to use a headless action to pipe project metadata to a VSCode extension to provide a Visual Studio like configuration / platform switcher, and wanted to ensure that the premake output would be valid json, without extra information being written to stdout.

starkos:
FWIW, I'm not a huge fan of this approach. I *am* good with rethinking what we output by default, I just don't think you should have to define a special kind of action to do it. Perhaps we move some of the default output over to verbose, and add a `--quiet` flag to suppress `stdout`?

SE2Dev:
@starkos Yes, that would also be potentially viable solution, provided users can still write to stdout / stderr through user defined actions.

samsinsane:
I tend to agree with @starkos here, this isn't a great solution. It's prone to failure as it's assuming that all future usages of `print`/`printf` will be wrapped in a check for `headless`. I'll admit that I'm probably going to forget about this thing in less than a month, and will likely approve and merge PRs that will break your expected functionality.

I highly recommend that you just emit a file to disk and pipe that into your VSCode extension. Alternatively, have your action *override* all of the output functions and have a copy of the originals for your action to call.

SE2Dev:
@samsinsane I did have an alternative version of newaction() that overrides various print functions until onStart() is called on the action and then disabled the print functions again after onEnd() completes, but I wasn't terribly fond of this method as if any new print functions were added, (e.g. eprint()) the hook would need to be manually updated.

```lua
-- Locals to keep track of the original functions
local _print = print
local _printf = printf
local _io_write = io.write

local _newaction = newaction

-- Enables support for headless actions
function newaction(a)
    if a.headless ~= true then
        return _newaction(a)
    end

    function io_initHooks()
        local _dummyFunc = function(...)
        end

        print = _dummyFunc
        printf = _dummyFunc
        io.write = _dummyFunc
    end

    function io_freeHooks()
        print = _print
        printf = _printf
        io.write = _io_write
    end

    -- Prevent any info from printing before the action starts
    if (a.trigger == _ACTION) then
        io_initHooks()
    end

    -- Re-enable printing before the action starts
    local _onStart = a.onStart
    a.onStart = function()
        io_freeHooks()
        if (_onStart ~= nil) then
            _onStart()
        end
    end

    -- Disable any summary info from printing
    local _onEnd = a.onEnd
    a.onEnd = function()
        if (_onEnd ~= nil) then
            _onEnd()
        end
        io_initHooks()
    end

    return _newaction(a)
end
```

samsinsane:
I'm going to close this off, please feel free to reopen.

PR #879, opened 2017-08-28
Fsharp support

tvandijck:
The Overwatch team has been working on adding FSharp Support to premake. This set of changes covers most of our usecases, but other changes may be required to get full coverage. Thanks to Mike Aversa for these changes.


tritao:
Interesting, we also started adding more F# code to one of our Premake projects.

For now we just reference the `.fsproj` in the solution file directly, but this could come handy :)

tvandijck:
The one gotcha in F# is that files or order dependent, so using:
```
files { '*.fs' } 
```
is most of the time not working, so you for F# projects you have to explicitly say:
```
files { 'a.fs', 'b.fs', etc }
```
other then that, this change works like a charm for us.

starkos:
Nice! I've only dabbled in F# myself so I'm not likely to be much help here.

realvictorprm:
Super awesome. If you need some help, I'm sure the  F# community and or me would happily help!

modules/vstudio/tests/fs2005/projectsettings.lua
```diff
@@ -0,0 +1,251 @@
...
+		<FileAlignment>512</FileAlignment>
+		<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+	</PropertyGroup>
+		]]
+	end
+
+
+	function suite.onVs2015_462()
+		p.action.set("vs2015")
+		framework "4.6.2"
```
samsinsane: `framework` is deprecated, I don't think we should explicitly test that the deprecation works for each API and just let the API test suite ensure that deprecation works.

modules/vstudio/vs2005_fsproj.lua
```diff
@@ -0,0 +1,86 @@
+--
+-- vs2005_fsproj.lua
+-- Generate a Visual Studio 2005-2010 F# project.
+-- Copyright (c) 2009-2014 Jason Perkins and the Premake project
```
samsinsane: First line should probably be the full path, but the other files above don't seem to do this like the test files do. Second line should probably not have a version range? 2005+ perhaps? Third line should be 2017-2017.

modules/vstudio/vs2005_fsproj.lua
```diff
@@ -0,0 +1,86 @@
...
+        dotnetbase.generate(prj)
+    end
+
+    function fs2005.tailCalls(cfg)
+        local tc = 'false'
+        if cfg.buildcfg == "Debug" then
+            tc = 'true'
+        end
+
+        _p(2, '<Tailcalls>%s</Tailcalls>', tc)
```
samsinsane: Perhaps this could just be `_p(2, '<Tailcalls>%s</Tailcalls>', iif(cfg.buildcfg == "Debug", 'true', 'false')`, there's also a function to determine if the build is a debug build, perhaps it should use that instead of `cfg.buildcfg`?

modules/vstudio/vs2005_fsproj_user.lua
```diff
@@ -0,0 +1,91 @@
+--
+-- vs2005_csproj_user.lua
+-- Generate a Visual Studio 2005/2008 C# .user file.
+-- Copyright (c) 2009-2015 Jason Perkins and the Premake project
```
samsinsane: First line should be `vs2005_fsproj_user.lua`, and probably full path. Second line should be `2005+ F# .user file`. Third line should be `2017-2017`.

src/_premake_init.lua
```diff
@@ -1650,7 +1651,7 @@
 
 	-- Windows and friends.
 
-	filter { "system:Windows or language:C#", "kind:ConsoleApp or WindowedApp" }
+	filter { "system:Windows or language:C# or language:F#", "kind:ConsoleApp or WindowedApp" }
```
samsinsane: The language part of this filter just doesn't seem right? If I build a C#/F# project on Linux, I expect the output file to not be `.exe` unless it's cross-compiling for Windows. Can anyone elaborate as to why this filter contains languages?

tritao: .NET executables end up in `.exe`, even in Linux.

realvictorprm: Agree, @samsinsane you end up feeding mono with the exe to start the application.

samsinsane: @tritao @realvictorprm Thanks for the clarification!

modules/vstudio/tests/fs2005/projectsettings.lua
tvandijck: is it? what is the alternative?

tvandijck:
@samsinsane thnx for the review, I'll see if I can "fix" things up today.

modules/vstudio/tests/cs2005/test_nuget_framework_folders.lua
```diff
@@ -1,42 +1,42 @@
 --
--- modules/vstudio/tests/cs2005/test_nuget_framework_folders.lua
+-- modules/vstudio/tests/dn2005/test_nuget_framework_folders.lua
```
samsinsane: The file location is still `cs2005`, was it supposed to move too?

realvictorprm:
I'll take a look into this now too.

PR #881, opened 2017-08-28
Object files should be marked as object files.

tvandijck:

### DIFF
--- a/modules/vstudio/tests/vc2010/test_files.lua
+++ b/modules/vstudio/tests/vc2010/test_files.lua
@@ -89,6 +89,16 @@
 		]]
 	end
 
+	function suite.none_onObjFile()
+		files { "docs/hello.obj" }
+		prepare()
+		test.capture [[
+<ItemGroup>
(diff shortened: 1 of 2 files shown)


samsinsane:
What benefits does this have over using `links` as you suggested in the other PR? I'm not 100% against this, but I think a project at work does have obj models in their project and I'd like to avoid issues if I can.

tvandijck:
> What benefits does this have over using links as you suggested in the other PR?
The files actually show up in the explorer... other then that, nothing.

After reading the other PR again, I think the original conclusion was to use the [buildaction](https://github.com/premake/premake-core/wiki/buildaction) API for this, which I think would be a good solution.

I think what we can do is if it's not set assume "none", so it doesn't change current behavior, and if you actually wanted them as "Object" in the project you can mark them with "buildaction("Object")". That way we all get cake and rainbows.

That said, I can indeed just 'link' these files... seems to work as well. So I don't really see this as high priority... let's just close all of these, and if I ever get to add the buildaction thing, I'll reopen.

Issue #884, opened 2017-08-31
filter with tags and files not working


tdesveaux:
I am using the tags filter to improve readability in some of the complex filters in our projects and came accross this issue.

Here is an example using the the base example from #789.

```
workspace 'foobar'
    configurations { 'release-std', 'debug-std', 'release-blz', 'debug-blz' }

   filter { 'configurations:*-std' }
        tags { 'use-std' }
   filter { 'configurations:*-blz' }
        tags { 'use-blz' }

   project 'test'
	kind 'ConsoleApp'
	platforms { 'x32', 'x64' }

	files
	{
		"src/main.c",
		"src/test.c"
	}

        filter { 'tags:use-std' } -- ok
            defines { 'USE_STD' }`

        filter { 'files:main.c' }  -- ok
            defines { 'MAIN' }

        filter { 'tags:use-blz', 'files:main.c' }  -- ignored
	    defines { 'BLZ_MAIN' }
```

I am willing to take a look but would appreciate if someone could point me in where I should be looking.

samsinsane:
My initial reaction would be to have [this](https://github.com/premake/premake-core/blob/f2f739d79cca259a13b5b27e828e243271119087/src/base/context.lua#L139-L143) function merge tables, but I'm concerned that this could result in all sorts of unexpected behaviour. Ultimately, the issue is that [this](https://github.com/premake/premake-core/blob/f2f739d79cca259a13b5b27e828e243271119087/src/base/oven.lua#L212)(project-level) has a higher precedence than [this](https://github.com/premake/premake-core/blob/f2f739d79cca259a13b5b27e828e243271119087/src/base/oven.lua#L605)(configurations-level), when applied to the file-level configurations [here](https://github.com/premake/premake-core/blob/master/src/base/fileconfig.lua#L103-L104)(note that `fcfg` already includes the project-level filters, and it's being "merged" into the `cfg` filters). That might actually be the intended behaviour, but `self.tags` is an empty table, which is why having `mergeFilters` merge tables would fix this.

@starkos and/or @tvandijck would have a better idea of what is the "better" way of fixing this, I haven't really touched this section of code before, so it would be a shot in the dark.

tvandijck:
I think this actually a bug indeed in the mergeFilters in that it assumes that values are 'nil' if not set... but for any API value that is a table, this is not true, it is an empty table instead. My guess would be that for any other value like this, the same issue would show up. 

[...]

PR #889, opened 2017-09-06
Fix context.mergeFilters

tdesveaux:
Fix related to #884.

This lead to a 10-15% performance drop for me.


tvandijck:
As much as I agree that this fixes #884, the performance implications are a little too severe... we should really look at ways to do this more conditionally, so the impact is minimized.. So I'm going to mark this as "request changes"...

tdesveauxPKFX:
@tvandijck I agree with this.
However, if someone can test on another project that would be good.
I tested this in our projects at work that were setup with a highly customized premake4 build and that I "adapted" to work in premake5.

I will still take a look to see what I can do to improve this.

tdesveauxPKFX:
With this, the performance are no longer an issue, difference with previous implementation are marginal.
However, this works only for tags and not other terms that are tables.
As far as I know, this is not an issue for now but it might become one in the future.

Would implementing mergeFilters in c improve performances?

tvandijck:
I don't know if it would be effective to do it in C... considering you then have to enumerate the table in C, you just end up with a whole lot of calls into lua api's anyway... I think the conditional approach is acceptable, and can always be extended.. I think it's really only the tags that suffer.

tdesveauxPKFX:
I forgot to mention, I also removed the call to table.joinunique as we don't care about duplicate and it takes more time to check duplicates than just copy the whole thing.

I did not remove the table.joinunique function though, I should remove it before merging this.

PR #896, opened 2017-09-19
fix disablewarnings

### LINKED ISSUE #795: C# disablewarnings not implemented
`disablewarnings` doesn't works in C# projects.

Differences in Visual C++ and Visual C# projects:
-  Visual C++ projects uses `<DisableSpecificWarnings>` tag
- Visual C# projects uses `<NoWarn>` tag

rNaire:
Fixes disablewarnings for C# and F#.  This resolves #795.


rNaire:
Didn't add F# tests as there were none similar.  Would be happy to do so if wanted though.

PR #901, opened 2017-10-02
Per file warning levels

tvandijck:


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2370,10 +2371,17 @@
...
 
-
 	function m.warningLevel(cfg)
 		local map = { Off = "TurnOffAllWarnings", Extra = "Level4" }
-		m.element("WarningLevel", nil, "%s", map[cfg.warnings] or "Level3")
+		m.element("WarningLevel", nil, map[cfg.warnings] or "Level3")
+	end
+
+
+	function m.warningLevelFile(cfg, condition)
```
samsinsane: Could this be merged with `m.warningLevel`? I'm pretty sure the other file-level functions do that?

tvandijck: This commit is a squash of 3 commits that tried that exact approach, and failed, so after going though those iterations, this is what ultimately worked.

tvandijck:
> Perhaps this PR should include support for gmake2, xcode4 and .NET, assuming they don't already support per file warnings.

* gmake2 already does the right thing with per file settings.
* xcode4 is not a thing here at blizzard.
* .NET doesn't support per file settings in the IDE.

samsinsane:
> xcode4 is not a thing here at blizzard.

This is a little bit off topic, but:
You guys still have your own xcode module right? Perhaps that could be worked into core in the same way that gmake2 was? Although, the xcode in your public repo appears to be for xcode6, so perhaps it will be more like the VS actions? Coexisting rather than as a future replacement, ultimately both will need to be reworked to share more of a common base, to enable xcode5, xcode7, and xcode8 files to be generated too.

tvandijck:
> You guys still have your own xcode module right?

Yes we do, and while the current "xcode" module in premake is called xcode4, it's really more about the format of the file, then the version of xcode. Which applies to "our" module as well. The format hasn't changed, some settings have come and gone, but most are essentially just environment variables, with little side effect if they are there. So ultimately the xcode4 as well as our xcode module just support xcode*, and not specifically xcode6, 7 or 8... it started out when xcode6 was latest, hence the name in our public repo... but we have since dropped the version, and have just moved along with whatever version of xcode our games ship with.. which is 8 right now I believe, but the projects still open in 6 and 7 as well, and so far work fine in the betas of 9.

> Coexisting rather than as a future replacement,

Yeah, I guess we can do that.. I don't really mind either way, the least I can do is update our public module repository with the latest of what we got, so people can download it and use it if they so please.


And just to give a little bit of history... When we started moving SC2 to use Premake, the Mac Team was really unhappy with the structure of what premake generated at the time. The xcode4 module created 200 xcodeproj folders, and a workspace. It took minutes to load that up, and compilation was taking hours. 

Our manually created project, was a single xcodeproj with multiple targets, which was just faster to load, easier to deal with in xcode itself, and while buildtimes are still not great, it was somewhere close to an hour orso.

So the mac team decided to write a new xcode module that basically just replicated the project that was until that time managed manually... a single xcodeproj file and nothing else. Our module as a result does not have a "OnProject" callback in the action... just a "OnWorkspace" callback, where it writes the entire thing in a single pass..

After that was working, the additional work that went into our module was to correctly use the override ability in xcode... in xcode you can have settings that are set on the project level, on the target level and on the configuration level.. So there is this somewhat complicated piece of code in our module to basically go around the premake configsets to fetch values only in available in the local configset, and not traverse down. So for example if you ask for "cfg.warnings" while in the premake script it was set at the project level, you'd get nil, while asking for "prj.warnings" would get you the setup values, and so in the xcode project it gets set at the right level.  Which reduced the size of the xcodeproj file by a lot, since it's actually rather rare that a setting is setup for a particular config specifically.

In addition it supports [Custom Rules](https://github.com/premake/premake-core/wiki/Custom-Rules), and per file configuration.

samsinsane:
@tvandijck Thanks for that, very interesting read! It sounds like these XCode modules might be situational (or perhaps everyone will always prefer to use targets instead), but maybe there's a way to bring the two modules together to give the end user more power over how projects are created for XCode? I don't know much about XCode, so I'm just assuming there are cases where users would want projects and targets in the same workspace.

starkos:
FWIW, I took the multiple projects approach because of [this Xcode bug](https://stackoverflow.com/questions/1456806/xcode-dependencies-across-different-build-directories).

PR #903, opened 2017-10-02
removed a no longer needed make pch race condition fix in gmake2 action

tvandijck:


samsinsane:
How has this issue been resolved in gmake2? Could it be applied to gmake too?

tvandijck:
> Could it be applied to gmake too?

Maybe, we don't use the gmake action (it's removed from our internal fork), and I don't know what exactly was fixed in the gmake2 action that makes this no longer a requirement... That said, a lot of the gmake2 code is very similar to the gmake code, so someone could figure things out probably. I think the race condition had to do with the creation of folders...

samsinsane:
> (it's removed from our internal fork)

May I ask why? I'm assuming that the names overlap? Doesn't this make modifying Premake painful?

> I don't know what exactly was fixed in the gmake2 action that makes this no longer a requirement

I kind of expected that to be the case, just thought I'd check so we could reduce the discrepancies between the two actions to assist with the transition from gmake -> gmake2.

tvandijck:
> May I ask why?

mostly because people are used to typing "premake gmake" on linux, all our build farms are setup that way, I guess I could have just renamed the gmake trigger to 'gmake-old' or something, but instead I choose to remove the code entirely, and just rename the gmake2 trigger to gmake. I'll think about this a little more.

> I don't know what exactly was fixed in the gmake2 action that makes this no longer a requirement

@bwhittle would know... he made the fixes, and maybe he can at least explain what is needed to fix it for the standard gmake action.

starkos:
It sounds like most of the good stuff is going into gmake2 at this point…is it possible to switch over?

tvandijck:
it is, but not without deleting the "D" module in the process... it was why we went this path, since the "D" module overrides a bunch of functionality in the gmake module, and with the big refactor I did in the gmake2 module a lot of that broke.

bwhittle:
samsinsane thank you for pointing this out.  Since I work off of the Bliz fork, I didn't realize gmake itself was forked so I can't speak to the reasons why the fork occurred.

I looked at the gmake generation and it does NOT have my prior pch race condition fixes.  As such, I don't think its appropriate to remove the .NOTPARALLEL.  However to be fair, I don't think the current .NOTPARALLEL in the gmake module prevents possible pch race conditions as the .NOTPARALLEL is only emiteds at the workspace level, which just means only one project can run at a time, however each project allows parallel execution.

Prior to an internal roll out of premake for our linux builds, Tom did warn me about possible issues with pch and using make with more than one job thread.  For us this was a big deal as we make heavy use of pch.

My fixes (which all appear to be in gmake2) were rather simple and made use of the pipe operator
https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

For example here an example of the code gen in make

```
ifneq (,$(PCH))
$(OBJECTS): $(GCH) $(PCH)
$(GCH): $(PCH)
	@echo $(notdir $<)
ifeq (posix,$(SHELLTYPE))
	$(SILENT) mkdir -p $(OBJDIR)
else
	$(SILENT) mkdir $(subst /,\\,$(OBJDIR))
endif
	$(SILENT) $(CXX) -x c++-header $(ALL_CXXFLAGS) -o "$@" -MF "$(@:%.gch=%.d)" -c "$<"
endif
```

and here it is gmake2

```
ifneq (,$(PCH))
$(OBJECTS): $(GCH) $(PCH) | $(OBJDIR) $(PCH_PLACEHOLDER)
$(GCH): $(PCH) | $(OBJDIR)
	@echo $(notdir $<)
	$(SILENT) $(CXX) -x c++-header $(ALL_CXXFLAGS) -o "$@" -MF "$(@:%.gch=%.d)" -c "$<"
$(PCH_PLACEHOLDER): $(GCH) | $(OBJDIR)
	$(SILENT) touch "$@"
else
$(OBJECTS): | $(OBJDIR)
endif
```

I can't promise the above fixes all possible race conditions (it does address the ordering of directory creation and pch creation).   I can report that we've done a bit of testing and have yet to hit a race condition after the removal of the .NOTPARALLEL (which was limiting our build performance.)

samsinsane:
> I'll think about this a little more.

@tvandijck It was mostly for my curiosity, you've mentioned that a few of the modules aren't built internally (could just be my perception and maybe it was just D and now gmake) but you make so many core changes that I've often wondered how you avoid breaking these modules. Perhaps we should just drop D and gmake, and rename gmake2 to gmake?

@bwhittle Thank you for taking the time to explain this! Based on the discussion above, making this change in gmake mightn't be required, and just making the jump to gmake2 would be the best option.

starkos:
> Perhaps we should just drop D and gmake, and rename gmake2 to gmake?

FWIW, I'm not opposed to this. Another option would be to remove or comment out the D module's makefile integration, and leave the rest intact.

Issue #480, opened 2016-04-19
Setting libtype to "None" makes VS2015 think the project is for VS2010


TrianglesPCT:
Setting **kind** to "None" disables the project from being built. That much works--

But it also confuses VS2015, and each project marked "None" shows up in the solution explorer as

name of project(Visual Studio 2010)

This causes warnings such as:

depo\common\build_premake\projects\shared\reflection\reflection.vcxproj : warning  : The build tools for Visual Studio 2010 (v100) cannot be found. Install Visual Studio 2010 (v100) to build using the Visual Studio 2010 (v100) build tools.

Also this doesn't happen every time-- seems to inconsistently happen...

DK22Pac:
In Visual Studio 2017 and 2015 (and maybe in 2012/2013 too) "PlatformToolset" node is required in MakeFile project configuration.
Here's how default MakeFile project configuration looks like in VS 2017:
'''<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
    <ConfigurationType>Makefile</ConfigurationType>
    <UseDebugLibraries>false</UseDebugLibraries>
    <PlatformToolset>v141</PlatformToolset>
  </PropertyGroup>'''
In current premake version, there's a special condition for "MakeFile" and "None" projects types ("None" projects are also VS MakeFile projects) to not add "PlatformToolset" node if project has no source (.cpp) files.
You can see this [here](https://github.com/premake/premake-core/blob/master/modules/vstudio/vs2010_vcxproj.lua#L2053).
If "PlatformToolset" node is not present, Visual Studio will set it to v100 "by default". Even if you open your solution/project in VS 2017.

To fix this, lines  2052 - 2056 must be removed.

DK22Pac:
Oh, I didn't notice it's already fixed 👍. You can ignore my changes.

PR #904, opened 2017-10-02
Fix makefile/utility project toolset under vs2015 and vs2017

tvandijck:
See: https://github.com/premake/premake-core/issues/480


tvandijck:
@samsinsane I don't know, we no longer use that.. if anyone can test that, I gladly modify the if statement.

samsinsane:
@tvandijck Same, sounds good to me!

PR #906, opened 2017-10-02
for vs2017 and up nuget is now handled by msbuild instead of premake

tvandijck:
via the new PackageReference nodes.

The upsides here are
1) the MS package downloading and validation is much faster than the premake version
2) since the config is in the project file instead of packages.config which means
   multiple nuget projects can sit in the same folder.

### DIFF
--- a/modules/vstudio/vs2010.lua
+++ b/modules/vstudio/vs2010.lua
@@ -85,16 +85,18 @@
 			end
 		end
 
-		-- Skip generation of empty packages.config files
-		local packages = p.capture(function() vstudio.nuget2010.generatePackagesConfig(prj) end)
-		if #packages > 0 then
-			p.generate(prj, "packages.config", function() p.outln(packages) end)
-		end
+		if not vstudio.nuget2010.supportsPackageReferences(prj) then
+			-- Skip generation of empty packages.config files
+			local packages = p.capture(function() vstudio.nuget2010.generatePackagesConfig(prj) end)
+			if #packages > 0 then
+				p.generate(prj, "packages.config", function() p.outln(packages) end)
+			end
 
-		-- Skip generation of empty NuGet.Config files
-		local config = p.capture(function() vstudio.nuget2010.generateNuGetConfig(prj) end)
-		if #config > 0 then
-			p.generate(prj, "NuGet.Config", function() p.outln(config) end)
+			-- Skip generation of empty NuGet.Config files
+			local config = p.capture(function() vstudio.nuget2010.generateNuGetConfig(prj) end)
+			if #config > 0 then
+				p.generate(prj, "NuGet.Config", function() p.outln(config) end)
+			end
 		end
 	end
 
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -1711,9 +1711,11 @@
 	end
 
 	function m.importNuGetTargets(prj)
-		for i = 1, #prj.nuget do
-			local targetsFile = nuGetTargetsFile(prj, prj.nuget[i])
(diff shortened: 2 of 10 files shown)


tvandijck:
@aleksijuvani since most of the nuget stuff came from you, would you be able to review this change?

aleksijuvani:
Nice. I've been meaning to do this for a while but I haven't had time. I can take a look later, maybe during the weekend?

tvandijck:
awesome, thank you ;)

bwhittle:
btw, i don't know whats up with the diff in modules/vstudio/tests/vc2010/test_nuget_package_references.lua

However there should only be a handful of changes.
1. adding a call to packageReferences around line 45
2. adding a `and _ACTION < "vs2017"` to the if clause at the start of nuGetReferences
3. the function impl of packageReferences

bwhittle:
oh it looks like some tab spacing issue :(

modules/vstudio/vs2010.lua
```diff
@@ -85,16 +85,18 @@
 			end
 		end
 
-		-- Skip generation of empty packages.config files
-		local packages = p.capture(function() vstudio.nuget2010.generatePackagesConfig(prj) end)
-		if #packages > 0 then
-			p.generate(prj, "packages.config", function() p.outln(packages) end)
-		end
+		if _ACTION < "vs2017" then
```
aleksijuvani: PackageReference is not yet supported for C++ projects, so this will have to be changed to check for that:

```lua
if _ACTION < "vs2017" or p.project.iscpp(prj) then
```

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1710,9 +1710,11 @@
 	end
 
 	function m.importNuGetTargets(prj)
-		for i = 1, #prj.nuget do
-			local targetsFile = nuGetTargetsFile(prj, prj.nuget[i])
-			p.x('<Import Project="%s" Condition="Exists(\'%s\')" />', targetsFile, targetsFile)
+		if _ACTION < "vs2017" then
```
aleksijuvani: This `if _ACTION < "vs2017" then` check will also have to be removed for the same reason.

src/base/oven.lua
```diff
@@ -667,7 +667,7 @@
 			-- packages.config file to the project. Is there a better place to
 			-- do this?
 
-			if #prj.nuget > 0 then
+			if #prj.nuget > 0 and _ACTION < "vs2017" then
```
aleksijuvani: This will also have to be changed to check for a C++ project.

src/base/validation.lua
```diff
@@ -225,7 +225,7 @@
 
 
 	function m.NuGetHasHTTP(prj)
-		if not http and #prj.nuget > 0 then
+		if not http and #prj.nuget > 0 and _ACTION < "vs2017" then
```
aleksijuvani: Also needs a check for C++ project.

src/base/validation.lua
```diff
@@ -255,6 +255,9 @@
 
 	function m.uniqueProjectLocationsWithNuGet(wks)
 		local locations = {}
+		if _ACTION >= "vs2017" then
```
aleksijuvani: We'll also need to check for C++ projects here.

aleksijuvani:
It might be a good idea to create a function that does the `_ACTION < "vs2017" or p.project.iscpp(prj)` check so that we can update it easily once/if they start supporting PackageReference for C++ projects.

tvandijck:
K that should all be taken care of... in addition I moved the nuget validation code to inside the visual studio module, since most of the nuget code lives the validation code is pretty specific to how Visual Studio expects those values. Any other backend could deal with nuget in a different way, and require different validation.

tvandijck:
@aleksijuvani ready for final review, whenever you have time...

PR #908, opened 2017-10-02
[core] path:join can now be deferred for those paths that execute lua

tvandijck:
prior to this change, paths that begin with ${ } were sometimes assumed to
be either absolute or relative.  new functions have been added in order
to reduce the number of systems that can break.

new functions are:

* path:absolutetype -> similair to isabsolute but can return an unknown or maybe result
* path:deferredjoin -> similair to path:join but creates a string for unknown absolute
	paths which we except to be generated at bake time
* path:hasdeferredjoin -> predicate to determine if a path needs to be evaled for
    	a deferred join
* path:resolvedeferredjoin -> resolves a deferredjoin path to an actual path
    	this is to be called after any %{} replacements occur.

right now all api path types use the deferredjoin code path and
detoken, solution:makerelative and project:makerelative have been updated
to use resolvedeferredjoin.

unit tests have been created to test the four new path functions.

### DIFF
--- a/src/host/path_isabsolute.c
+++ b/src/host/path_isabsolute.c
@@ -4,24 +4,14 @@
  * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
  */
 
- /**
- * \file   path_isabsolute.c
- * \brief  Determines if a path is absolute or relative.
- * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
- */
-
 #include "premake.h"
 #include <ctype.h>
 #include <string.h>
-
+#include "path_isabsolute.h"
 
 #if PLATFORM_WINDOWS
 #define strncasecmp _strnicmp
 #endif
-
-#define JOIN_RELATIVE 0
-#define JOIN_ABSOLUTE 1
-#define JOIN_MAYBE_ABSOLUTE 2
 
 int do_absolutetype(const char* path)
 {
@@ -84,9 +74,7 @@ int do_absolutetype(const char* path)
 		{
 			closing = strchr(path + 2, '}');
 			if (closing != NULL)
-			{
 				return JOIN_MAYBE_ABSOLUTE;
-			}
 		}
 		// find the second closing %
 		path += 1;
--- a/src/host/path_isabsolute.c
+++ b/src/host/path_isabsolute.c
@@ -4,10 +4,117 @@
  * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
  */
 
+ /**
+ * \file   path_isabsolute.c
+ * \brief  Determines if a path is absolute or relative.
+ * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
+ */
+
 #include "premake.h"
 #include <ctype.h>
 #include <string.h>
 
+
+#if PLATFORM_WINDOWS
+#define strncasecmp _strnicmp
+#endif
+
+#define JOIN_RELATIVE 0
+#define JOIN_ABSOLUTE 1
+#define JOIN_MAYBE_ABSOLUTE 2
+
+int do_absolutetype(const char* path)
+{
+	char c;
+	const char* closing;
+	size_t length;
+
+	while (path[0] == '"' || path[0] == '!')
+		path++;
+	if (path[0] == '/' || path[0] == '\\')
+		return JOIN_ABSOLUTE;
+	if (isalpha(path[0]) && path[1] == ':')
+		return JOIN_ABSOLUTE;
+
+	// $(foo) and %(foo)
+	if ((path[0] == '%' || path[0] == '$') && path[1] == '(')
+	{
+		char delimiter = path[0];
+		closing = strchr(path + 2, ')');
+		if (closing == NULL)
+			return JOIN_RELATIVE;
+
+		path += 2;
+		// special case VS macros %(filename) and %(extension) as normal text
+		if (delimiter == '%')
+		{
+			length = closing - path;
+			switch (length) {
+			case 8:
+				if (strncasecmp(path, "Filename)", length) == 0)
+					return JOIN_RELATIVE;
+				break;
+			case 9:
+				if (strncasecmp(path, "Extension)", length) == 0)
+					return JOIN_RELATIVE;
+				break;
+			default:
+				break;
+			}
+		}
+
+		// only alpha, digits, _ and . allowed inside $()
+		while (path < closing) {
+			c = *path++;
+			if (!isalpha(c) && !isdigit(c) && c != '_' && c != '.')
+				return JOIN_RELATIVE;
+		}
+
+		return JOIN_ABSOLUTE;
+	}
+
+	// $ORIGIN.
+	if (path[0] == '$')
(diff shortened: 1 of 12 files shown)


tvandijck:
This is a pretty big change, and I'm not 100% capable of explaining it as it was introduced by the World of Warcraft team... I'll let @bwhittle explain things further to get the discussion started...

bwhittle:
This change came after observing inconsistent with odd behavior when specific tokens were used in paths (sometimes a specific token worked fine in one context but not another.)
At one point I realized that the issue was that a specific premake token I was using (file.basename, I believe) was being treated by premake as an absolute path.  At first I tried changing the behavior of path_isabsolute to treat tokens as relative paths and this approach caused issues since most of the premake tokens expand to full paths.

I realized the source of the issue is that false positives or false negatives with path_isabsolute can lead to several problems.

The core of the issue is if a path is joined with another path which starts off with a premake token, path_join can’t correctly form the correct path since token expansion occurs later in the process.

One solution (which is implemented in this change) is to extend path_isabsolute to return a "maybe" if a premake token is at the start of the path (either '&{' or '%[' .)  When joining paths, if the 2nd path has a “maybe” result then both parameters are concatenated with a delimiter (currently the pipe character.)   Multiple “maybe absolute” paths can be concatenated together with essentially treats it as a path stack.   

Finally, after token expansion, we then double check to see if the path is a deferred join (quick scan for the delimiter) and if the path is a deferred join then we break the strings apart and perform a normal path_join.   

I’ve added a number of unit tests to handle all of the deferred join cases I can think.  Additionally, we've been using this using this code in production for several months now.

I still need to create a couple unit tests to highlight the original path_join problem in context of a full premake project with tokens and token expansion.  Such unit tests would result in unit test failures without a fix (either this fix or another.)

Potential issues with this commit are
The implementation could likely be cleaner.  
For example, it may make more sense to have a single code path in path_join to always create the deferred join path if need be, however I added a parameter to path_join (allowDeferredJoin) to make it explicit which type of behavior is desired.   
However, the only time allowDeferredJoin should not be desirable are when paths are joined AFTER token expansion AND the second path has a token (in which case I’d argue the path is incorrect either way and is the result of a premake script bug.)  So given that, it may make more sense to have a single code path for path_join instead of two for the sake of simplicity.  

This commit does change behavior (as most bug fixes do).  This means it’s possible that some premake projects are currently working around the path_join + token bug via string concatentation or other methods.  If so, then it’s possible this fix could change the final paths in their generated project files than compared to what they are getting now (in which case the original work around likely needs to be removed.)

There is special handling for paths with tokens spread throughout the code base.  I’ve fixed one of the case (in api.lua.)  Likely other code can be simplified to no longer inspect if a given path has a token or not (i.e. just join any two paths to be resolved after token expansion.)

bwhittle:
btw, while it can go up in a different commit here is a new unit test which highlights the problem you can get with any api path object that begins with a premake macro.  These tests fail without the fix, but succeed with the above changes.

--
-- test_tokens.lua
-- Copyright (c) Jason Perkins and the Premake project
--


	local p = premake
	local suite =  test.declare("vstudio_vs2010_tokens")
	local vc2010 = p.vstudio.vc2010


--
-- Setup
--

	local wks, prj

	function suite.setup()
		p.action.set("vs2010")

		wks = test.createWorkspace()
	end

	local function prepare()
		prj = test.getproject(wks, 1)
		vc2010.files(prj)
	end



	function suite.customBuild_onBuildRuleMultipleBuildOutputs()
		location "projects"
		files { "hello.cg" }

		filter "files:**.cg"
			buildcommands { "cgc %{file.relpath}" }
			buildoutputs { "%{file.basename}.a", "%{file.basename}.b" }
		prepare()
		test.capture [[
<ItemGroup>
	<CustomBuild Include="..\hello.cg">
		<FileType>Document</FileType>
		<Command>cgc %(Identity)</Command>
		<Outputs>../%(Filename).a;../%(Filename).b</Outputs>
	</CustomBuild>
</ItemGroup>
		]]
	end

	function suite.customBuild_onBuildRuleWithMessage()
		location "projects"
		files { "hello.cg" }
		filter "files:**.cg"
			buildmessage "Compiling shader %{file.relpath}"
			buildcommands { "cgc %{file.relpath}" }
			buildoutputs { "%{file.basename}.obj" }
		prepare()
		test.capture [[
<ItemGroup>
	<CustomBuild Include="..\hello.cg">
		<FileType>Document</FileType>
		<Command>cgc %(Identity)</Command>
		<Outputs>../%(Filename).obj</Outputs>
		<Message>Compiling shader %(Identity)</Message>
	</CustomBuild>
</ItemGroup>
		]]
	end

	function suite.customBuild_onBuildRuleWithAdditionalInputs()
		location "projects"
		files { "hello.cg" }
		filter "files:**.cg"
			buildcommands { "cgc %{file.relpath}" }
			buildoutputs { "%{file.basename}.obj" }
			buildinputs { "common.cg.inc", "common.cg.inc2" }
		prepare()
		test.capture [[
<ItemGroup>
	<CustomBuild Include="..\hello.cg">
		<FileType>Document</FileType>
		<Command>cgc %(Identity)</Command>
		<Outputs>../%(Filename).obj</Outputs>
		<AdditionalInputs>../common.cg.inc;../common.cg.inc2</AdditionalInputs>
	</CustomBuild>
</ItemGroup>
		]]
	end

src/host/path_isabsolute.c
```diff
@@ -4,74 +4,128 @@
  * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
  */
 
+ /**
+ * \file   path_isabsolute.c
+ * \brief  Determines if a path is absolute or relative.
+ * \author Copyright (c) 2002-2016 Jason Perkins and the Premake project
+ */
+
```
samsinsane: Bad merge/rebase?

src/host/path_isabsolute.c
```diff
@@ -4,74 +4,128 @@
...
+	// either %ORIGIN% or %{<lua code>}
 	if (path[0] == '%')
 	{
+		if (path[1] == '{') //${foo} need to defer join until after detokenization
+		{
+			closing = strchr(path + 2, '}');
+			if (closing != NULL)
+			{
+				return JOIN_MAYBE_ABSOLUTE;
+			}
```
samsinsane: Based on other code around here, this doesn't need to have the curly braces.

src/host/path_join.c
```diff
@@ -1,62 +1,56 @@
...
+* \brief  Join two or more pieces of a file system path.
+* \author Copyright (c) 2002-2013 Jason Perkins and the Premake project
+*/
 
 #include "premake.h"
 #include <string.h>
 
+#define JOIN_RELATIVE 0
+#define JOIN_ABSOLUTE 1
+#define JOIN_MAYBE_ABSOLUTE 2
```
samsinsane: These are also defined in `path_isabsolute.c`, perhaps they should exist in a shared header to avoid issues later?

tvandijck:
we'll clean up those style issues.

tvandijck:
Failure is only due to http test timing out... merging since Mac and Windows builds passed.

PR #916, opened 2017-10-11
Added asmoptions [BUT WHY DOESN'T IT WORK???]

hatcat:
Just as you can add additional build options to the compile tool through buildoptions, so I want to be able to add additional options to the assembler tool. I registered an API called asmoptions added a function to vs2010_vcxproj.lua called m.additionalASMOptions and added it to the Masm category.

Now, when I make an asmoptions declaration at workspace scope, those declarations aren't forwarded through to the assembly files. I think I'[m missing something, and as I'm writing this I'm getting an idea that I may have missed the distinction between file specific and non-file specific declarations.

Any ideas from anyone?

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -794,6 +794,7 @@
 			local fileCfgFunc = function(fcfg, condition)
 				if fcfg then
 					return {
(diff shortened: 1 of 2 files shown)


hatcat:
False assumption: compile is a built in tool while masm is a custom tool. I'll come up with another solution.

PR #917, opened 2017-10-11
MASM preprocessor definitions

hatcat:


tvandijck:
So, it seems you reverted the asmoptions in this PR, but instead added defines...

Either way, looking at the individual changes, and including the asmoptions, you are indeed correct that this change only added them to the per-file config settings. So unless you set the asmoptions in a `filter { 'files:*.asm' }` they would indeed not apply...

To make it work for the "workspace/project" you have to also emit the <AdditionalOptions> element in a <Masm> group where the ClCompile group is emitted. It seems that when I made the categories I didn't really think about this one to be honest... but the insertion point is here:

https://github.com/premake/premake-core/blob/master/modules/vstudio/vs2010_vcxproj.lua#L277

you can look at the `m.clCompile` function as an example, or `m.resourceCompile` is an easier example.

hatcat:
I realised that what I particularly wanted to do was set preprocessor definitions for particular assembly files, so that's what I've enabled in this pull request. The asmoptions is not necessary for us right now so I decided against adding it to our fork. I have yet to get the hang of the github mechanisms, so pardon the extraneous portions.

Should we need to set additional Masm options, I'll revisit this part of the code.

PR #923, opened 2017-10-13
Add Visual Studio ARM64 support. 

tvandijck:


tvandijck:
We might indeed yes... this really only applies if the target is windows, but I don't think we've really encountered any issues with this.

PR #925, opened 2017-10-16
Exception handling completed for vcxproj

hatcat:
Added synchronous and C-throw exceptions to complete the set.


starkos:
Would it be possible to clean up the branch so it contains on the commit(s) related to this change? It looks like some extra baggage got pulled in.

hatcat:
Right, new PR in the queue.

PR #926, opened 2017-10-16
Fixes translateCommandsAndPaths, and debugargs.

tvandijck:


premake5.lua
```diff
@@ -155,8 +155,8 @@
 
 		filter "configurations:Debug"
 			targetdir   "bin/debug"
-			debugargs   { "--scripts=%{prj.location}/%{path.getrelative(prj.location, prj.basedir)} test"}
-			debugdir    "%{path.getrelative(prj.location, prj.basedir)}"
+			debugargs   { "--scripts=.", "test" }
```
tvandijck: I don't think this line is correct yet...

PR #927, opened 2017-10-16
bug fix for recursive token expansion

tvandijck:
added test in gmake2.


starkos:
Would it be possible to get a test that should the recursive expansion? The gmake tests on this PR don't seem to show it?

PR #930, opened 2017-10-17
Completed exception handling

hatcat:
Added synchronous and C-Throw to VS projects.


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1475,6 +1475,12 @@
 			m.element("ExceptionHandling", condition, "false")
 		elseif cfg.exceptionhandling == "SEH" then
 			m.element("ExceptionHandling", condition, "Async")
+		elseif cfg.exceptionhandling == "On" then
+			m.element("ExceptionHandling", condition, "Sync")
+		elseif cfg.exceptionhandling == "Default" then
```
tvandijck: the idea is that default doesn't emit the field, so it becomes the 'default' of visual studio...
This is probably also why the tests are failing now.

PR #931, opened 2017-10-17
Added synchronous and C-Throw to VS projects.

hatcat:


tvandijck:
Awesome.. thank you.... For what it is worth... you don't have to close and make a new pull request. When you push your updated branch to the same branch as the existing pull request (master in your case), the pull request will automatically update. That way it also keeps prior discussions and reviews in the same context.

tvandijck:
An on another note, it is often easier to make pull requests from side branches...
that way it is easier to keep your master in sync with the official master. It's obviously up to you to pick your workflow, but I found that making a clean branch from the official master, and then 'cherry-pick' the changes I want to commit as a pull request into that new branch, it is easier to update those branches and fix things in isolation. It also allows you to make multiple pull requests at once.

hatcat:
Yes, I really need to bite the bullet and RTFM for git and github. BRB...

PR #937, opened 2017-10-30
Add table.combine()

SE2Dev:
Add table.combine(), a function which combines the behaviors of table.merge() and table.join() to combine multiple tables, but append indexed elements rather than overwrite them.

### DIFF
--- a/src/base/table.lua
+++ b/src/base/table.lua
@@ -373,6 +373,25 @@
 	end
 
 
+--
+-- Combines multiple tables by adding the key-value associations from one table to another
+-- then adding indexed values (essentially a combined table.merge() & table.join()
+-- Returns the resulting merged table
+--
+
+	function table.combine(...)
+	    -- Merge the keyed values
+	    local result = table.merge(...)
+	    -- Append the indexed values
+	    local joined = table.join(...)
+	    for i, v in ipairs(joined) do
+	        result[i] = v
+	    end
+
+	    return result
+	end
+
+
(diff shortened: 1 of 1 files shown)


starkos:
I think I would be inclined to add this to one of the other functions, instead of having three. @tvandijck and @samsinsane, any thoughts?

tvandijck:
My only hesitance is that table is getting big and hard to follow with function that have somewhat 'clear' names, but not very self-documenting...

both `join`, `merge` and `combine` all by name to me do the same thing, but none of them actually do, and are slightly different. It's already bad that `join` and `merge` exist, adding a third just makes it worse... 

Looking at the implementation of all:
- `join` should be `table.concat`, or something like that.
- `merge` is actually doing a  join in the 'mathematical' sense.
- `combine` is somewhat hard to describe to be honest, I'm not exactly sure.

if we take a step further, I think some table operations are 'array' operations, and some are 'map/table' operations. `table.join` is currently an array operation, so it could maybe even be `array.concat`.

starkos:
I would like to create more "strict" collection classes/namespaces at some point. Array is definitely one; I think there may be opportunities to optimize performance for array operations. I'd also like to formalize a "set", which we use quite a lot (an array that can only have one of each unique value).

So it sounds like this would be better just dropped right into `table.merge()` then?

samsinsane:
I'm going to close this off, please feel free to reopen.

PR #944, opened 2017-11-09
C++0x dialect support

TurkeyMan:
Useful on machines with old compilers


src/tools/gcc.lua
```diff
@@ -135,9 +135,12 @@
 		},
 		cppdialect = {
 			["C++98"] = "-std=c++98",
+			["C++0x"] = "-std=c++0x",
 			["C++11"] = "-std=c++11",
+			["C++1y"] = "-std=c++1y",
 			["C++14"] = "-std=c++14",
-			["C++17"] = "-std=c++1z",
+			["C++1z"] = "-std=c++1z",
+			["C++17"] = "-std=c++1z", -- TODO: will the actual standard be 'c++17'?
```
samsinsane: According to [this](https://gcc.gnu.org/onlinedocs/gcc/Standards.html):
> The C++ language was further revised in 2017 and ISO/IEC 14882:2017 was published. This is referred to as C++17, and before publication was often referred to as C++1z. GCC supports all the changes in the new specification. For further details see https://gcc.gnu.org/projects/cxx-status.html#cxx1z. Use the option -std=c++17 to select this variant of C++. 

So yes, it should be changed to `c++17`.

src/_premake_init.lua
```diff
@@ -754,8 +754,11 @@
 		allowed = {
 			"Default",
 			"C++98",
+			"C++0x",
 			"C++11",
+			"C++1y",
 			"C++14",
+			"C++1z",
```
samsinsane: Unless I'm mistaken, don't these all have `gnu++` equivalents that should be supported too?

TurkeyMan: Dun fixed it.

PR #947, opened 2017-11-12
Fixed a bug with gmake2 not generating compiler macros when using gcc.

Gereld:
gcc.gettoolname returns nil for gcc toolset. As a result gmake2 doesn't generate compiler macros (CC=gcc, CXX=g++, etc...).

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -502,5 +502,5 @@
 		if (cfg.gccprefix and gcc.tools[tool]) or tool == "rc" then
 			return (cfg.gccprefix or "") .. gcc.tools[tool]
 		end
-		return nil
+		return gcc.tools[tool]
 	end


Gereld:
On a second thought, I think not defining CC, CXX and AR in the makefile is done by design. It makes sense on Linux where those environement variables are usually set. But on Windows, they are not. As a result, invoking "premake5 gmake2" or "premake5 gmake2 --cc=gcc" will still leave them unset. Thus you can't use the generated makefile on Windows.
In my opinion, the solution is to sort this out in the makefile by testing, for example, if CC is set, if not output the "CC=gcc" statement.

tvandijck:
Tests are failing, so until that time we can't merge this... other then that, I don't really have an opinion about whether this is a good change or not... I have no issues with it personally.

PR #949, opened 2017-11-23
Added DPI awareness support to VS2010+ projects

### LINKED ISSUE #783: DPI Awareness
Not sure about earlier versions of Visual Studio. But in VS2017 there is an option in the Property Pages under Manifest Tool to enable DPI Awareness. It doesn't appear that premake supports this.

The options appear to be:
None
High DPI Aware
Per Monitor High DPI Aware

Output example:
```
    <Manifest>
      <EnableDpiAwareness>PerMonitorHighDPIAware</EnableDpiAwareness>
    </Manifest>
```

samsinsane:
Resolves #783

### DIFF
--- a/modules/vstudio/tests/vc2010/test_manifest.lua
+++ b/modules/vstudio/tests/vc2010/test_manifest.lua
@@ -52,3 +52,37 @@
 		prepare()
 		test.isemptycapture()
 	end
+
+--
+-- Check that DPI Awareness emits correctly
+--
+
+	function suite.dpiAwareness_None()
+		dpiawareness "None"
+		prepare()
+		test.capture [[
+<Manifest>
+	<EnableDpiAwareness>false</EnableDpiAwareness>
+</Manifest>
+		]]
+	end
+
+	function suite.dpiAwareness_High()
+		dpiawareness "High"
+		prepare()
+		test.capture [[
+<Manifest>
+	<EnableDpiAwareness>true</EnableDpiAwareness>
+</Manifest>
+		]]
+	end
+
+	function suite.dpiAwareness_HighPerMonitor()
+		dpiawareness "HighPerMonitor"
+		prepare()
+		test.capture [[
+<Manifest>
+	<EnableDpiAwareness>PerMonitorHighDPIAware</EnableDpiAwareness>
+</Manifest>
+		]]
+	end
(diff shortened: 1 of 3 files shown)


samsinsane:
@tvandijck Not overly sure to be honest, it was a feature request. However, it [looks](https://msdn.microsoft.com/en-us/library/windows/desktop/mt843498(v=vs.85).aspx) like it enables various functions to respond differently based on the setting. (I skimmed this page, so I could be incredibly wrong)

samsinsane:
Added [documentation](https://github.com/premake/premake-core/wiki/dpiawareness), is the availability right? Are we drilling down into the alpha releases? I feel like I saw that the other day, but I'm not sure what it was on.

tvandijck:
Availability should be alright...
In my opinion our next release should just be 5.0, or preferably 5.1 (see below), we've been in alpha for 3 years now, things seem pretty stable overall.

All the ideas @starkos has for the query system seem a rather large refactor and deviation from what we have currently that I'd almost argue that that is a 5.5 or even a 6.0. We are internally experimenting with that too btw, based on his ideas and test branches.

The other thing I really think we should be doing is automate the release number.
We use this modification: 
- https://github.com/Blizzard/premake-core/commit/f93c769b5849847b19cacdd28d1d6177bc7cd1e5

which basically just takes the tag number from git, and embeds that into the source code. I proposed it once as a PR, but I think at the time there was some argument about the dependency on git, which I feel is somewhat moot since we're on github, and it's just the embed script for automation, we can easily extent it to support other source control systems, or use a command line argument to use the git system, and default to the current behavior.. 

We could start at 5.0.0, although it'll screw us up a little bit, as we by now are at 5.0.41, but we'll figure that part out here then. We could also just start at 5.1.0, which would help us out.

pdoane:
The Windows 10 Creators Update made additional changes to DPI awareness.  There are now 4 values that are interesting:

- Unaware
- System Aware
- Per Monitor Aware V1
- Per Monitor Aware V2

For PerMonitorV2, the manifest should generate:

```
<dpiAware>True/PM</dpiAware>
<dpiAwareness>PerMonitorV2, PerMonitor</dpiAwareness>
```

I don't think PerMonitorV2 has been reflected into .vcxproj yet.

It might be a good idea to change the enumeration names as "High" isn't really clear as to whether it should map to V1 or V2 rules.

More details here: https://blogs.windows.com/buildingapps/2017/04/04/high-dpi-scaling-improvements-desktop-applications-windows-10-creators-update/

PR #951, opened 2017-11-28
Add support for `targetextension` in xcode4

tempura-sukiyaki:


modules/xcode/xcode_common.lua
```diff
@@ -892,11 +892,15 @@
 			settings['EXECUTABLE_PREFIX'] = cfg.buildtarget.prefix
 		end
 
-		--[[if cfg.targetextension then
+		if cfg.kind ~= "ConsoleApp" and cfg.targetextension then
```
samsinsane: Is there any particular reason why a `ConsoleApp` can't specify a `targetextension`? `app.out` for example?

src/_premake_init.lua
```diff
@@ -1655,6 +1655,9 @@
 
 	-- Add variations for other Posix-like systems.
 
+	filter { "system:MacOSX", "kind:WindowedApp" }
+		targetextension ".app"
```
samsinsane: Could this also apply to `system:ios`?

src/base/config.lua
```diff
@@ -49,10 +49,14 @@
 		local bundlepath = ""
 
 		if cfg.system == p.MACOSX and kind == p.WINDOWEDAPP then
```
samsinsane: Could this also apply to `p.iOS`?

src/base/config.lua
```diff
@@ -49,10 +49,14 @@
...
-			bundlename = basename .. ".app"
+			bundlename = basename .. extension
 			bundlepath = path.join(bundlename, "Contents/MacOS")
 		end
 
+		if cfg.system == p.MACOSX and kind == p.CONSOLEAPP then
```
samsinsane: Same comment as before, is there any reason why we're being so hostile towards `ConsoleApp`?

modules/xcode/xcode_common.lua
tempura-sukiyaki: In xcode there is practically no way to add extensions at ConsoleApp.
Even if you put values in `WRAPPER_EXTENTION` or` EXECUTABLE_EXTENSION`, the product has no extension.

samsinsane: Oh ok, interesting. Thanks for the clarification!

src/base/config.lua
tempura-sukiyaki: It was the same reason I wrote above.
However, it is not good as it fails the test in this part.

src/_premake_init.lua
tempura-sukiyaki: I did not specifically think about `system: ios`.
Probably it is better to apply.

src/base/config.lua
tempura-sukiyaki: I did not specifically think about `system: ios`.
Probably it is better to apply.

src/_premake_init.lua
tempura-sukiyaki: Even if `_TARGET_OS` is set to `ios`, the generated project seems to be correct, but should I modify something by myself?
(e.g. `fliter { "system:macosx or ios", "kind:WindowedApp" }`)

PR #952, opened 2017-11-28
Add support for `targetextension` in xcode4

tempura-sukiyaki:
I'm sorry I closed the last pull request, even though you commented.

> Is there any particular reason why a `ConsoleApp` can't specify a `targetextension`? `app.out` for example?

There is no way to add extensions to `ConsoleApp` in Xcode.
Even if you put values in `WRAPPER_EXTENSION` or `EXECUTABLE_EXTENSION`, the product has no extension.

> Could this also apply to `system:ios`?

I did not specifically think about `system:ios`.
This can also apply to `system:ios`.


TurkeyMan:
LGTM, but I don't know about XCode...

samsinsane:
Why didn't I merge this...

PR #955, opened 2017-11-29
change objdir to use more unique variant

tvandijck:
This change the objdir for each configuration to use a more consistent and more descriptive value.
It allows for multiple platforms, architectures, toolsets and configurations to all build concurrently.

We've ran into issues where two machines building to the same network share clobbered each others folders with .o & .obj files, because when running premake on linux it no longer considers the windows configurations, and so generates the same output folders.

By creating a more 'unique' variant name, we fixed that issue.

### DIFF
--- a/contrib/luashim/luashim.c
+++ b/contrib/luashim/luashim.c
@@ -851,7 +851,7 @@ void shimInitialize(lua_State* L)
 
 	// Find the 'SHIM' entry in the registry.
 	const Table* reg = hvalue(&G(L)->l_registry);
-	const Node* n = findNode(reg, 'SHIM');
+	const Node* n = findNode(reg, 0x5348494D /* 'SHIM' */);
 	assert(n != NULL);
 
 	g_shimTable = (const LuaFunctionTable_t*)n->i_val.value_.p;
--- a/modules/codelite/tests/test_codelite_config.lua
+++ b/modules/codelite/tests/test_codelite_config.lua
@@ -16,6 +16,8 @@
 	local wks, prj, cfg
 
 	function suite.setup()
+		_TARGET_OS="linux"
+
 		p.action.set("codelite")
 		p.indent("  ")
 		wks = test.createWorkspace()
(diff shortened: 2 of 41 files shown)


TurkeyMan:
So... this seems fine, except that it deviates from VS default project output. These paths are nothing like VS would ever emit by default.
I think we need everyone's buy-in to make this change...

tvandijck:
oh I agree, it's somewhat controversial for sure... it just fixes a bunch of issues we've been seeing, and overall we've been using this particular 'schema' for the last 3 years with great success...

ratzlaff:
Here at the company I work for, we specify non-default output locations for every platform as well (for different reasons), the configuration to do so is just not baked into the premake5 executable.

I am not a fan of those directory names as default, but that is just a personal opinion. As long as I can still override the defaults, I would not expect my current premake5 configuration to be any different.

samsinsane:
Yeah, I'm not sure about this for defaults. For example, our SDK emits to `Lib` and `Lib/Debug`, then the files themselves are named based on things, `udSDK.vc$(PlatformToolsetVersion).dll`, or something like this. So, there's people out there who don't put things in folders and instead have 20 binaries for all the different platforms in a single folder.

Personally, I prefer the properly named approach. However, I feel like your solution is great for a package server, but in the event you've just dump the SDK into your repo, specifying a single `libdirs` entry isn't really possible? Windows has the toolset specified, Linux doesn't. I'd prefer to just specify a single entry, `%{os}-%{toolset}-%{platform}-%{configuration}` is nice and simple. (Assume each of those is the correct token.)

Also, would be nice to have a way to know what Linux distro the binary is built for, but that's probably outside the scope of this PR.

starkos:
I agree that the change is needed, but also don't like these defaults.

For object directories, I'd be happy enough if you changed `obj/windows-clang-debug` to `obj/Debug/Windows/Clang` to better match Visual Studio's defaults. Most people bootstrapping a new project aren't going to care too much about the obj folder tree, and those that have a standard will need to configure it themselves anyway.

For binary directories, I think the Premake default should really be doing the minimum necessary. Per the other comments, most people will already have opinions on how to configure this anyway.

Perhaps, for the binary directories, there is a way to better support company-wide settings? So someone could create a script like (untested):

```lua
-- Defaults for project binary outputs
targetdir 'bin/%{cfg.platform}-%{cfg.toolset}-%{cfg.configuration}'
```

…and place it somewhere in the environment and have it picked up by all developers, so it doesn't need to be manually added to every project? We have a system script and search paths now, but maybe there is a better solution?

starkos:
There seems to be consensus that better defaults would be good, but something other than proposed by this PR. Can this be closed, or would someone like to run with it?

tvandijck:
@TurkeyMan can you delete this branch from the Blizzard repo (Blizzard:variants)

PR #956, opened 2017-11-29
Usage proof-of-concept.

tvandijck:
This new 'using' API, and a slight addition to the storage of all properties, allows us to mark settings as either public, private or interface.

`using` then allows you to use import public & interface properties from other projects, effectively creating a way for a project to say... "if you use me, you need these settings".

This is currently in a proof of concept stage, and any feedback or additional unit-tests would be extremely helpful.

### DIFF
--- a/src/base/configset.lua
+++ b/src/base/configset.lua
@@ -310,6 +310,7 @@
 		local block = {}
 		block._criteria = crit
 		block._origin = cset
+		block._export = {}
 
 		if basedir then
 			block._basedir = basedir:lower()
@@ -360,16 +361,29 @@
 --    an error message.
 ---
 
-	function configset.store(cset, field, value)
+	function configset.store(cset, field, value, visibility)
 		if not cset.current then
 			configset.addblock(cset, {})
 		end
 
 		local key = field.name
 		local current = cset.current
+		local vis = visibility or 'private' -- private is the default behavior.
 
 		local status, result = pcall(function ()
-			current[key] = p.field.store(field, current[key], value)
+			if vis == 'public' then
+				-- public means both for outselfs and for those who use me.
+				current._export[key] = p.field.store(field, current._export[key], value)
+				current[key]         = p.field.store(field, current[key], value)
+			elseif vis == 'interface' then
+				-- interface means, only for those who use me.
+				current._export[key] = p.field.store(field, current._export[key], value)
+			elseif vis == 'private' then
+				-- private means just for me.
+				current[key] = p.field.store(field, current[key], value)
+			else
+				p.error('invalid visibility '%s', options are `private`, `public` & `interface`.', vis)
+			end
 		end)
 
 		if not status then
--- a/src/base/configset.lua
+++ b/src/base/configset.lua
@@ -493,6 +493,7 @@
 
 		local abspath = filter.files
 		local basedir
+		local kind
 
 		for i = 1, n do
 			local block = blocks[i]
@@ -511,7 +512,10 @@
 				table.insert(result.blocks, block)
 
 				if cset.workspace then
-					configset.processUsing(result, cset.workspace, block, filter)
+					if block.kind then
+						kind = block.kind
+					end
+					configset.processUsing(result, cset.workspace, block, filter, kind)
 				end
 			end
 		end
@@ -522,15 +526,15 @@
 		return result
 	end
 
-	function configset.processUsing(result, wks, block, filter)
+	function configset.processUsing(result, wks, block, filter, kind)
 		if block.using and #block.using > 0 then
 			for _, name in ipairs(block.using) do
-				configset.useExportsFromProject(result, wks, name, filter)
+				configset.useExportsFromProject(result, wks, name, filter, kind)
 			end
 		end
 	end
 
-	function configset.useExportsFromProject(result, wks, name, filter)
+	function configset.useExportsFromProject(result, wks, name, filter, kind)
 		local prj = p.workspace.findproject(wks, name)
 		if not prj then
 			p.error("Project '%s' is not in the solution '%s'.", name, wks.name)
@@ -551,11 +555,16 @@
 				newBlock._basedir = block._basedir
 				newBlock._origin  = result
 
+				-- links don't propagate onto static libraries.
+				if kind ~= p.STATICLIB then
+					newBlock.links = nil
+				end
+
 				-- insert resulting block
 				table.insert(result.blocks, newBlock)
 
 				-- recurse
-				configset.processUsing(result, wks, newBlock, filter)
+				configset.processUsing(result, wks, newBlock, filter, kind)
 			end
 		end
 	end
--- a/src/base/configset.lua
+++ b/src/base/configset.lua
@@ -556,7 +556,7 @@
 				newBlock._origin  = result
 
 				-- links don't propagate onto static libraries.
-				if kind ~= p.STATICLIB then
+				if kind == p.STATICLIB then
 					newBlock.links = nil
 				end
 
--- a/src/base/configset.lua
+++ b/src/base/configset.lua
@@ -482,6 +482,12 @@
 			result = configset.new()
 		end
 
+		-- tag ourselves as used.
+		result._imports = result._imports or {}
+		result._imports[cset] = true
+
+		-- now iterator through all blocks and extract only those
+		-- that apply to the current filter.
 		local blocks = cset.blocks
 		local n = #blocks
 
@@ -503,6 +509,10 @@
 
 			if criteria.matches(block._criteria, filter) then
 				table.insert(result.blocks, block)
+
+				if cset.workspace then
+					configset.processUsing(result, cset.workspace, block, filter)
+				end
 			end
 		end
 
@@ -510,4 +520,42 @@
 
 		result.compiled = true
 		return result
+	end
+
+	function configset.processUsing(result, wks, block, filter)
+		if block.using and #block.using > 0 then
+			for _, name in ipairs(block.using) do
+				configset.useExportsFromProject(result, wks, name, filter)
+			end
+		end
+	end
+
+	function configset.useExportsFromProject(result, wks, name, filter)
+		local prj = p.workspace.findproject(wks, name)
+		if not prj then
+			p.error("Project '%s' is not in the solution '%s'.", name, wks.name)
+		end
+
+		-- if it was already imported, skip.
+		if result._imports[prj] then
+			return
+		end
+
+		-- mark us as done.
+		result._imports[prj] = true
+
+		for _, block in ipairs(prj.blocks) do
+			if criteria.matches(block._criteria, filter) then
+				-- copy the block entirely.
+				local newBlock = table.shallowcopy(block._export)
+				newBlock._basedir = block._basedir
+				newBlock._origin  = result
+
+				-- insert resulting block
+				table.insert(result.blocks, newBlock)
+
+				-- recurse
+				configset.processUsing(result, wks, newBlock, filter)
+			end
+		end
 	end
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1229,6 +1229,13 @@
 	}
 
 	api.register {
+		name = "using",
+		scope = "config",
+		kind = "list:string",
+		tokens = true,
+	}
+
+	api.register {
 		name = "usingdirs",
 		scope = "config",
 		kind = "list:directory",
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1232,7 +1232,6 @@
 		name = "using",
 		scope = "config",
 		kind = "list:string",
-		tokens = true,
 	}
 
 	api.register {
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -11,11 +11,15 @@
 	local api = p.api
 	local configset = p.configset
 
-
+---
+-- Create the public&interface namespaces for the api.register functions.
+---
+	public = {}
+	interface = {}
 
 ---
 -- Set up a place to store the current active objects in each configuration
--- scope (e.g. wprkspaces, projects, groups, and configurations). This likely
+-- scope (e.g. workspaces, projects, groups, and configurations). This likely
 -- ought to be internal scope, but it is useful for testing.
 ---
 
@@ -277,10 +281,19 @@
 		end
 
 		-- create a setter function for it
-		_G[name] = function(value, visibility)
-			return api.storeField(field, value, visibility)
+		_G[name] = function(value)
+			return api.storeField(field, value)
 		end
 
+		public[name] = function(value)
+			return api.storeField(field, value, 'public')
+		end
+
+		interface[name] = function(value)
+			return api.storeField(field, value, 'interface')
+		end
+
+		-- TODO, do we need the ability to remove things from public&interface?
 		if p.field.removes(field) then
 			_G["remove" .. name] = function(value)
 				return api.remove(field, value)
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -277,8 +277,8 @@
 		end
 
 		-- create a setter function for it
-		_G[name] = function(value)
-			return api.storeField(field, value)
+		_G[name] = function(value, visibility)
+			return api.storeField(field, value, visibility)
 		end
 
 		if p.field.removes(field) then
@@ -484,7 +484,7 @@
 -- gets parceled out to the individual set...() functions.
 --
 
-	function api.storeField(field, value)
+	function api.storeField(field, value, visibility)
 		if value == nil then
 			return
 		end
@@ -507,7 +507,7 @@
 			error(err, 3)
 		end
 
-		local status, err = configset.store(target, field, value)
+		local status, err = configset.store(target, field, value, visibility)
 		if err then
 			error(err, 3)
 		end
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -36,6 +36,11 @@
 			prj.group = ""
 		end
 
+		-- projects using me, should link against me.
+		local f = p.field.get("links")
+		p.configset.store(prj, f, prj.name, 'interface')
+
+		-- return new project.
 		return prj
 	end
 
--- a/tests/_tests.lua
+++ b/tests/_tests.lua
@@ -42,6 +42,7 @@ return {
 	-- Baking tests
 	"oven/test_filtering.lua",
 	"oven/test_objdirs.lua",
+	"oven/test_using.lua",
 
 	-- API tests
 	"api/test_boolean_kind.lua",
--- a/tests/oven/test_using.lua
+++ b/tests/oven/test_using.lua
@@ -12,31 +12,29 @@
 -- Setup
 ---
 
-	local wks, prj1, prj2
-
 	function suite.setup()
-		wks = workspace("MyWorkspace")
-		configurations { "Debug", "Release" }
+		workspace("MyWorkspace")
+			configurations { "Debug", "Release" }
 	end
 
 	function suite.testPublicProperty()
 		-- create two projects.
-
-		prj1 = project "MyProject1"
+		local prj1 = project "MyProject1"
 			kind 'StaticLib'
-
 			includedirs {"private"}
 			includedirs ({"public"}, "public")
 
-		prj2 = project "MyProject2"
+		local prj2 = project "MyProject2"
 			kind 'ConsoleApp'
-
 			using { "MyProject1" }
 
 		-- test if prj2 contains the public includedirs, but not the private ones.
-		cfg = test.getconfig(prj2, "Debug")
-		test.print(table.tostring(cfg.includedirs))
+		local cfg = test.getconfig(prj2, "Debug")
 		test.isequal(#cfg.includedirs, 1)
 		test.contains({ path.getabsolute("public") }, cfg.includedirs)
+
+		-- it must also link against it.
+		test.isequal(#cfg.links, 1)
+		test.contains({ "MyProject1" }, cfg.links)
 	end
 
--- a/tests/oven/test_using.lua
+++ b/tests/oven/test_using.lua
@@ -1,0 +1,42 @@
+---
+-- tests/oven/test_using.lua
+-- Test the using and import of exported values.
+-- Copyright (c) 2014-2015 Jason Perkins and the Premake project
+---
+
+	local p = premake
+	local suite = test.declare("oven_using")
+	local oven = p.oven
+
+---
+-- Setup
+---
+
+	local wks, prj1, prj2
+
+	function suite.setup()
+		wks = workspace("MyWorkspace")
+		configurations { "Debug", "Release" }
+	end
+
+	function suite.testPublicProperty()
+		-- create two projects.
+
+		prj1 = project "MyProject1"
+			kind 'StaticLib'
+
+			includedirs {"private"}
+			includedirs ({"public"}, "public")
+
+		prj2 = project "MyProject2"
+			kind 'ConsoleApp'
+
+			using { "MyProject1" }
+
+		-- test if prj2 contains the public includedirs, but not the private ones.
+		cfg = test.getconfig(prj2, "Debug")
+		test.print(table.tostring(cfg.includedirs))
+		test.isequal(#cfg.includedirs, 1)
+		test.contains({ path.getabsolute("public") }, cfg.includedirs)
+	end
+
--- a/tests/oven/test_using.lua
+++ b/tests/oven/test_using.lua
@@ -22,7 +22,7 @@
 		local prj1 = project "MyProject1"
 			kind 'StaticLib'
 			includedirs {"private"}
-			includedirs ({"public"}, "public")
+			public.includedirs {"public"}
 
 		local prj2 = project "MyProject2"
 			kind 'ConsoleApp'


ratzlaff:
Instead of

    includedirs ({"public"}, "public")

What do you think of instead something like:

    configuration {"usage", "public"}
        includedirs ("public")

Reason being is that using `configuration {...}` now looks like any other scope-limited thing and the existing functions wont have to change.

tvandijck:
`configuration` is a deprecated API... 
`filter` should be used instead..

I guess, we could indeed do something like:
```lua
filter({ 'system:windows'}, 'public')
```

To mark a block as a public vs private block, which I think would be a little more memory efficient then my current implementation.

TurkeyMan:
In first impression, I felt like this was a cool improvement, although logically awkward... since I don't really think this usage has anything to do with 'filter'-ing.
Thinking on it more though, I think it's the case that, given a typical filter block, the properties won't all have matching exposure, and that will result in breaking typical filter blocks into multiple filter blocks, possibly leading to a torrent of filter blocks each with very few properties.
I suspect scripts would get very "filter" and indentation heavy... not sure that's a good thing.

tvandijck:
Just as an example, with this PR you can now write:
```lua
project "MyProject1"
    kind 'StaticLib'
    includedirs {"private"}
    public.includedirs {"public"}

project "MyProject2"
    kind 'ConsoleApp'
   using { "MyProject1" }
```

The result for MyProject2 is that it links against MyProject1, as well as takes the includedirs from MyProject1 that are marked as 'public'.
It becomes more interesting with deeper hierarchies, since things propagate, although again, `using` is just another API, so in order for a using to propagate to however uses that library, you'd write
```
public.using { 'MyProject1' }
```

etc...

TurkeyMan:
This is definitely superior to the original patch, and this improvement has taken wind out of the suggested "use filter" sails..

samsinsane:
Out of curiosity, why `using { "MyProject1" }` and not have `links { "MyProject1" }` perform the same action? It's how most things are currently linked anyway, or is that the reason? To ensure that people are aware of the change?

starkos:
FWIW, I personally prefer splitting out the "how to build" from "how to use". Tossing this out there as a discussion point.

```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'private' }

usage 'MyProject1'
    includedirs { 'public' }

project 'MyProject2'
    kind 'ConsoleApp'
   using { 'MyProject1' }  -- or `links { 'MyProject1' }`, per @samsinsane
```
 
A usage of this style can be written for a library that you use but don't build.

```lua
usage 'ThirdPartyLibrary'
   includedirs { '../vendor/thirdPartyLib/include' }
   libsdirs { '../vendor/thirdPartyLib/lib' } 
   links { 'thirdParty1", "thirdParty2' }

project 'MyProject2'
    kind 'ConsoleApp'
   links { 'ThirdPartyLibrary' }
```

Selfishly, this also lines up a little better with my own mental model of the configuration system, but I'm MIA so much these days I don't know if that's worth any consideration.

Thanks @tvandijck for putting this together and getting things going!

tvandijck:
@starkos Yeah, the problem I'm having with the usage container is that containers need to be named. They effectively have no relation at all to projects whatsoever. So you can't have a single property that applies to both the project itself, as well as to anyone who uses it.

In this implementation, there is three options... private, interface and public. Where public is effectively private+interface... all settings applied, apply to both the project itself, as well as anyone who uses the project. 
To accomplish something similar with the 'usage' pattern, you would have to make two usage containers, where one uses the other, and then use the private one in the project itself. like so

```lua
usage 'myproject_private'
   includedirs { 'private' }

project 'myproject'
    uses 'myproject_private'

usage 'myproject'
   uses 'myproject_private'
   includedirs { 'public' }
```

it's not an improvement..

All that aside, the concept introduced in this PR, is the same concept as CMake, so for those transitioning between the two, or trying to convince people that premake > Cmake, this is a familiar concept that is easy to explain.. "it's the same as cmake"... the Usage/Use pattern is quite a bit more complex.

All that said, I like the fact that you can make completely unrelated usage containers, and use those. particularly helpful for shared settings, but this PR does not prevent that from being added in addition. I think both have value. I just dislike that I have to make a usage container that has the same name as my project to get anything exported. It's also quite a bit more lines of code to export 1 setting, which is what ends up beeing the case 9 our of 10 times. My solution is 1 line of code, the usage pattern is at least 2, and probably 3 because of the separation syndrome, which needs an extra empty line.

ie
```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'private' }

usage 'MyProject1'
    includedirs { 'public' }
```
vs
```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'private' }
    public.includedirs { 'public' }
```

ratzlaff:
Conceptually, I prefer the clear separation between "how to build" and "how to use".

This PR seems much more flexible than the usage system we have implemented here at my company.

The convention we use is to declare the import/export symbols for dll linkage on windows explicitly, so we have filter blocks inside of using functions like so:
```
someProject = {
  configurations = {
    windows = {
      exports = {"PROJECT_DLL=__declspec(dllexport)"	}
    ,  imports = {"PROJECT_DLL=__declspec(dllimport)"	}
    }
  }
, using = function (inProject)
    configuration "windows"
      defines(inProject.configurations.windows.imports)
    configuration {}
  end
}
```
How would I configure a usage unique to windows with this PR?

tvandijck:
that example would collapse down to:

```lua
project 'someproject'
...
filter { 'system:windows' }
    defines {  "PROJECT_DLL=__declspec(dllexport)" }
    interface.defines {"PROJECT_DLL=__declspec(dllimport)"}
```

tvandijck:
> Conceptually, I prefer the clear separation between "how to build" and "how to use".

I don't disagree. but I don't like the "say everything twice" result of the 'usage' containers. 
Also, 'conceptually' I like everything that concerns my project to be in my project.
They are basically two competing opinions/concepts.

Again however, I think the 'usage' container stuff, is certainly something we could do in addition. It's not hard to imagine a world where both could work together to allow for even more flexibility. I think both patterns have their uses, and I can certainly imagine a case where a usage container makes more sense. I just don't see it for the general case.

src/base/configset.lua
```diff
@@ -360,16 +361,29 @@
...
 		end
 
 		local key = field.name
 		local current = cset.current
+		local vis = visibility or 'private' -- private is the default behavior.
 
 		local status, result = pcall(function ()
-			current[key] = p.field.store(field, current[key], value)
+			if vis == 'public' then
+				-- public means both for outselfs and for those who use me.
```
ratzlaff: Typo: 'outselfs'

ratzlaff:
> that example would collapse down to:

Nice! It was not obvious to me that the filter would play nice with `interfaces`

> They are basically two competing opinions/concepts

Totally valid, I do believe that good features come out of constructive discussion (like this) where you are able to effectively defend your position. With your API I can certainly see how one would go about using it to implement the more verbose case in my own scripts.

It appears that your PR now takes over the 'public', 'interface' and 'private' tables. What about putting those somewhere under the `premake` table?

tvandijck:
@ratzlaff we could indeed, you'd have to write premake.public.*** everywhere, where you currently don't have to use the premake prefix for everything, because everything goes simply into _G...

I can go either way, but I do agree that taking ownership of 'public' and 'interface', may cause issues. I'm open to suggestions..

ratzlaff:
>  I'm open to suggestions..

Let the end-user name it whatever they want, they know better whether or not 'public' is used for something else

```
local public = premake.public
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'private' }
    public.includedirs { 'public' }
```

echopoint:
Curious, why not just reverse it by making it look like an includedirs extension then?
i.e.
 `includedirs.public { 'public' }` 
vs 
`public.includedirs { 'public' }`

starkos:
One more try, then :) What about:

```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'build-only' }  -- i.e. "private", is the default

   filter { stage = 'build or link' }
      includedirs { 'build-or-link' } -- i.e. "interface"…or is that "public"?

   filter { stage = 'link' }
      includedirs { 'link-only' } -- i.e. …whatever the other one is
```

So yes, you still need the extra line for the filter. But build and usage are clearly separated, it handles the use cases of "interface" and "public", easily expands to new values, and it doesn't introduce any new syntax variations. I think that's the best I've got.

If you decide to take the field prefix/postfix approach, my poor brain would appreciate different names. I just looked at the code last night and I've already forgotten which is "private-public" and which is "public-public", plus the confusion of overloading existing programming language terms with very different meanings.

The term "build" makes sense to me. I'm not sure what word to use for "use" other than "link", even though you might not necessarily be linking?

```lua
build.includedirs { … }
buildAndLink.includedirs { … }
link.includedirs { … }
```

Anything worthwhile there?

starkos:
Should have mentioned: using `filter` would let you use the existing baking/context stuff to pull out the different settings for builds and usages, by setting “stage” to “build” or “link” (or whatever) when you do the fetch.

aleksijuvani:
The "private" and "public" prefixes look *really* janky. I like the syntax @starkos proposed above, but I'm not sure if calling them "stages" is very intuitive.

starkos:
@aleksijuvani: That's cool; can you suggest any alternatives? Maybe "phase"? (This is assuming @tvandijck agrees with you…)

aleksijuvani:
I feel like "stage" or "phase" suggests like these relate to a build stage, which is not the case..

Instead these are a bunch of settings that will have to be used (or not used) when building projects that "use" this project. Perhaps something like this would make more sense:

```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'build-only' }  -- i.e. "private", is the default

   filter "building or linking"
      includedirs { 'build-or-link' } -- i.e. "interface"…or is that "public"?

   filter "linking"
      includedirs { 'link-only' } -- i.e. …whatever the other one is
```

Or perhaps something like this would be more consistent with the other filters?

```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'build-only' }  -- i.e. "private", is the default

filter "building:MyProject1 or links:MyProject1"
  includedirs { 'build-or-link' } -- i.e. "interface"…or is that "public"?

filter "links:MyProject1"
  includedirs { 'link-only' } -- i.e. …whatever the other one is
```

aleksijuvani:
Actually the latter could have

```lua
filter "project:MyProject1 or links:MyProject1"
```

instead of

```lua
filter "building:MyProject1 or links:MyProject1"
```

tvandijck:
Both @TurkeyMan and my argument is that we're not filtering... we're exporting data... it's very unnatural and confusing to abuse filter for this...

besides, when your new query system makes the cut..
```
filter { 'project:foobar' }
```
would already allow filtering based on project..

Then the other confusing part to me is... "build", "link", are all stages... premake does not have a concept a phases, and it seems somewhat adhoc to add it now. besides, it's a concept that only C/C++ has... many other languages just don't have these or even similar stages... C# for example, the compiler gets the full list of .cs files, and simply spits out an executable.. not linking, nothing... just one stage.

Then, while I am totally open to alternative naming, "public", "interface" and "private" are concepts used in many many languages to express exported fields, properties, etc.. I mean, C#, C++, Python, you name it... In addition, it's a concept used in CMake. While we have as premake no chance at beating their momentum and becoming the thing CMake currently is to the C++ community, I feel that adding concepts that are completely different from CMake will just increase the barrier to entry. 

Anyway, then as my final point, which is no argument for or against any of this.. December 15, is my last day at Blizzard. I'll be leaving to move me and my family closer to the rest of my family, and will be starting a new adventure in Sweden. As a result, it is likely that my involvement with Premake will also come to an end, or at least in a much reduced capacity, since it goes from work to hobby, and well, I have 200 hobbies, so Premake will have to timeshare.

I'm not trying to push things through or anything, but these last 2 work weeks I won't be able to make additional commits to the PR's I've made so far. If further adjustments are needed/required, others will have to branch these off, and make those changes, as between December 15 and January 15 I'll be completely out of reach.

Anyway, it's been a fun ride these last 3 years. I'm totally convinced the premake > CMake in many aspects, but it also still needs a lot of work. I hope Blizzard stays involved, @TurkeyMan is not working for Blizzard, so he may have some time here and there to keep merging things upstream like I have in the past few years.

ratzlaff:
Thanks for all your contributions to premake @tvandijck Dont be a stranger!

aleksijuvani:
I don't like the weird object-oriented analogy. It's unnecessary, and it's not very intuitive to me that specifying that project A links against project B would cause A to "inherit" B. The prefix/suffix syntax isn't particularly pretty either, but I guess that's subjective.

The concept of filters already exists in Premake and it'd be trivial to extend them to support two new fields.

In any case, if you do decide to go this way, I think "interface" should be called "protected" instead. "interface" doesn't make sense in this context (at least in the C++ and C# land).

Good luck in your future endeavors, @tvandijck!

tvandijck:
> and it's not very intuitive to me that specifying that project A links against project B would cause A to "inherit" B

That's why it's not done through "links", but through another API. I don't like the conflated use of links either.

> The concept of filters already exists in Premake and it'd be trivial to extend them to support two new fields.

it does... to filter for stuff... not to export properties.. I never has had that ability, and in my opinion neither should it be extended for that purpose. The underlying block storage, certainly could be extended to provide this ability. It's a pretty natural extension, but at the same time, it would increase the number of blocks per project quite a bit, and hence have an impact on the compile performance of the blocks, but that is all besides the point.

filter is to filter for stuff, not for creating an export block.
containers have been used for this purpose in the past (see [this](https://github.com/Meoo/premake-export) module). But containers require a name, which again makes no sense at all for this use-case.

> The prefix/suffix syntax isn't particularly pretty either, but I guess that's subjective.

I welcome any alternatives. We've tried two different things in this PR, the prefix is defintely better then what we tried first `includedirs( {'...'}, "public")`
The suffix syntax isn't easy to accomplish, you can't make a table and a function in the same spot in _G, so you would have to do some funky trickery through a metatable. It wouldn't make things any faster.

aleksijuvani:
I see.

tvandijck:
So one of the alternatives to look at, and maybe experiment with, is still the container approach. but require the name of the container to be fixed... 

To stick with my previous examples, and naming:
```lua
project "MyProject1"
    kind 'StaticLib'
    includedirs {"private"}
    export 'public'
        includedirs {"public"}

project "MyProject2"
    kind 'ConsoleApp'
   using { "MyProject1" }
```

the downside beeing that a new container resets filters, so if you wanted to make things specific to platforms, you'd have to do:
```lua
project "MyProject1"
    kind 'StaticLib'
    includedirs {"private"}

    export 'public'
        filter { 'system:windows' }
             includedirs {"public/windows"}
        filter { 'system:linux' }
             includedirs {"public/linux"}
```

we can simply assert for the name of an export container to be either private,public,interface, or whatever other names are agreed upon. It's not that much of a diversion from the export module I linked to before. It's just that we assign semantics to the name of an export block, which I think is really the important part.

starkos:
> December 15, is my last day at Blizzard. I'll be leaving to move me and my family closer to the rest of my family, and will be starting a new adventure in Sweden.

Congrats, and best of luck on your adventure! we'll certainly miss your contributions and your insights! I certainly understand the bit about the hobby projects, but we'll leave your account turned on…just in case, you know. :)

TurkeyMan:
I'm way late to this party, but I just wanted to drop my thoughts :)

There was some discussion above about concepts like "build or link"; I don't think that's the right way to think about this feature. This is about projects that depend on other projects inheriting state from the dependencies. 'building'/'linking'/etc are C/C++/compiled-language concepts. Lots of languages aren't like that, and also/rather, lots of projects don't express a conventional build environment. It's common for things like utility projects to do something that's not part of a traditional 'build' process, and for that reason, I think it's important to make this inheritance of general state not expressed in those terms.
I think the proper conceptual structure here is "importing 'state' from dependent projects", or "exporting state to consumer projects", whichever way around makes more sense.

From that perspective, I like Tom's original proposal, that allows general state to be marked 'public' or 'interface' and then it will be available to consumers.
I personally like the 'public'/'interface' terminology. I find it intuitive, although I understand this is subjective, but I also support Tom's argument that it's established by cmake, and I think that's a strong argument to stick with established nomenclature.

We discussed this pattern:
```lua
project "MyProject1"
    kind 'StaticLib'
    includedirs {"private"}

    export 'public'
        filter { 'system:windows' }
             includedirs {"public/windows"}
        filter { 'system:linux' }
             includedirs {"public/linux"}
```
Other than leading to very deep nesting, I'm concerned that stateful api's kinda suck, and `filter` is already problematic in that way. `filter { ... }` much be followed by `filter {}` to reset to un-filtered... this state retention pattern will also exist for some other API that sets the visibility mode, and the cadence of filter changes and visibility changes will not match. This will either lead to very noisy code, or very difficult to understand code.

```lua
project "MyProject1"
    includedirs {"private"}

    visibility 'public'
        filter { 'system:windows' }
             includedirs {"public/windows"}
        filter { 'system:linux' }
             includedirs {"public/linux"}
        filter {} -- !!! be careful to reset !!!

    visibility 'interface'
        includedirs { 'public/api' } -- otherwise this line will still be filtered by linux, even though it doesn't look like it!
        filter { 'system:windows' }
             includedirs {"public/windows-api"}
        filter { 'system:linux' }
             includedirs {"public/linux-api"}
        filter {}
```

The indentation looks like python scoping, but remember, it's NOT, and it's a big gotcha for premake users that don't really think that they're writing a Lua program, which is a concept premake strongly tries to encourage.

The above, for contrast, according to the current PR:
```lua
project "MyProject1"
    includedirs {"private"}
    interface.includedirs { 'public/api' }

    filter { 'system:windows' }
         public.includedirs {"public/windows"}
         interface.includedirs {"public/windows-api"}

    filter { 'system:linux' }
         public.includedirs {"public/linux"}
         interface.includedirs {"public/linux-api"}
```

I know which one I find more readable...

TurkeyMan:
Also, we had a discussion about making separate containers for the public stuff... which seems cool, but we quickly realised that in practise, it would lead to a very high number of basically-empty containers, with maybe 1 or 2 options. The container count is a good heuristic to estimate our performance, which is a serious problem. So we concluded that adding heaps of superfluous containers might not be the best idea.

samsinsane:
@tvandijck Sad to see you go! I agree with @starkos, definitely going to miss your contributions and insight, not to mention gleaning random bits of knowledge off you, like I have over the last couple of years. Best of luck in Sweden!

starkos:
Wow, lots to process here! I'll respond to a few of the points that were batted my way, then I'll follow up with another post to refocus the discussion a bit.

> Both @TurkeyMan and my argument is that we're not filtering... we're exporting data... it's very unnatural and confusing to abuse filter for this...

For reference, here is what I proposed, with improved (?) naming:

```lua
project 'MyProject1'
   includedirs { 'private/include' }
   
   filter { when = 'building or using' }
      includedirs { 'private-and-public/include' }
      
   filter { when = 'using' }
      includedirs { 'public/include' }
```
   
Whether or not it is the best approach, it is a stretch to call it "abuse". Filters say "apply these settings when these conditions are met", and that's what we're doing: apply these settings when this project is built; apply these settings when it is referenced/used/linked/whatever.


> @tvandijck: Then the other confusing part to me is... "build", "link", are all stages... premake does not have a concept a phases, and it seems somewhat adhoc to add it now.

> @TurkeyMan: There was some discussion above about concepts like "build or link"; I don't think that's the right way to think about this feature.

That was me, and I made it clear that I didn't like the term "link" for "use" but needed something for the sake of example. I never suggested tying it to a toolset's compile/link stages. I've cleaned up my example to make that more clear.


> @tvandijck: while I am totally open to alternative naming, "public", "interface" and "private" are concepts used in many many languages to express exported fields, properties, etc.. I mean, C#, C++, Python, you name it...

I'll be more specific then: "public" belongs to the set of public, private, protected, internal/package. "interface" belongs to the set of interface, class, enum, struct. They are orthogonal concepts, and both of them could be construed to mean "public".

If you didn't know Premake, but knew software development, you could read this and pretty much know what it is doing:

```lua
project 'MyProject1'
    kind 'StaticLib'
    includedirs { 'include', 'build/include', 'vendor/include' }
    files { '**.h', '**.cpp' }
```

If you didn't know Premake, but knew software development, you couldn't interpret this without external documentation:

```lua
project 'MyProject1'
    kind 'StaticLib'    
    includedirs { 'vendor/include' }
    public.includedirs { 'build/include' }
    interface.includedirs { 'include' }
    files { '**.h', '**.cpp' }
```

We're trying to distinguish between "use these settings while building this project", and "apply these settings when some project tries to use/reference/link me". Whatever approach gets chosen, I'd like terms that make that clear without having to dig up the docs (and I'll admit my own first proposals weren't great).


> @tvandijck: In addition, it's a concept used in CMake

> @TurkeyMan: I also support Tom's argument that it's established by cmake, and I think that's a strong argument to stick with established nomenclature.

This keeps being brought up, so I guess it needs to be addressed. I personally find it a very weak argument. The right approach should stand on its own merits and not because someone else is doing it. That doesn't mean it is a bad approach or the wrong approach, and it is fair to point out the prior art, but this shouldn't be considered the compelling argument.

And now I have to point out: the people I've talked with who have switched away from CMake to Premake (and if you're listening, feel free to chime in) all did so because they they preferred Premake's syntax. Let's be honest: that is really Premake's biggest (and perhaps only) selling point against a project with as much momentum as CMake.

I do think you've both made your point, and now I'd really love to let that particular thread rest.

> I don't like the weird object-oriented analogy. It's unnecessary, and it's not very intuitive to me that specifying that project A links against project B would cause A to "inherit" B.

FWIW, I agree with this, it should be possible to link or reference a project *without* pulling in whatever settings it decided to "export". But I think we can make that work with any of these approaches, we just need a separate call for uses/imports. An example purely to demonstrate what I mean for discussion:

```lua
project 'MyProject'
   links { 'SomeLibrary' }  -- just links
   imports { 'OtherLibrary' } -- links and imports settings
```


> @TurkeyMan: we had a discussion about making separate containers for the public stuff... which seems cool, but we quickly realised that in practise, it would lead to a very high number of basically-empty containers, with maybe 1 or 2 options. The container count is a good heuristic to estimate our performance, which is a serious problem.

A valid point. I don't think we should choose one approach over another on performance alone (I'm actively trying to reduce the per-block cost in another effort), but it shouldn't be discounted.

Okay, one last thing:

> @TurkeyMan: Other than leading to very deep nesting, I'm concerned that stateful api's kinda suck, and filter is already problematic in that way. filter { ... } must be followed by filter {} to reset to un-filtered... this state retention pattern will also exist for some other API that sets the visibility mode, and the cadence of filter changes and visibility changes will not match.

His example, note the comments:

```lua
project "MyProject1"
    includedirs {"private"}

    visibility 'public'
        filter { 'system:windows' }
             includedirs {"public/windows"}
        filter { 'system:linux' }
             includedirs {"public/linux"}
        filter {} -- !!! be careful to reset !!!

    visibility 'interface'
        includedirs { 'public/api' } -- otherwise this line will still be filtered by linux, even though it doesn't look like it!
        filter { 'system:windows' }
             includedirs {"public/windows-api"}
        filter { 'system:linux' }
             includedirs {"public/linux-api"}
        filter {}
```

So first: umm...Premake makes pretty heavy use of stateful, declarative APIs, yeah? But putting that aside, the "fix" for this is trivial: `visibility` should reset the filter, just like any other container (whether it is a container or not). Then you no longer need the `filter {}` call.

FWIW, I don't think approach offers anythig over the prefix or filter-based approaches, but I'll get to that in a follow-up.

starkos:
The story so far. Please use your imagination to scale these minimal examples up to meaningful project sizes, with additional filters for platforms and all of the stuff you would normally need to script.

Please keep in mind that the names of things can be changed.

When expressing a preference, it would be helpful if you could provide some explanation. And if you don't like these names, please try to provide some alternatives you *do* like.

### Option A: Prefixes

```lua
project "MyProject1"
    includedirs {"private"}
    public.includedirs {"public-and-private"}
    interface.includedirs { "public" }
```

### Option B: Filters

```lua
project "MyProject1"
    includedirs {"private"}

    filter { "when:building or using" }
      includedirs { "public-and-private" }

   filter { "when:using" }
      includedirs { "public" }
```

### Option C: Exports

```lua
project "MyProject1"
    includedirs {"private"}

   export "public"
      includedirs { "public-and-private" }

   export "interface"
      includedirs { "public" }
```

In all cases, we'll assume the values are "used" with a syntax like:

```lua

project "MyProject2"
    kind 'ConsoleApp'
   using { "MyProject1" }
```

Discuss!

starkos:
At this point, I'm still in favor of the filters (option B). 

- I like that it follows the existing "apply these settings when…" model and doesn't require any new concepts.
- I like that you can say "building or using" and don't need a synthetic term to describe settings that are used for both.
- I think the implementation will end up being cleaner under the hood, since it can use the existing filtering code to separate out "building" from "using", rather than requiring an additional code path to figure out which is which.
- I think the performance hit of the additional settings blocks can be easily mitigated with an early out test against the `when` clause, skipping any further criteria checking.

My second choice is prefixes (option A):

- I do like the brevity of it.
- I really dislike the names "public" and "interface", which [I find almost impossibly unintuitive](https://github.com/premake/premake-core/pull/956#issuecomment-349313516), but I'm really struggling to come up with meaningful alternatives. With better names I would like this approach more (but probably still less than Option B).

Exports (option C) feel like a half-measure. It is more verbose than prefixes, but less flexible than the filters. I'd rather see one of the other options chosen.

ratzlaff:
The big thing for me about this change is about how much 'new' stuff I'll have to learn, read documentation on and teach others about when they have questions here locally. 

Like: Option B
- No new keywords, extends `filter` with more options
- Reads nicely

Dislike: Option C
- New keyword `export` would probably be deprecated in favor of `filter`
- Otherwise the same as B

Dislike: Option A
- I really don't like having two competing ways of using a premake function. I could be wrong, but I expect this Option to require far more documentation than Option B.
- Premake scripts are not written in the style of `table.function`
```python
prj = project "MyProject1"
prj.includedirs {"private"}
prj.public.includedirs {"public-and-private"}
prj.interface.includedirs { "public" }
```

samsinsane:
Oh boy, this PR is out of control haha! Thanks for compiling (heh) the list of options @starkos!

Option A:
 - I vote against this
 - I feel introducing "namespaced" APIs will result in more being added, "we have them for this, why not for this?" and before we know it we have `dotnet.framework` and `macosx.frameworks`. While there's some benefits to making it obvious that some APIs are just very specific to one thing, if we overhaul things (i.e. csharp -> dotnet for C# and F#) we can end up destroying most projects that are targeting/using this one thing. Having said that, the example that @ratzlaff presented for `prj` actually looks pretty sweet but I'm sure there are caveats to that system that wouldn't be overly desirable. However, it would allow for something like `prj.filters["windows"] = filter { "system:windows" }` and then you could add sub filters `prj.filters["windows"].filters["debug"] = filter { "configuration:debug" }`. I think that's kind of cool, but it's definitely a totally different direction from where we are now.
 - I find it weird that there's three concepts "private", "public" and "both", and that `public` is "both" and `interface` is "public". No idea what would be "better", avoiding needing a third term is my "better idea".

Option B:
 - I like this for the simplicity and consistency
 - I dislike this for the verbosity that will likely end up being generated. `filter { "when:*" }` is better than some "both" terminology that doesn't quite make sense. However, if you have `platforms` specific, `configurations` specific or `system` specific settings, or a combination of these, for "public", "private" and "both", you end up with a mess of filters. `#platforms * #configurations * #system * 3` - worst case, but most of our projects end up with a filter for each `system` at the very least.

Option C:
 - I dislike that there's a new API
 - I dislike the "public", "private" and "both" terminology - see point in A.
 - Do I `filter` and then `export` or do I `export` and then `filter`? Is there a way to achieve both depending on how the user wishes to do this? Being forced to duplicate something because it only works one way is kind of lame, but achieving both ways is probably "impossible".

Overall, I think Option B could do well and I vote for it. I think our `filter` system needs a bit of an overhaul, complex filter queries having to be duplicated for "private", "public" and "both" will be annoying. Sub-filters could assist with reducing this, but that kind of leads into Option C with a new kind of API or just a complete rewrite of how `filter` works - both require learning which is not overly desirable.

tvandijck:
I guess I'll add my pros/cons vote myself...

## Option A:
- I clearly like the simplicity of the approach,
- I'm not a fan of the prefix/suffix, but am open to suggestions.
- I'm totally open to better names... while I agree that "because CMake" is not a major argument, I choose these names because rather then inventing new ones, I went with something established.

## Option B: 
- I have no objections, and it's a super tiny change to the current PR. "when" has to be dealt with a little special, and not be included in the final criteria I think, but it's pretty easy to get that to work.
- I don't like the name "building", it's not what we're doing, the concept of building is an unknown concept to premake... we're just generating a makefile/project after all. An alternative could be "internal" or "local".. 
Besides, "local" is really only there to help you from repeating yourself:
```lua
project "MyProject1"
    includedirs {"private"}

    filter { "when:building or using" }
      includedirs { "public-and-private" }
```
vs
```lua
project "MyProject1"
    includedirs {"private", "public-and-private"}

    filter { "when:using" }
      includedirs { "public-and-private" }
```
I have no objection to just dropping "local/building/private" entirely.

##  Option C:
is not that far off from Option B, it just puts a container in the middle except for a filter.
I see it as basically the same as Option B, since in both Option B and C you have to repeat all your filters for the different platforms/configs/etc. Which in Options A isn't the case.

tvandijck:
> I feel introducing "namespaced" APIs will result in more being added, "we have them for this, why not for this?" and before we know it we have dotnet.framework and macosx.frameworks

@samsinsane If only I could do api.register("namespace.name")... I mean, I've been wanting that so badly for a long time...

For example, just in the above discussions, we seen the use of `import`, `use`, `usage`, `using`... If any of those would be used by premake core... at least 5 modules that we currently list would break, because of reuse of the same api keywords. It's a flaw in the extensible of premake. Api's should be registered in the modules namespace... and only core premake API's should be in _G.

It's an entirely different discussion, way off topic... but since you brought it up...

TurkeyMan:
> visibility should reset the filter, just like any other container (whether it is a container or not). Then you no longer need the filter {} call.

This seems obvious, but then this is impossible:
```lua
    filter { "system:linux" }
        visibility "private"
            api "blah"
        visibility "interface"
            includedirs { "whatever" }
```

I guess I just presumed you would also be allowed to specify a filter, and then nominate visibilities within the applied filter.
Resetting the filter when visibility changes will lead to noisier scripts, and I suspect most users would presume this should be possible.

TurkeyMan:
Responding to @starkos 3 possibilities.
A - I like it because of it's brevity, and that it doesn't lead to indentation spam. I don't personally find it unintuitive.

B - My issue is that filters are already used for filtering. Your example is over-simple, this is what I predict will happen in reality:
```lua
project "MyProject1"
    includedirs {"private"}

    filter { "system:linux" }
        includedirs {"sys/linux"}
    filter { "system:windows" }
        includedirs {"sys/win"}

    filter { "when:building or using" }
      includedirs { "public-and-private" }
    filter { "when:building or using", "system:linux" }
      includedirs { "sys/linux/public-and-private" }
    filter { "when:building or using", "system:windows" }
      includedirs { "sys/win/public-and-private" }

   filter { "when:using" }
      includedirs { "public" }
   filter { "when:using", "system:linux" }
      includedirs { "sys/linux/public" }
   filter { "when:using", "system:windows" }
      includedirs { "sys/win/public" }
```
My point is, there are some sets of filter criteria that already to a a kind of combinatorial explosion. I argue that conflating 'visibility' into filtering will almost always lead to combinatorial filter statement explosion.

C - I just don't like it at all; I don't really feel it's superior to either A or B, and introduces additional issues, like visibility resets filter, meaning you can't nest changes of visibility beneath an active filter, which will lead to unnecessarily noisy scripts.

tvandijck:
@TurkeyMan I'm with you, but I think your example is also a bit of an exacuration. 

It's an example that is not happening very often in practice... If we look at the 'packages' we have in Blizzard I'd say the majority only `export src/include`, because the header files are generally fairly platform independent.... I think the only exception I can find of the top of my head is curl, because it uses some stupid autoconf to generate headers specifically for windows/linux/mac. But even that one if properly engineered could just be merged with a couple of #ifdefs to select the right file for the right platform. 

So while I agree that in a few cases there is certainly a possibility for filter explosions, the reality is that we're exporting very few things and rarely do we have a need to export based on platform/system. We almost never export based on configuration, since those are defined at the workspace level, and are rarely consistent between all projects.

The majority of exported items at least in our cases are going to be looking like this:

```lua
project 'foobar'
    kind 'staticlib'
    includedirs { '../include' }
    files { '**.cpp', '../include/**.h' }

    filter {'when:using'}
       includedirs { '../include' }
       links { 'foobar' }
```

I could be wrong obviously... but I think we certainly need to prioritize this discussion on the practical and stay away from hypothetical. I think internally we have a lot of practical examples we can look at.
image

TurkeyMan:
Regarding A; recall that a prior incarnation of this PR went with this:
```lua
   includedirs({ "include", "dirs" }, "interface")
```
As an alternative to `interface.includedirs()`. I don't think that's superior, but just to remind that the approach does have some flexibility in the API design, and maybe further alternatives that people don't hate are possible...?

samsinsane:
> @TurkeyMan I'm with you, but I think your example is also a bit of an exacuration.

@tvandijck I would have to disagree here. We don't use a package server and what Manu has suggested is pretty close to what I'm expecting to see for some of our projects. Curl and SDL2 are wild, Linux uses the standard apt/yum packages, macOS uses a framework, Windows builds curl and has the SDL2 binaries, iOS does something, and the particular project I'm thinking of doesn't hit Android yet. I'd love to move to a package server when I get time to look into one (yours looks nice, it's a custom one isn't it?).

Similarly, our SDK has a unique `targetname` for each system, platform and toolset combination, it also will dump debug builds into the "debug" folder beside the release binaries. This project was setup and configured long before Manu introduced Premake into the company, changing everything to be "Premake compatible" might not be possible as we ship this to our customers.

I don't know much about how your packaging system works, but I'm going to assume that it eliminates most of the system/platform specific magic. For example, our SDK could be `libudSDK.a` for Android x86 and ARM, Linux x64 (CentOS) and Linux x64 (Ubuntu), this can't really be configured without filters, again because it was setup and configured prior to Premake, so those concepts didn't exist. This might be unique to our usage, and I'm happy to have our stuff be ugly because of this, but it seems like it would exist elsewhere if it exists here.

> @samsinsane If only I could do api.register("namespace.name")... I mean, I've been wanting that so badly for a long time...

@tvandijck I see what you mean, having the modules not polluting the global namespace would be nice. However, when things move from specific to global, there's dramas to deal with there. For example, VS2015+ are getting more and more concepts that would originally be considered part of the xcode or gmake modules. Specifically, VS2015+ targeting iOS requires an xcode project/workspace (because reasons) and the `frameworks` concept would likely end up being required for VS2015+. Having said that, I'm sort of coming around to the idea, but I still have reservations about diving into it.

tvandijck:
@samsinsane the majority of our "packages" just build from source... we don't really do much binary distributions, and while I agree that for binary distributions the filters might be somewhat heavier, in all of SC2/WoW & Overwatch I think we use about 6 of them... and not because they need to be binary, but just because things like OpenSSL are just stupidly complicated buildscripts, DirectX, PS4SDK, etc are just not supplied in source form, or in cases like the chromium SDK, where it would take about an hour to build where it's just not practicle...

curl, I get what you are saying, but like I said, we just build it from source, it takes about 10 seconds, and once it's build it generally doesn't build again locally, so it's not a big deal compile time wise, and because it builds from source, there is no differences in targetnames, etc... those are just not important settings, premake resolves all those automatically anyway when you are talking about project to project dependencies.

In you case, the libudSDK.a if it was a package that just build from source, why do you care where it is, and what it is called... you just do `'links { 'SDK' }` and let premake figure out what to link, where it is, and what to call it. We generally don't care, as long as the final executable is in the right spot.

tvandijck:
@samsinsane 

>  (yours looks nice, it's a custom one isn't it?).

it is, but it's the simplest ASP.NET thing in the world... it looks fancy and all, but in the end it just has a few http endpoints that premake talks to, and that is it...

you can see it all here:
https://github.com/blizzard/premake-packagemanager

and what we're hoping is that with this PR, that entire system just goes away mostly, except for that part where we download a zip file and unpack it, and load it's premake scripts. Right now the whole meta-data system is effectively a system to accomplish the same as what this PR provides natively to projects. It would simplify our stuff a lot.

TurkeyMan:
I would also add that I don't think I'm exaggerating; we have large solutions here with a lot of projects, but we don't support many platforms, and we tend to use the same build config across all projects in the solution (at least on Overwatch).
Scripts tend to get noisy with lots of platforms, or where the build has more platform-specific details than we have, and I've worked on a lot of projects like that in the past.

As I've said in person though, I think a good solution to that problem in general is introducing a filter stack with push/pop-filter support. But that's totally separate development/debate ;)

samsinsane:
> the majority of our "packages" just build from source...

Oh ok, that's interesting.

> In you case, the libudSDK.a if it was a package that just build from source, why do you care where it is, and what it is called...

Long story, but source access to the renderer (basically the SDK) isn't generally available within the company. Perhaps this is unique to my company, and I'm already looking into binary package servers for all the things, so it won't be a huge issue for us.

tvandijck:
If binary distribution is your default... https://www.conan.io/ is the shizzle ;)

tvandijck:
rebased against master, no additional changes.

noresources:
I am very interested to see that implemented in premake. It's the missing feature I am waiting to continue my work on writing build recipes for some common libraries I am using.
As many, I like the filter approach because it doesn't add new API and respect the general writing style of premake files.

@tvandijck I hope your new adventure in Sweden is going well and you will find some time to continue your great work on premake.

tvandijck:
As this PR is based out of the Blizzard org branches, I have to pull these over to my own fork, close this PR and make a new one. Here at DICE we're not using premake, so this will have to become a hobby kind of thing. My computer at home is in a container somewhere on the ocean right now, so I won't have access to that until mid february... so please have a little patience.

samsinsane:
@tvandijck glad to see you resurface! Hope everything has been going well, and you're settling in at DICE. That's a shame about your computer, hope it all goes well! I'm sure you'll have convinced DICE to use Premake before mid February though, so you'll probably be working on this again much sooner. 😛

TurkeyMan:
It's disappointing this has stalled... let's revisit with fresh feelings on the matter.
Tom presented a few API options; there appears to be no consensus. Have anyone's opinions changed in the meantime? Maybe we can agree on a path forwards?
Perhaps some creative alternative options?

ratzlaff:
Just throwing ideas out there. No idea what it takes to implement something like this:

```lua
project "MyProject1"
    includedirs {["usage"] = "private", "src/include"}
    includedirs {["usage"] = "public", "include" }

project "MyProject2"
  using "MyProject1"
  -- adds src/include from MyProject1 to includedirs
```

noresources:
Not a new idea but another thought about the already presented approaches. 
To me, it is important to stick to the premake "look and feel" and avoid multiple way of doing things. My coding style will naturally go in favor of option A, but in the premake world, the filter way (option B) seems more accurate. 

@TurkeyMan, when you said
> My point is, there are some sets of filter criteria that already lead to a kind of comabinatorial explosion. 

Don't you think a part of this combinatorial could be solved with some sort of dedicated functions. Even if a premake script is purely declarative, Lua can still be used for specific/complex tasks.

In addition, I think the usage API is mainly dedicated to people who write libraries and want to provide a build script to other. The final purpose is to ease the library integration task for the "end user". I have nothing against something complex under the hood if I can use it simply. (I say that as both a "library writer" and a "end user")

samsinsane:
@starkos I think option B is good, but as @TurkeyMan points out, the filter system gets out of control once you hit `platform`, `system` and `configuration` specific requirements for the different values of `when`.

aleksijuvani:
Revisiting this now, I was initially in favor of option B (filters), but I think I'm leaning more towards option C (exports) now. I initially disliked how there was a new API, but if the combinatorial explosion of filters is truly an issue, it does look like the more elegant solution.

There was some sort of scoping issue pointed out with option C (exports), but I feel like the issue is avoided if `export`s work in filter scope, and apply to the project that the filter is in.

Did we ever reach a consensus on the naming of `export`'s arguments? I think we were at "public" (using), "private" (local) and "interface" (local and using)?

tvandijck:
I think the filter system itself can be improved to alleviate that issue as well...
just a few example options:

### Idea 1.
while it's not scoping, we could at least allow adding to the current filter scope. It's not hard to amend the terms of a preexisting filter.. we could even just store it as a tree.
```
-- add something to the current filter.
filter { '*', 'system:windows' }

-- override the filter as usual.
filter { 'system:windows' }
```

### Idea 2.
push/pop - we can assert, or just pop all when we switch to another container.
```
beginfilter { 'system:windows' }
endfilter()
````

Either way, while I agree that filter is currently somewhat of a nightmare, there certainly is ways to improve that issue on it's own, completely separate from the usage ideas. Maybe we should open a new "issue" about this one in particular.

**Comment by aleksijuvani** [comment]
I initially commented here about starting to prefer option C (exports) over option B (filters), but after hitting "send" I realized that you'd still have to repeat the export for every filter you have.

I agree that the filter system could be improved, but I also feel like this is a separate issue and shouldn't necessarily be what's holding back this PR.

**Comment by starkos** [comment]
In agreement that improving the filters would be a good thing regardless of what's decided here. FWIW, I've tinkered with an approach like:

```
filter ({ "configurations:Debug" }, function()
	-- do some stuff
end)
```

**Comment by WorldofBay** [comment]
Reading this PR and seeing how long it's been without a final decision and without an approach I as user like I'll add my opinion.

From those three options provided i prefer option A because the other two don't feel right while A is just overly verbose for a lot of things in the same "visibility" which doesn't happen that often.

Option B which seems to be the most popular amongst you fell down for me right away. The reason is the way I see filters as a user:
When project A has a filter, that filter affects project A.
In option B project A has a filter that **can't affect project A at all**, because project A can't import itself.

At the same time a filter in project A is enclosed by the project and doesn't affect other projects.
In option B the filter in project A **affects every project linking to A**.

I also dislike opening another scope as in option C, mainly because that scope has to end somewhere. I fell into that hole when adopting rules where i have to call "solution()" after every rule to ensure the rule scope is left.
I don't want to have to add a "project()" whenever i want to ensure I'm clear of visibilities.

I liked starkos' approach of the interface pattern: splitting public interface from implementation. Imo it would fit imports but that pattern has downsides as mentioned. One downside is that you have to say things twice, another downside is the reusability: say you have two base projects using the same stuff, needing to export the same stuff but being disjunct.

The solution would be my suggestion: do the interface pattern but complete and fully included in premake syntax:
```lua
interface "public_base"
  includedirs "public"

project "base1"
  interfaces "public_base"
  includedirs "private1"

project "base2"
  interfaces "public_base"
  includedirs "private2"
```
This option would take a syntax already established in premake (see rules), it would not need things said twice as any of the other suggestions used, it could work both with filters inside the interface block and inside the project offering the interface and if anyone comes up with a name suited better than 'interface' it could as well be self-explanatory.

If i get that right, "public" in the original suggestion meant "me and all that use me", "interface" meant "only those that use me" and private "only me".
With my option the "private" would stay as-is, any existing would be just equally valid and would not need any assumption of a "default". "public" would be everything in the 'interface' block, again without any further pollution of existing code (you can simply drag&drop your old code into the interface and it works).
The third value "interface" is possible with a syntax like
```lua
project "base1"
  interfaces { extern = "public_base" }
```
, which could also be extended to use the fields 'intern' for private reuse and 'export' for public use
or
```lua
project "base1"
  interfaces "public_base:extern"
```
or
```lua
project "base1"
  externalinterfaces "public_base"
```
etc

I don't know how this would work out internal but from the user point of view I think it is a great way of doing it.
Again, the name "interface" for that is suboptimal, it works in the declaration of the "interface" as many people will recognize it as a shared information but it would not use in the project as it doesn't look like an export.
Thinking about that the "interfaces" call in the project could also be "export_interfaces".

WorldofBay:
The "interface" keyword works in workspace scope, the "interfaces" statement works in interface and configuration scope.

starkos:
Getting back to this, much belatedly. I've given it a good think and [written up a new proposal](https://github.com/premake/premake-core/issues/1346) that I *think* solves all the pain points of the approaches presented here (including mine).

I'm going to go ahead and close this PR for now, and see if the discussion brings us back. Many thanks to @tvandijck for taking the time to put this together, and for spurring the best discussion on the topic so far. :thumbsup:

PR #959, opened 2017-11-30
Multiple gmake2 fixes

tvandijck:


modules/gmake2/_preload.lua
```diff
@@ -15,6 +15,7 @@
 		trigger         = "gmake2",
 		shortname       = "Alternative GNU Make",
 		description     = "Generate GNU makefiles for POSIX, MinGW, and Cygwin",
+		toolset         = "gcc",
```
TurkeyMan: How is this default limited to linux?

tvandijck: it's not.... but I can't make it optional, and not setting it will result in the toolset beeing 'nil' when using gmake... unless you're project sets it... in which case this change is moot anyway...

In my opinion it's better to have a default, then to have 'nil', but if you want to make this conditional on some other setting instead please feel free to make this even better... as long as nil doesn't show up...

filter { 'toolset:gcc' }  currently doesn't work on linux, even when using gmake/gmake2, and even though gcc is indeed the default because there is stupid code everywhere that does `toolset = toolset or 'gcc'`

TurkeyMan: Would it be better to put something like
```
filter { "system:linux" }
   toolset "gcc"
```
somewhere in bootup?

tvandijck: That is indeed the alternative for linux.... for any other platform the action would still be broken.

TurkeyMan: I agree... this should be merged, but I lines like those above should probably also exist for windows/osx/freebsd, etc.

PR #965, opened 2017-12-05
Move part of the workspace baking up

tvandijck:
@starkos, can you think of any reason why this wouldn't be correct.
I don't see any code in the oven.bakeConfigs that actually depends on any output from projects or the bakeChildren.
I might be wrong, but I'd like to make this change if at all possible?

### DIFF
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -105,6 +105,13 @@
 
 		self.location = self.location or self.basedir
 		context.basedir(self, self.location)
+	
+		-- Build a master list of configuration/platform pairs from all of the
+		-- projects contained by the workspace; I will need this when generating
+		-- workspace files in order to provide a map from workspace configurations
+		-- to project configurations.
+
+		self.configs = oven.bakeConfigs(self)	
 
 		-- Now bake down all of the projects contained in the workspace, and
 		-- store that for future reference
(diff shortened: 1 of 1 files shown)


tvandijck:
So parts of the bakeFiles, I want to refactor (or I already did here) to do the `compilebuildoutputs` flag in place, rather then as a step after the p.workspace.bake. And as part of that refactor I want to add the ability for those generated files to go into another project. This would be easy if `configmap` wasn't a thing, but it is, and so I need to map from project config to the others project config, and the only way to do that is by mapping both from the workspace config... Unfortunately without this change, the workspace configs don't exist yet when baking the projects.

So moving this up, bakes the workspace configs first, allowing you to enumerate them during the project bake.. 

I may not make much sense, PR incoming..

tvandijck:
This loop in particular wouldn't work without this change: 
https://github.com/premake/premake-core/pull/968/commits/4da400e0a138bf75eb03b1d58fde93783cfefe99#diff-47294235a21ef45b5104ae2328145578R615

starkos:
Again, I think this looks fine. I'm hitting "Update branch" to catch it up to master, but feel free to merge when it's ready (if I don't do it first).

tvandijck:
mac is always so far behind... linux + windows are good.

samsinsane:
> mac is always so far behind...

I know right! There's only a couple of hours a day where there isn't a backlog of macOS builds - they also *love* performing maintenance on those things all the time.

tvandijck:
Yeah, I wonder if there is an alternative, or a way to provide them with hardware for specific jobs... I don't mind setting up a mac here specifically for those builds, so we get somewhat quicker turn around...
AppVeyor is always super quick, and the Travis Linux builds are generally OK too, but I guess iOS development is just super popular, and hence always extremely backlogged.

PR #979, opened 2017-12-19
Define _HAS_EXCEPTIONS=0 from vs2010 for exceptionhandling off

tdesveauxPKFX:
Not defining _HAS_EXCEPTIONS=0 when ExceptionHandling is off cause warning.
`warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc`

Ref: (VSINTALL)\VC\include\xstddef

### DIFF
--- a/modules/vstudio/tests/vc2010/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2010/test_compile_settings.lua
@@ -574,20 +574,6 @@
 <ClCompile>
 	<PrecompiledHeader>NotUsing</PrecompiledHeader>
 	<WarningLevel>Level3</WarningLevel>
-	<Optimization>Disabled</Optimization>
-	<ExceptionHandling>false</ExceptionHandling>
-		]]
-	end
-
-
-	function suite.exceptions_onNoExceptionsVS2013()
-		exceptionhandling "Off"
-		p.action.set("vs2013")
-		prepare()
-		test.capture [[
-<ClCompile>
-	<PrecompiledHeader>NotUsing</PrecompiledHeader>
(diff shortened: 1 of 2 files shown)


starkos:
Thanks for this! You will also need to run the unit tests (`premake5 test`) and fix the ones that are failing because of the new flag.

tdesveauxPKFX:
My bad, totally forgot. Should be good now.

Also I took the time to track the commit that added the vs2013 condition.
5a4ad51aecdb4b9bec95cd769be59fc6e61f9c07
It was made by @tvandijck , we may want to have his opinion on this?

samsinsane:
@tdesveauxPKFX He's currently unavailable for the next month(?) perhaps longer.

@TurkeyMan do any of your projects build with VS2010 still? Additionally, do you guys disable the warning 4530? That number looks really familiar, and I'm assuming that's because it was disabled here (probably by you).

Just to provide a little more insight into the reason this was attached to VS2013+, for those that don't already know. VS2013 had an issue that resulted in warnings regarding `throw`/`noexcept` in various C/C++/STL headers, and if I recall correctly, these warnings couldn't be disabled - or it just failed to disable them. So using this was much easier, I only came across this issue *after* upgrading to VS2013, our VS2010 builds didn't suffer from the 4530 warning (probably because it was disabled). I'm going to assume everyone just disabled that warning and moved on, and then VS2013 reared it's head.

starkos:
I'm not at all familiar with this issue, I'll leave it up to @samsinsane and @TurkeyMan to make the call on this one.

samsinsane:
Let's merge it! It resolves an issue, and makes the code less "condition-y" and that's just perfect :+1:

PR #993, opened 2018-01-10
Fix /usr/lib64 being used as search dir before project's libdir

### LINKED ISSUE #992: Premake is generating makefile with -L/usr/lib64, leading to link errors
Hi,

On Linux, Premake's behaviour is to add ``-L/usr/lib64`` before any other linking path when compiling for x86-64, as  written and test (for instance) in the following file : 

https://github.com/premake/premake-core/blob/ef9b341d6f9cb1208d60d2dcdea399adf2014b8b/src/tools/gcc.lua#L356

https://github.com/premake/premake-core/blob/3f225eb7f2c7db81060a6141d31e0d4e10fec3a4/modules/gmake2/tests/test_gmake2_ldflags.lua#L59

The behaviour is the same on gmake and gmake2, according to the generated makefiles.

However, it means that if you have a library installed in ``/usr/lib``, a symlink between ``/usr/lib64`` and ``/usr/lib`` and you try to compile this library and an executable linking to this library, it will fail to link to the correct library. See the following issue for details : 

DigitalPulseSoftware/NazaraEngine#149

It is worth adding that even a simple premake5.lua file with the ``Architecture x86-64`` option will generate a flawed makefile. The real issue comes from the following behaviour of ``-L`` option in gcc (extracted from the man page):

```
Order does matter when you use several options of the same kind; for example, if you specify -L more than once, the directories are searched in the order specified.  Also, the placement of the -l option is significant.
```

I believe that premake should follow the same behaviour as in #543 and NOT include the ``-L/usr/lib64`` flag. However, if you believe you should add it, it should at least be injected as the last library search directory.

alexandre-janniaux:
Move the /usr/lib64 added by the compiler script (src/tools/gcc.lua) after every other -L
declaration.

The other fix is to completely remove the -L /usr/lib64 entry. See issue #992 for details.

fixes #992


alexandre-janniaux:
It might be possible to write a test to assert this behaviour doesn't happen by, for example, writing a libm mock, compiling it into a library, compile an executable and link against libm and finally check that the function in the program runs the mock behaviour and not the libm behaviour.

starkos:
Updated to latest master branch, just waiting for the checks to pass…

alexandre-janniaux:
Thank you for the quick integration of this fix!

Issue #935, opened 2017-10-28
vs2017 action results in solution that targets Windows SDK 8.1


JohannesMP:
On windows 10 with a c++ project, using `premake5 vs2017` will result in a solution that targets the Windows SDK 8.1, not the new 10.0.* SDK Versions.

Example error message when trying to build a bog standard 'Hello World' C++ solution created with Premake5 (compiled from source on commit 66362c0 ):

`1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Platforms\Win32\PlatformToolsets\v141\Toolset.targets(34,5): error MSB8036: The Windows SDK version 8.1 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".`

This will always happen on a machine that only has VS2017 installed, without the 8.1 SDK, which is the default state. On such a machine, the 'Retarget Projects' right click option shows the following:
![](https://i.imgur.com/Yflukv7.png).

While this is a quickfix, I feel that the default behavior for the `vs2017` action should be to use the 10.0 SDK that is guaranteed to be installed, not the 8.1 SDK that will become less and less used as more devs upgrade to using only 2017.

We should not assume that someone running the `vs2017` action has VS2015 or the older SDK installed.

------------

Edit: I just saw #450 and the addition of the [`systemversion`](https://github.com/premake/premake-core/wiki/systemversion), so at least I have a nicer workaround now with `systemversion("10.0.15063.0")`, but I still feel that it would be a more reasonable default than 8.1.

dcourtois:
I have the same problem. And even worse in my opinion is that bootstraping Premake with vs2017 doesn't work : dev /upgrade will not update the SDK version which will still point to 8.1 :/

Basically doing this will not work:

```
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat"
nmake -f Bootstrap.mak windows
```

JohannesMP:
If it helps, I ended up adding this to my premake5.lua file: 

````lua
function os.winSdkVersion()
    local reg_arch = iif( os.is64bit(), "\\Wow6432Node\\", "\\" )
    local sdk_version = os.getWindowsRegistry( "HKLM:SOFTWARE" .. reg_arch .."Microsoft\\Microsoft SDKs\\Windows\\v10.0\\ProductVersion" )
[...]

PR #994, opened 2018-01-16
added a 'latest' systemversion for vs2017

dcourtois:
Sorry for the delay, but here is the pull request for issue #935 and the other related ones.

You can now specify `systemversion "latest"`. If the action is `vs2017`, it will query the registry to retrieve the latest SDK 10 version number. If anything's wrong (action less than vs2017, the registry does not exist, etc.) no system version will be written in the vcxproj.

There are 2 basic unit tests, but I couldn't run them, tests currently fail on `test_http.lua` on my setup...
I tested this on my personal project, and also to build Premake itself using vs2017, and it works fine (just needed to add `systemversion "latest"` on the various Premake scripts for the Windows config.

### DIFF
--- a/modules/vstudio/tests/vc2010/test_globals.lua
+++ b/modules/vstudio/tests/vc2010/test_globals.lua
@@ -253,3 +253,48 @@
 </PropertyGroup>
 		]]
 	end
+
+--
+-- Check that the "latest" systemversion works.
+-- note: we override the os.getWindowsRegistry method for reliable tests.
+--
+
+	function suite.windowsTargetPlatformVersionLatest_on2017()
+		p.action.set("vs2017")
+		systemversion "latest"
+		prepare()
+		local oldRegistry = os.getWindowsRegistry
+		os.getWindowsRegistry = function (key)
+			return "10.0.11111"
+		end
+		test.capture [[
+<PropertyGroup Label="Globals">
+	<ProjectGuid>{42B5DBC6-AE1F-903D-F75D-41E363076E92}</ProjectGuid>
+	<IgnoreWarnCompileDuplicatedFilename>true</IgnoreWarnCompileDuplicatedFilename>
+	<Keyword>Win32Proj</Keyword>
+	<RootNamespace>MyProject</RootNamespace>
+	<WindowsTargetPlatformVersion>10.0.11111.0</WindowsTargetPlatformVersion>
+</PropertyGroup>
+		]]
+		-- restore the registry function
+		os.getWindowsRegistry = oldRegistry
+	end
+
+--
+-- Check that the "latest" systemversion will not add <WindowsTargetPlatformVersion>
+-- when the action is less than 2017
+--
+
+	function suite.windowsTargetPlatformVersionLatest_on2015()
+		p.action.set("vs2015")
+		systemversion "10.0.10240.0"
+		prepare()
+		test.capture [[
+<PropertyGroup Label="Globals">
+	<ProjectGuid>{42B5DBC6-AE1F-903D-F75D-41E363076E92}</ProjectGuid>
+	<IgnoreWarnCompileDuplicatedFilename>true</IgnoreWarnCompileDuplicatedFilename>
+	<Keyword>Win32Proj</Keyword>
+	<RootNamespace>MyProject</RootNamespace>
+</PropertyGroup>
+		]]
+	end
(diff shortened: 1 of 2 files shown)


modules/vstudio/tests/vc2010/test_globals.lua
```diff
@@ -253,3 +253,48 @@
...
+	end
+
+--
+-- Check that the "latest" systemversion will not add <WindowsTargetPlatformVersion>
+-- when the action is less than 2017
+--
+
+	function suite.windowsTargetPlatformVersionLatest_on2015()
+		p.action.set("vs2015")
+		systemversion "10.0.10240.0"
```
samsinsane: Is this supposed to be `latest`?

dcourtois: Ah ! You're totally right ! Is there a way to run specific tests ? (Couldn't run the whole test suite because of some other tests failing before this one)

tvandijck:
I'm with @samsinsane here...

tvandijck:
Yeah, I ran into similar issues with some of the http test failures... it's kind of why I added the --insecure option.... I run my tests locally with that option, so the ssl tests pass... it's often because of some firewall stuff in corporate organizations...

dcourtois:
Ok I could run the tests, and I have a problem: I need to override `os.getWindowsRegistry` and restore to be able to return a testable version number, but the code I'm using is not working. I'm a bit rusty in Lua, do you have any idea why it's not working ? And any way to achieve this ?

modules/vstudio/tests/vc2010/test_globals.lua
```diff
@@ -253,3 +253,48 @@
...
+
+--
+-- Check that the "latest" systemversion works.
+-- note: we override the os.getWindowsRegistry method for reliable tests.
+--
+
+	function suite.windowsTargetPlatformVersionLatest_on2017()
+		p.action.set("vs2017")
+		systemversion "latest"
+		prepare()
```
samsinsane: `prepare` is what will call `os.getWindowsRegistry`, so you'll need to put this just above the `test.capture`

samsinsane:
> Hi, like I said in the first comment

Woops, sorry about that, I completely forgot that you mentioned the tests failing in your original comment. I was referring to the failed builds with appveyor and travis, in my head it made sense at the time, sorry for the confusion!

> do you have any idea why it's not working ? And any way to achieve this ?

I've added a comment for this, that *should* fix it.

dcourtois:
@samsinsane thanks, that was it ! some other tests still fail on my machine, but at least appveyor seems to be happy now :)

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2484,6 +2491,15 @@
...
 	end
 
+--
+-- Get the latest installed SDK 10 version from the registry.
+--
+
+	function m.latestSDK10Version()
+		local arch = iif(os.is64bit(), "\\WOW6432Node\\", "\\")
+		local version = os.getWindowsRegistry("HKLM:SOFTWARE" .. arch .."Microsoft\\Microsoft SDKs\\Windows\\v10.0\\ProductVersion")
+		return iif(version ~= nil, version .. ".0", nil)
```
samsinsane: This took me a little bit to figure out but these tests fail on Linux and macOS because `version .. ".0"` is being evaluated here. To resolve this issue, you'll need to break this out into a proper if statement.

samsinsane:
@dcourtois No worries, glad it sorted it out, just noticed another issue when I checked the Travis builds. I've added a comment for that, hopefully that's the last thing! I'm sure you're just as keen to merge this as I am at this point. 😄 Thanks for doing this!

dcourtois:
My worst PR ever, thanks for the help :D Fingers crossed ...

samsinsane:
Everything looks good now! We'll merge this when the mac build finally runs - seems like it's time for TravisCI to buy some more macs haha.

tvandijck:
That could be days ;) Their mac builds are so overloaded all the time, it often takes 4-5 hours for that queue to clear... I wouldn't wait... if linux is passing, bsd/mac will pass too on this one, there is no difference in codepath for this particular test between the two...

samsinsane:
@tvandijck you raise some good points. Merging.

dcourtois:
Now I can suggest to update the Premake and modules' scripts to use `systemversion "latest"` to finally be able to bootstrap Premake using vs2017 :p

PR #1003, opened 2018-02-12
Xcode 9.2 -std=c++17, MACOSX_DEPLOYMENT_TARGET

joeriedel:
Xcode 9.2 doesn't work correctly with premake5 generated projects when settings c++17. Also premake wasn't using systemversion() on macosx targets.

### DIFF
--- a/modules/xcode/xcode_common.lua
+++ b/modules/xcode/xcode_common.lua
@@ -963,6 +963,11 @@
 			if family then
 				settings['TARGETED_DEVICE_FAMILY'] = family
 			end
+		else
+			local minOSVersion = project.systemversion(cfg)
+			if minOSVersion ~= nil then
+				settings['MACOSX_DEPLOYMENT_TARGET'] = minOSVersion
+			end
 		end
 
 		--ms not by default ...add it manually if you need it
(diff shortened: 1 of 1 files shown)


samsinsane:
Hmm, this is a tricky change. Firstly, given that `c++11` isn't supported in earlier versions of XCode but `c++0x` is, I have a feeling that this is a very similar situation - perhaps it's time to start making new actions for XCode for these breaking changes? Secondly, you'll need to update the unit tests *and* add unit tests for `systemversion`. In the future, it would be better if these unrelated changes were split into their own PRs, while your C++17 changes are potentially contentious, your `systemversion` changes are not and are just going to be held up while we discuss what to do with C++ support in XCode.

pdoane:
There are similar discussions going on for the Visual Studio side.  I think the best answer here is that 'c++11' only maps to a compiler that claims full compliance and not to 'c++0x, so this seems like a change in the right direction but it needs to be done universally.  Additional options for 'c++0x' would make sense as well.

starkos:
Seeing as the systemversion() change has already been merged by #1337, I'm going to go ahead and close this one. I agree with @pdoane that this fits into the large discussion for handing the "fluid" toolset releases we're seeing these days.

PR #1005, opened 2018-02-16
Fix gmake dependencies

### LINKED ISSUE #981: Makefile target dependencies aren't correct
I've noticed that the make output has lots of errors with rule dependencies, and this leads to problems with multi-core builds (ie, `make -j2` or more)

Consider:
```make
all: prebuild prelink $(TARGET) | $(TARGETDIR) $(OBJDIR)
	@:
```
What this says is that `all` depends on `prebuild` `prelink` and `$(TARGET)`, but it misses the hierarchy of dependency.
Truth is, `all` depends on `$(TARGET)`, which depends on `prelink`, which depends on `$(OBJECTS)`, which may-or-may-not depend on `$(PCH)`, which depends on `prebuild`.
We need to restructure the dependency resolution into the proper dependency structure, otherwise multi-threaded doesn't work.

Currently, `-j3` may cause `prebuild`, `prelink`, and `$(TARGET)` to all be built at the same time in parallel, which is wrong, since each step may depend on successful completion of the prior steps.

TurkeyMan:
Attempt to fix #981 

Note, this patch doesn't implement any logic to prevent execution of the prebuild step if the project is not dirty... it will ALWAYS run prebuild, just like gmake output always has, but the deps should now be structured that parallel builds work.

### DIFF
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -267,7 +267,7 @@
 ---------------------------------------------------------------------------
 
 	function gmake2.phonyRules(prj)
-		_p('.PHONY: clean prebuild prelink')
+		_p('.PHONY: clean prebuild')
 		_p('')
 	end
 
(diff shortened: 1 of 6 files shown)


tvandijck:
Make sure you discuss this change with @bwhittle (at World of Warcraft).
He did most of the PCH related fixes earlier last year.

Other then the test failures, the change looks fine to me.

TurkeyMan:
This has been tested now on a large project.

TurkeyMan:
If someone wants to bless, this makes parallel builds work with make.

TurkeyMan:
Sure, I can approve til the cows come home (I created this github account!) :P
But I totally wanted someone else to give this one a good once-over, since it feels pretty dangerous to me.

PR #1011, opened 2018-02-27
Fixes an issue with custom rules tracking files.

Dandielo:
Using 'Build Rules' would sometimes cause the tracker ignore changes in a source file, because of the tracker file being in the wrong directory.


tvandijck:
Seems like a good change to me.

PR #1018, opened 2018-03-10
Added release info badges to the readme

samsinsane:


starkos:
These are cool, thanks! Do we need to do anything to keep them up to date? If so, can you add the necessary steps to `scripts/RELEASE.txt`, so I don't forget them?

neico:
@starkos you might want to visit https://shields.io/ to see how the "API" works,  
in other words, most of it is grabbed and probably cached for a sane amount of time from the repo directly.

The only thing that might need adjustments is the actual "commits since last tag release", unless you opt to tag the latest release with a special tag that'll get moved (remove & add) with every tag. (as changing tags *shouldn't* affect git history)

starkos:
> The only thing that might need adjustments for each release is the actual "commits since last tag release", unless you opt to tag the latest release with a special tag

If that is something that needs to be done, can you add it to `scripts/RELEASE.txt`?

neico:
I assume @samsinsane can include this in this PR, as it's a rather minimal change as well (in case you were talking to me and not him)

samsinsane:
Hey, sorry, was AFK for a couple weeks. As @neico has already indicated, the Shields.io API does everything for us except for the "commits since" shield. We can alternatively use `latest` instead of `v5.0.0-alpha12` or whatever it is, but to my understanding `latest` only works with proper releases and not pre-releases like we're doing. The repo they use for the example latest doesn't have a tag "latest" and it seems to only be GitHub keeping track of which release is "latest", so it looks like it's entirely dependent on GitHub-isms.

FWIW this is what the latest shield looks like: [Commits since last release]()

I'll update the RELEASE.txt with this information when I get a chance!

starkos:
I don't mind setting a "latest" tag (in addition to the version tag) if that helps?

samsinsane:
I don't think it's a "latest" tag, I think it has to be a release where you don't tick the "pre-release" box. No idea to be honest, you can try pushing a git tag "latest" and see if that badge I commented updates?

samsinsane:
@starkos RELEASE.txt changes are ready for you to look over, I'm more than happy to change it if you want.

PR #1019, opened 2018-03-16
Remove git submodule commands

Passw:
It seems that there is no git submodule any more


starkos:
Good catch, thanks for this!

samsinsane:
This is a fair enough change, personally I would have left them in. I'd like to see some of the external things become submodules instead, given that our premake5.lua for "contribs" can exist anywhere and modify which files are included, we *could* have working examples of how you "premake-ify" external projects you don't control. We might have modifications to some config file, but we can just exclude the provided ones and include our required ones. We would probably need GitHub to eventually fix the Releases source package to pull in submodule source code too, since it's pretty useless without it.

PR #1021, opened 2018-03-19
Clear additional global state between unit test runs

starkos:
Clears out some global state that was being left over between unit tests runs, causing indeterminate results when running multiple tests against data at the global scope.


starkos:
FWIW, I have no idea why the Travis macOS test has failed…I'm unable to reproduce the problem here on my macOS machine.

tvandijck:
The macOS test failed because of one other 'order' dependency... so this commit is a step in the right direction, but I think there is some other global state that has a somewhat rare effect on that particular macOS test.

either way... LGTM... merge it...

PR #1023, opened 2018-03-26
Include library directories in Codelite projects

Gaztin:


samsinsane:
@tvandijck you removed this [here](https://github.com/premake/premake-core/commit/21bcefe892788449eef0991fd082aec4a359077a#diff-a7a4daf0118f250a58a977de622a37df), are you able to recall why?

tvandijck:
I tried to get Starcraft 2 working in CodeLite on Linux for Deepmind, so the changes I made are based on that attempt. 
@TurkeyMan would have to confirm if Blizzard is still using this and object to these changes or not.

starkos:
Assigned @TurkeyMan as a reviewer.

Gaztin:
Yeah it's correct in my use case, which is just a typical C++ workspace with one application project and a couple of static library projects that the application uses.

It wasn't finding the library files, so I overrode the linker function and added the same changes as in this PR. Afterwards, it worked fine!

Issue #1032, opened 2018-04-03
vstudio static lib pdb


tdesveauxPKFX:
Hi,

I have a use case where we redist static libs and we want to redist pdbs for debug builds.
For this, we want to output the pdb next to our lib but the symbolspath api does not meet this use.

symbolspath will add a `ProgramDatabaseFile` element to the linker block, however staticlib projects use the librarian who does not use `ProgramDatabaseFile`.

What we do is add a `ProgramDatabaseFileName` element in the compile block for our static libs with the desired path. In the end, we have a pdb containing the needed debug infos.

This solution does not feel that "clean" to me so I'm not sure you'd want this in Premake.

If you are interested in a pull request that implement this behavior in the vstudio module, I'd be happy to submit one.

starkos:
I have no objection to this approach, and off the top of my head am not aware of another workaround. +1 from me for a PR!

PR #1033, opened 2018-04-03
vstudio staticlib symbolspath

tdesveauxPKFX:
ref: #1032 

Also fixed names from tests I added in #1030

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -313,7 +313,7 @@
 --
 
 	m.elements.clCompile = function(cfg)
-		return {
+		local calls = {
 			m.precompiledHeader,
 			m.warningLevel,
 			m.treatWarningAsError,
@@ -348,6 +348,12 @@
 			m.callingConvention,
 			m.languageStandard,
 		}
+
+		if cfg.kind == p.STATICLIB then
+			table.insert(calls, m.programDatabaseFile)
+		end
+
+		return calls
 	end
 
 	function m.clCompile(cfg)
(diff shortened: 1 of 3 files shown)


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2175,14 +2181,15 @@
 		end
 	end
 
-
 	function m.programDatabaseFile(cfg)
 		if cfg.symbolspath and cfg.symbols ~= p.OFF and cfg.debugformat ~= "c7" then
-			m.element("ProgramDatabaseFile", nil, p.project.getrelative(cfg.project, cfg.symbolspath))
+
+			local key = iif(cfg.kind == p.STATICLIB, "ProgramDataBaseFileName", "ProgramDatabaseFile")
```
tvandijck: Personally, I would prefer this to be split into two methods, one for "ProgrameDatabaseFileName", and one for "ProgramDatabaseFile"..

While I get that they will be almost identical except for their name, it allows for easier override functionality...

tvandijck:
Thank you for all the added tests, much appreciated.

modules/vstudio/vs2010_vcxproj.lua
tdesveauxPKFX: That was my first implementation but that would duplicate the condition and would be a pain to maintain IMO.
But I get your point about override, if you feel that's the way to go, I have no objection.
I will make the changes later today.

tdesveauxPKFX:
This solution sounds good to me, I will implement it later today when I have some time.

PR #1037, opened 2018-04-07
HLSL Shader compilation in Visual Studio

tdesveauxPKFX:
Visual Studio can compile HLSL shaders with the FxCompile block.

As far as I know, this is available at least from vs2012. I did not test for earlier versions.

There is one thing bothering me with this implementation, I had to create shaderdefines and shaderoptions even though the resulting tag is the same as ClCompile.
This is to avoid doing something like
```
filter { "files:**.hlsl" }
   defines { "SOME_DEFINE" }
```
as it would be treated as a per-file config instead of a block one.

If you have any idea to improve this, I'm up to implement it.

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -274,6 +274,7 @@
 		else
 			return {
 				m.clCompile,
+				m.fxCompile,
 				m.resourceCompile,
 				m.linker,
 				m.manifest,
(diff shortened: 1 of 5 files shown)


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -696,13 +733,52 @@
 		end
 	}
 
+---
+-- ClCompile group
+---
+	m.categories.FxCompile = {
+		name	   = "FxCompile",
+		extensions = { ".hlsl" },
+		priority   = 3,
```
samsinsane: Does this need to be priority 3?

tdesveauxPKFX: Not really.
I implemented it some time ago so I don't remember it that clearly but it might be that vs add this block just after the ClCompile block so I wanted to reproduce this.

I can update this to keep the previous priority order if you'd prefer.

samsinsane: Fair enough, I see no reason to hold the PR up over this. What you have is fine.

PR #1039, opened 2018-04-08
Various new API

tdesveauxPKFX:
This is what we added to Premake where I work and I found that these have merits to be merged.

These might not be implemented across all possible tools/actions.


tdesveauxPKFX:
Sure, I won't be able to do this today but I will take care of this later this week.

samsinsane:
Sounds good, I look forward to going over them!

PR #1040, opened 2018-04-09
Get 'compileas' working when using a 'filter "files: ..."' scope.

ratzlaff:
Extends `compileas` to be used in file filters
```lua
  project "main"
      language "C++"

      files  { "src/file.c" }
      filter { "files:src/file.c" }
          compileas "C++"
```

### DIFF
--- a/modules/d/actions/gmake.lua
+++ b/modules/d/actions/gmake.lua
@@ -73,6 +73,18 @@
 		end
 	end)
 
+--
+-- Don't create $(OBJDIR) in the d-file rules
+--
+
+	p.override(make, "objDirInFileRules", function(oldfn, prj, node)
+		-- node is nil when making pch rules
+		if not node or not path.isdfile(node.abspath) then
+			oldfn(prj, node)
+		end
+	end)
+
+
 
 ---
 -- Add namespace for element definition lists for p.callarray()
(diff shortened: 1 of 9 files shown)


modules/d/tests/test_gmake.lua
```diff
@@ -105,6 +105,11 @@ endif
 		test.capture [[
 $(OBJDIR)/blah.o: blah.d
 	@echo $(notdir $<)
+ifeq (posix,$(SHELLTYPE))
+	$(SILENT) mkdir -p $(OBJDIR)
+else
+	$(SILENT) mkdir $(subst /,\\,$(OBJDIR))
+endif
```
samsinsane: This seems to be an unrelated change? I think this is how it used to be but it was causing issues when running multiple jobs.

tvandijck:
LGTM, I'm deferring to someone else to review/test this however.

samsinsane:
To be honest, everything except for the gmake changes LGTM. I don't understand enough about the gmake generator to know if something will break - I'll mark this as approved, but hopefully @starkos or @TurkeyMan can chime in on the gmake changes. (For what it's worth, the gmake2 changes make sense!)

ratzlaff:
@tdesveauxPKFX rebased as requested!

PR #1042, opened 2018-04-09
Add UnsignedChar API

tdesveauxPKFX:


samsinsane:
Could Clang, MSC, VS, XCode and CodeLite support this too?

tdesveauxPKFX:
I thought clang used shared flags from gcc but while it does it's not as I though so yeah I will implement it.

I will have to look for the others.

samsinsane:
@tdesveauxPKFX yeah it's a tricky one, *some* things aren't supported on Clang. Best example is `floatingpoint "Strict"`, but `floatingpoint "Fast"` is, which was super annoying when we didn't have a `Default` option for it.

samsinsane:
Amazing stuff! Thanks for adding in support for all of the other actions and toolsets!

PR #1044, opened 2018-04-09
Add Visibility API for gcc/clang toolsets

tdesveauxPKFX:


tdesveauxPKFX:
It seems that Clang copy all cxx flags from gcc [here](https://github.com/premake/premake-core/blob/cb1f642485c4c4d3f39f0a12f47161d48051a057/src/tools/clang.lua#L96).

I also added the other possible values from what running `gcc -fvisibility=x` displayed.

And I'm wondering about removing `-fvisibility-inlines-hidden` from this. It could be misleading in the present state.

tdesveauxPKFX:
From what I read, it seems indeed that `visibility-inlines-hidden` can be useful on it's own only.
I will setup an other API for it.

samsinsane:
Sounds good!

tdesveauxPKFX:
Inlines visibility moved to another PR (#1049)

tdesveauxPKFX:
@tvandijck why did you close this?
Although I move inlines visibility to another PR, the visibility API is still useful.

samsinsane:
I'm going to assume this was an accident.

tvandijck:
No, not an accident, just confused..  It said "Inlines visibility moved to another PR (#1049)", so I assumed this one was no longer needed...

At the very least the PR should be renamed to describe what it is now doing though, considering it no longer "Adds VisibilityHidden to flags"...

PR #1045, opened 2018-04-09
Add debuggerflavor API

tdesveauxPKFX:


modules/vstudio/_preload.lua
```diff
@@ -21,6 +21,19 @@
...
+
+	p.api.register {
+		name = "debuggerflavor",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"WindowsLocalDebugger",
+			"WindowsRemoteDebugger",
+			"WebBrowserDebugger",
+			"WebServiceDebugger"
```
samsinsane: These could simply be:
```lua
"Local",
"Remote",
"WebBrowser",
"WebService",
```
I would recommend not using the VS element values for the API values. VS changes the values sometimes and you'll end up needing a lookup table anyway, others adding support for VS extensions might need to use different values too.

modules/vstudio/tests/vc2010/test_debug_settings.lua
```diff
@@ -101,3 +101,22 @@ foo=bar</LocalDebuggerEnvironment>
...
+		test.capture [[
+<DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
+		]]
+	end
+
+	function suite.debuggerFlavor_OnWindowsRemote()
+		debuggerflavor "WindowsRemoteDebugger"
+		prepare()
+		test.capture [[
+<DebuggerFlavor>WindowsRemoteDebugger</DebuggerFlavor>
```
samsinsane: Does this just set the "default" debugger?

tdesveauxPKFX: Not sure what you mean here.

samsinsane: I think this is a good start for fleshing out the debugger options. If I get a chance I might try to add the remote values. :+1:

Issue #936, opened 2017-10-29
Improve systemversion on Visual Studio


redorav:
After reading through https://github.com/premake/premake-core/pull/450, I thought we might just be stuck by Visual Studio not supporting different Windows versions as targets, but I tinkered about with the vcxproj and it seems like this can indeed be done, it's just that by default Visual Studio seems not to make it available through their UI (that, or it's a bug for them too)

The **WindowsTargetPlatformVersion** property is generally found in the Globals group, but can be modified in the Configuration, so conditionally setting it like the following

```xml
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
    <ConfigurationType>Application</ConfigurationType>
    <UseDebugLibraries>true</UseDebugLibraries>
    <PlatformToolset>v141</PlatformToolset>
    <CharacterSet>MultiByte</CharacterSet>
    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
</PropertyGroup>
```

works well. I tested this in a simple project that included the DX12 header files. In one configuration it compiles, in the other (8.1) it doesn't. So, essentially, we need to remove the warning on vs actions that warn against using systemversion out of scope and include this. Thanks!

Issue #1038, opened 2018-04-08
Question about Android in premake


redorav:
I was looking at the recent merge where Android support is added into premake, and quite excited about it so I went and tried it out. I created a workspace like

```lua
filter { "platforms:Android32"}
		system("android")
		toolset("clang")
		architecture("arm")
```

and the project is

```lua
project "projectname"
	kind("StaticLib")
	language("C++")
```

~~This creates a solution but the result is not the same as going to the New Project .. Cross Platform .. Native Activity Application in Visual Studio. It doesn't create a .androidproj or packaging file and also points to the Clang 3.8 compiler which is reported as not installed for me. I think this is because the ApplicationType is not set to Android in the vcxproj.~~

~~Am I doing something wrong? After exploring a bit more it seems android support in premake actually needs a Visual Studio plugin? Am I wrong in thinking this?~~

After playing a bit with it, it seems I need to add system("android") at project scope and I can't do it at workspace scope under a filter("platforms:PlatformName") condition. This is probably a bug. I've also managed to figure out kind("androidproj") is used to create the .androidproj file that builds the apk. I've gone and added that to the wiki.

samsinsane:
> After playing a bit with it, it seems I need to add system("android") at project scope and I can't do it at workspace scope, even unconditionally.

Do you mean:
```lua
workspace "workspace"
  system "android"
```
or
```lua
system "android"

workspace "workspace"
```

I do the former in a personal project and it seems to work fine, but I'm still using the module separately. Looks like you figured out the other issues, sorry for the confusion, I should have updated the readme when I added VS2015 support.

There's still more to clean up with this module, and if I can get a chance, I'll add *some* `gmake`/`gmake2` support too - I can't replicate the `androidproj` behaviour, but in my incredibly limited experience, running ant was enough, though it required a change or two in one or more of the ant files. I haven't used the Gradle version yet, so I have no clue what's involved there.

samsinsane:
Yeah, that's a limitation with VS. It requires a project to be classified as Android, which means it can't target any other platform at all. In the `vcxproj`, it's part of the "project" elements rather than any configuration block, like ARM/ARM64/x86/x64 would be.

redorav:
I wonder if this is a little bit like my other question https://github.com/premake/premake-core/issues/936 where it looks like it's a global thing because VS doesn't do it through the UI but it can actually be done. I'll experiment a bit further.

samsinsane:
[...]

PR #1048, opened 2018-04-10
Add conditional behavior to global variables, add systemversion as first implementation

redorav:
The UI can't do this but setting it manually actually works. Fixes https://github.com/premake/premake-core/issues/936


redorav:
Sure, I'm all for having a discussion about it. To me it doesn't sound so much "unsupported" as "not exposed" in the sense that there seem to be many things the .vcxproj can do that aren't in the UI. I totally see your point that there's a kind of risk in exposing these kinds of things, but my feeling is it looks more like a limitation of the UI than a limitation of the xml system they have in place. Perhaps it would be worthwhile to see what cmake does or try to find some better documentation coming from MS.

Regarding the changes, I'll make it more flexible so it can go in Globals or Config depending on settings.

redorav:
Thanks for the feedback @starkos and @samsinsane , I'm currently investigating how to expand the Globals section to include conditionals as needed. The interesting thing about these modifications is that they are actually tweakable via the UI provided the condition was present there to begin with, so it's not like we're taking anything away from the user.

redorav:
I've moved systemversion back to Globals, and also create conditional global entries in the vcxproj. One thing I've not found a straightforward way of doing is to not create the property groups if no entries are going to be populated. I think the assumption is that there's always something that's going to populate it. Any suggestions? I've added an example test, but I mean to expand it once we agree on a final implementation.

starkos:
> One thing I've not found a straightforward way of doing is to not create the property groups if no entries are going to be populated

If you search the source for "io.capture" you can find some places where we do that. You run the code and capture the output, and then check to see if the output is empty before writing it out. Let me know you need help with it and I'll throw together a proper example.

redorav:
I think I managed to do it via premake.capture() which I guess was renamed at some point from io.capture.

redorav:
As soon as you're happy with the changes I've also fixed #1038 locally which relies on this PR to work, since this one allows setting conditions on globals and the other fix extends the function via premake.override.

starkos:
@samsinsane, are you okay with this one? If so, feel free to merge.

PR #1049, opened 2018-04-12
Add InlinesVisibility API

tdesveauxPKFX:
As said in #1044, inlines visibility should have it's own API.


TurkeyMan:
Why does this need its own API? Why can't this be expressed with visibility?
It feels very unlikely that you would ever supply `-fvisibility-inlines-hidden` without also supplying `-fvisibility=hidden`.

tvandijck:
@TurkeyMan do you still have objections? otherwise go ahead and merge this one, LGTM.

Issue #1051, opened 2018-04-13
Generated Visual Studio 2005 projects have conflicting properties


natrist:
Using Premake 5.0.0-alpha-12 to build for vs2005 targeting Windows 7:
Everything seems to work and then VS warns that:
Warning	1	Command line warning D9025 : overriding '/MT' with '/MD'	cl
Warning	2	Command line warning D9025 : overriding '/MT' with '/MD'	cl

for both my projects...

When I debug, it says that my program wasn't built with debug information and when I try to run any of the binaries it says that I'm missing MSVCR80.dll, which I clearly have.

If I put the DLL in the bin folder I get the following error:
R6034: An application has made an attempt to load the C runtime library incorrectly.

natrist:
premake5.lua:

`-- Generate the 'Storm' workspace (a.k.a. Visual Studio solution)
workspace "Storm"
	configurations { "Debug", "Release" }
	location "Build"

-- Generate the 'ConsoleStorm' project
-- This project is used for testing the Storm library from within the IDE
project "ConsoleStorm"
	kind ("ConsoleApp")
	language "C++"

	files "Source/ConsoleStorm.cpp"
	includedirs { "Source/H" }
	links { "StormLib" }

-- Generate the 'StormLib' project
-- This is the main library project, containing the preciouuus
project "StormLib"
	kind ("StaticLib")
	language "C++"

	-- Generic build steps
	-- This needs to be set before platform-specific steps or it will not run
	-- Possible Premake bug?
	files { "Source/*.cpp" }
	-- We include all files under Source/H to create the Visual Studio filter for header files
	files { "Source/H/**" }
	removefiles { "Source/ConsoleStorm.cpp" }
	includedirs { "Source/H" }

	-- Windows ONLY build steps
	filter "system:windows"
		files { "Source/W32/*.cpp" }
	
	-- Linux ONLY build steps
	filter "system:linux"
		files { "Source/Linux/*.cpp" }
	
	-- MacOS ONLY build steps
	filter "system:macosx"
		files { "Source/Mac/*.cpp" }
`

samsinsane:
I don't have VS2005, so I can't verify much locally. Can you put this at the top of your premake5.lua and let me know if that resolves the issue.
```lua
require "vstudio"
premake.override(premake.vstudio.vc200x, "VCCLCompilerTool", function(oldfn, cfg, toolset)
	oldfn(cfg, iif(toolset == premake.tools.msc, nil, toolset))
end)
```

[...]

PR #1054, opened 2018-04-16
Fixes for various VS2005

robinwatts:
We are wanting to adopt premake5 within the mupdf project, but we've hit some problems with the VS2005 generation.

I have fixes (well, workarounds at least) implemented in the referenced commits. If the fixes can't be taken as-is, hopefully they'll a) illustrate the issues, and b) allow someone to point the way to fixing them properly.

The problems that this works around:

1) Normal generation of VS2005 C projects includes a 'ProgramDataBaseFileName=""' entry. This has the effect of confusing VS2005 into thinking that everything is always out of date. Any builds of the solution do a complete rebuild. Simply omitting this line solves it.

2) The standard debug information format these days (certainly for us) is to use /Zi, not /Z7. Also (if I'm reading the lua right), it is output at one point using the index within the list. Accordingly, I've introduced an empty 'Off' value.

3) DebugInformationFormat appears to always be being set to 0 (because toolset is never non-NULL in any of the cases in which I've hit the code that outputs it. This is the wrong value, so I've just disabled the code.

4) Finally, the approach taken by the code appears to be to just bung the flags for the different options in an "AdditionalOptions" entry.

The solution doesn't parse these flags, just passes them on, after setting the flags that it thinks should be used.

So, for instance, if you output a project that is to be a debug project, with the "/Od" option in the AdditionalOptions field, and load that into VS, the properties window will show the Optimisation field as being the default (which IIRC is "/Ot"). When built, both flags are passed to the compiler. The right one is used, but a warning is given. It's also very confusing for people to not see their options reflected in the project settings.

Accordingly, I've added code that spots the flags in the AdditionalOptions field, removes them and outputs the proper flags as appropriate.

It is quite possible that there is a neater solution to all of this, but I am not a lua speaker, so this is the best I could come up with on short notice.

### DIFF
--- a/modules/vstudio/vs200x_vcproj.lua
+++ b/modules/vstudio/vs200x_vcproj.lua
@@ -819,10 +819,54 @@
 		end
 	end
 
-
+	function tablefindandremove(tab, el)
+		for index, value in pairs(tab) do
+			if value == el then
+				table.remove(tab, index)
+				return true
+			end
+		end
+		return false
+	end
 
 	function m.additionalExternalCompilerOptions(cfg, toolset)
 		local buildoptions = table.join(toolset.getcxxflags(cfg), cfg.buildoptions)
+		if tablefindandremove(buildoptions, "/MT") then
+			p.x('RuntimeLibrary="0"');
+		elseif tablefindandremove(buildoptions, "/MTd") then
+			p.x('RuntimeLibrary="1"');
+		elseif tablefindandremove(buildoptions, "/MD") then
+			p.x('RuntimeLibrary="2"');
+		elseif tablefindandremove(buildoptions, "/MDd") then
+			p.x('RuntimeLibrary="3"');
+		end
+		if tablefindandremove(buildoptions, "/Od") then
+			p.x('Optimization="0"');
+		elseif tablefindandremove(buildoptions, "/Os") or
+		       tablefindandremove(buildoptions, "/O1") then
+			p.x('Optimization="1"');
+		elseif tablefindandremove(buildoptions, "/Ot") or
+		       tablefindandremove(buildoptions, "/O2") then
+			p.x('Optimization="2"');
+		elseif tablefindandremove(buildoptions, "/Ox") then
+			p.x('Optimization="3"');
+		end
+		if tablefindandremove(buildoptions, "/Z7") then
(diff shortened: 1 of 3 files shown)


samsinsane:
You should check out #1051, I've got a little hack in there to resolve the `toolset` related issues. It's not the best solution, but there's a lot of code that makes the assumption that the `toolset` isn't set and this is a pretty quick way to get it back into that state.

robinwatts:
And in the latest commit

5) a fix for a quoting issue with predefinitions.

    We have a project file that needs to set a symbol to include quoted
    strings. For instance:

       defines {
         'FT_CONFIG_OPTIONS_H="slimftoptions.h"'
       }

    The source .c file then does:

       #include FT_CONFIG_OPTIONS_H

    With the definition like this, gnu make can accept the generated
    files and does the wrong thing. VS project files have the quotes
    appear in them as &quot; but these disappear on being passed to
    the actual compiler.

    Having modified premake5 to output \quot; we get exactly the
    results we need.

    An alternative workaround we tried was to use:

       defines {
         'FT_CONFIG_OPTIONS_H=\\"slimftoptions.h\\"'
       }

    which produced working VS2005 files, but broken gnu make files.

modules/vstudio/vs200x_vcproj.lua
```diff
@@ -819,10 +819,54 @@
...
+			if value == el then
+				table.remove(tab, index)
+				return true
+			end
+		end
+		return false
+	end
 
 	function m.additionalExternalCompilerOptions(cfg, toolset)
 		local buildoptions = table.join(toolset.getcxxflags(cfg), cfg.buildoptions)
```
tdesveauxPKFX: I think a simpler solution would be to replace this with `local buildoptions = cfg.buildoptions`.
Them pulling only the wanted options from the toolset and having functions called to output the correct tags like the vs2010 generator.

starkos:
This one hasn't moved in a long time, I'm going to go ahead and close it. If someone wants to take it on:

- It's always better to break multiple fixes out into multiple PRs
- It needs unit tests to show that it is doing what it says
- The `tablefindandremove`-related logic should be reworked to use a `config.mapFlags()` approach similar to the VS201x actions

PR #1062, opened 2018-04-17
Allow system("android") per configuration

redorav:


TurkeyMan:
Yeah, I'm not sure what the right process is anymore. That "Update branch" button is the spawn of satan... but to avoid it, you need admin privileges!
I can't believe Github thought this was okay :/
I hope Github are aware they're idiots, and making options to configure this flow for projects.

pdoane:
> Yeah, I'm not sure what the right process is anymore. That "Update branch" button is the spawn of satan... but to avoid it, you need admin privileges!

It's a part of the branch protection settings so it can be controlled per-branch/per-repository as to whether code needs to be up to date.

Issue #1060, opened 2018-04-17
Not sure about `debugflavour`


TurkeyMan:
@redorav recently introduced a new api which I'm not so sure about.
I feel there's already comprehensive set of api's to give sufficient debugger configuration, and the new API feels poorly defined. I feel like it can be expressed in terms of the existing API's...

Perhaps we could get some comment about the reason for this approach?

redorav:
@TurkeyMan I haven't added any new apis to premake (as far as I know!). I've looked to see who added it but I can't find it in the repo.

tdesveauxPKFX:
Hey, I did.

And now that you mention it, I see other already existing APIs that I could have used, like `debugger`.
I don't know how I missed it. I will wait a little to see if someone find a reason I should not extend `debugger` and submit a PR fixing this if not.

TurkeyMan:
Oh I'm sorry, I was browsing the recent commit log, and you both have a lot of commits near the top. Sorry for the confusion!

TurkeyMan:
I see how I made the confusion... you have both merged master into your working branches a few times, and then produce PR's which have been accepted. Those merges are pulling each others patches into your working branches. The top of the revision graph is a mess! ;)

For future reference, can we please be sure to sanitise branches before creating PR's?
Rebase on master, eliminate all the merges into working branch from the timeline, squash incremental changes if it makes sense to do so, or massage into a small number of patches for convenient review; ie, one patch implements feature, next patch updates unit-tests, next refactors existing code to use the new feature. It's nice to present PR's as a series of patches that the reviewer should understand and digest in series, and it preserves much better understanding of history.

I'll try and keep my eye on it better in the future. Sorry I missed it!

redorav:
I apologize for any mess I may have caused in the history. I'm starting to learn how git works and I'm sure I'm doing a few things wrong on the way. My first PR ever has been in this project if that serves as an indicator :) I'll be sure to follow your advice for the next.

TurkeyMan:
Regarding the `debugflavour` thing. Options are `Local`, `Remote`, `WebBrowser`, `WebService`.
I think local/remote distinction can be implied by other existing options, like `debugremotehost`. Safe to say, if that's set, we're remote-debugging.
The the 2 web options can probably be `debuggertype`'s? Does that feel right?

tdesveauxPKFX:
[...]

PR #1063, opened 2018-04-18
Replace debuggerflavor with debugger

tdesveauxPKFX:
Ref: #1060 

Remove duplicate in APIs as `debugger` can be used to set `DebuggerFlavor` in Visual Studio.

### DIFF
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -23,6 +23,11 @@
 
 	-- Initialize Specific API
 
+	p.api.addAllowed("debugger", "VisualStudioLocal")
+	p.api.addAllowed("debugger", "VisualStudioRemote")
+	p.api.addAllowed("debugger", "VisualStudioWebBrowser")
+	p.api.addAllowed("debugger", "VisualStudioWebService")
+
 	p.api.register {
 		name = "shaderoptions",
 		scope = "config",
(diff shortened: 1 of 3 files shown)


TurkeyMan:
I'm still not sure I'm into this... it feels like we've confused 2 concepts now. Busy ATM, let me get back to this...

tdesveauxPKFX:
Yeah, I was too quick in making this. I'm pretty busy as well but should be able to look into it this week end.

src/_premake_init.lua
```diff
@@ -285,6 +285,10 @@
 			"Default",
 			"GDB",
 			"LLDB",
+			"Local",
+			"Remote",
```
TurkeyMan: I don't like `Local` & `Remote` here, I think those states are implied by `debugremotehost` being set or not no?

tdesveauxPKFX:
I somewhat forgot about this.

Just updated the branch and the only notable change was to rename Local to VisualStudioLocal (same for the others).

I think this PR can be merged. If there is further concerns about how to handle `DebuggerFlavor` for Visual Studio, there should be another issue/PR.

starkos:
I defer to @TurkeyMan, who had some concerns about the previous versions. Will add as a reviewer.

samsinsane:
@TurkeyMan Do you have any objections to merging this?

starkos:
Closing this one in favor of #1357, which deprecates rather than removes `debuggerflavor()`.

PR #1067, opened 2018-04-25
Add 'Default', 'Dwarf' and 'SplitDwarf' arguments to 'debugformat'

ratzlaff:
Adds 'Default', 'Dwarf' and 'SplitDwarf' options to 'debugformat' when using the xcode, gmake or gmake2 modules.

Wiki updates are at: https://github.com/ratzlaff/premake-core/wiki/debugformat
wiki repo URL: https://github.com/ratzlaff/premake-core.wiki.git (branch: xcode_debugformat)

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -32,6 +32,18 @@
 
 
 --
+-- Returns string to be appended to -g
+--
+	function gcc.getdebugformat(cfg)
+		local flags = {
+			Default = "",
+			Dwarf = "dwarf",
+			SplitDwarf = "split-dwarf",
+		}
+		return flags
+	end
+
+--
 -- Returns list of C compiler flags for a configuration.
 --
 	gcc.shared = {
@@ -93,9 +105,11 @@
 			High = "-Wall",
 			Off = "-w",
 		},
-		symbols = {
-			On = "-g"
-		},
+		symbols = function(cfg, mappings)
+			local values = gcc.getdebugformat(cfg)
+			local debugformat = values[cfg.debugformat] or ""
+			return { On = "-g" .. debugformat }
+		end,
 		unsignedchar = {
 			On = "-funsigned-char",
 			Off = "-fno-unsigned-char"
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -108,7 +108,11 @@
 		symbols = function(cfg, mappings)
 			local values = gcc.getdebugformat(cfg)
 			local debugformat = values[cfg.debugformat] or ""
-			return { On = "-g" .. debugformat }
+			return {
+				On       = "-g" .. debugformat,
+				FastLink = "-g" .. debugformat,
+				Full     = "-g" .. debugformat,
+			}
 		end,
 		unsignedchar = {
 			On = "-funsigned-char",
--- a/modules/gmake/tests/cpp/test_flags.lua
+++ b/modules/gmake/tests/cpp/test_flags.lua
@@ -53,6 +53,39 @@
 	end
 
 --
+-- symbols default to 'off'
+--
+	function suite.symbols_default()
+		symbols "default"
+		prepare { "cFlags", "cxxFlags" }
+		test.capture [[
+  ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS)
+  ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS)
+		]]
+	end
+
+--
+-- All other symbols flags also produce -g
+--
+	function suite.symbols_fastlink()
+		symbols "FastLink"
+		prepare { "cFlags", "cxxFlags" }
+		test.capture [[
+  ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -g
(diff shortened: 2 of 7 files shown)


ratzlaff:
I am not familiar with the debug formats that gcc handles, which is the only reason I did not put them in _premake_init.lua. I thought this would be better encapsulation for module-specific parameters. 

If gcc does handle dwarf, I'll move it over. Does gcc do dwarf-with-dsym also?

starkos:
Xcode uses Clang under the hood, so Clang must support them at least?

ratzlaff:
rebased onto master

modules/xcode/_preload.lua
```diff
@@ -40,6 +40,10 @@
 		}
 	}
 
+	p.api.addAllowed("debugformat", {
+		"dwarf-with-dsym",
```
samsinsane: There's a similar option on Linux that does the same thing, it's called something else though. Given that they're the same thing, but have different names, it would be good to use generic terminology - I believe `dsym` is very much an Apple term, I only got macOS hits when I googled it the other night.

ratzlaff: The idea here was to not introduce a layer of mapping between premake values and what is generated in the xcode project file (ie: what you enter is what you get).

I (mistakenly) thought that by adding to `debugformat` in the xcode action, those options would not be available outside of the module, whoops!

It looks like the equivalent gcc flag is be `-gsplit-dwarf`. Ill make that the value instead.

starkos:
Another option is to make "dwarf-with-dsym" an alias of "split-dwarf", or vice versa.

ratzlaff:
@starkos I just squashed some commits just now to do that. Updating my copy of the documentation now...

src/tools/gcc.lua
```diff
@@ -93,9 +93,10 @@
 			High = "-Wall",
 			Off = "-w",
 		},
-		symbols = {
-			On = "-g"
-		},
+		symbols = function(cfg, mappings)
+			local debugformat = cfg.debugformat or ""
+			return { On = "-g" .. debugformat }
```
TurkeyMan: What does `-gc7` mean?

TurkeyMan:
I feel like there probably needs to be a `Default` value allowed here too, and all the allowed values have a weird `lower-case` convention. This is unlike any other API in premake, especially with the `-` separator, which may cause inconveniences with lua syntax when indexing explicitly.
I think there should probably be `Default`, `C7`, `Dwarf`, `SplitDwarf`, and map them to proper values in the function (ignoring C7 for GCC)

ratzlaff:
Added 'Default', changed capitalization of values.

Did not change c7 to C7 - should be a separate PR to properly deprecate the flag.

src/tools/gcc.lua
```diff
@@ -93,9 +105,11 @@
 			High = "-Wall",
 			Off = "-w",
 		},
-		symbols = {
-			On = "-g"
-		},
+		symbols = function(cfg, mappings)
+			local values = gcc.getdebugformat(cfg)
+			local debugformat = values[cfg.debugformat] or ""
+			return { On = "-g" .. debugformat }
```
TurkeyMan: Is it a supported feature that we return a table from this function?
I thought this function was just meant to return the options string (the value part)...?

Assuming this works, it should also support `FastLink` and `Full` values for `symbols`. Docs allege that if they are undefined for a platform, they are a synonym for `On`.

TurkeyMan:
Sorry to push back on this again, but `symbols` is such a fundamental API ;)
I recognise that the code was already stale (only `On` was present in the code before you modified it), but this is a good opportunity to de-stale it ;)

src/tools/gcc.lua
ratzlaff: Returning a table from this function is to preserve as much of the existing behavior prior to this change as possible.

The test for 'function' type and executing the function [here](https://github.com/premake/premake-core/pull/1067/files/95d1c021f6345558291aba7ea71530f9b94eba37#diff-00c8206c40c5891453b9c823076a1c54R495) is what triggers this logic to take place.

ratzlaff: Ah. I see what you mean about `FastLink` and `Full`. Will add those in when I get home later.

ratzlaff: How do I add a default value to the map so I can future-proof these values? Specifically I am referring to [this](https://github.com/premake/premake-core/pull/1067/files/95d1c021f6345558291aba7ea71530f9b94eba37#diff-00c8206c40c5891453b9c823076a1c54R524) check. I do not know what I would google for in order to get map._ working.

ratzlaff:
@TurkeyMan Next iteration ready to go!

samsinsane:
@TurkeyMan are you happy with the latest changes?

tdesveauxPKFX:
I think all concerns @TurkeyMan had were addressed and this was not updated in a while.

I think this is a nice addition and looking through the changes feels fine with merging.
Should we override the pending requested changes?

samsinsane:
@tdesveauxPKFX I'm happy for this to go in, I think we've let this PR stagnate for long enough. If there's problems, we can just address them as they pop up.

samsinsane:
I'm just going to merge this now, thanks for fixing it up multiple times! Sorry it's taken so long to get this through.

ratzlaff:
Thanks!

PR #1069, opened 2018-04-26
Support successfully bootstrapping on Solaris (OpenIndiana).

lanurmi:
Some support already existed, but not enough. Also the "linkoptions" provided was incompatible with the Solaris ld.

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -91,6 +91,16 @@ bsd: $(SRC)
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake
 	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN` config=$(CONFIG)
 
+solaris: $(SRC)
+	$(SILENT) rm -rf ./bin
+	$(SILENT) rm -rf ./build
+	$(SILENT) rm -rf ./obj
+	mkdir -p build/bootstrap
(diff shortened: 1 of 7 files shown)


samsinsane:
Does this fix Solaris, or just OpenIndiana? I wouldn't want to approve a PR that breaks the official OS so that an open source derivative works instead. Are you able to provide any additional context to this for my peace of mind?

@starkos You appear to be the one who originally added Solaris support, but on behalf of someone else? Do you have access to a Solaris box?

lanurmi:
> Does this fix Solaris, or just OpenIndiana?

I hope both, but I don't currently have access to a genuine Solaris to verify that.

> I wouldn't want to approve a PR that breaks the official OS so that an open source derivative works instead. Are you able to provide any additional context to this for my peace of mind?

The two `linkoptions` lines were the only explicit reference to solaris within the bootstrap configuration, so I suspect that current master would not build on a genuine Solaris either. The rest of the configuration would treat Solaris as either Linux or BSD.

As for the `linkoptions`, [this post](https://stackoverflow.com/questions/596076/solaris-linker-equivalent-to-the-gnu-ld-export-dynamic-flag) suggests that `--export-dynamic` is a feature of GNU ld only. (GNU ld may be available on a Solaris system, but it is not the default ld.)

samsinsane:
@lanurmi True, sorry about that! I didn't go back over your changes again after reading up on Solaris and OpenIndiana, I should have since you haven't made any changes that look like they would cause issues.

I'll wait for Jason to respond to see if he's got access to a genuine Solaris machine, otherwise, we can probably just merge this. Thanks for changes!

starkos:
> @starkos You appear to be the one who originally added Solaris support, but on behalf of someone else? Do you have access to a Solaris box?

Umm…wow, no. I haven't even seen a Solaris box this century, I'm pretty sure. The original Solaris support was contributed by someone else, back in the dark days before GitHub :)

PR #1070, opened 2018-04-26
Support successfully bootstrapping on OpenBSD.

lanurmi:
This PR depends on PR #1068, which should be applied first.

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -89,7 +89,7 @@ bsd: $(SRC)
 	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_USE_POSIX -DLUA_USE_DLOPEN -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lm
 	./build/bootstrap/premake_bootstrap embed
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake
-	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN` config=$(CONFIG)
+	$(MAKE) -C build/bootstrap -j`getconf NPROCESSORS_ONLN` config=$(CONFIG)
 
 windows-base: $(SRC)
 	$(SILENT) if exist .\bin rmdir /s /q .\bin
--- a/src/host/premake.c
+++ b/src/host/premake.c
@@ -365,7 +365,7 @@ int premake_locate_executable(lua_State* L, const char* argv0)
 	}
 #endif
 
-#if PLATFORM_BSD
+#if PLATFORM_BSD && !defined(__OpenBSD__)
 	int len = readlink("/proc/curproc/file", buffer, PATH_MAX - 1);
 	if (len < 0)
 		len = readlink("/proc/curproc/exe", buffer, PATH_MAX - 1);


src/host/premake.c
```diff
@@ -365,7 +365,7 @@ int premake_locate_executable(lua_State* L, const char* argv0)
 	}
 #endif
 
-#if PLATFORM_BSD
+#if PLATFORM_BSD && !defined(__OpenBSD__)
```
samsinsane: This seems not great - is there *something* that can be done to support OpenBSD in this block instead? What does this block do on OpenBSD? I got it working on FreeBSD by adding that `sysctl` call, is there something similar for OpenBSD?

Sorry to bombard you with questions! I'm keen to see what we can do to improve the BSD situation!

Bootstrap.mak
```diff
@@ -89,7 +89,7 @@ bsd: $(SRC)
 	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_USE_POSIX -DLUA_USE_DLOPEN -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lm
 	./build/bootstrap/premake_bootstrap embed
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake
-	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN` config=$(CONFIG)
+	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN || getconf NPROCESSORS_ONLN` config=$(CONFIG)
```
samsinsane: Would it be possible to remove `_NPROCESSORS_ONLN` or does this depend on the BSD flavour?

lanurmi: Actually, it seems `_NPROCESSORS_ONLN` doesn't work on FreeBSD either; so the underscore-less variant would be fine for both OpenBSD and FreeBSD. But as someone else added the variant with underscore in the first place, dunno if it actually works on some *BSD.

lanurmi:
I tried compiling also on NetBSD, but encountered way more fundamental errors (such as compiler not supporting `long long` -- not sure if there's something badly wrong with my installation, or is it supposed to be like that). In any case, I haven't tested other BSD variants besides OpenBSD and FreeBSD, but of course supporting all of them would be preferable.

src/host/premake.c
samsinsane: That's fair enough, perhaps wrapping that block instead would be better? Not required, if the none of it will work on OpenBSD it's probably pointless reducing it down to what doesn't compile.

Bootstrap.mak
samsinsane: That would have been me, I just copied the Linux one and it built the project, I didn't investigate the value. If it doesn't work on your FreeBSD it likely didn't work on mine, I think we should remove the underscore version. :+1:

lanurmi: Yeah, when it doesn't work it is not easily noticeable with GNU Make, as it accepts `-j` without a numeric value too, and the build doesn't fail. I was (unintentionally) using a non-gnu make on OpenBSD, which then failed because it requires a value for `-j`. And yes I agree, the underscore version should be removed.

src/host/premake.c
samsinsane: I meant that entire `if` block that contains the call to `sysctl`, but it's not really worth the effort. Thanks for the link, it's a pretty interesting discussion.

Bootstrap.mak
samsinsane: Cool, can you remove it in this PR? We can merge this once you've done that. 😄

lanurmi: Done. 😄

PR #1009, opened 2018-02-20
Visual Studio added explicit C++17 option

Enhex:
More correct than staying with `latest`.

PR #1071, opened 2018-04-26
Add C++17 support for Visual Studio 2017.

manxorist:
This also adds support for 'cppdialect "C++latest"'.

### DIFF
--- a/modules/vstudio/tests/vc2010/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2010/test_compile_settings.lua
@@ -1183,6 +1183,48 @@
 	<PrecompiledHeader>NotUsing</PrecompiledHeader>
 	<WarningLevel>Level3</WarningLevel>
 	<Optimization>Disabled</Optimization>
+	<LanguageStandard>stdcpp17</LanguageStandard>
+</ClCompile>
+		]]
+	end
+
+	function suite.onLanguage_CppLatest_VS2010()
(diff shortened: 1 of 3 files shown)


manxorist:
This pull request is based on discussion in #1009.

pdoane:
There are two directions that can work here:

  1. Use the specific Visual Studio version (e.g. 15.6) in the action.  VS2017 is no longer enough information to figure out how to map some options and only supporting the latest release isn't practical for how VS2017 is being developed.
  2. Use options that directly map to Visual Studio usage (in this case C++latest).

More details of this were discussed on #1009.

manxorist:
> Use the specific Visual Studio version (e.g. 15.6) in the action. VS2017 is no longer enough information to figure out how to map some options and only supporting the latest release isn't practical for how VS2017 is being developed.

It never was. As already mentioned, minor versions of Visual Studio did add features also in earlier versions.

> Use options that directly map to Visual Studio usage (in this case C++latest).

Which is precisely what I have implemented in this pull request.

> More details of this were discussed on #1009.

I just **need** C++17 support, and I will happily change the pull request to whatever minimal solution is desired. In #1009 it was requested to add `"C++latest"` (which I think totally makes sense to have), which I did.

pdoane:
Is there anything left to go over here? We talked through this a bit on the previous PR and this is implementing the decision made there.

PR #1072, opened 2018-04-27
Add color to failing test output

ratzlaff:
1) use fptus in the implementation of term.setTextColor to have better newline control
2) change the color of the test name when tests fail so that they are easier to visually parse


tvandijck:
test failures are unrelated to this change.. sometimes the http server for the http tests is rate limiting us, and starts timing out.

PR #1074, opened 2018-04-30
Fixed issues with duplicate tests

samsinsane:
- Test suite will now fail when test names are reused
- Fixed several unit tests that reused test names

### DIFF
--- a/modules/self-test/test_declare.lua
+++ b/modules/self-test/test_declare.lua
@@ -36,7 +36,14 @@
 			error('Duplicate test suite "'.. suiteName .. '"', 2)
 		end
 
-		local suite = {}
+		local _suite = {}
+		-- Setup a metatable for the test suites to use, this will catch duplicate tests
+		local suite = setmetatable({}, {
+			__index = _suite,
+			__newindex = function (table, key, value) if m.detectDuplicateTests and _suite[key] ~= nil then error('Duplicate test "'.. key .. '"', 2) end _suite[key] = value end,
+			__pairs = function (table) return pairs(_suite) end,
+			__ipairs = function (table) return ipairs(_suite) end,
+		})
 
 		suite._SCRIPT_DIR = _SCRIPT_DIR
 		suite._TESTS_DIR = _TESTS_DIR
--- a/modules/codelite/tests/test_codelite_config.lua
+++ b/modules/codelite/tests/test_codelite_config.lua
@@ -196,7 +196,7 @@ cmd2</StartupCommands>
 		]]
 	end
 
-	function suite.OnProject_PreBuild()
+	function suite.OnProject_PostBuild()
 		postbuildcommands { "cmd0", "cmd1" }
 		prepare()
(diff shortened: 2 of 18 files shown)


modules/self-test/test_declare.lua
```diff
@@ -36,7 +36,14 @@
 			error('Duplicate test suite "'.. suiteName .. '"', 2)
 		end
 
-		local suite = {}
+		local _suite = {}
+		-- Setup a metatable for the test suites to use, this will catch duplicate tests
+		local suite = setmetatable({}, {
+			__index = _suite,
+			__newindex = function (table, key, value) if m.detectDuplicateTests and _suite[key] ~= nil then error('Duplicate test "'.. key .. '"', 2) end _suite[key] = value end,
```
tvandijck: can we use normal multi-line code... I know it's just a styling thing, but it makes it a bit more readable.
also `p.error`.

tvandijck:
could this be the dreaded random fail with just `next()` in the log ?

modules/self-test/test_declare.lua
samsinsane: Ah, my bad, I wasn't supposed to leave it like that.

Not sure why I used `error` over `p.error`, weird.

samsinsane:
> could this be the dreaded random fail with just `next()` in the log ?

I've never seen this error before, I kind of doubt that it's related to this though. The duplicate tests just override the previous test, it's all in order so they always replace the same function. I would imagine that error is probably much more likely to be part of the alias fixes I made, there *were* 1-3 extra "tests" and 2 of them are set to `nil`. If, a big if here, `for testName, testFunction in pairs(test.suite) do` has an internal iterator, it's likely that the `next()` function fails because the table entry was removed when it was set to `nil`. No idea how the `pairs` function works, but that's my guess - that's assuming that either of my PRs resolves that `next()` issue.

modules/self-test/test_declare.lua
samsinsane: Oh, the top of this function uses `error`, that's why I used it. `p.error` doesn't provide the file location either, whereas `error` does, which is likely why `error` is used instead here.

samsinsane: Example using `error`:
```
C:\github\premake-core>bin\release\premake5.exe test
Running action 'test'...
Error: ** Error: Duplicate test "referencesAreRelative_onDifferentProjectLocation"
```
Example using `p.error`:
```
C:\github\premake-core>bin\release\premake5.exe test
Running action 'test'...
Error: ...-core/modules/vstudio/tests/vc2010/test_project_refs.lua:83: Duplicate test "referencesAreRelative_onDifferentProjectLocation"
```

samsinsane:
> Either way, this, and that alias test thing, are pretty darn good catches...

Yeah, I was trying to improve the output of the test runner and my numbers weren't lining up, so I sort of just stumbled across these two bugs. Potentially fixing that random fail sounds like a nice bonus! :+1:

PR #1076, opened 2018-05-01
Add support for mixed C++/D project

EmmeKappaErre:
Modified the D module to support the C++/D mixed mode. This allows to have D files in a C++ project and provide means to set the compiler options.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1366,6 +1366,18 @@
 		}
 	}
 
+	api.register {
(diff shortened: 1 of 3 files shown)


EmmeKappaErre:
Hi Samuel,
Thanks for your feedback. I will try to rearrange the change and make it easier to read.
I am not entirely sure how to do the tests, any recommendation on which other test I could use as a template?
Thanks,

Marco

PR #1080, opened 2018-05-04
Fix curl compilation on Debian PowerPC.

lanurmi:
For some reason, __ppc__ is not defined by GCC, but __PPC__ is.

### DIFF
--- a/contrib/curl/include/curl/curlbuild.h
+++ b/contrib/curl/include/curl/curlbuild.h
@@ -528,7 +528,7 @@
 
 #elif defined(__GNUC__)
 #  if !defined(__LP64__) && (defined(__ILP32__) || \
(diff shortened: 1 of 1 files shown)


TurkeyMan:
I feel like this patch should probably be submit against CURL...

lanurmi:
CURL upstream [seems to check for `__powerpc__`](https://github.com/curl/curl/pull/1774) now, which solves the same problem.

tvandijck:
can we just take the fix from curl then?

lanurmi:
I guess, but if you do, please pay attention that the PR I linked to caused a regression, so their correct fix is actually two commits. Also my fix includes a blind fix for PPC64, and theirs doesn't.

PR #1087, opened 2018-05-15
Fix missing DebugInformationFormat when symbols are set to full

rorydriscoll:
When symbols are set to "full" then the DebugInformationFormat attribute is not emitted no matter what the other settings are. For me, this manifested itself as the editandcontinue function appearing not to work.


rorydriscoll:
It would be less error prone to explicitly list out the default case too and make the final else emit an error. That would have saved me a couple of hours. In another language this would be a switch statement with an error in the default case.

PR #1091, opened 2018-05-20
Fix mkdir command on msdos if directory already exists

S6066:
On windows `mkdir -p foo` creates two folders: -p and foo. When using mkdir to create a folder already existing, useless logs appear in the shell (even with the `@` before the command)

### DIFF
--- a/modules/gmake/gmake.lua
+++ b/modules/gmake/gmake.lua
@@ -145,7 +145,7 @@
 	function make.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
-		_p('else ifeq(msdos),$(SHELLTYPE))')
+		_p('else ifeq (msdos,$(SHELLTYPE))')
 		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)', dirname, dirname)
 		_p('else')
 		_p('\t$(SILENT) mkdir $(subst /,\\\\,%s)', dirname)
--- a/modules/gmake/gmake.lua
+++ b/modules/gmake/gmake.lua
@@ -146,7 +146,7 @@
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
 		_p('else ifeq(msdos),$(SHELLTYPE))')
-		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)')
+		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)', dirname, dirname)
 		_p('else')
 		_p('\t$(SILENT) mkdir $(subst /,\\\\,%s)', dirname)
 		_p('endif')
--- a/modules/gmake/gmake.lua
+++ b/modules/gmake/gmake.lua
@@ -145,6 +145,8 @@
 	function make.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
+		_p('else ifeq(msdos),$(SHELLTYPE))')
+		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)')
 		_p('else')
 		_p('\t$(SILENT) mkdir $(subst /,\\\\,%s)', dirname)
 		_p('endif')
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -124,7 +124,7 @@
 	function gmake2.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
-		_p('else ifeq(msdos),$(SHELLTYPE))')
+		_p('else ifeq (msdos,$(SHELLTYPE))')
 		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)', dirname, dirname)
 		_p('else')
 		_p('\t$(SILENT) mkdir $(subst /,\\\\,%s)', dirname)
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -125,7 +125,7 @@
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
 		_p('else ifeq(msdos),$(SHELLTYPE))')
-		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)')
+		_p('\tif not exist "$(subst /,\\\\,%s)" $(SILENT) mkdir $(subst /,\\\\,%s)', dirname, dirname)
 		_p('else')
 		_p('\t$(SILENT) mkdir $(subst /,\\\\,%s)', dirname)
 		_p('endif')
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -124,6 +124,8 @@
 	function gmake2.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
+		_p('else ifeq(msdos),$(SHELLTYPE))')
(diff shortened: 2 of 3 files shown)


tdesveauxPKFX:
I can't reproduce the issue on my side.
MINGW is identified as a `posix` and handle `-p`.
MSDOS command line go through the `else` which already does a `mkdir` without `-p`.

Seeing your PR, is the fact that the command fail if the directory already exists an issue to you?

S6066:
When running that `@echo $(SHELLTYPE)` rules, I get... nothing

I ran `make` with that `-n` argument, and now things get weird
I'm not using default windows cmd, I'm using Cmder, and running mingw into it makes `make` think that SHELLTYPE is posix (even if the `echo $(SHELLTYPE)` I ran did nothing). When running in a standard cmd prompt, I don't have any `-p`.

But still, It would make useless logs without that `-p`, so the PR stays useful

> On another note, _p() works the same way as printf(), meaning it will format %s in the first argument with the following arguments. The modification you made don't work as the first argument contains 2 %s but no arguments to substitute them. You can look at the lines above and below for a correct example.

Whoops. Committing the fix ASAP

modules/gmake/gmake.lua
```diff
@@ -145,6 +145,8 @@
 	function make.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
+		_p('else ifeq(msdos),$(SHELLTYPE))')
```
tdesveauxPKFX: There is an extra `)` after msdos here.

modules/gmake2/gmake2.lua
```diff
@@ -124,6 +124,8 @@
 	function gmake2.mkdir(dirname)
 		_p('ifeq (posix,$(SHELLTYPE))')
 		_p('\t$(SILENT) mkdir -p %s', dirname)
+		_p('else ifeq(msdos),$(SHELLTYPE))')
```
tdesveauxPKFX: Same here.

tdesveauxPKFX:
> When running that @echo $(SHELLTYPE) rule, I get... nothing

Just to check, where did you add it?
I added it in the prebuild rule after the project gen like this:
```
prebuild: | $(OBJDIR)
	@echo $(SHELLTYPE)
	$(PREBUILDCMDS)
```

`SHELLTYPE` should never be empty.

> I ran make with that -n argument, and now things get weird

Not sure what you mean by things get weird.

>I'm not using default windows cmd, I'm using Cmder, and running mingw into it makes make think that SHELLTYPE equals posix (even if the echo $(SHELLTYPE) I ran did nothing). When running in a standard cmd prompt, I don't have any mkdir -p, just mkdirs.

I will try with Cmder. To be sure, the issue happen when running the mingw32-make bin from Cmder or from mingw?

Sorry to be such a pain, I'm trying to find the correct cause of the issue to have the best fix possible.
Thanks for taking the time!

tdesveauxPKFX:
So I tried with Cmder and the issue is that it is detected as a posix shell but use Windows mkdir.

I don't see how your PR would solve the issue though.

This should be fixed here for gmake2: 
https://github.com/premake/premake-core/blob/ff732af5f6cec5832d6e1e789ee8632ef2e32e7f/modules/gmake2/gmake2.lua#L275-L284

The only way I see right now to do make the difference is that with cmd.exe, Cmder and Powershell, ComSpec is defined with `$(ComSpec)` and MinGW has `$(COMSPEC)`. That seems pretty weak too me but it worked where I tried.

An implementation to replace the above function:
```
SHELLTYPE := posix
ifeq (.exe,$(findstring .exe,$(ComSpec)))
  SHELLTYPE := msdos
endif
```

I'd like someone else to weight in on this. No sure where to go.

S6066:
Wait wait wait! The PR isn't about Cmder not detected as msdos, but msdos code not verifying folders for mkdir

> Just to check, where did you add it?
> I added it in the prebuild rule after the project gen like this:

I added it in the .PHONY thing, then made a rule with the same name. I know the problem isn't from the makefile code since at first I didn't put the @ before the echo, so I had `echo `, with SHELLTYPE empty

I'm fixing the extra parenthesis, didn't see that too.

tdesveauxPKFX:
Wow, I don't know why I though it was about that.

The reason why there is no check is that Makefile will take care of this for us.
The rule `$(OBJDIR)` is interpreted to the path to the directory and Makefile will check if it needs to execute the rule or not.
I can't see a reason for Makefile to execute the rule if the directory exists.

Could you provide a Makefile that cause the issue?

tdesveauxPKFX:
Wow, I don't know why I though it was about that.

The reason why there is no check is that Makefile will take care of this for us.
The rule `$(OBJDIR)` is interpreted to the path to the directory and Makefile will check if it needs to execute the rule or not.
I can't see a reason for Makefile to execute the rule if the directory exists.
By chance, does the logs not appears the first time you run make on a clean directory but appear only if you run it, run clean then run it again?

If not, could you provide a Makefile that cause the issue? Your Makefile might contains something that mine don't that cause the issue.

Also, you can run the unit-tests locally with the `test` action so that you can fix them before pushing your modifications.

S6066:
> By chance, does the logs not appears the first time you run make on a clean directory but appear only if you run it, run clean then run it again?

These are the steps I do:
1. Generate makefile with gmake action, on windows
2. Run normal make => no logs appear
3. Run make clean
4. Run normal make again => no logs appear
5. Run normal make a last time => obviously, "folder already exist"

> If not, could you provide a Makefile that cause the issue? Your Makefile might contains something that mine don't that cause the issue.

Sure, but first, I have to say there's something else really weird
I ran `make` after doing a `make clean`:
```
> C:\PortablePrograms\mingw-w64_x64-7.3.0-seh\mingw64\bin\mingw32-make
"==== Building efgh (debug_x64) ===="
main.cpp
g++: error: CreateProcess: No such file or directory
mingw32-make[1]: *** [efgh.make:177: obj/x64/Debug/main.o] Error 1
mingw32-make: *** [Makefile:33: efgh] Error 2```
I got that CreateProcess error too on a real project. I thought it was the project's premake's fault. Looks like it's not. And this is a serious thing, since it breaks build

And, for the makefile/project: https://transfer.sh/N08g0/premakefile.zip

> Also, you can run the unit-tests locally with the test action so that you can fix them before pushing your modifications.

I was editing from GitHub and hoping everything would work correctly, I'll try to clone this time

**Comment by tdesveauxPKFX** [comment]
Thanks for the Makefile, I will look into it.

>I got that CreateProcess error too on a real project. I thought it was the project's premake's fault. Looks like it's not. And this is a serious thing, since it breaks build

Indeed that's worrying. I think I got the same error once or twice when testing but did not look into it yet.

Can you try this PR #1103 and tell me if that fixes your issues?

**Comment by tdesveauxPKFX** [comment]
Just tested your Makefile and indeed I see the same issue with `gmake` but not with `gmake2`.

Following are the rule for `TARGETDIR` and how Makefile will interpret them.

`gmake`:
```
$(TARGET): $(GCH) ${CUSTOMFILES} $(OBJECTS) $(LDDEPS) $(RESOURCES)
	@echo Linking efgh
ifeq (posix,$(SHELLTYPE))
	$(SILENT) mkdir -p $(TARGETDIR)
else
	$(SILENT) mkdir $(subst /,\\,$(TARGETDIR))
endif
	$(SILENT) $(LINKCMD)
	$(POSTBUILDCMDS)
```

`gmake2`:
```
$(TARGETDIR):
	@echo Creating $(TARGETDIR)
ifeq (posix,$(SHELLTYPE))
	$(SILENT) mkdir -p $(TARGETDIR)
else
	$(SILENT) mkdir $(subst /,\\,$(TARGETDIR))
endif

$(TARGET): $(GCH) $(OBJECTS) $(LDDEPS) | $(TARGETDIR)
	@echo Linking efgh
```
As you can see, `gmake2` has a standalone rule for `TARGETDIR` to be created. This way, once it has been create, Makefile doesn't try to create it again.
However, `gmake` will try to create `TARGETDIR` each time it execute `TARGET` even if the directory already exists.

I am (really this time) closing this PR as the changes are not necessary nor make sense regarding the issue.

I invite you to use the `gmake2` generator as we are moving away form `gmake`.

I will still submit a PR to fix the issue as it does not require that much efforts.

I invite you to open an issue / PR regarding the CreateProcess error if you find this is due to Premake.

Anyway, thanks again for the time you spent to help me track down the cause!

S6066:
>Indeed that's worrying. I think I got the same error once or twice when testing but did not look into it yet.
>Can you try this PR #1103 and tell me if that fixes your issues?

I would like to build https://github.com/premake/premake-core/pull/1103 but can't, because of that CreateProcess error.
I tried to build Bootstrap.mak, and use a binary of premake5 to generate a gmake2 makefile with the premake5.lua, failed in both cases

tdesveauxPKFX:
#1103 does fix an issue but the PR you will want to try is #1105 instead.

I don't mind helping you further but as it seems unrelated to Premake for now, please contact me through my email (you can find it on my GitHub profile).

PR #1100, opened 2018-05-30
gmake2: Remove pch from FORCE_INCLUDE to allow NoPCH per-file

tdesveauxPKFX:


modules/gmake2/gmake2_cpp.lua
```diff
@@ -603,23 +600,25 @@
...
-			elseif p.languages.iscpp(fcfg.compileas) then
-				return 'ALL_CXXFLAGS'
-			end
+		if cfg.pchheader and not cfg.flags.NoPCH and (not fcfg or not fcfg.flags.NoPCH) then
+			table.insert(flags, "-include $(PCH_PLACEHOLDER)")
 		end
 
-		if path.iscfile(file.name) then
-			return 'ALL_CFLAGS'
+		if fcfg and fcfg.flagsVariable then
```
samsinsane: I'm not entirely sure what `flagsVariable` is, why that would override the `$(ALL_*FLAGS)` flag, or how it's related to PCH files?

tdesveauxPKFX: This is the per-file arguments variable name.

With gmake2, if files have specific arguments, the generator will create a Makefile var `PERFILE_FLAGS_*` to group them.

As an example, You can have this:
`PERFILE_FLAGS_0 = $(ALL_CXXFLAGS) -mf16c`

then your file rule will be
`$(SILENT) $(CXX) $(PERFILE_FLAGS_0) $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"`
instead of
`$(SILENT) $(CXX) $(ALL_CXXFLAGS) $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"`

The way it is implemented is that `ALL_*FLAGS` is contained in `PERFILE_FLAGS_*` if necessary so it doesn't need to be added as well.

PR #1101, opened 2018-05-30
xbox360 module

redorav:
I have migrated all of the xbox 360 code from premake core into this module, plus extended it with new apis with everything I could find publicly. Some things I originally added I have removed since I couldn't find any public references to it. It mostly had to do with Kinect and other flags that I don't think are too important anyway. There is a references.txt file with links. I have verified it works locally.


redorav:
I have an issue with the last test that is failing, I don't know how to properly set it up. If I create a configuration + platform combination like Win32 + Xbox 360 Debug and Release, I get the expected output in the resulting vcxproj. However through the test it always fails to generate it. If you have any suggestions they would be appreciated.

tdesveauxPKFX:
From what I saw, the issue is that `suite.setup` use `test.createWorkspace` which create a workspace and a project. Then, you start your test function in the project space.
That mean that the project does have `Win32` and `Xbox 360` but the workspace (here the visual solution) only has `Win32` (because no platforms were added so that default to `x32`)

Here is how to implement a working test:
```
		local testWorkspace = workspace("MyWorkspace")

		platforms { "Win32", "Xbox 360" }

		filter {"platforms:Xbox 360"}
			system("xbox360")
		filter{}

		project "MyProject"
			uuid "C9135098-6047-8142-B10E-D27E7F73FCB3"
		testWorkspace = test.getWorkspace(testWorkspace)
		sln2005.configurationPlatforms(testWorkspace)
```

I don't know if this is the best way / best practice here but this produce the intended result.

redorav:
Thanks, @tdesveauxPKFX that did the trick

tdesveauxPKFX:
@redorav Just waiting for you to update to master. If you could rebase and skip the merge commit that would be perfect.

modules/xbox360/LICENSE.txt
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2013-2015 Emilio López and individual contributors.
```
samsinsane: Hmm, this is quite interesting to me. This module has never lived as a unique (like `gmake2`, `raw` and `self-test`) and probably should just default to the project license. I don't really know how the legal side of all this works, but it seems strange to me that you can drop files into a project with a different copyright holder. Additionally, a (large?) portion of this code already existed under the project license and I'm pretty sure you can't just change the license on it because you moved it to another file in a different folder.

Also, the copyright years is wrong, it should probably be `2018-2018` or just `2018`.

I'm not trying to be a jerk here, I really don't know how the legality of all this works but it just seems really strange to me. @starkos I would imagine you probably have a better understanding of this? I've always just put it down as `Jason Perkins and …` when I create new files, I assumed it was supposed to be the project copyright notice, not who creates the file?

modules/xbox360/_preload.lua
```diff
@@ -0,0 +1,284 @@
+--
+-- Name:        xbox360/_preload.lua
+-- Purpose:     Define the Xbox 360 APIs
+-- Author:      Emilio López
+-- Copyright:   (c) 2013-2015 Manu Evans and the Premake project
```
samsinsane: Related to previous comment. Year and "main" copyright holder.

modules/xbox360/xbox360_sln2005.lua
```diff
@@ -0,0 +1,10 @@
+--
+-- xbox360/xbox360_sln2005.lua
+-- Xbox 360 integration for vstudio.
+-- Copyright (c) 2012-2015 Manu Evans and the Premake project
```
samsinsane: Related to first comment. Year and "main" copyright holder.

modules/xbox360/xbox360_vcxproj.lua
```diff
@@ -0,0 +1,439 @@
+--
+-- xbox360/xbox360_vcxproj.lua
+-- Xbox 360 integration for vstudio.
+-- Copyright (c) 2012-2015 Manu Evans and the Premake project
```
samsinsane: Related to first comment. Year and "main" copyright holder.

modules/xbox360/xbox360_vstudio.lua
```diff
@@ -0,0 +1,33 @@
+--
+-- xbox360/xbox360_vstudio.lua
+-- Xbox 360 integration for vstudio.
+-- Copyright (c) 2012-2015 Emilio López and the Premake project
```
samsinsane: Related to first comment. Year and "main" copyright holder.

redorav: I will reply to this comment but it relates to all the previous ones. 

>I'm not trying to be a jerk here, I really don't know how the legality of all this works but it just seems really strange to me. @starkos I would imagine you probably have a better understanding of this? I've always just put it down as Jason Perkins and … when I create new files, I assumed it was supposed to be the project copyright notice, not who creates the file?

To be honest, I didn't know what to do here, and as you can see I have been a bit inconsistent about it too. All of that was copied from the android module which I used as baseline, and used to say Manu Evans. At some point I went through the code and thought putting his name would mean @TurkeyMan is now somehow responsible for the code and maintenance, which I didn't want to impose, but on the other hand I surely don't own any copyright to the code, by just contributing to an existing codebase.

I would be happy with just putting author and keeping the Jason Perkins line for the copyright, the idea being that anyone with a question or change knows who to address. The Android module does contain a different name, though, which got me thinking about changing it.

starkos: Certainly assigning the copyright to me makes it easier, even if I am only one of several authors. I've never been clear on the correct way to do it all from a laws and lawyers standpoint.

And as you suggested, perhaps this should be split out to a standalone module and list on the  [Modules](https://github.com/premake/premake-core/wiki/Modules) page instead? We could still host it under the Premake team account if that's easier, or @redorav can host it under his own account if they'd like?

modules/xbox360/xbox360_vcxproj.lua
redorav: I did indeed. Amended 👍

modules/xbox360/xbox360_vstudio.lua
redorav: Like I said, I don't mind the copyright aspect of it, it was more a way of specifying author that wasn't too thought through. I've changed it already.

In terms of core or separate module, I was inclined to keep in core where it has always been, but I'm not opposed to hosting it myself. I just thought that it has been there all along anyway doing no harm.

I don't know what the criteria are for having something in core or not. One thing I've always wondered is if premake could reference them as git submodules, compiled in the final distribution of premake but not a part of the repository. That way you get maximum out of the box support within premake without the burden of maintenance. I kind of like the fact that you have a single executable that can generate everything, without having extra lua files lying around.

starkos: > One thing I've always wondered is if premake could reference them as git submodules

We did take that approach initially, but it ended up being a maintenance nightmare. Things are much easier now.

> In terms of core or separate module, I was inclined to keep in core where it has always been, but I'm not opposed to hosting it myself. I just thought that it has been there all along anyway doing no harm.

There is no hard or fast rule for this. Separate repositories helps spread out the maintenance load, which helps keep the core developers from getting overwhelmed. And since console dev kits are pretty specialized things, I think it might make sense to split it out. But I'll defer to the folks who are actually doing console development, I'm not currently.

redorav: I think you guys are right, for me the reason to have it as a core module was more convenience and ease of use (single executable), but I understand it's an old console and more of a hobby thing these days. I think I'll redo this PR over the weekend, where I keep the removal of xbox 360 from core and don't commit any part of the module, and then create a repository for it that I maintain.

In terms of the single executable, I think I'll create a new repository that pulls in premake core as a submodule, and the different modules I'm interested in as submodules too, which I can then compile and use that as the executable for my personal stuff, keeping things separate.

PR #1104, opened 2018-06-01
Xcode Fixes

erincatto:
Fix cppdialect=C++11 (Xcode 9.3.1 does not like c++0x)
Fix angle bracket include complaints for building Premake in Xcode.


tdesveauxPKFX:
Out of curiosity, what is xcode complaining about with `#include <>`?

Otherwise, this looks fine to me, will still leave approval to someone more familiar with XCode.
Might also make sense to split this in two PR?

modules/xcode/xcode_common.lua
```diff
@@ -1032,7 +1032,7 @@
 	xcode.cppLanguageStandards = {
 		["Default"] = "compiler-default",  -- explicit compiler default
 		["C++98"] = "c++98",
-		["C++11"] = "c++0x",      -- Xcode project GUI uses c++0x, but c++11 also works
+		["C++11"] = "c++11",      -- Xcode project GUI uses c++0x, but c++11 also works
```
samsinsane: This comment is no longer relevant.

erincatto:
Here is an example error:
/premake-core/binmodules/example/main.c:1:10: 'luashim.h' file not found with <angled> include; use "quotes" instead

modules/xcode/xcode_common.lua
erincatto: Done

PR #1105, opened 2018-06-01
gmake: Add rules and dependencies for directory creation

tdesveauxPKFX:


tdesveauxPKFX:
This was one of the issues related to #1091. 
I agree with you on making people go with `gmake2`, however this was a minor fix and was trivial to fix.

I also noticed that we still use `gmake` in Bootstrap.mak. I will submit a PR to use `gmake2` instead.

PR #1106, opened 2018-06-02
Object files should be marked as object files.

TurkeyMan:


tdesveauxPKFX:
Hum, #671 and #881 already brought that up.

Have something changed since then?

TurkeyMan:
Truestory...

PR #1109, opened 2018-06-02
Add 'excludefrombuild' API.

TurkeyMan:


src/_premake_init.lua
```diff
@@ -482,6 +482,13 @@
 		kind = "boolean"
 	}
 
+	api.register {
+		name = "excludefrombuild",
```
tdesveauxPKFX: Do we need a new API here? Can't this be done through `excludes` (`removefiles`) like vstudio?

samsinsane: That wouldn't result in the file being in the "solution explorer", it literally removes the file. This API just stops the file from being built, although I'm fairly certain that this already exists somewhere.

samsinsane: Oh yeah, it's a flag `flags { "ExcludeFromBuild" }`.

samsinsane: This looks like the exact same code as #243 which was closed as it wasn't the desired solution to the problem.

tdesveauxPKFX: I was confused because in the scripts we use at work we have some
```
filter { "platforms:not x" }
    excludes { "file" }
```
for platform dependent code and it behave as you said. File are in explorer view but not built.
It's actually us that does the wrong thing here.

I'd say that leveraging file filter and the flag/boolean API would make more sense but I do not have a strong opinion either way.

starkos:
There is already an `ExcludeFromBuild` flag, which this PR sets under the hood. I'd rather see the flag deprecated, and any under-the-hood logic referencing it switched over to check this API value instead (all flags must die).

TurkeyMan:
@tvandijck might want to update this patch... or not? :)

tvandijck:
the 'exludes' api already does a similar thing, which is what I never realized when making this change. It's in use by WoW and HotS I beleive, but they should just change to using pre-existing API's. either through the filter, or through excludes.

I don't feel this API as is has a right to exist.

starkos:
`excludes` (or `remove files`) is different, that removes the file from the project entirely. The `ExcludeFromBuild` flag leaves the file in the project, so it will appear in the IDE and can still be easily edited, but doesn't include it when building. The feature is definitely useful, I'd just like to see the flag go away.

tvandijck:
True, but if you exclude it from only one configuration/platform, it acts like the "ExcludeFromBuild" in practice. the file will still appear in the IDE, but disabled for the configuration where you removed it. So I think that covers the use-case for which I introduced this particular API.

starkos:
That makes sense. So that plus improved `buildaction` support should cover all of the cases, got it.

PR #1110, opened 2018-06-02
Add warning for deprecated toolset identifiers.

TurkeyMan:


src/base/tools.lua
```diff
@@ -31,6 +31,7 @@
 
 	function p.tools.normalize(identifier)
 		if identifier:startswith("v") then -- TODO: this should be deprecated?
+			p.warnOnce(identifier, "toolset '%s' is deprecated, toolset 'msc-%s' instead.", identifier, identifier)
```
samsinsane: I'm fairly certain we've had this discussion before, and it was decided that we wanted `vXXX` to be accepted forever, so deprecation made absolutely no sense.

starkos: We did that have discussion; I wanted to provide that as a Visual Studio friendly value. The quorum of core developers is larger now, and I don't mind if I'm overruled.

tdesveauxPKFX: I see the value in keeping this.

What are the arguments to deprecate it?

starkos:
I believe the argument against was that it would be easier to understand if all of the values followed a consistent format.

tdesveauxPKFX:
`msc` is appended when the `toolset` API is used, so we don't have to worry about it elsewhere. Any inconsistency would be user side and this is mentioned in the wiki. I don't think it will one day collide with something else and it provide user with something that is familiar to them.

I personally think we should keep it.

PR #1111, opened 2018-06-02
Fix building Premake using mingw

tdesveauxPKFX:
Only works with MinGW64 due to an error in luasocket.

I'd appreciate if someone that have used MinGW before could take a look.

### DIFF
--- a/contrib/libzip/premake5.lua
+++ b/contrib/libzip/premake5.lua
@@ -16,6 +16,8 @@ project "zip-lib"
 
 	filter "system:windows"
 		defines { "_WINDOWS" }
+	filter { "system:windows", "toolset:mingw" }
+		defines { "HAVE_SSIZE_T_LIBZIP" }
 
 	filter "system:macosx"
 		defines { "HAVE_SSIZE_T_LIBZIP" }
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -52,13 +52,13 @@ none:
 	@echo "   osx linux bsd"
 
 mingw: $(SRC)
-	$(SILENT) rm -rf ./bin
-	$(SILENT) rm -rf ./build
-	$(SILENT) rm -rf ./obj
-	mkdir -p build/bootstrap
-	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lole32
+	$(SILENT) if exist .\bin rmdir /s /q .\bin
+	$(SILENT) if exist .\build rmdir /s /q .\build
+	$(SILENT) if exist .\obj rmdir /s /q .\obj
+	if not exist build\bootstrap (mkdir build\bootstrap)
+	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lole32 -lversion
 	./build/bootstrap/premake_bootstrap embed
-	./build/bootstrap/premake_bootstrap --os=windows --to=build/bootstrap gmake
+	./build/bootstrap/premake_bootstrap --os=windows --to=build/bootstrap gmake2 --cc=mingw
 	$(MAKE) -C build/bootstrap config=$(CONFIG)_$(PLATFORM)
 
 osx: $(SRC)
--- a/premake5.lua
+++ b/premake5.lua
@@ -126,7 +126,11 @@
 			defines     { "_CRT_SECURE_NO_DEPRECATE", "_CRT_SECURE_NO_WARNINGS", "_CRT_NONSTDC_NO_WARNINGS" }
 
 		filter { "system:windows", "configurations:Release" }
-			flags       { "NoIncrementalLink", "LinkTimeOptimization" }
+			flags       { "NoIncrementalLink" }
+
+		-- MinGW AR does not handle LTO out of the box and need a plugin to be setup
+		filter { "system:windows", "configurations:Release", "toolset:not mingw" }
+			flags		{ "LinkTimeOptimization" }
 
 	project "Premake5"
 		targetname  "premake5"
@@ -169,6 +173,8 @@
 
 		filter "system:windows"
 			links       { "ole32", "ws2_32", "advapi32" }
+		filter "toolset:mingw"
+			links		{ "version", "crypt32" }
 
 		filter "system:linux or bsd or hurd"
 			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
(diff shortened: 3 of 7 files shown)


tdesveauxPKFX:
Well, I installed both MinGW32 and 64 and it works with MinGW64.

MinGW32 has an error related to curl if I remember correctly.

I expected a review on this part:
```
filter { "system:windows", "action:gmake*" }
```

I think it is pretty safe to assume that Makefiles on Windows are used with MinGW but just wanted to confirm it.

starkos:
>  think it is pretty safe to assume that Makefiles on Windows are used with MinGW but just wanted to confirm it.

Could be Cygwin too, but I would still expect those libraries to be present. I think this should be okay.

tdesveauxPKFX:
I though Cygwin used MinGW, I'm really confused about all these things. Let me test Cygwin to be sure;

If everything works fine, I will merge this myself if that's ok with you.

tdesveauxPKFX:
Well, it seems to be an issue on cygwin side. If someone want to build with MinGW, they should get MinGW not from cygwin.

If someone want to look into cygwin, I had to add change 
https://github.com/premake/premake-core/blob/d6dfaa3bade3b6017f77aedd6728aadcf2311b2c/src/host/premake.h#L20
to
```
#if defined(__linux__) || defined(__CYGWIN__)
```

Anyway, that still fix an issue so I'm inclined to merge as is. I will merge this tomorrow to let the opportunity for people to disagree.

tdesveauxPKFX:
> I also have an issue with mingw64 gcc in cygwin. Here is the log:

Disregard this... I just forgot to add PLATFORM=x64 so it was building and looking for 32bit libraries.

I just added a fix in os_stat for MinGW32 if the curl problem is ever fixed and the Cygwin identification.

I tried to build with Cygwin for a correctly (with the PLATFORM=x64) and I currently have a seg fault on linking. Otherwise everything build.

TurkeyMan:
> I think it is pretty safe to assume that Makefiles on Windows are used with MinGW but just wanted to confirm it.

I'd like to think that's not a reasonable assumption; msc should work with makefiles too, but it's definitely not used often.

contrib/libzip/premake5.lua
```diff
@@ -16,6 +16,8 @@ project "zip-lib"
 
 	filter "system:windows"
 		defines { "_WINDOWS" }
+	filter { "system:windows", "action:gmake**" }
```
TurkeyMan: Yeah, I don't like this strategy... I feel like mingw/cygwin might be systems, but I'm not entirely sure :/

src/host/premake.h
```diff
@@ -17,7 +17,7 @@
 
 /* Identify the current platform I'm not sure how to reliably detect
  * Windows but since it is the most common I use it as the default */
-#if defined(__linux__)
+#if defined(__linux__) || defined(__CYGWIN__)
```
TurkeyMan: Why do you want `PLATFORM_STRING` to be `"linux"` under cygwin?

TurkeyMan: cygwin == linux is kinda like saying bsd == linux no?

tvandijck: we don't... but the problem for example is that curl uses autoconf to generate most of it config header files. Since premake does not have such a facility, we pre-generated these header files. and checked them in.. and well, obviously they are wrong for some platforms. That's why for example mingw32 doesn't seem to work, because most of those header files have been generated on 64-bit linux variants....

it's a common recurring issue with 3rd party libraries that do this kind of thing, and we basically have to hand craft a header file that satisfies all platforms. It's why I created an autoconf module to try and somewhat remedy this. It was mostly a prototype proof of concept, and it's certainly not impossible to add such facilities to premake. I just don't like it. I feel like if you have to resort to autoconf tools, your doing something wrong.

tdesveauxPKFX:
> I'd like to think that's not a reasonable assumption; msc should work with makefiles too, but it's definitely not used often.

One thing to consider is that Premake generated Makefiles cannot be used with NMake. So we would be talking about using a MinGW/Cygwin makefile binary to run a msc Makefile.

However, I agree that it's a somewhat bold assumption but I didn't see an alternative.
I never used MinGW nor Cygwin so this is a result of some hacking around on my part.
If someone want to take this on, I'd be grateful. Otherwise, I think I need to take another look at this.

src/host/premake.h
tdesveauxPKFX: > That's why for example mingw32 doesn't seem to work

Not sure this is the cause but I should check it out. I think I saw a GitHub issue somewhere where someone had the same issue and the cause was MinGW32.

> Why do you want `PLATFORM_STRING` to be `"linux"` under cygwin?

Building with Cygwin uses posix libraries. But this is an awful solution and should absolutely not be merged into master.
I had second though about this PR yesterday and definitely want to take another look at it now.

contrib/libzip/premake5.lua
samsinsane: > I feel like mingw/cygwin might be systems

I disagree, is there anything stopping you from using either of these to target Android or Linux? I'd be more inclined to say that these should fall under `toolset`, or a subtype of `toolset`/`action`.

TurkeyMan: Maybe... the distinction between mingw/cygwin and 'windows' is that they have a different runtime, different ABI, different libs. They are *definitely* a different platform than 'windows' (assuming an MSVC windows world).

I have often wanted to generate a makefile that uses MSVC because I want to share the same build system across platforms. I don't usually do it, but I feel like it should be possible to produce a makefile that supports the normal 'windows' ecosystem, in addition to mingw/cygwin.

TurkeyMan: Yeah nar. It's definitely not a toolset thinking on it. Mingw/cygwin are undeniably different platforms... or 'systems'?
Building mingw from windows is no different than building android from windows. The entire pipeline is incompatible with 'windows' by premakes conventional assumptions.

TurkeyMan:
> So we would be talking about using a MinGW/Cygwin makefile binary to run a msc Makefile.

That's entirely feasible. Although supporting nmake is something I'd like to clean up too one day!

Yeah, I'm not happy with the assumption. It's rarely used, so I have no doubt that when it's committed, it'll 'stick', and it'll be that way forever :P

contrib/libzip/premake5.lua
samsinsane: I really don't know enough about either of them to push for anything in particular, they just don't seem like `system`s to me.

tdesveauxPKFX:
Finally some time to look into this.

After some thoughts, I'd say that MinGW should be handled as a toolset.

In the same way you can build either with gcc or clang on Linux, you should be able to build with cl or MinGW on Windows with a Makefile. Well not really in the same way as MinGW and cl can't use the same options like gcc and clang does.

This way, the bootstrap can just use `--cc=mingw`.

samsinsane:
@tdesveauxPKFX @TurkeyMan are we good to merge this?

starkos:
I rebased this to master, made some additional fixes, and created PR #1420. It isn't perfect, but it does provide a build path for MinGW users. Thanks to @tdesveaux for doing the hard parts.

PR #1112, opened 2018-06-04
Android fixes from bliz

TurkeyMan:


modules/android/tests/test_config_props.lua
```diff
@@ -0,0 +1,73 @@
...
+
+
+--
+-- Check the structure with the default project values.
+--
+
+	function suite.structureIsCorrect_onDefaultValues()
+		prepare()
+		test.capture [[
+<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Android'" Label="Configuration">
```
samsinsane: This doesn't seem right to me, the value for `$(Platform)` was always `x86`, `x64`, `ARM` or `ARM64` for me, why is it `Android` now?

TurkeyMan:
@tvandijck It was getting out of hand... something had to be done >_<
I was not responsible for any of these changes (I make changes to github first, bliz second), and there's quite a few in here that I'm not sure about the intent. Kinda hard to know what should be pushed back or not.

modules/android/tests/test_config_props.lua
TurkeyMan: Not my code! The tests run ;)

TurkeyMan: How can Android use the arch as the platform, and not be confused with windows configurations (that are also named for the arch)?

pdoane: There's something wrong with the test cases in general.  We already had test cases for rtti and there a bug fixes here specifically for rtti.  Somehow the tests are working differently in isolation and it might be why the platform is weird here too.

samsinsane: Yeah, this looks like the vs-android configuration. Microsoft uses the platform to indicate the architecture, not the system - they use `ApplicationType` to indicate the system. I would say that the action isn't being set correctly for the tests, it does seem to set `p.action.set("vs2015")` in the `suite.setup() ` which mightn't be the correct way of doing this in the unit tests. Might be worth giving "vs2015" a search in the tests folder and seeing how it's done elsewhere. :+1:

samsinsane:
@TurkeyMan have you had a chance to fix this up?

starkos:
I think this one needs to be closed. It looks like some of the issues it addresses have already been fixed in other PRs, and there hasn't been any movement here.

Happy to see this one re-opened, but the individual fixes really should be broken out into separate PRs so they can be approved independently. Going forward, I'll be encouraging more pushback in that direction.

PR #1113, opened 2018-06-07
Support buildaction for C++ projects

TurkeyMan:
Had to remove 'allowed', since the list of possible build actions is extensible and kinda arbitrary.


TurkeyMan:
Yeah, it probably needs to result in `None` if a bad action name name was given.

I think C# also has an arbitrary/extensible set of buildactions just the same as C++ no?
One lame thing about the C# build actions is that the existing list of actions are MAPPED to the real action names. I think C# should accept the real action names, and not re-name them as it has.

Also, I think the C# code should really be refactored into rules, just like the C++ project code.

TurkeyMan:
It's already handled: https://github.com/premake/premake-core/blob/master/src/tools/dotnet.lua#L63
If buildaction is given some value that isn't supported, it may end out with `None`, which is what the new C++ implementation does.
That said, there's also a bunch of other logic and matching going on in that code for other conditions, but it wouldn't be right to change any of that.

PR #1117, opened 2018-06-15
Fix Xcode autocompletion with excluded files

sp-jordi-vilalta:
Xcode's autocompletion gets disabled when a file is added to the exclude list. Unfortunately, the compilation units module adds the ExcludeFromBuild flag to all the source files, and the Xcode module translates it into adding them to the Xcode exclude list.

This commit checks if a file must be excluded in all the configurations and in that case it removes the file from the target, so there's no need to add it to any exclusion list (letting Xcode index the file). In case a file is excluded in just some configurations, it prints a warning and keeps the old behavior for that file.


modules/xcode/xcode_project.lua
```diff
@@ -15,6 +15,39 @@
...
+
+		local value
+		for cfg in premake.project.eachconfig(prj) do
+			local filecfg = premake.fileconfig.getconfig(node, cfg)
+			if filecfg then
+				local newValue = not not filecfg.flags.ExcludeFromBuild
+				if value == nil then
+					value = newValue
+				elseif value ~= newValue then
+					print("WARNING: " .. node.name .. " is excluded in just some configurations. Autocompletion will not work correctly on this file in Xcode.")
```
samsinsane: This should probably use `p.warn` instead?

sp-jordi-vilalta: Ok, changed.

PR #1119, opened 2018-06-19
Fix C# buildaction

TurkeyMan:
`buildaction` should be able to override what the file extension says.
This also generalises support for `buildaction` to complement the recent change to support it in C/C++ projects.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -205,6 +205,18 @@
 	}
 
 	api.register {
+		name = "copytooutputdirectory",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"Never",
+			"Always",
+			"PreserveNewest"
+		}
+	}
+
+	api.register {
 		name = "debugargs",
 		scope = "config",
 		kind = "list:string",
@@ -540,6 +552,17 @@
 	}
 
 	api.register {
+		name = "includeinvsix",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"On",
+			"Off"
+		}
+	}
+
+	api.register {
 		name = "inlining",
 		scope = "config",
 		kind = "string",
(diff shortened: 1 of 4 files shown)


TurkeyMan:
I added additional work on the end.

src/_premake_init.lua
```diff
@@ -876,6 +916,12 @@
 		}
 	}
 
+	api.register {
+		name = "resourcename",
```
samsinsane: This is unused.

src/_premake_init.lua
```diff
@@ -204,6 +204,18 @@
 		tokens = true,
 	}
 
+	api.register {
+		name = "copytooutputdirectory",
```
samsinsane: Could this be an extension to the `buildaction` instead? Some of the other new APIs also seem like they could be extensions of `buildaction`. Additionally, there isn't really any way to override the usage of these APIs, they appear to only be used in a single, ever growing function.

src/tools/dotnet.lua
```diff
@@ -41,26 +41,51 @@
...
-					info.action = "None"
-				else
-					info.action = "ApplicationDefinition"
-				end
-			else
-				info.action = "Page"
-			end
-		else
-			info.action = "None"
+		local actionMap = {
```
samsinsane: This and `extensionMap` would probably be better stored outside of the function, and correctly scoped. Being able to override things easily has been one of the best things about the VStudio module.

src/_premake_init.lua
starkos: Is this different than `copylocal`?

src/tools/dotnet.lua
TurkeyMan: Good point.

TurkeyMan:
I'm just trying to chase down the author for comment on those points...

samsinsane:
@TurkeyMan were you able to catch up with the author?

starkos:
@TurkeyMan - what's the status on this one? Looks like there was some follow-up that may have stalled?

starkos:
I'm going to go ahead and close this one as stalled for now. Feel free to reopen if/when you get a chance to review.

PR #1131, opened 2018-07-05
Fixed custom build rules loosing their dependency information when building up-to-date projects / files.

Dandielo:
Fixes invalid behaviour while using custom build rules generated with premake5. 
If a build was forced on a up-to-date project, or no changes where reported for the custom build rule, the tlog files associated with the rule would be cleared from any dependency information. 
After that happens changing files which should be set as 'additional dependencies' wouldn't trigger a rebuild of files associated with custom build rules.


Dandielo:
Okay, i found one new issue with this one. Now it does look like it properly checks timestamps of all input files, but it does also mark the project as always out of date. Will update this requests soon

Dandielo:
Okay, so generally the problem was that msbuild does not like when TLog files store a relative path. So I've managed to transform all relative paths to fullpaths just using their xml api.
I've also added a condition to write to a TLog file only if the given lines which we want to write do not exist there already. 
Now rules always properly see changes to any input dependency, and do not cause a out-of-date false-positive.

Dandielo:
Sure, I will add the tests and squish the commits, just atm can't tell when it's gonna be. 
Anyway thanks for the review

Issue #1121, opened 2018-06-22
Solution project paths are mapped incorrectly


JSandusky:
Occurs in master built for Win-x64, does not occur in the last release (5.0.0 Alpha 12 Windows).

Edit: for generating VS2015 sln/projects.

Given code like the following:

    workspace "Pylon"
        location "../."

    project "MathGeoLib"
        location "../Generated"

with a folder structure of:

    - Root
        - premake5.exe
        - premake5.lua
        - Premake
            - _Pylon.lua
             - ... more build scripts ...
            - MathGeoLib.lua
            - ... more build scripts ...

The intent is for the solution file to be placed in the project's root directory and the projects into a *Generated* folder in the *root*. 

With the last release of Premake5 this worked. Checked out yesterday and built from source this now emits a solution file whose paths are all written along the lines of:

    Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "MathGeoLib", "..\Generated\MathGeoLib.vcxproj", "{C19A7F90-AD52-C1DD-165A-9A3902315CCE}"
    EndProject

when it should be:

    Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "MathGeoLib", "Generated\MathGeoLib.vcxproj", "{C19A7F90-AD52-C1DD-165A-9A3902315CCE}"
    EndProject

Only the solution paths are incorrect. Everything else is working fine (include paths, link targets, etc). Manually fixing the paths works as a work-around.

tdesveauxPKFX:
It looks like there is an issue in path.normalize with path ending with `.`

I will try to patch it but in the mean time you can simply remove the trailing `.` in your workspace location.

samsinsane:
This should be fixed in master, @JSandusky are you able to confirm that this has resolved your issue?

JSandusky:
@samsinsane, yes - it does fix it.

PR #1132, opened 2018-07-06
Refactor path.normalize and handle path ending with .

tdesveauxPKFX:
Related to #1121

### DIFF
--- a/src/host/path_normalize.c
+++ b/src/host/path_normalize.c
@@ -8,128 +8,124 @@
 #include <ctype.h>
 #include <string.h>
 
-// isspace custom version (visual c++'s one doesn't like utf8 characters)
-static int is_space(char c)
-{
-	return  (c >= 9 && c <= 13) || c == 32;
-}
+#define	IS_SEP(__c)			((__c) == '/' || (__c) == '\\')
 
-static void* normalize_substring(const char* str, const char* endPtr, char* writePtr) {
-	const char* const source = str;
-	const char* const writeBegin = writePtr;
-	const char* ptr;
-	char last = 0;
-	char ch;
+#define IS_UPPER_ALPHA(__c)	((__c) >= 'A' && (__c) <= 'Z')
+#define IS_LOWER_ALPHA(__c)	((__c) >= 'a' && (__c) <= 'z')
+#define IS_ALPHA(__c)		(IS_UPPER_ALPHA(__c) || IS_LOWER_ALPHA(__c))
 
-	while (str != endPtr) {
-		ch = (*str);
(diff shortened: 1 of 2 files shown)


src/host/path_normalize.c
```diff
@@ -8,128 +8,124 @@
...
 
-		/* filter out /./ */
-		if (ch == '/' && last == '.') {
-			ptr = str - 2;
-			if (*ptr == '/') {	// there is no need to check whether ptr >= source since all the leading ./ will be skipped in path_normalize
-				if (ptr - 1 < source || *(ptr - 1) != ':') {
-					--writePtr;
-				}
+	const char * const dstRoot = dstPtr;
+	unsigned int folderDepth = 0;
```
samsinsane: These probably need to be at the top of the function - some people build Premake with compilers that don't support this. I can't recall what, but we've had issues opened about it before - it might be older versions of VS.

src/host/path_normalize.c
```diff
@@ -8,128 +8,124 @@
...
-	const char* path = luaL_checkstring(L, 1);
-	const char* readPtr = path;
-	char buffer[0x4000] = { 0 };
-	char* writePtr = buffer;
-	const char* endPtr;
+	const char	*path = luaL_checkstring(L, 1);
+	const char	*readPtr = path;
+	char		buffer[0x4000] = { 0 };
+	char		*writePtr = buffer;
+	const char	*endPtr;
```
samsinsane: This spacing seems inconsistent with the rest of the code base?

src/host/path_normalize.c
```diff
@@ -8,128 +8,124 @@
...
+	}
 
-			++str;
-			continue;
+	// Handle path starting with a sep (C:/ or /)
+	if (IS_SEP(*srcPtr))
+	{
+		++srcPtr;
+		*(dstPtr++) = '/';
+		// Handle path starting with //
```
samsinsane: There doesn't appear to be any unit tests for this.

src/host/path_normalize.c
samsinsane: The issue I was thinking about is #696, so VS2010 will most like have the issue if it's a problem in VS2012.

tdesveauxPKFX: Indeed, this is an issue with C89 standard... I will fix that whenever I have some time.

There are other part of the code that have this issue, maybe set appveyor to build with vs2012 if possible?

samsinsane: Oh, if it's elsewhere, let's just merge this! Thanks for looking into it!

PR #1140, opened 2018-07-26
Remove extra semicolons

carusogabriel:


carusogabriel:
@samsinsane I did in PHP’s Core, but, as we use it as a library there, we came with the same issue. Maybe we can find where these files come from and remove them?

samsinsane:
- [curl](https://github.com/curl/curl/blob/e78f2cfe56c39a6c32191c207aae683de0e9a042/lib/x509asn1.c#L715) has been fixed
- [libzip](https://github.com/nih-at/libzip/blob/d90df6cf4bbc96442469104e0c6a717fa231a92b/lib/zip_source_crc.c#L142-L152) was probably modified by someone accidentally. I see no reason to modify it again.
- The mbedtls files are test files so I see no reason to modify them, and they haven't been updated in the main repo; [here](https://github.com/ARMmbed/mbedtls/blob/b363382ba4c0489b6045112dbaf100758ec72cc2/tests/scripts/gen_ctr_drbg.pl#L41), [here](https://github.com/ARMmbed/mbedtls/blob/b363382ba4c0489b6045112dbaf100758ec72cc2/tests/scripts/gen_gcm_decrypt.pl#L61), and [here](https://github.com/ARMmbed/mbedtls/blob/b363382ba4c0489b6045112dbaf100758ec72cc2/tests/scripts/gen_gcm_encrypt.pl#L44).

I'm kind of inclined to just leave our copies as they are.

PR #1142, opened 2018-07-27
Support for CSDialect

TurkeyMan:
@Evilsmevil This is your great work in the wild...

### DIFF
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -687,7 +687,6 @@
 		end
 	end
 
-
 	function dotnetbase.targetFrameworkVersion(cfg)
 		local action = p.action.current()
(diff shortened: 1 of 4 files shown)


modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -687,7 +687,6 @@
 		end
 	end
 
-
```
samsinsane: I'm fairly certain that there's normally two lines between functions. This also applies to the new lines below for `dotnetbase.csversion`.

src/_premake_init.lua
```diff
@@ -590,6 +590,12 @@
 		kind = "string",
 	}
 
+	api.register {
+		name = "csversion",
```
samsinsane: These are supposed to be in alphabetical order, I think `dotnetframework` was renamed from `framework` which is why it's here and not up with the other `d*` APIs.

PR #1144, opened 2018-07-31
Add `sharedlibtype` in xcode4

tempura-sukiyaki:
Add support for CocoaBundle(.bundle) and Framework(.framework).

### DIFF
--- a/src/base/config.lua
+++ b/src/base/config.lua
@@ -50,7 +50,7 @@
 
 		if cfg.system == p.MACOSX and (kind == p.WINDOWEDAPP or (kind == p.SHAREDLIB and cfg.sharedlibtype)) then
 			bundlename = basename .. extension
-		  bundlepath = path.join(bundlename, iif(kind == p.SHAREDLIB and cfg.sharedlibtype == "OSXFramework", "Versions/A", "Contents/MacOS"))
+			bundlepath = path.join(bundlename, iif(kind == p.SHAREDLIB and cfg.sharedlibtype == "OSXFramework", "Versions/A", "Contents/MacOS"))
 		end
 
 		local info = {}
--- a/src/base/config.lua
+++ b/src/base/config.lua
@@ -48,7 +48,7 @@
 		local bundlename = ""
 		local bundlepath = ""
 
-		if cfg.system == p.MACOSX and kind == p.WINDOWEDAPP then
+		if cfg.system == p.MACOSX and (kind == p.WINDOWEDAPP or (kind == p.SHAREDLIB and cfg.sharedlibtype)) then
 			bundlename = basename .. extension
 			bundlepath = path.join(bundlename, "Contents/MacOS")
 		end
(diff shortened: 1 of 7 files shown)


src/base/config.lua
```diff
@@ -48,7 +48,7 @@
 		local bundlename = ""
 		local bundlepath = ""
 
-		if cfg.system == p.MACOSX and kind == p.WINDOWEDAPP then
+		if cfg.system == p.MACOSX and (kind == p.WINDOWEDAPP or (kind == p.SHAREDLIB and cfg.sharedlibtype)) then
```
samsinsane: I might be mistaken here, but this will place the DyLib in `proj.framework/Contents/MacOS` correct? At work the framework I generate doesn't have that path, it's laid out like this:
```
- Headers/
  - ...
- Versions/
  - A/
    - Headers/
    - <proj dylib>
  - Current/
    - Headers/
    - <proj dylib>
- <proj dylib>
```
I think the `Versions` folder is just a bunch of symlinks though. Am I doing it wrong, or are there multiple ways of doing this? I'm certain that I copied an existing Framework when doing this, perhaps it was doing it the old way?

tempura-sukiyaki: You're right.
I would like to modify it to "proj.framework/Versions/A".
Do you think that is good?

samsinsane: I don't think my hacks at work to generate a Framework did anything special to generate that structure, I believe it was automatically done by XCode. I'll have another look at what I do tomorrow if I get a chance, but you might only need to do this when `cfg.system == p.MACOSX and (kind == p.WINDOWEDAPP or (kind == p.SHAREDLIB and cfg.sharedlibtype == "OSXBundle"))`? If you have a Mac available you might be able to test this out faster than I can get back to you about what I do. :)

tempura-sukiyaki: `info.bundlepath` is a token like `% {cfg.buildtarget.bundlepath}`, isn't it?

Xcode outputs the entity of dylib to "proj.framework/Versions/A".
And create a symbolic link for dylib in "proj.framework".
"proj.framework/Versions/Current" is a symbolic link for "proj.framework/Versions/A".

As such, when `sharedlibtype` is` OSXFramework`, I think it is better to make it "proj.framework/Versions/A".

```lua
		if cfg.system == p.MACOSX and (kind == p.WINDOWEDAPP or (kind == p.SHAREDLIB and cfg.sharedlibtype)) then
			bundlename = basename .. extension
			bundlepath = path.join(bundlename, iif(kind == p.SHAREDLIB and cfg.sharedlibtype == "OSXFramework", "Versions/A", "Contents/MacOS"))
		end
```

samsinsane: I'm so sorry, I though that these variables were used to output something specific into the XCode project files. I didn't realize it was part of the token system instead, what you're suggesting makes sense!

starkos:
Nice!

PR #1145, opened 2018-08-01
Add support `kind:Utility` in xcode4

tempura-sukiyaki:
`kind:Utility` is output as PBXAggregateTarget.


starkos:
I'm willing to defer…these are good contributions, and they can always be fixed up.

samsinsane:
@starkos There are unit tests in the commit already, was there a specific thing you wanted them to test?

PR #1159, opened 2018-09-05
! allow an action to return an error code

neo2buha:
since I'm using premake to do some scripting it's useful for some actions to return error code 
let me know if this can break anything else


samsinsane:
@starkos any thoughts? I feel that if there's any kind of return value, all of the functions should be able to return errors. Additionally, knowing which function failed would probably be required, `return ret, "<function> failed"` or something?

samsinsane:
@neo2buha it looks like you've pushed the pepperfish commit into this branch.

neo2buha:
Thanks @saminsane.. I didn't noticed it..
"grumbles something something git"
"grumbles back in my day we were doing version control on 1.44Mb disks"
"google how to backout a commit in git"

neo2buha:
Does anyone else has any thoughts on this :D ?

tdesveauxPKFX:
As @samsinsane, I feel if you return a result code here, it should do so for every function but it makes little sense for the other ones. At least not returning `nil` if there is not an execute execute function would be better.

I do not use this feature personally, what are you reasons to not handle errors inside the execute?

neo2buha:
@tdesveauxPKFX Thanks for the quick feedback 
Not sure I understand "At least not returning nil if there is not an execute "
I think this is what i'm doing ...
```
		if a.execute then
			return ret
		end
```
or am I missing something ?
I use this feature because I have scripts that I run on a build server so I want premake to return an error code so the build can fail if something goes wrong.  "handle errors inside the execute?" Is there any other way to make premake return an error code ?

tdesveauxPKFX:
(my comment was not clear, hope this one will be better)

With you current implementation, if the action does not have an `execute` function, `action.call` will return `nil`.
Whereas, if it does, it will return whatever the `execute` function returns (I assume an int in your case?). This create a discrepancy between actions.
Although we could consider `nil` as the "good" return value.

As for handling errors inside execute, I use `premake.error` which raise an exception. (I think you can catch it with `pcall` if needed, but never tried myself).

neo2buha:
never mind this will not work ...
I had this change in an older version of premake and just backported it here ...
But i now noticed that _premake_main will always return 0 
```
function _premake_main()
		p.callArray(m.elements)
		return 0
	end 
```
I'll give premake error a try and see if it works ! 
Thanks ! 
I'll close this !

Issue #1157, opened 2018-09-05
rule propertydefinition separator conflicting with vs201x Separator


WorldofBay:
The optional propertydefinition field "separator" for list properties is conflicting with the optional vs201x list Separator.

They do different things but you can only use both or none.

Propertydefinition separator:
That field is the string used to write the table down so that it is read as table. For vs it needs to be a ';', but the default is ' '. So to use the list functionality in vs it has to be set. **This is not a conversion** (could be abused to be though).

Vs list Separator:
This field is used to concatenate lists to pass them as input. **This is a conversion** to command-line arguments.

Clarification:
We have a cumstom build tool, let's call it foo. Foo can have includes, using the -I switch. Foo can have multiple -I options specified. The type of -I parameters is "string".

We want to call foo -Ibar -Ibat.
We use vs201x (language c++), have a proper premake environment and a custom rule "foo" containing the following:
```lua
propertydefinition {
  name = "Includes",
  kind = "list",
  switch = "-I",
}
```
We have a project that calls foo like that:
```lua
rules "foo"
fooVars { Includes = { "bar", "bat" } }
```

This results in the following lines in the .vcxproj:
```xml
<foo>
  <Includes>bar bat</Includes>
</foo>
```

This is invalid code and will not be read by VS, because arrays are separated by ';'.
Foo on file "file.foo" is called like that:
`foo file.foo`

Now we change our property:
```lua
propertydefinition {
  name = "Includes",
  kind = "list",
  switch = "-I",
  separator = ";",
}
```

This results in the following lines in the .vcxproj:
```xml
<foo>
  <Includes>bar;bat</Includes>
</foo>
```

Valid!

But we have a new problem, now the foo.xml caintains the following code:
```xml
    <StringListProperty
      Name="Includes"
      HelpContext="0"
      Switch="-I"
      Separator=";"
      />
```
And foo is called like that:
`foo -Ibar;bat file.foo`

How do we call foo correctly?
Scrap lists and use strings in the project:
```lua
rules "foo"
fooVars { Includes = table.concat({ "bar", "bat" },";") }
```

But with that we can only add items once. Every subsequent call will use the wrong separator at some point.

So how to fix that?
Easiest way would be to change the name of the vs2010 module looks for from "separator" to "Separator" (as vs syntax is uppercase anyways) or a more verbose "vsseparator". I don't know how many projects break with that change but i'm sure there are more that finally work.
[...]

PR #1160, opened 2018-09-05
use ';' as default separator for vs

WorldofBay:
This is a possible minimal workaround for #1157 .

Gmake2 keeps using " " as default list separator and VS201x gets the ";" it needs as default. It is still changeable in propertydefinitions. The field separator still does two different things at once but now it is at least possible to use list properties in VS201x projects.

### DIFF
--- a/modules/gmake2/tests/test_gmake2_file_rules.lua
+++ b/modules/gmake2/tests/test_gmake2_file_rules.lua
@@ -36,9 +36,20 @@
 			value = false,
 			switch = "-p2"
 		}
+	
+		propertydefinition {
+			name = "TestListProperty",
+			kind = "list"
+		}
+	
+		propertydefinition {
+			name = "TestListPropertySeparator",
+			kind = "list",
+			separator = ","
+		}
 
 		buildmessage 'Rule-ing %{file.name}'
-		buildcommands 'dorule %{TestProperty} %{TestProperty2} "%{file.path}"'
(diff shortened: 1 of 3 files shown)


WorldofBay:
Updated the test, they pass now.

WorldofBay:
I changed the test and added one for a custom separator.

I have no idea what to capture in the gmake2 version though. Maybe someone else can help?

WorldofBay:
i added your test and one for no separator

for some reason the actual output without separator adds an '\'

for now i add it to the test though i don't know if that is right

WorldofBay:
updated my branch now, all tests pass again

samsinsane:
The changes look good, are you able to squash the commits? Once squashed it should be good to merge, thanks!

WorldofBay:
first time for me, so i'm unsure what i should do with the merged commits.
squash, keep, drop?

btw do i have to update the branch before (7 commits behind)?

samsinsane:
You don't need to update the branch, you can rebase it if you'd like.

As for the merged commits, dropping them seems like the right thing to do here. I think squashing might pull in the changes from master, which wouldn't be weird, keeping the commit isn't the end of the world but it's kind of a useless merge since you're several commits behind master now. That's my thought process behind this, if you feel differently, I'm happy to discuss. 😄

WorldofBay:
Squashed my commits and dropped the others.

samsinsane:
Thanks!

PR #1177, opened 2018-10-12
Enable property categories in VS projects

WorldofBay:
Allow the usage of Categories in VS to structure properties for big rules.

Items without Category go in "General", also the Category "Command Line" is invalid.

Usage:

```lua
propertydefinition {
  name = "O0",
  switch = "/O0",
  category = "Optimizations",
}
```

### DIFF
--- a/modules/vstudio/tests/vc2010/test_rule_xml.lua
+++ b/modules/vstudio/tests/vc2010/test_rule_xml.lua
@@ -179,3 +179,46 @@
 </EnumProperty>
 		]]
 	end
+
+	function suite.properties_onStringWithCategory()
+		createVar { name="MyVar", kind="string", category="Custom Category" }
+		local r = test.getRule("MyRule")
+		m.properties(r)
+		test.capture [[
+<StringProperty
+	Name="MyVar"
+	HelpContext="0"
+	DisplayName="MyVar"
+	Category="Custom Category"
+	/>
+		]]
+	end
+  
+  function suite.categories_onStringWithCategory()
+		createVar { name="MyVar", kind="string", category="Custom Category" }
+		local r = test.getRule("MyRule")
+		m.categories(r)
+		test.capture [[
+<Rule.Categories>
+	<Category
+		Name="General">
+		<Category.DisplayName>
+			<sys:String>General</sys:String>
+		</Category.DisplayName>
+	</Category>
+	<Category
+		Name="Custom Category">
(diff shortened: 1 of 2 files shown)


WorldofBay:
idk if 2 captures in 1 test are possible

i would like to add
```lua
		test.capture [[
<Category
	Name="Custom Category">
	<Category.DisplayName>
	  	<sys:String>Custom Category</sys:String>
	</Category.DisplayName>
</Category>
		]]
```
to the test to also test if the category itself is added correctly

samsinsane:
Off the top of my head I'm not sure if it would work, you can try it? Locally, you should be able to run `bin\debug\premake5.exe test --scripts=.` in the repo directory and that will run the tests with the raw Lua files instead of the embedded ones.

Also, once you've got this PR into a state you're happy with, can you please squash the commits? Thanks!

WorldofBay:
I'm currently using this code via premake.override and it works.
I searched through the repo for usage of "category" and it's used but in unrelated context in base that as far as i see it does not conflict anywhere.

It's a change in the VS_2010 module, so it doesn't affect anything outside of that. I can only really test it in VS 2015 and VS 2017 though.

It's quite helpful in the GUI then (fortran as a rule):
grafik

WorldofBay:
The test maybe can capture twice but it can't capture the <Category> tag because it skips that part of the xml. So unless you know a way to add that test, this pull request is complete as-is.

samsinsane:
I'm not sure I'm 100% following what the test needs to use, but I think something like this might work?
```lua
	function suite.categories_onStringWithCategory()
		createVar { name="MyVar", kind="string", category="Custom Category" }
		local r = test.getRule("MyRule")
		m.categories(r)
		test.capture [[
<Category
	Name="Custom Category">
	<Category.DisplayName>
	  	<sys:String>Custom Category</sys:String>
	</Category.DisplayName>
</Category>
		]]
```

The `m.categories` function emits the `<Category>` elements, while the `m.properties` function emits the `<StringProperty>` elements. You may need to do some more massaging of the `test.capture` but I think that's what you're after?

WorldofBay:
It ... works but it demands exact indentation and also compares to the whole `<Rule.Categories>` tag instead of just the custom part.

WorldofBay:
Now it contains the categories test, so this pull request is complete.

Once it is merged i will edit the corresponding wiki entries.

PR #1178, opened 2018-10-15
Avoid impl dependent multi-char constants

cengizIO:
Hello!

I was receiving this warning while compiling for linux:

```
src/host/premake.c: In function ‘premake_init’:
src/host/premake.c:197:36: warning: multi-character character constant [-Wmultichar]
  lua_rawseti(L, LUA_REGISTRYINDEX, 'SHIM');
                                    ^~~~~~
```

Instead of disabling the warning, I've decided to convert that array key to
an expression.

And I've tested with following compilers:

gcc version 7.3.0 (Ubuntu 7.3.0-27ubuntu1~18.04)
gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)

Thank you

### DIFF
--- a/contrib/luashim/luashim.c
+++ b/contrib/luashim/luashim.c
@@ -851,7 +851,7 @@ void shimInitialize(lua_State* L)
 
 	// Find the 'SHIM' entry in the registry.
 	const Table* reg = hvalue(&G(L)->l_registry);
-	const Node* n = findNode(reg, 'SHIM');
+	const Node* n = findNode(reg, 'S' + 'H' + 'I' + 'M');
 	assert(n != NULL);
 
 	g_shimTable = (const LuaFunctionTable_t*)n->i_val.value_.p;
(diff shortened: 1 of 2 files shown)


WorldofBay:
You change the array index from 1397246285 to 305 for gcc, clang, msvc and icc. I doubt that this points at the correct value.

```c
const Node* n = findNode(reg,((((('S' << 8) + 'H') << 8) + 'I') << 8) + 'M');
```

cengizIO:
Good point!

However I don't think the actual value has any effect on node retrieval. 

What I read from code, we're creating a node with a fixed index key and retrieving it afterwards.

Unless of course, the value was chosen to be something that won't collide with other node indexes that are < 1.3m.

samsinsane:
I'd prefer that this was changed to the hex value, `0x5348494D` with a comment indicating that it's equivalent to `'SHIM'`.

The value is important, existing binary modules use `'SHIM'` and won't work if Premake5 is using `'S' + 'H' + 'I' + 'M'`. The current change will require that all existing binary modules be rebuilt against the latest version of luashim, we have no way of knowing how many binary modules our users have, if any. It's best to avoid breaking them where we can.

PR #1180, opened 2018-10-18
Add `os.ios` option in xcode4

tempura-sukiyaki:


modules/xcode/_preload.lua
```diff
@@ -6,6 +6,25 @@
 
 	local p = premake
 
+--
+-- Register the ios extension
+--
+
+	local osoption = p.option.get("os")
```
samsinsane: This should be part of the `os` option by default, I've used `gmake`, `gmake2` and `xcode4` for iOS builds. Same thing goes for the `os.systemTags`.

modules/xcode/_preload.lua
```diff
@@ -6,6 +6,25 @@
...
+			ios = ios or allowed[1] == "ios"
+		end)
+		if not ios then
+			table.insert(osoption.allowed, { "ios",  "iOS" })
+		end
+	end
+
+	-- add system tags for ios.
+	if not os.systemTags[p.IOS] then
+		os.systemTags[p.IOS] = { "ios", "mobile" }
```
samsinsane: For both this and the `p.MACOSX` entry, it might be worth adding an `"apple"` entry? This would allow all of the checks you changed below to be `apple` instead and they'd work for tvOS and watchOS - and whatever comes next. Thoughts?

modules/xcode/_preload.lua
tempura-sukiyaki: I see. I will do so.

modules/xcode/_preload.lua
tempura-sukiyaki: I see. I think that it looks very good.

tempura-sukiyaki: I feel it is better to define `darwin` instead of `apple`.
`macosx` already has that value.

```lua
	os.systemTags =
	{
		["aix"]      = { "aix",     "posix" },
		["bsd"]      = { "bsd",     "posix" },
		["haiku"]    = { "haiku",   "posix" },
		["ios"]      = { "ios",     "darwin", "mobile" }
		["linux"]    = { "linux",   "posix" },
		["macosx"]   = { "macosx",  "darwin", "posix" },
		["solaris"]  = { "solaris", "posix" },
		["windows"]  = { "windows", "win32" },
	}
```

It may be better to include `posix`.

samsinsane: I agree, perhaps both `darwin` and `posix`?

samsinsane:
This is really cool! Thanks for this!

PR #1189, opened 2018-11-04
Host path.normalize: Fix normalization for paths containing tokens

### LINKED ISSUE #1186: Alpha13 generates different paths from alpha12
In our premake file over at https://github.com/recastnavigation/recastnavigation we use the following to copy a dependency to the output directory:
```lua
postbuildcommands {
    -- Copy the SDL2 dll to the Bin folder.
    '{COPY} "%{wks.location}../../Contrib/SDL/lib/%{cfg.architecture:gsub("x86_64", "x64")}/SDL2.dll" "%{cfg.targetdir}"'
}
```

In alpha12 this generated (for vs2017):
```
IF EXIST "$(SolutionDir)..\..\Contrib\SDL\lib\x86\SDL2.dll"\ (xcopy /Q /E /Y /I "$(SolutionDir)..\..\Contrib\SDL\lib\x86\SDL2.dll" "..\..\Bin" &gt; nul) ELSE (xcopy /Q /Y /I "$(SolutionDir)..\..\Contrib\SDL\lib\x86\SDL2.dll" "..\..\Bin" &gt; nul)
```

But in alpha13 it generates:
```
IF EXIST "Contrib\SDL\lib\x86\SDL2.dll"\ (xcopy /Q /E /Y /I "Contrib\SDL\lib\x86\SDL2.dll" "..\..\Bin" &gt; nul) ELSE (xcopy /Q /Y /I "Contrib\SDL\lib\x86\SDL2.dll" "..\..\Bin" &gt; nul)
```
the latter path is incorrect and fails the build. Do we need to fix something or is this a bug?

tdesveauxPKFX:
Fixes #1186 

I'm not 100% happy with this fix but that's the best I could come up now.

### DIFF
--- a/src/host/path_normalize.c
+++ b/src/host/path_normalize.c
@@ -17,6 +17,18 @@
 
 #define IS_SPACE(__c)		((__c >= '\t' && __c <= '\r') || __c == ' ')
 
+#define IS_WIN_ENVVAR_START(__c)	(*__c == '%')
+#define IS_WIN_ENVVAR_END(__c)		(*__c == '%')
+
+#define IS_VS_VAR_START(__c)		(*__c == '$' && __c[1] == '(')
+#define IS_VS_VAR_END(__c)			(*__c == ')')
+
+#define IS_UNIX_ENVVAR_START(__c)	(*__c == '$' && __c[1] == '{')
+#define IS_UNIX_ENVVAR_END(__c)		(*__c == '}')
+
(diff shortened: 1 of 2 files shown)


samsinsane:
Your reasoning makes sense; I personally dislike multiline macros, but I think what you've done is probably the best option. Ignore my previous comment :+1:

PR #1190, opened 2018-11-06
Add support for `externalproject` with multiple targets in xcode4

tempura-sukiyaki:
xcodeproj may have multiple targets.
`externalproject` must be able to explicitly specify one of those targets.

I think that this can be solved by introducing the following description.

```lua
externalproject "targetname@projectname"
```

It may be better to judge whether the project is `external`, but I think that there are no practical problems in particular.

### DIFF
--- a/modules/xcode/tests/_tests.lua
+++ b/modules/xcode/tests/_tests.lua
@@ -5,5 +5,6 @@ return {
 	"test_xcode4_project.lua",
 	"test_xcode4_workspace.lua",
 	"test_xcode_dependencies.lua",
+	"test_xcode_externalproject.lua",
 	"test_xcode_project.lua",
 }
--- a/modules/xcode/tests/test_xcode_externalproject.lua
+++ b/modules/xcode/tests/test_xcode_externalproject.lua
@@ -1,0 +1,372 @@
+--
+-- tests/actions/xcode/test_xcode_dependencies.lua
+-- Automated test suite for Xcode project dependencies.
+-- Copyright (c) 2009-2011 Jason Perkins and the Premake project
+--
+
+	local suite = test.declare("xcode_external")
+	local p = premake
+	local xcode = p.modules.xcode
+
+
+---------------------------------------------------------------------------
+-- Setup/Teardown
+---------------------------------------------------------------------------
+
+	local wks, prj, eprj1, etrg1, etrg2, tr
+
+	function suite.teardown()
+		wks = nil
+		prj = nil
+		eprj1 = nil
+		etrg1 = nil
+		etrg2 = nil
+		tr = nil
+	end
+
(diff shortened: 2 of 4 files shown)


tempura-sukiyaki:
I may not understand what you are saying correctly...

"xcodeproj" created with "premake" always includes only one product (target) per project.

```lua
configurations { "Release" }

workspace "Workspace" do

  project "StaticLib" do
    kind "StaticLib"
  end

  project "SharedLib" do
    kind "SharedLib"
  end

  project "ConsoleApp" do
    kind "ConsoleApp"
    links {"StaticLib", "SharedLib"}
  end

end
```

"SharedLib.xcodeproj" created from the above script looks something like this.
SharedLib

"StaticLib.xcodeproj" looks something like this.
StaticLib

"Products" and "TARGETS" have only one item.

However, if you create "xcodeproj" with Xcode, one project may contain more than one product (target).

ExternalProject

Each of these targets has completely independent project settings.

Currently there is no way to specify such "xcodeproj" with [`externalproject`](https://github.com/premake/premake-core/wiki/externalproject).

samsinsane:
My point was that instead of doing:
```lua
externalproject "targetname@projectname"
```

Something like this would be better:
```lua
externalproject "projectname"
  configmap { ["*"] = "targetname" }
  -- or a new API like this:
  targetmap { ["*"] = "targetname" }
  -- or perhaps a simpler API:
  xcodetargetname "targetname"
```

I don't think the `externalproject` call is the correct place to have the target name information. At some point Premake should probably support the concept of a target but I'm not really sure how it would fit.

tempura-sukiyaki:
Actually, I do not think that writing "targetname" to `externalproject` is an excellent way.
I examined ways like `xcodetargetname` before this implementation.

However, in that case, designation of `links` and `dependson` seems difficult, so I gave up.
When using an API such as `xcodetargetname`, how should we describe `links` and `dependson`?

samsinsane:
I would imagine that the `links` and `dependson` code paths would give access to the other project at some point, then it would be a matter of querying the project for the `xcodetargetname` entry. I think initially this can just be part of `externalproject`s, since Premake doesn't support generating targets but when it does we'll need to revisit this assumption.

I think this should suffice for now:
```lua
workspace "wks"
  project "prj"
    links { "ext" } -- This would detect the `xcodetargetname` in the ext project
  externalproject "ext"
    xcodetargetname "targetname"
```

tdesveauxPKFX:
Using existing APIs, you might be able to do something like this:
```lua
externalproject("targetname")
	location	"location/to/project/file"
	filename 	"projectfile"
```
this way, you don't have anything to do to link with `targetname`.
Well, I'm neither familiar with XCode nor the module so I might be wrong.

PR #1192, opened 2018-11-11
Update path.lua

dsvi:
added *.c++ extension

### DIFF
--- a/src/base/path.lua
+++ b/src/base/path.lua
@@ -171,7 +171,7 @@
 	end
 
 	function path.iscppfile(fname)
-		return path.hasextension(fname, { ".cc", ".cpp", ".cxx" })
+		return path.hasextension(fname, { ".cc", ".cpp", ".cxx", ".c++" })
 			or path.isobjcppfile(fname)	-- is this really right?
 			or path.iscfile(fname)
 	end


ratzlaff:
Is this a common extension?

samsinsane:
This was requested in #612, however, I'm fairly certain that this change won't work for `gmake2` based on this snippet: https://github.com/premake/premake-core/blob/ed483dd9bc5b5e357d210781ab2e44243c90eb5e/modules/gmake2/gmake2_cpp.lua#L59-L63

dsvi:
> Is this a common extension?

Relatively. qmake/qtcreator recongnizes it as C++ file.


> I'm fairly certain that this change won't work for `gmake2` based on this snippet:

Nothing will, since gmake2 does not rely on the path. functions. gmake works fine with this change.

samsinsane:
> Relatively. qmake/qtcreator recongnizes it as C++ file.

Do you have any examples of what uses this extension? I'm kind of curious about where this extension comes from, things seem to support it but nobody seems to use it. I find this really weird.

> Nothing will, since gmake2 does not rely on the `path.` functions. gmake works fine with this change.

I didn't go any deeper than the `rule` definition, but I kind of assumed the `path.` functions weren't called further down. FWIW, this change will need to work in `gmake2` before we can merge it.

dsvi:
> Do you have any examples of what uses this extension? 

https://www.google.com/search?q=.c%2B%2B+file+extension
gcc toolset recognizes it as well
https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html
Anyway, it doesn't hurt to add, does it?

>  FWIW, this change will need to work in `gmake2` before we can merge it.

I believe the best way to achieve it, is to redesign `gmake2` to rely on the detection functions from `path.`, like `gmake` does. 
This change as it is doesn't break anything in `gmake2`

ratzlaff:
Can you implement any new tests with the gmake and gmake2 generators for this extension?

dsvi:
[Hmm...](http://www.quickmeme.com/img/8f/8fc3162d32c7e051f086e0b5b18c443365deb6892f3fe5bc6f2677bba3447afd.jpg)

tdesveauxPKFX:
> 
> > Do you have any examples of what uses this extension?
> 
> https://www.google.com/search?q=.c%2B%2B+file+extension
> gcc toolset recognizes it as well
> https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html
> Anyway, it doesn't hurt to add, does it?
> 
> > FWIW, this change will need to work in `gmake2` before we can merge it.
> 
> I believe the best way to achieve it, is to redesign `gmake2` to rely on the detection functions from `path.`, like `gmake` does.
> This change as it is doesn't break anything in `gmake2`

Indeed, it does not hurt and recognizing the same extensions as GCC looks fine to me.
However, I agree with @samsinsane here. The `gmake` generator will be deprecated in favor of `gmake2` in the future so we'd need to have this works for `gmake2` to merge.
From a quick look, I think you only need to make changes [https://github.com/premake/premake-core/blob/master/modules/gmake2/gmake2_cpp.lua#L60](here) and [https://github.com/premake/premake-core/blob/master/modules/gmake2/gmake2_cpp.lua#L102-L113](here).

And as @ratzlaff mentioned, a simple unit test that confirm that the correct rule is applied would be great in case a future change break this. There don't seem to be one at the moment but it would have it's place in *_file_rules.lua file test.

PR #1193, opened 2018-11-13
Add tests for xcode id generator

ratzlaff:
the 'long' type is not always of size four bytes - which is what the current code assumes. Changed to uint32_t and added tests for the xcode id-generator.

### DIFF
--- a/modules/xcode/tests/test_xcode4_project.lua
+++ b/modules/xcode/tests/test_xcode4_project.lua
@@ -18,6 +18,7 @@
 
 	xcode.used_ids = {}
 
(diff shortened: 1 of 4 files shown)


ratzlaff:
Yeah, I have been making local tests here for some of the custom modifications we have made to premake (nothing major, most of which ends up as pull requests). Had to figure out why the windows executable would create an xcode project differently than when we ran the same scripts on the macs. Eventually nailed it down to the size differences of 'long' between the 32 and 64-bit versions of the executable.

So right now I believe we are able to generate the exact same output from all the major platforms.

Issue #1172, opened 2018-09-25
.rc files exclusion for non-windows systems


yuyoyuppe:
Hi all!

My project currently has single .sln for multiple platforms(Windows and Xbox). I need *.rc files for Windows build, but must exclude it for Xbox one.  Typically we would use flags for that:
```lua

filter { "files:**.rc", "platforms:Xbox" }
  flags{"ExcludeFromBuild"}
```

But it doesn't work, because in `modules\vstudio\vs2010_vcxproj.lua` we're setting `checkFunc` for resources like that:
```lua
			m.emitFiles(prj, group, "ResourceCompile", nil, fileCfgFunc, function(cfg)
				return cfg.system == p.WINDOWS
			end)

```
...which prevents `m.excludedFromBuild` from being invoked, since I'm relying on `.system` for my Xbox module.

I've fixed the issue like that:

```diff
diff --git a/modules/vstudio/tests/vc2010/test_files.lua b/modules/vstudio/tests/vc2010/test_files.lua
index 51019fad..89ab5164 100644
--- a/modules/vstudio/tests/vc2010/test_files.lua
+++ b/modules/vstudio/tests/vc2010/test_files.lua
@@ -278,7 +278,9 @@
 		prepare()
 		test.capture [[
 <ItemGroup>
-	<ResourceCompile Include="hello.rc" />
+	<ResourceCompile Include="hello.rc">
+		<ExcludedFromBuild>true</ExcludedFromBuild>
+	</ResourceCompile>
 </ItemGroup>
 		]]
 	end
@@ -295,6 +297,7 @@
 <ItemGroup>
 	<CustomBuild Include="hello.cg">
 		<FileType>Document</FileType>
+		<ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ExcludedFromBuild>
 		<Command Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">cgc $(InputFile)</Command>
 		<Outputs Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(InputName).obj</Outputs>
 	</CustomBuild>
diff --git a/modules/vstudio/vs2010_vcxproj.lua b/modules/vstudio/vs2010_vcxproj.lua
index 4ba96c6d..cbcab19a 100644
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -1167,8 +1167,11 @@
[...]

PR #1196, opened 2018-11-18
Emit files check feature

yuyoyuppe:
See #1172. Also I've added [.ixx extension](https://blogs.msdn.microsoft.com/vcblog/2017/05/05/cpp-modules-in-visual-studio-2017/) support, since they were "not participating in build" by default(so no `buildaction` required)


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -744,7 +744,7 @@
 ---
 	m.categories.ClCompile = {
 		name       = "ClCompile",
-		extensions = { ".cc", ".cpp", ".cxx", ".c", ".s", ".m", ".mm" },
+		extensions = { ".cc", ".cpp", ".cxx", ".c", ".s", ".m", ".mm", ".ixx" },
```
samsinsane: I'm not overly keen on this change. `.ixx` can be used for inline definitions which should be included by another file, rather than compiled themselves. VS2017 recognizes (and seems to use) the extension for C++ Modules but this isn't part of any standard and I can't really see anything that indicates that it needs to be explicitly compiled? I think this is better suited as a local override, using [`buildaction'](https://github.com/premake/premake-core/wiki/buildaction) like this:
```lua
filter { "files:**.ixx" }
  buildaction "ClCompile"
filter {}
```

starkos:
Based on the conversion on #1172, it sounds like this is not the preferred way to solve this problem. I'm going to go ahead and close this PR and let that issue play out. Feel free to reopen as needed.

yuyoyuppe:
@starkos sure, ping me if you decide on a different implementation - I can still do it :)

starkos:
It [sounds like](https://github.com/premake/premake-core/issues/1172#issuecomment-441021586) the `checkFunc` should just be removed entirely.

PR #1201, opened 2018-12-04
Updated code lite to properly add LD_LIBRARY_PATH is set if libdirs is set

pintodragon:
Also added a change to the tree.traverse to make it generate the base virtual directories. Without this change I was getting projects that had files that were not part of a virtual directory. These files would not show up in the editor and there was no way for me to easily find them and move them to a virtual directory for them to show up.


modules/codelite/codelite_project.lua
```diff
@@ -268,9 +267,21 @@
 		local pauseexec  = iif(prj.kind == "ConsoleApp", "yes", "no")
 		local isguiprogram = iif(prj.kind == "WindowedApp", "yes", "no")
 		local isenabled  = iif(cfg.flags.ExcludeFromBuild, "no", "yes")
-
-		_x(3, '<General OutputFile="%s" IntermediateDirectory="%s" Command="%s" CommandArguments="%s" UseSeparateDebugArgs="%s" DebugArguments="%s" WorkingDirectory="%s" PauseExecWhenProcTerminates="%s" IsGUIProgram="%s" IsEnabled="%s"/>',
-			targetname, objdir, command, cmdargs, useseparatedebugargs, debugargs, workingdir, pauseexec, isguiprogram, isenabled)
+    local ldPath = ''
+    
+    for _, libdir in ipairs(cfg.libdirs) do
+			ldPath = ldPath .. ";" .. project.getrelative(cfg.project, libdir)
```
pintodragon: The ; should be platform specific.

modules/codelite/codelite_project.lua
```diff
@@ -111,7 +111,7 @@
 
 		-- TODO: ...
 	end
-
+
```
samsinsane: This can probably be reverted.

modules/codelite/codelite_project.lua
```diff
@@ -268,9 +267,21 @@
 		local pauseexec  = iif(prj.kind == "ConsoleApp", "yes", "no")
 		local isguiprogram = iif(prj.kind == "WindowedApp", "yes", "no")
 		local isenabled  = iif(cfg.flags.ExcludeFromBuild, "no", "yes")
-
-		_x(3, '<General OutputFile="%s" IntermediateDirectory="%s" Command="%s" CommandArguments="%s" UseSeparateDebugArgs="%s" DebugArguments="%s" WorkingDirectory="%s" PauseExecWhenProcTerminates="%s" IsGUIProgram="%s" IsEnabled="%s"/>',
-			targetname, objdir, command, cmdargs, useseparatedebugargs, debugargs, workingdir, pauseexec, isguiprogram, isenabled)
+    local ldPath = ''
```
samsinsane: Somehow I managed to not notice this before, the indentation here isn't right. It looks like you're using spaces while the rest of the file is using tabs.

pintodragon: Ah, I never use tabs.

Also the LD_LIBRARY_PATH may be linux specific. So I will have to figure out how to put the correct value in there for windows.

modules/codelite/codelite_project.lua
pintodragon: Adding the extra line here actually conforms to the rest of the file so I will leave it.

pintodragon:
This should be ready to merge now.

pintodragon:
I did and in the process ended up having the process closed. I can raise a new one if needed

PR #1218, opened 2019-01-09
Add support & tests for XCTest sharedlibtype

Dingobloo:
.xctest files are OSXBundles that have executable code in them that Xcode uses to populate its integrated [unit testing UI](https://developer.apple.com/documentation/xctest). Since premake supports OSXBundles through the [sharedlibtype API](https://github.com/premake/premake-core/wiki/sharedlibtype) I extended that support to also include xctest. 

My use case was wanting to use premake rather than cmake with [CUTI](https://github.com/k-brac/CUTI) a C++ unit testing wrapper that integrates with XCode & Visual Studio, but others may find it helpful for swift or objective-c projects that include unit tests.

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -306,6 +306,8 @@
 		if table.contains(os.getSystemTags(cfg.system), "darwin") then
 			if cfg.sharedlibtype == "OSXBundle" then
 				return "-bundle"
+			elseif cfg.sharedlibtype == "XCTest" then
+				return "-bundle"
 			elseif cfg.sharedlibtype == "OSXFramework" then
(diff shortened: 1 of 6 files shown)


src/tools/gcc.lua
```diff
@@ -306,6 +306,8 @@
 		if table.contains(os.getSystemTags(cfg.system), "darwin") then
 			if cfg.sharedlibtype == "OSXBundle" then
 				return "-bundle"
+			elseif cfg.sharedlibtype == "XCTest" then
```
samsinsane: This probably could have been an `or` on the previous line, but the consistency is good. :+1:

PR #1220, opened 2019-01-09
Implement the androidapilevel() API in Android packaging projects

ifarbod:
The `androidapilevel` API is only implemented for shared library projects. Typically, one would want to change the SDK version for the APK project as well. When you don't specify it, VS picks `android-21` as the default `compileSdkVersion` for you. This PR adds support for `AndroidAPILevel` for APK projects (the code is simply taken from `vsandroid_vcxproj.lua`).

### DIFF
--- a/modules/android/vsandroid_androidproj.lua
+++ b/modules/android/vsandroid_androidproj.lua
@@ -80,10 +80,18 @@
 	end
 
 
+	function android.androidAPILevel(cfg)
+		if cfg.androidapilevel ~= nil then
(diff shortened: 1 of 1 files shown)


modules/android/vsandroid_androidproj.lua
```diff
@@ -80,10 +80,18 @@
 	end
 
 
+	function android.androidAPILevel(cfg)
```
samsinsane: I'm pretty sure you can just call the function from the other file without duplicating the code in this file. The `android` variable is the same in both files, so it should work without any other modifications.

ifarbod: Sure, I'll fix this.

PR #1223, opened 2019-01-11
Allow using Clang 5.0 for VSAndroid projects

ifarbod:
This is only supported in VS2017+, should I add a check for it or is it ok like this?

PS: Can the `androidproj` value for `kind` be renamed to something else? Maybe something like `AndroidApp`, `Packaging`, or `Application`? It doesn't really feel inline with `WindowedApp` and other values.


ifarbod:
Looks like #1112 already includes this.

Issue #1010, opened 2018-02-21
.NET Core


BlackOfWorld:
I have read the wiki that premake5 supports c#, but does it support .NET Core? If not will it be implemented in the future?

samsinsane:
Having a quick look, it doesn't look like we support specifying the `TargetFramework` element, which is how it's done according to [this](https://docs.microsoft.com/en-us/dotnet/standard/frameworks). I'm not sure there's much required if you'd want to give it a go.

tritao:
I think there's more changes required than just `TargetFramework` for .NET Core projects. The whole `.csproj` structure was simplified. For instance, files are not explicitly specified now, and NuGet package references are specified in the project directly, and not in a separate `packages.config` file.

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <LangVersion>7.1</LangVersion>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <AssemblyName>Name</AssemblyName>
    <PackageReferenceId>Name</PackageReferenceId>
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
    <DefineConstants>$(DefineConstants);UNIX</DefineConstants>
  </PropertyGroup>
  <PropertyGroup Condition="$(Environment) == 'Production'">
    <DefineConstants>$(DefineConstants);PRODUCTION</DefineConstants>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Package.NetStandard" Version="0.5.0" />
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\Reference\To.csproj" />
  </ItemGroup>
  <ItemGroup>
    <None Update="Dockerfile">
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
  </ItemGroup>
  <ItemGroup>
    <Folder Include="Exchanges\" />
  </ItemGroup>
  <Import Project="..\PackageReferences\Reference\To.targets" Condition="Exists('..\PackageReferences\Reference\To.targets')" />
</Project>
```

samsinsane:
@tritao do these same changes impact F# projects too? I imagine they would.

tritao:
I just checked, yes.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
[...]

PR #1225, opened 2019-01-14
Added support for generating .NET Core projects.

tritao:
This adds a new action `netcore` that generates new style MSBuild `.csproj` projects for .NET Core.

This has been tested on CppSharp project, and it gets us far enough for a working build.

Also snuck in a fix to prevent us from generating empty property groups.

Related issue: https://github.com/premake/premake-core/issues/1010

### DIFF
--- a/modules/vstudio/_manifest.lua
+++ b/modules/vstudio/_manifest.lua
@@ -23,4 +23,5 @@ return {
 	"vs2013.lua",
 	"vs2015.lua",
 	"vs2017.lua",
+	"netcore.lua",
 }
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -20,6 +20,7 @@
 	include("vs2013.lua")
 	include("vs2015.lua")
 	include("vs2017.lua")
+	include("netcore.lua")
 
 	-- Initialize Specific API
 
@@ -140,5 +141,6 @@
 			_ACTION == "vs2012" or
 			_ACTION == "vs2013" or
 			_ACTION == "vs2015" or
-			_ACTION == "vs2017";
+			_ACTION == "vs2017" or
+			_ACTION == "netcore";
 	end
--- a/modules/vstudio/netcore.lua
+++ b/modules/vstudio/netcore.lua
@@ -1,0 +1,66 @@
+--
+-- netcore.lua
+-- Extend the existing exporters with support for .NET Core targets.
+-- Copyright (c) Jason Perkins and the Premake project
+--
+
+local p = premake
+local vstudio = p.vstudio
+
+---
+-- Define the .NET Core export action.
+---
+
+newaction {
+    -- Metadata for the command line and help system
+
+    trigger     = "netcore",
+    shortname   = ".NET Core (MSBuild)",
+    description = "Generate .NET Core MSBuild project files.",
+
+    -- Visual Studio always uses Windows path and naming conventions
+
+    targetos = "windows",
+    toolset  = "dotnet",
+
+    -- The capabilities of this action
+
+    valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None", "Utility" },
+    valid_languages = { "C#", "F#" },
+    valid_tools     = {
+        dotnet = { "msnet" },
+    },
+
+    -- Workspace and project generation logic
+
+    onWorkspace = function(wks)
+        p.vstudio.vs2005.generateSolution(wks)
+    end,
+    onProject = function(prj)
+        p.vstudio.vs2010.generateProject(prj)
+    end,
+    onRule = function(rule)
+        p.vstudio.vs2010.generateRule(rule)
+    end,
+
+    onCleanWorkspace = function(wks)
+        p.vstudio.cleanSolution(wks)
+    end,
+    onCleanProject = function(prj)
+        p.vstudio.cleanProject(prj)
+    end,
+    onCleanTarget = function(prj)
+        p.vstudio.cleanTarget(prj)
+    end,
+
+    pathVars = vstudio.vs2010.pathVars,
+
+
+    -- This stuff is specific to the Visual Studio exporters
+
+    vstudio = {
+        solutionVersion = "12",
+        versionName     = "15",
+        targetFramework = "netstandard2.0",
+    }
+}
--- a/modules/vstudio/tests/cs2005/projectsettings.lua
+++ b/modules/vstudio/tests/cs2005/projectsettings.lua
@@ -7,8 +7,8 @@
 	local p = premake
 	local suite = test.declare("vstudio_dn2005_projectsettings")
 	local dn2005 = p.vstudio.dotnetbase
+	local netcore = p.vstudio.netcore
 	local cs2005 = p.vstudio.cs2005
-
 
 --
 -- Setup
@@ -26,7 +26,8 @@
 	local function prepare()
 		prj = test.getproject(wks, 1)
 
-		dn2005.prepare(cs2005)
+		local cs = _ACTION == "netcore" and netcore or cs2005
+		dn2005.prepare(cs)
 		dn2005.projectProperties(prj)
 	end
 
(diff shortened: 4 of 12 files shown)


tritao:
I started this work by relying on `dotnet.isnetcoretarget(prj)`, but realized that might not be the best approach.

.NET Core projects support multiple target frameworks (using `<TargetFrameworks>`), and even supports targeting old .NET Framework versions, so just relying on the `dotnetframework` seems like it will bring some limitations in the future.

The other approach that I thought was by re-using the `clr` API, but that also comes with limitations, since right now we use `clr "Unsafe"` to specify if the code is unsafe, and that would conflict with a new `clr "netcore"` option.

Another advantage of the action is easy filtering with the `filter` API.

```lua
  filter { "action:not netcore"}
    links { "Lib" }

  filter { "action:netcore"}
    nuget { "Lib:1.2.3" }
```

All in all, I think it makes sense to be its own action, since it's not really related to any of the VS actions, as it is cross platform and completely independent of Visual Studio.

If we want to clean this up in the future, I think coming up with a unified `msbuild` action, that could take a `--flavor=vs20xx|netcore` parameter, would be more elegant. Then we could map the "legacy" VS actions to this new unified action.

samsinsane:
I have a few questions based on your reply:
- Can you target an old framework and still be a .NET Core project?
  - If so, what makes it a .NET Core project if not the framework?
- You've indicated that the .NET Core projects are disconnected from Visual Studio, so these open exactly the same in VS2017 and VS2019 without issue? It doesn't display `(Visual Studio 2017)` in VS2019 and ask to be upgraded and all that?
  - And I guess by extension, how do new elements get added or values get added/changed/removed when there's no versioning information? Surely, we would need some way for a user to specify they're using VS2017 and not VS2019.
- How does this work for Makefile generation since we have C# support there?
- The VS generators are already largely unified, maybe I've misunderstood your goal? It just seems like the only change will be `premake5 vs2019` will become `premake5 msbuild --flavor=vs2019` which is way more to type than should be necessary?

Sorry for all the questions, I just thought .NET Core was the next release of the .NET framework after 4.6, so I'm kind of confused haha. I haven't used C# day to day for nearly 5 years, so I'm well and truly out of the loop on what's going on.

tritao:
1) Yes, you can target .NET Framework projects with .NET Core style projects.

From what I've found, If you have `<TargetFramework>net461</TargetFramework>`, then you get a regular .NET Framework app that simply makes use of the newer csproj project model, and which can be easily built with the dotnet command-line tool.

2) I think it should work like you suggested, without an upgrade dialog for VS upgrades. However I have not tested this yet myself. The main difference than what came before is that now MSBuild is being used as a cross-platform build tool and I guess won't evolve tied with VS, but more as an independent build system.

3) I'm still trying to figure this out, but it's not as clear-cut as it once was. `csc` was easy to call from the command-line, but from what I've seen, with the Roslyn compiler, it's not as easy. The new .NET Core, while it simplifies things from the POV of the developer (with `dotnet` tool), makes things way more complicated and convoluted internally, especially since now more of the base class libraries are distributed as Nuget packages.

Some work will need to be done to make this all work using Make. I might look into it at a later time.

4) Aye, they're unified from a code perspective. What I mean is that it might make sense to stop tying the action name to VS versions. I can see eventually MSBuild on Linux being able to build C++ projects for instance (in fact I build a prototype of this a few years ago for Mono). If that happens, and Premake supports it, then it makes no sense to call it `vs201x` or `netcore`. `msbuild` would be a more appropriate name for it. And you'd need a flag to signal which version of MSBuild to target.

tritao:
@samsinsane Have you had a chance to think about this?

samsinsane:
@tritao I still have reservations about this, firstly, I'd really like to confirm the VS2017 vs VS2019 issue. Additionally, I can't really see how a user would be able to generate a VS2017 compatible .NET Core project if there's no way to specify the VS version. This leads me to believe that it being an API instead of an action makes the most sense.

Additionally, being locked to VS/MSBuild for .NET Core projects just seems wrong to me. The general trend I've seen is that people want to use whatever setup they can imagine, and having .NET Core projects controlled at the action level would restrict this.

Pitj3:
Is there any update on this?

tritao:
@Pitj3 We need as a community to agree on a design that makes sense going forward.

Re-reading the previous posts, I think we should probably do this:

1) Flag .NET Core projects as a new kind of `dotnetframework` value
Actually we should extend `dotnetframework` to support multiple values, such as: 
   `dotnetframework { "netcoreapp2.2", "net471" }` 

2) Decide what action triggers .NET Core style projects
We can just go with the status quo and extend the `vs2017` and friends to check if
the project is targeting a .NET Core framework. If that is the case, then we would
go with generation of modern csproj projects.

We need (or at least my projects do) some good way to filter .NET Core projects to be able to target generate both .NET Framework or .NET Core depending on configuration. I was using `netcore` as an action for this, if we go with different approach, then maybe we extend `filter {}` to support `dotnetframework`.

3) Optionally, add an alias action called `msbuild` (and get rid of the`netcore` action on this PR)
The idea here is that on non-Windows platforms, where there is no VS, having to use
`vs2xxx` Premake actions to generate what is in fact MSBuild build logic feels like Windows/VS-specific legacy.

It raises another question though, which is how to flag which particular version/flavor of MSBuild is generated. Right now we've used the VS version for this, as traditionally, people just used MSBuild as distributed by VS releases. But in a cross-platform world, and with MSBuild being released independently from VS, I don't think it makes sense anymore to tie it to VS versions.

Anyway just my two cents, happy to hear other thoughts about this.

dpeter99:
Hi!
Just reading throught this PR and i have one question. Would this allow for projects that have non dot net core parts?
My current project has c++ main code and a precompiler written in c# uising dot net (so it can be built on any platform). They are setup so VS builds the c# part and runs it before compiling the c++ main code.

samsinsane:
@tritao Regarding your points:
1. I think this is a good idea, if you can target multiple .NET frameworks this makes sense.
2. I think extending the existing actions makes the most sense. As @dpeter99 has pointed out, mixing project types is required and having a new set of actions will likely lead to incompatibilities between the projects for various systems (Windows, VSLinux, Android, iOS) and languages (C#, F#, C++, D).
3. This seems kind of tricky. Having `msbuild` and `vs*` using the same code paths would require _all_ of the `_ACTION` checks to be changed, and as mentioned above, modules that extend the VS actions won't work and may generate bad data in the `msbuild` action. It sounds like it would be better if the `vs*` actions were the aliases instead, specifying very specific versions of MSBuild - which don't seem like they would be overly easy to determine? The projects say `15.0`, but it looks like my MSBuild (VS2017) is `15.9` so I'm not sure how those out of the loop on this VS/MSBuild situation would know what to put in?

As for your comment between point 2 and 3, I don't really understand. Are you wanting `filter { "dotnetframework:..." }` or are you wanting do something like this:
```lua
filter { "configuration:something" }
  dotnetframework "net471"
filter { "configuration:another" }
  dotnetframework "netcoreapp2.2"
filter {}
```

ClxS:
Adding to this: Don't forget .NET Standard projects!

I agree with @samsinsane on making the .net style projects be generated based on the targetframework would be much better. We have mixed language solutions at work where we use C++ exclusively for our native run-time, with C# being used for tools. These exist in the same project. To further complicate this we utilize .NET Standard (for all libraries), .NET Core (web stuff), and .NET Framework (WPF) projects all inside the same solution.

tritao:
I have been busy and not sure when I'll have time to finish this one.
Would anyone like to take over this one and get it ready for merging?

ClxS:
I'm short on time right myself but I'll implement this in the future if no one beats me to it. [Introducing .NET 5](https://devblogs.microsoft.com/dotnet/introducing-net-5/) announcement means there's not much choice in having this added if C# is to continue being a supported language. As of the release of .NET 5, .NET Framework is gone and it's not yet decided if .NET Standard will also be gone. There's a chance that what is currently .NET Core will be all that's left.

ClxS:
This PR is outdated now, #1386 has been merged.

samsinsane:
Closing based on previous comment.

PR #1229, opened 2019-01-20
Generated makefiles no longer error when a configuration can be missing

### LINKED ISSUE #1185: Per-config file lists for gmake
I'm in need of per-config file lists for my project and I was wondering if there were any plans to support it for gmake? Is it at all possible to support it for gmake? If there are no current plans and it is in fact possible, would you mind pointing me in the right direction to implement it? I'd be interested in making a pull request if I'm successful!

Also I tried looking for issues mentioning it but nothing came up, sorry if there's already an issue I missed!

samsinsane:
Resolves #1185

### DIFF
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -251,8 +251,6 @@
 		end
 
 		if not first then
-			p.outln('else')
-			p.outln('  $(error "invalid configuration $(config)")')
 			p.outln('endif')
 			p.outln('')
 		end
--- a/modules/gmake2/tests/test_gmake2_file_rules.lua
+++ b/modules/gmake2/tests/test_gmake2_file_rules.lua
@@ -157,8 +157,6 @@ $(OBJDIR)/hello.o: src/hello.c
 	@echo $(notdir $<)
 	$(SILENT) $(CC) $(ALL_CFLAGS) $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
@@ -194,8 +192,6 @@ obj/Release/hello.obj: hello.x
 	$(SILENT) cxc -c "hello.x" -o "obj/Release/hello.xo"
 	$(SILENT) c2o -c "obj/Release/hello.xo" -o "obj/Release/hello.obj"
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
@@ -227,8 +223,6 @@ obj/Release/hello.obj: hello.x hello.x.inc hello.x.inc2
 	$(SILENT) cxc -c "hello.x" -o "obj/Release/hello.xo"
 	$(SILENT) c2o -c "obj/Release/hello.xo" -o "obj/Release/hello.obj"
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
@@ -261,8 +255,6 @@ obj/Release/hello.obj: hello.x
 	$(SILENT) c2o -c "obj/Release/hello.xo" -o "obj/Release/hello.obj"
 obj/Release/hello.other obj/Release/hello.another: obj/Release/hello.obj
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
--- a/modules/gmake2/tests/test_gmake2_objects.lua
+++ b/modules/gmake2/tests/test_gmake2_objects.lua
@@ -90,8 +90,6 @@ OBJECTS += $(OBJDIR)/hello_debug.o
 else ifeq ($(config),release)
 OBJECTS += $(OBJDIR)/hello_release.o
 
-else
-  $(error "invalid configuration $(config)")
 endif
 
 		]]
@@ -156,8 +154,6 @@ OBJECTS += $(OBJDIR)/hello111.o
 else ifeq ($(config),release)
 OBJECTS += $(OBJDIR)/hello1.o
 
-else
-  $(error "invalid configuration $(config)")
 endif
 
 		]]
@@ -190,8 +186,6 @@ CUSTOM += obj/Debug/hello.luac
 else ifeq ($(config),release)
 CUSTOM += obj/Release/hello.luac
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
@@ -224,8 +218,6 @@ OBJECTS += obj/Debug/hello.obj
 else ifeq ($(config),release)
 OBJECTS += obj/Release/hello.obj
 
-else
-  $(error "invalid configuration $(config)")
 endif
 		]]
 	end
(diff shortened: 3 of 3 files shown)


tdesveauxPKFX:
I run into this issue at work two weeks or so ago.
I fixed it while keeping the error in the `else` of the config block to perform a check on the `config` parameter.
I implemented it as to have a smallest change in behavior as possible so it may not be the best approach and I didn't have time to look into it.

Here is the diff if you are interested.
If you agree that keeping the error there is valuable but have an issue with the implementation, I should be able to look into it this week.
```diff
diff --git a/modules/gmake2/gmake2.lua b/modules/gmake2/gmake2.lua
index e63669ab15..63a8ecac4b 100644
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -195,7 +195,7 @@
 	end
 
 
-	function gmake2.outputSection(prj, callback)
+	function gmake2.outputSection(prj, callback, hasconfigcheck)
 		local root = {}
 
 		for cfg in project.eachconfig(prj) do
@@ -251,8 +251,10 @@
 		end
 
 		if not first then
-			p.outln('else')
-			p.outln('  $(error "invalid configuration $(config)")')
+			if hasconfigcheck then
+				p.outln('else')
+				p.outln('  $(error "invalid configuration $(config)")')
+			end
 			p.outln('endif')
 			p.outln('')
 		end
diff --git a/modules/gmake2/gmake2_cpp.lua b/modules/gmake2/gmake2_cpp.lua
index e89706163d..0af98f450c 100644
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -329,7 +329,7 @@
 		_p('# Configurations')
 		_p('# #############################################')
 		_p('')
-		gmake2.outputSection(prj, cpp.elements.configuration)
+		gmake2.outputSection(prj, cpp.elements.configuration, true)
 	end
 
 
diff --git a/modules/gmake2/gmake2_utility.lua b/modules/gmake2/gmake2_utility.lua
index d8d965e9ff..9a061afc0a 100644
--- a/modules/gmake2/gmake2_utility.lua
+++ b/modules/gmake2/gmake2_utility.lua
@@ -225,7 +225,7 @@
 		_p('# Configurations')
 		_p('# #############################################')
 		_p('')
-		gmake2.outputSection(prj, utility.elements.configuration)
+		gmake2.outputSection(prj, utility.elements.configuration, true)
 	end
 
 
```

samsinsane:
I can't see any reason for these checks, the `config` is validated in the workspace makefile: https://github.com/premake/premake-core/blob/00c8d53c5355326520cceb6f112f69eedbc5ec18/modules/gmake2/gmake2_workspace.lua#L39-L65

I'm pretty much 100% confident it was a copy-paste error, I just can't see any reason that someone would want this:
```
# File sets
# #############################################

OBJECTS :=

ifeq ($(config),release)
OBJECTS += $(OBJDIR)/hello.o

else
  $(error "invalid configuration $(config)")
endif
```
which is generated with:
```lua
files { "hello.cpp" }
filter "Debug"
  removefiles { "hello.cpp" }
```

Similarly with your diff, I believe the following would fail to build `make config=debug`:
```lua
defines { "_DEBUG" }
filter "Release"
  removedefines { "_DEBUG" }
```
as it would generate a single `ifeq` block, and then error for all other configurations, despite their validity.

I could be misunderstanding the code, so I'm more than happy to discuss further.

tdesveauxPKFX:
The `OBJECTS` is indeed invalid at the moment and that is the issue I encountered.

However, it seems ALL variation of config_platform are created even if identical.

From your example:
```
ifeq ($(config),debug_x32)
TARGETDIR = bin/x32/Debug
TARGET = $(TARGETDIR)/testProj.exe
OBJDIR = obj/x32/Debug
DEFINES += -D_DEBUG
ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -m32
ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS) -m32
ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib32 -m32 -s

else ifeq ($(config),debug_x64)
TARGETDIR = bin/x64/Debug
TARGET = $(TARGETDIR)/testProj.exe
OBJDIR = obj/x64/Debug
DEFINES += -D_DEBUG
ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -m64
ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS) -m64
ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib64 -m64 -s

else ifeq ($(config),release_x32)
TARGETDIR = bin/x32/Release
TARGET = $(TARGETDIR)/testProj.exe
OBJDIR = obj/x32/Release
DEFINES +=
ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -m32
ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS) -m32
ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib32 -m32 -s

else ifeq ($(config),release_x64)
TARGETDIR = bin/x64/Release
TARGET = $(TARGETDIR)/testProj.exe
OBJDIR = obj/x64/Release
DEFINES +=
ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -m64
ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS) -m64
ALL_LDFLAGS += $(LDFLAGS) -L/usr/lib64 -m64 -s

else
  $(error "invalid configuration $(config)")
endif
```

I find the "redundant" check useful as the project Makefiles can be called directly without using the workspace Makefile.

samsinsane:
Can you call the project Makefiles directly? I tried this once quite a long time ago and it didn't work, that was with `gmake` though, so I just assumed it didn't work.

Anyway, I looked at it a bit closer and there's extra logic for `OBJECTS` so it won't do the same thing with the configurations. It still feels kind of wrong to use the configurations to perform these checks, but I think your diff is probably the better solution out of the two. :+1:

starkos:
So what's the verdict on this one? Do we keep the `else` and the error, or remove it?

starkos:
What's up with this one? I just tried to reproduce the problem with this:

```lua
workspace "MyWorkspace"

project "MyProject"
    kind "ConsoleApp"
    language "C++"

    files { "hello.cc" }

    filter { "configurations:Debug" }
        removefiles "hello.cc"
```

And this was the result:

```
$ make config=debug
==== Building MyProject (debug) ====
MyProject.make:75: *** "invalid configuration debug".  Stop.
make: *** [MyProject] Error 2
```

I'm inclined to merge this as-is and remove the error condition. Thoughts?

samsinsane:
@starkos It would take me a bit to get back into state on this, but if you're happy to merge as-is then we can probably revisit this if someone runs into issues with the side effects? I'm not really sure how common it is to call the individual makefiles though, as I said above, it didn't work for me but I easily could have been doing something wrong.

Issue #1205, opened 2018-12-12
managed/unmanaged linking


WorldofBay:
Why does premake explicitly refuse to link projects with `clr "on"` and projects without?

I can't find a reason for that and we got some cases where we link non-clr projects to clr-projects via pragma comments because premakes doesn't do it.

I found that `clr "pure"` and `clr "safe"` might have problems but `clr "on"` not, so why is premake not checking for pure/safe instead (which aren't supported any more for c++ anyways)?

samsinsane:
I have absolutely no clue why, but if you can provide a minimal Premake script to reproduce the problem, and explain how one would manually link the two projects correctly via the UI, then we should be able to get a fix going for this. Unless you wish to submit a PR to fix this, in which case I'm more than happy to review and merge it. 😃

WorldofBay:
Well linking them is just adding the dependency, it works (otherwise our build would fail) but it is impossible to do it via premake because of this check in base/config.lua/config.canLink(cfg, target, linkage):

```lua
-- Can't link managed and unmanaged projects

local cfgManaged = project.isdotnet(cfg.project) or (cfg.clr ~= p.OFF)
local tgtManaged = project.isdotnet(target.project) or (target.clr ~= p.OFF)
return (cfgManaged == tgtManaged)
``` 

in our build i added

```lua
if (cfg.clr == premake.ON) or (target.clr == premake.ON) then
  return true
end
```

before that to make it work but i want to know first why that check is there in the first place before i flesh out a PR that works in our case but possibly breaks other people's stuff.

@starkos you added that in d3ec20d1c3689d8e7f39dd4d61ecb24bbb9c7511, maybe you can help?

clr "pure" and clr "safe" unsopported in vs2017:
https://docs.microsoft.com/en-us/cpp/dotnet/mixed-native-and-managed-assemblies?view=vs-2017

WorldofBay:
PR is merged, so this is closed, thanks sam

PR #1240, opened 2019-02-07
allow linking of mixed c++ assemblies

WorldofBay:
#1205 
been over a month, yet nobody yelled at me so here's the PR

Note that all tests still pass on the changed behaviour.


WorldofBay:
i don't want this to enter the void. is anything blocking the merge or are you just short in time?

PR #1252, opened 2019-03-05
GameBoy build support

rlewicki:
Added new kind for `gmake` target which is `GameBoyApp`. It produces Makefile that outputs executable file with `.gb` extension and uses `lcc` compiler. For now you have to use `--cc` flag set to `lcc` during project generation in order to work properly. I'm pretty sure there is a way to set `lcc` compiler as default `cc` value when project kind is set to `GameBoyApp` same way target extension is set.

 I'm open to any suggestions and improvements.

### DIFF
--- a/modules/gmake/_preload.lua
+++ b/modules/gmake/_preload.lua
@@ -20,7 +20,7 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Utility", "Makefile", "GameBoyApp" },
 		valid_languages = { "C", "C++", "C#" },
 		valid_tools     = {
-			cc     = { "clang", "gcc" },
+			cc     = { "clang", "gcc", "lcc" },
 			dotnet = { "mono", "msnet", "pnet" }
 		},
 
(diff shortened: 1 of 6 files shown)


samsinsane:
There's a couple of things I would recommend before merging into core.
- Support `gmake2` instead of `gmake`, eventually we'll be replacing `gmake` with `gmake2` and this will just stop working.
- Instead of using `kind` to target Gameboy, use `system`
- I would recommend splitting all of this out into a module. Some of the other modules we have should be able to guide you through this, as well as the wiki.

I'd suggest starting out with this in a separate repo, `rlewicki/premake-gameboy` for example. Add it to the list of modules [here](https://github.com/premake/premake-core/wiki/Modules) and if there's enough interest in having it in core we can look into that.

rlewicki:
Thank you for the response and valuable feedback. I will work on the mentioned module in my spare time then.

starkos:
It sounds like this one is now be worked on else where (be sure to add a link to the [Modules page](https://github.com/premake/premake-core/wiki/Modules) if so!), so I'm going to go ahead and close this one.

PR #1258, opened 2019-03-17
Added support for Visual Studio 2019 solutions

CosmicRey:
Uses new action `vs2019`

### DIFF
--- a/modules/vstudio/tests/fs2005/projectsettings.lua
+++ b/modules/vstudio/tests/fs2005/projectsettings.lua
@@ -153,6 +153,24 @@
 		]]
 	end
 
+	function suite.onVs2019()
+		p.action.set("vs2019")
+		prepare()
+		test.capture [[
+	<PropertyGroup>
+		<Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+		<Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+		<ProjectGuid>{AE61726D-187C-E440-BD07-2556188A6565}</ProjectGuid>
+		<OutputType>Exe</OutputType>
+		<AppDesignerFolder>Properties</AppDesignerFolder>
+		<RootNamespace>MyProject</RootNamespace>
+		<AssemblyName>MyProject</AssemblyName>
+		<TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
+		<FileAlignment>512</FileAlignment>
+		<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+	</PropertyGroup>
+		]]
+	end
 
 	function suite.onVs2015_462()
 		p.action.set("vs2015")
--- a/modules/vstudio/tests/vc2010/test_globals.lua
+++ b/modules/vstudio/tests/vc2010/test_globals.lua
@@ -277,6 +277,22 @@
 		]]
 	end
 
+	
+	function suite.windowsTargetPlatformVersionLatest_on2019()
+		p.action.set("vs2019")
+		systemversion "latest"
+		prepare()
+		test.capture [[
+<PropertyGroup Label="Globals">
+	<ProjectGuid>{42B5DBC6-AE1F-903D-F75D-41E363076E92}</ProjectGuid>
+	<IgnoreWarnCompileDuplicatedFilename>true</IgnoreWarnCompileDuplicatedFilename>
+	<Keyword>Win32Proj</Keyword>
+	<RootNamespace>MyProject</RootNamespace>
+	<WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+</PropertyGroup>
+		]]
+	end
+
 --
 -- Check that the "latest" systemversion will not add <WindowsTargetPlatformVersion>
 -- when the action is less than 2017
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -63,8 +63,12 @@
 
 	function m.project(prj)
 		local action = p.action.current()
-		p.push('<Project DefaultTargets="Build" ToolsVersion="%s" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">',
-			action.vstudio.toolsVersion)
+		if _ACTION >= "vs2019" then
+			p.push('<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">')
+		else
+			p.push('<Project DefaultTargets="Build" ToolsVersion="%s" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">',
+				   action.vstudio.toolsVersion)
+		end
 	end
 
 
@@ -2507,7 +2511,13 @@
 			-- handle special "latest" version
 			if min == "latest" then
 				-- vs2015 and lower can't build against SDK 10
-				min = iif(_ACTION >= "vs2017", m.latestSDK10Version(), nil)
+				-- vs2019 allows for automatic assignment to latest
+				-- Windows 10 sdk if you set to "10.0"
+				if _ACTION >= "vs2019" then
+					min = "10.0"
+				else
+					min = iif(_ACTION == "vs2017", m.latestSDK10Version(), nil)
+				end
 			end
 
 			return min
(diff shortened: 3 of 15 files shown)


fsfod:
I guess you beat me to punch to get a PR up for this https://github.com/fsfod/premake-core/commits/vs2019. You can just take any of my changes if you want for this PR most of them are same as yours.
I was just diffing between Visual Studio generated console projects between 2017 and 2019 for most of what i figured out. Something i did miss was the the ToolsVersion attribute changed to "Current" for vcxproj.user files instead of "15" idk how much it really matters to visual studio through.

CosmicRey:
Awesome thanks, I'll take a look!

modules/vstudio/tests/vc2010/test_globals.lua
```diff
@@ -277,6 +277,25 @@
 		]]
 	end
 
+	
+	function suite.windowsTargetPlatformVersionLatest_on2019()
+		p.action.set("vs2019")
+		systemversion "latest"
+		local oldRegistry = os["getWindowsRegistry"]
+		os["getWindowsRegistry"] = function (key) return "10.0.11111" end
```
samsinsane: Keeping this override makes the unit test confusing, it's probably best to remove this and make it clear here that VS2019 uses `10.0` to indicate "use latest SDK".

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2507,7 +2511,13 @@
 			-- handle special "latest" version
 			if min == "latest" then
 				-- vs2015 and lower can't build against SDK 10
-				min = iif(_ACTION >= "vs2017", m.latestSDK10Version(), nil)
+				-- vs2019 allows for automatic assignment to latest
+				-- Windows 10 sdk if you set to "10.0"
```
samsinsane: I haven't had a chance to try out VS2019 yet, is this a fully documented value? Can it be chosen in the dropdown in the GUI? If not, perhaps this should be `auto` and `latest` should stay as it was?

modules/vstudio/vs2010_vcxproj_user.lua
```diff
@@ -54,8 +54,12 @@
 
 	function m.userProject()
 		local action = p.action.current()
-		p.push('<Project ToolsVersion="%s" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">',
-			action.vstudio.toolsVersion)
+		if _ACTION >= "vs2019" then
+			p.push('<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">')
```
samsinsane: Why couldn't the `toolsVersion` in the VS2019 action be set to `"Current"` instead? The point of the `toolsVersion` is to avoid multiple code paths, so it would be good to maintain this if we can.

modules/vstudio/vs2019.lua
```diff
@@ -0,0 +1,68 @@
...
+		end,
+
+		pathVars = vstudio.vs2010.pathVars,
+
+		-- This stuff is specific to the Visual Studio exporters
+
+		vstudio = {
+			solutionVersion = "12",
+			versionName     = "16",
+			targetFramework = "4.7.2",
```
samsinsane: Did you confirm this from generating the project manually? It's just kind of weird that VS2017 was `4.5.2` and VS2019 is `4.7.2`.

modules/vstudio/tests/fs2005/projectsettings.lua
```diff
@@ -153,6 +153,24 @@
...
+		test.capture [[
+	<PropertyGroup>
+		<Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+		<Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+		<ProjectGuid>{AE61726D-187C-E440-BD07-2556188A6565}</ProjectGuid>
+		<OutputType>Exe</OutputType>
+		<AppDesignerFolder>Properties</AppDesignerFolder>
+		<RootNamespace>MyProject</RootNamespace>
+		<AssemblyName>MyProject</AssemblyName>
+		<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
samsinsane: Did you confirm this by manually creating the project? It's a bit strange the version didn't change between VS2017 and VS2019.

modules/vstudio/vs2010_vcxproj.lua
CosmicRey: Yes and it is set by default in projects. I think this move was due to the consistent addition of Windows SDK versions and how it would conflict with builds on different machines. 
Since its reasonable to expect the same (if not more) amount of sdk changes, they added this as a choice.

And yes this a choice you can make in the drop down menu 😁.

modules/vstudio/vs2019.lua
CosmicRey: Yes this is what is generated when you create a C# project.

modules/vstudio/tests/vc2010/test_globals.lua
CosmicRey: Good point. Fixed!

modules/vstudio/tests/fs2005/projectsettings.lua
CosmicRey: Yea, that's an oversight by me, fixed!

CosmicRey:
@samsinsane Just pinging to see if there is anything else needed.

modules/vstudio/vs2010_vcxproj_user.lua
samsinsane: Interesting, perhaps allowing `action.vstudio.toolsVersion` to be a function? I'm fairly certain we have this in a few other places, checking the type is a function, otherwise assuming it's the expected type. The functions tend to have a single parameter `wks`, `prj` or `cfg` whichever is appropriate. This would allow you to modify the usages of `action.vstudio.toolsVersion` but only convert VS2019 to a function.

CosmicRey: Yea okay, but would wks, prj, or cfg know that its making a vcxproj.user file? My inclination would be it isn't, since the function would basically have to do something like so:
```
if projectType == "user" then
    return "Current"
else
    return "15.0"
end
```
It would be "simpler" to just pass in whether it is the `vcxproj.user` file as a single parameter.

CosmicRey: An alternative could be if we changed toolsVersion into a table... So instead we could have
```
toolsVersion = {
    project = "15.0",
    user = "Current"
}
```
The downside would be we would need to change all the instances where toolsVersion was used. More up-front work to get it working, but it would solve the problem relatively elegantly.

Or we just change the `vstudio` table to have a `userToolsVersion` variable, although the result would be the same.

CosmicRey: I came up with a better alternative, I just added a table value to 2019 only, userToolsVersion, which is just optionally checked by this:
```
p.push('<Project ToolsVersion="%s" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">',
				action.vstudio.userToolsVersion or action.vstudio.toolsVersion)
```
Hope this resolves the issue for you

CosmicRey: @samsinsane ping

WorldofBay:
lgtm but if you want to fulfill style things i'd have 3 little things:
- end tables with comma so that future edits are easily copypastable/single-line changes (modules/vstudio/_manifest.lua)
- maybe for the same reason add an or false in modules/vstudio/_preload.lua:
```lua
_ACTION == "vs2019" or
false;
```
- there is no trailing newline in modules/vstudio/tests/vc2010/test_header.lua


however, those are not necessary, just nice to have

at the very end you should squash your commits when you're ok then sam will merge once he's got the time :)

modules/vstudio/vs2010_vcxproj_user.lua
samsinsane: You could have used `project.isdotnet(prj)` in the function to swap between the values. However, I think splitting them into the two keys in the action is probably for the best since there is actually three values; nothing in `vcxproj` (omitting the `ToolsVersion` attribute), `Current` in the .NET projects, and `15.0` in `vcxproj.user` (Shouldn't this be `16.0`? VS2017 is 15.0). Your way is the least complicated, and ultimately is far more flexible if Microsoft decide to read `ToolsVersion` to the `vcxproj` in VS2021.

modules/vstudio/vs2019.lua
```diff
@@ -0,0 +1,69 @@
...
+			toolsVersion    = "15.0",
```
samsinsane: I asked this in another reply, but just to ensure it isn't missed: Should this be `16.0`?

CosmicRey: As it stands, they did not increase the `toolsVersion` it is still set to `15.0`. I just updated the RC to the official release and created a new project to confirm. It is still set to `15.0`.

samsinsane: Weird, thank you for confirming this for me!

PR #1259, opened 2019-03-20
Adds shaderincludedirs to fxcompile configuration

tekgoblin:
Added shaderincludedirs to the visual studio configuration for fxcompile. Kept name inline with the premake includedirs

### DIFF
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -39,6 +39,14 @@
 	}
 
 	p.api.register {
+		name = "shaderincludedirs",
+		scope = "config",
+		kind = "list:string",
+		tokens = true,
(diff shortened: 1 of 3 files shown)


modules/vstudio/_preload.lua
```diff
@@ -38,6 +38,14 @@
 		tokens = true,
 	}
 
+	p.api.register {
+		name = "shaderincludedirs",
+		scope = "config",
+		kind = "list:string",
```
samsinsane: Why not `list:directory` like `includedirs`?

starkos:
Shame this one stalled so close to the finish line. Closing this in favor of #1355, which changes the field type to "list:directory" as suggested.

Issue #1251, opened 2019-03-02
Issues with premake-android


ifarbod:
I've been using the official `android` module way before it made its way to premake-core, I already got sick of gradle/CMake/ninja at the time but thankfully this extension was a lifesaver, it's nice to have everything in one IDE.
Having used it for more than a year, I came along with many issues that I had to manually fix and work around, eventually, I switched to managing manual VS solutions for this sole reason. In fact I only do so for Android and iOS while everything else (win32/macos) is covered by premake.
After I saw it getting merged into upstream `premake-core`, I've noticed that it has received some fixes, So I started to switch back slowly. While it's more stable than ever, it's far enough to be feature complete. Here's a short list of my current issues and concerns:

- [x] Rework the `stl` API, have `staticruntime` to decide instead. Example:
```lua
stl "libc++"
staticruntime "On"
-- Would produce: "c++_static"
-- Or:
stl "gnustl"
staticruntime "Off"
-- Would produce: "gnustl_shared"
```
- [x] The `MultiProcessorCompilation` flag has no effect. This is because Android MDD projects use `UseMultiToolTask` instead.
- [x] The `kind` value for packaging projects `androidproj` doesn't really feel inline with the built-in ones like `SharedLib`, `ConsoleApp` and such, my suggestion would be to change it to something like `AndroidApp`, `Packaging` or `WindowedApp` if it's guaranteed to not conflict with anything.
- [x] Some of the windows/VS-specific attributes can easily end up in Android projects if you use them at the workspace scope, ex: `characterset "Unicode"`
- [ ] A simple way to change the Ant build target, so compiling signed APKs becomes possible. `signapk "On"`, I don't know of a better name for this.
- [ ] `toolchainversion`'s functionality could be merged with `toolset`, maybe something like `toolset "clang-5"`
- [x] `thumbmode` only works for the old vs-android extension, also could change its name to `instructionmode`.
- [x] A lot of the well-established premake APIs don't have any effect in VSAndroid, examples include but not limited to: `disablewarnings`, `pic`, `vectorextensions`, `rtti`, `exceptionhandling`.
- [x] `exceptionhandler "UnwindTables"`
- [x] The value `C++latest` for `cppdialect` should generate a `-std=c++1z`, Meanwhile `C++17` generates `AdditionalBuildOptions` instead of `CppLanguageStandard`. `cdialect` simply doesn't work.
[...]

PR #1263, opened 2019-03-28
Android module changes, stage 1

ifarbod:
As discussed in #1251, this is a PR that contains merely fixes and minor changes, I'll do the second PR with the new functions and features once this gets merged.

Here's a quick brief of what I did:
- Reworked the `stl` API, using `staticruntime` to decide on shared/static.
```lua
stl "libc++"
staticruntime "On"
-- Would produce: "c++_static"
-- Or:
stl "gnustl"
staticruntime "Off"
-- Would produce: "gnustl_shared"
```
- Fixed the `MultiProcessorCompilation` flag.
- Renamed `androidproj` to `Packaging`.
- Prevented some Windows-exclusive attributes from ending up in Android projects, such as CharacterSet.
- Added `NEON` support for `vectorextensions`.
- Added `UnwindTables` support for `exceptionhandling`.
- Fixed `targetName` not working in Packaging projects.
- Included #1112 (minus the unit tests) which fixes `disablewarnings`, `exceptionhandling` and `rtti`.
- Unify `cppdialect` and `cdialect`, added support for `C++latest` which maps to C++17 at the moment.
- Fixed `pic`.
- Fixed `thumbmode`.

### DIFF
--- a/modules/android/_preload.lua
+++ b/modules/android/_preload.lua
@@ -13,11 +13,12 @@
 --
 
 	p.ANDROID = "android"
-	p.ANDROIDPROJ = "androidproj"
+	p.ANDROIDPROJ = "Packaging"
 
 	api.addAllowed("system", p.ANDROID)
-	api.addAllowed("architecture", { "armv5", "armv7", "aarach64", "mips", "mips64", "arm" })
+	api.addAllowed("architecture", { "armv5", "armv7", "aarch64", "mips", "mips64", "arm" })
 	api.addAllowed("vectorextensions", { "NEON", "MXU" })
+	api.addAllowed("exceptionhandling", {"UnwindTables"})
 	api.addAllowed("flags", { "Thumb" })
 	api.addAllowed("kind", p.ANDROIDPROJ)
 
@@ -65,6 +66,7 @@
 			"3.5",
 			"3.6",
 			"3.8",
+			"5.0",
 		},
 	}
 
@@ -74,15 +76,10 @@
 		kind = "string",
 		allowed = {
 			"none",
-			"minimal c++ (system)",
-			"c++ static",
-			"c++ shared",
-			"stlport static",
-			"stlport shared",
-			"gnu stl static",
-			"gnu stl shared",
-			"llvm libc++ static",
-			"llvm libc++ shared",
+			"gabi++",
+			"stlport",
+			"gnu",
+			"libc++",
 		},
 	}
 
--- a/modules/android/vsandroid_vcxproj.lua
+++ b/modules/android/vsandroid_vcxproj.lua
@@ -17,6 +17,25 @@
 
 
 --
+-- Utility functions
+--
+
+	function table.findAndRemove(t, value)
+		for i, v in ipairs(t) do
+    		if v == value then
+      			table.remove(t, i)
+				break
+    		end
+  		end
+	end
+
+	local function setBoolOption(optionName, flag, value)
+		if flag ~= nil then
+			vc2010.element(optionName, nil, value)
+		end
+	end
+
+--
 -- Add android tools to vstudio actions.
 --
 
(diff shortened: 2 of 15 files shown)


ifarbod:
Looks like some of the tests are failing, is the order of these really important?

```xml
stack traceback:
	...make-core/modules/android/tests/test_android_project.lua:35: in function 'testFunction'
[  FAILED  ] test_android_project.rttiOn
...make-core/modules/android/tests/test_android_project.lua:46: (4) expected:
	<ExceptionHandling>Enabled</ExceptionHandling>
...but was:
	<RuntimeTypeInfo>true</RuntimeTypeInfo>
fulltext:
<ClCompile>
	<RuntimeTypeInfo>true</RuntimeTypeInfo>
</ClCompile>
```

modules/android/vsandroid_vcxproj.lua
```diff
@@ -78,6 +97,11 @@
 
 	premake.override(vc2010.elements, "configurationProperties", function(oldfn, cfg)
 		local elements = oldfn(cfg)
+
+		table.findAndRemove(elements, vc2010.characterSet)
```
WorldofBay: wrap those removes in a check for cfg.system == p.ANDROID or else a premake call can't deal with vsandroid and vs2010 both present

WorldofBay:
the other wrong test aren't because of order but because of default values in `_premake-init.lua`

both `rtti` and `exceptionhandling` are set to "Default" if not specified.

if the new behaviour is better you need to edit the tests.
i have no idea why the current test `suite.exceptionHandlingOff` has no rtti set as it should get from the defaults but for the rest the defaults seem to trigger defaults if any configuration is done.

modules/android/vsandroid_vcxproj.lua
ifarbod: Sure thing, [fixed](https://github.com/ifarbod/premake-core/commit/dfdaf2ca160b3a4adddc5743dccb16e0f0ddcc48).

modules/android/vsandroid_vcxproj.lua
```diff
@@ -326,10 +391,16 @@
...
+		if cfg.vectorextensions == "NEON" then
+			vc2010.element("EnableNeonCodegen", nil, "true")
+		end
+	end
+
 	premake.override(vc2010, "runtimeTypeInfo", function(oldfn, cfg, condition)
 		if cfg.system == premake.ANDROID then
 			-- Note: Android defaults to 'off'
-			if cfg.rtti then
+			if cfg.rtti and cfg.rtti ~= premake.OFF then
```
WorldofBay: `rtti`, if not set, defaults to "Default" which will be a `true` here, while the comment says the default is `off`, so `false`.

i don't know which one is correct but either the comment is wrong or you need to add a check for "Default".

also you'll need to change almost every test in the `test_android_project` suite to look for the correct behaviour for both `rtti` and `exceptionhandling`.

modules/android/_preload.lua
```diff
@@ -13,11 +13,12 @@
 --
 
 	p.ANDROID = "android"
-	p.ANDROIDPROJ = "androidproj"
+	p.ANDROIDPROJ = "Packaging"
```
WorldofBay: there is a test in `test_android_files` that crashes because it uses "androidproj".

is it ok to make it crash for users?

anyways you'll have to change the test to use the correct "Packaging".

WorldofBay:
the rest looks clean but since i know what sam will say i can say it now before he reads it, so you're already done by then:
once you're finished, squash those commits into a single one

modules/android/vsandroid_vcxproj.lua
ifarbod: I believe this bit is from @TurkeyMan's PR.

modules/android/_preload.lua
ifarbod: All right, I'm gonna fix those in a second.

ifarbod:
Thanks for the quick reviews, will do once I figure out the rest of the unit test issues

ifarbod:
@WorldofBay All of the tests pass now, the GCC builds on Travis didn't start for some reason, I'll squash my commits into one if everything else is okay.

ifarbod:
I messed up my local checkout beyond repair after squashing with premake-core's changes from master, I don't like having a lot of merge commits so here's another PR: #1264

PR #1264, opened 2019-03-28
Android changes, stage 1 (#1263)

ifarbod:
As discussed in #1251, this is a PR that contains merely fixes and minor changes, I'll do the second PR with the new functions and features once this gets merged.

Here's a quick brief of what I did:
- Reworked the `stl` API, using `staticruntime` to decide on shared/static.
```lua
stl "libc++"
staticruntime "On"
-- Would produce: "c++_static"
-- Or:
stl "gnustl"
staticruntime "Off"
-- Would produce: "gnustl_shared"
```
- Fixed the `MultiProcessorCompilation` flag.
- Renamed `androidproj` to `Packaging`.
- Prevented some Windows-exclusive attributes from ending up in Android projects, such as CharacterSet.
- Added `NEON` support for `vectorextensions`.
- Added `UnwindTables` support for `exceptionhandling`.
- Fixed `targetName` not working in Packaging projects.
- Included #1112 (minus the unit tests) which fixes `disablewarnings`, `exceptionhandling` and `rtti`.
- Unify `cppdialect` and `cdialect`, added support for `C++latest` which maps to C++17 at the moment.
- Fixed `pic`.
- Fixed `thumbmode`.


modules/android/tests/test_android_project.lua
```diff
@@ -41,7 +42,9 @@
 	end
 
 	function suite.rttiOn()
+		exceptionhandling "On"
 		rtti "On"
+
```
ratzlaff: This newline breaks consistency with the other tests

modules/android/vsandroid_vcxproj.lua
```diff
@@ -45,10 +64,10 @@
 
 		return elements
 	end)
-	
+
 	premake.override(vc2010.elements, "globalsCondition", function (oldfn, prj, cfg)
 		local elements = oldfn(prj, cfg)
-		
+
```
ratzlaff: Can you revert these whitespace changes?

modules/android/vsandroid_vcxproj.lua
```diff
@@ -16,6 +16,25 @@
...
+--
+
+	function table.findAndRemove(t, value)
+		for i, v in ipairs(t) do
+    		if v == value then
+      			table.remove(t, i)
+				break
+    		end
+  		end
+	end
```
ratzlaff: This should go in table.lua
There needs to be regression tests for its functionality
The indentation is inconsistent

modules/android/vsandroid_vcxproj.lua
```diff
@@ -95,42 +119,40 @@
...
-				["stlport shared"] = "stlport_shared",
-				["gnu stl static"] = "gnustl_static",
-				["gnu stl shared"] = "gnustl_shared",
-				["llvm libc++ static"] = "c++_static",
-				["llvm libc++ shared"] = "c++_shared",
+				none = "system",
+				["gabi++"] = "gabi++",
+				stlport = "stlport",
+				gnustl = "gnustl",
+				["libc++"] = "c++",
```
ratzlaff: Can you fix the style so that it is consistent? ["..."] is missing from 'none', 'stlport' and 'gnustl'

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -616,7 +616,7 @@
 		if kind == "list:path" then
 			return table.concat(vstudio.path(cfg, value), ';')
 		end
-	
+
```
ratzlaff: Can you revert this whitespace change?

modules/android/vsandroid_vcxproj.lua
ifarbod: These are caused by the editorconfig from the project so it seemed normal to me, but sure.

modules/android/tests/test_android_project.lua
ifarbod: Fixed.

modules/android/vsandroid_vcxproj.lua
ifarbod: [Done](https://github.com/premake/premake-core/pull/1264/commits/b01f495a258218e3d56f57807c029056b5c5a44f)

modules/android/vsandroid_vcxproj.lua
ifarbod: Shouldn't I do that in another PR?

ratzlaff: @samsinsane would have the final word on that one.
My feelings is that since this is a new utility function written for a specific use case, it would make sense to be added in this PR to give it context for anyone who later asks 'when/why was this put in'.

But I dont have the final word on that, so I'll defer to sam =)

ifarbod: Right, thanks for the review either way.

WorldofBay: i looked into `table.lua` now and saw you can use `table.difference` instead.

```lua
-- instead of
-- table.findAndRemove(elements, vc2010.characterSet)
-- table.findAndRemove(elements, vc2010.wholeProgramOptimization)
-- table.findAndRemove(elements, vc2010.windowsSDKDesktopARMSupport)
-- do
elements = table.difference(elements, {
  vc2010.characterSet,
  vc2010.wholeProgramOptimization,
  vc2010.windowsSDKDesktopARMSupport,
})
-- or
elements = table.filter(elements, function(entry) {
  local tbl = {
    [vc2010.characterSet] = true,
    [vc2010.wholeProgramOptimization] = true,
    [vc2010.windowsSDKDesktopARMSupport] = true,
  }
  return not tbl[entry]
})
```

both of them do not change in-place but copy the table.

i suggest you use one of those for now and in a separate standalone pull request you (or me or anyone else) adds the in-place removal to `table.lua` and changes those calls to your call - but you should really think about allowing a table and maybe also a function (beware of clashes, as you can't remove a single function without wrapping it in a table then) as input instead of just a single element.

ifarbod: @WorldofBay I guess `table.difference` will do it for now - and feel free to do it yourself :)

modules/vstudio/vs2010_vcxproj.lua
ifarbod: Not sure if I can do that with EditorConfig intervening upon commit.

modules/android/vsandroid_vcxproj.lua
```diff
@@ -314,9 +380,15 @@
 	premake.override(vc2010, "exceptionHandling", function(oldfn, cfg, condition)
 		if cfg.system == p.ANDROID then
 			-- Note: Android defaults to 'off'
-			if cfg.exceptionhandling then
+			if cfg.exceptionhandling and cfg.exceptionhandling ~= "Default" then
```
samsinsane: Instead of adding this check, you could instead check `if exceptions[cfg.exceptionhandling] ~= nil then` before calling `vc2010.element`. You might even be able to remove this whole if statement, but you would need to add a check for the non-VS2015 version: `if cfg.exceptionhandling == premake.ON then`.

modules/android/vsandroid_vcxproj.lua
```diff
@@ -326,10 +398,16 @@
...
+		if cfg.vectorextensions == "NEON" then
+			vc2010.element("EnableNeonCodegen", nil, "true")
+		end
+	end
+
 	premake.override(vc2010, "runtimeTypeInfo", function(oldfn, cfg, condition)
 		if cfg.system == premake.ANDROID then
 			-- Note: Android defaults to 'off'
-			if cfg.rtti then
+			if cfg.rtti and cfg.rtti ~= premake.OFF then
```
samsinsane: `if cfg.rtti == premake.ON then`

modules/android/vsandroid_vcxproj.lua
```diff
@@ -211,19 +236,38 @@
...
---				android.ShortEnums,
+				android.shortEnums,
+				android.cStandard,
+				android.cppStandard,
 			})
 			if _ACTION >= "vs2015" then
 				table.remove(elements, table.indexof(elements, vc2010.debugInformationFormat))
-				table.remove(elements, table.indexof(elements, android.thumbMode))
+
+				elements = table.difference(elements, {
```
samsinsane: Why not just use `table.remove` as above?

modules/android/vsandroid_vcxproj.lua
```diff
@@ -211,19 +236,38 @@
...
+					-- Ignore multiProcessorCompilation for android projects, they use UseMultiToolTask instead.
+					vc2010.multiProcessorCompilation,
+					-- minimalRebuild also ends up in android projects somehow.
+					vc2010.minimalRebuild,
+				})
+
+				-- VS has NEON support through EnableNeonCodegen.
+				table.replace(elements, vc2010.enableEnhancedInstructionSet, android.enableEnhancedInstructionSet)
+				-- precompiledHeaderFile support.
+				table.replace(elements, vc2010.precompiledHeaderFile, android.precompiledHeaderFile)
```
samsinsane: This isn't wrong, but it feels inconsistent with how this and other modules work, which is to override the existing functions and change or extend their behaviour - there's plenty of examples of this in this file, including this function.

modules/android/_preload.lua
```diff
@@ -13,11 +13,12 @@
 --
 
 	p.ANDROID = "android"
-	p.ANDROIDPROJ = "androidproj"
+	p.ANDROIDPROJ = "Packaging"
```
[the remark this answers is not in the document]
ifarbod: Right, should I move p.ANDROID too?

modules/android/_preload.lua
```diff
@@ -13,11 +13,12 @@
...
 
 	api.addAllowed("system", p.ANDROID)
-	api.addAllowed("architecture", { "armv5", "armv7", "aarach64", "mips", "mips64", "arm" })
+	api.addAllowed("architecture", { "armv5", "armv7", "aarch64", "mips", "mips64", "arm" })
 	api.addAllowed("vectorextensions", { "NEON", "MXU" })
+	api.addAllowed("exceptionhandling", {"UnwindTables"})
```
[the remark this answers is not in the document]
ifarbod: GCC does have `-funwind-tables` https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html (so does clang) - Could leave this here for now until another PR.

modules/android/vsandroid_androidproj.lua
```diff
@@ -132,6 +132,9 @@
 		if cfg.kind == p.ANDROIDPROJ then
 			return {
 				android.outDir,
+				vc2010.intDir,
+				vc2010.targetName,
+				--targetExt,
```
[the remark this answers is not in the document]
ifarbod: Done.

modules/android/_preload.lua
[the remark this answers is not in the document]
ifarbod: After replacing all occurrences of `ANDROIDPROJ` to `PACKAGING`, I can't compile premake anymore:

`E:\projects\premake-core>premake5 embed
Error: [string "src/base/api.lua"]:147: attempt to index a nil value (local 'value')`

The error is not that helpful unfortunately.

modules/android/vsandroid_vcxproj.lua
ifarbod: [Fixed](https://github.com/ifarbod/premake-core/commit/8c7ccc99d4f2d913f1bbd4e30e6063e6a7f4bf87).

modules/android/_preload.lua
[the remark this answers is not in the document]
ifarbod: , https://github.com/ifarbod/premake-core/commit/4821372bfc0967f5f8f3fab8afa1686f416c5795

modules/android/tests/test_android_project.lua
```diff
@@ -30,6 +30,7 @@
 	end
 
 	function suite.rttiOff()
+		exceptionhandling "On"
```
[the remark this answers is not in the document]
ifarbod: https://github.com/ifarbod/premake-core/commit/4821372bfc0967f5f8f3fab8afa1686f416c5795

modules/android/vsandroid_vcxproj.lua
ifarbod:
```lua
local exceptions = {
				On = "Enabled",
				Off = "Disabled",
				UnwindTables = "UnwindTables",
			}
			if exceptions[cfg.exceptionhandling] ~= nil then
				if _ACTION >= "vs2015" then
					vc2010.element("ExceptionHandling", condition, exceptions[cfg.exceptionhandling])
				else
					if cfg.exceptionhandling == premake.ON then
						vc2010.element("GccExceptionHandling", condition, "true")
					end
				end
			end```

I hope this is good enough :)

Comment (ifarbod) [reply modules/android/vsandroid_vcxproj.lua]: https://github.com/ifarbod/premake-core/commit/6b06e408e69fb56770860fb26335ab2ac68697a2

File: modules/android/_preload.lua [hunk 14]
```diff
@@ -65,6 +66,7 @@
 			"3.5",
 			"3.6",
 			"3.8",
+			"5.0",
```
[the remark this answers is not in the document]
ifarbod: I'll try doing it myself in the next PR.

modules/android/vsandroid_vcxproj.lua
samsinsane: It's a tad wonky having the action check inside the LUT check, but it's fine.

modules/android/_preload.lua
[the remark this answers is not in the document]
samsinsane: I'm not sure how that could happen, perhaps try `premake5 --scripts=. embed`?

modules/android/vsandroid_vcxproj.lua
```diff
@@ -16,6 +16,25 @@
 	local config = p.config
 
 
+--
+-- Utility functions
+--
+
+	function table.findAndRemove(t, value)
```
samsinsane: This appears to be unused and can probably be removed.

modules/android/vsandroid_vcxproj.lua
ifarbod: I hope this is better: https://github.com/ifarbod/premake-core/commit/d8742697d345325fccf0ec8eb445c0698439b65b

modules/android/vsandroid_vcxproj.lua
ifarbod: Removed the function for this PR, but as @WorldofBay said, I hope we can get it in `table.lua` someday.

modules/android/_preload.lua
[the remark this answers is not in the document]
ifarbod: There's even more weird errors now:

```lua
Error: [string "src/base/_foundation.lua"]:57: cannot open vsandroid_vcxproj.lua: No such file or directory
```

Which doesn't make sense, again.

modules/android/vsandroid_vcxproj.lua
WorldofBay: is the Off entry even necessary here as it matches the default?

also @samsinsane what about `p.ON` instead of `"On"` ?

with `iif` it then boils down to that:

```lua
if _ACTION >= "vs2015" then
	local exception = iif(cfg.exceptionhandling == p.ON, "Enabled", cfg.exceptionhandling)
	if exception then
		vc2010.element("ExceptionHandling", condition, exception)
	end
else
	if cfg.exceptionhandling == p.ON then
		vc2010.element("GccExceptionHandling", condition, "true")
	end
end
```

ifarbod:
I think I messed something up here, gonna look into it now.

modules/android/vsandroid_vcxproj.lua
ifarbod: Should I switch over to the class `premake.override` calls for this? I went with `table.replace` as it seemed more convenient to me and I've seen some modules (not the official ones) do it like this.

modules/android/_preload.lua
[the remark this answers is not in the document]
ifarbod: I committed the replacements, but as soon as I move `premake.PACKAGING` to _foundation, I get that error.

modules/android/vsandroid_vcxproj.lua
samsinsane: Hmm, looks like GitHub ate my reply. This is probably fine, I prefer `premake.override` because I'm used to it but there isn't really much of a difference between the two approaches. I've marked this as resolved.

modules/android/_preload.lua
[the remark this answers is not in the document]
samsinsane: That's really weird, let's keep it here for now then.

samsinsane:
Once you squish the commits down it should be good to merge! 👍

modules/android/_preload.lua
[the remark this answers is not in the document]
ifarbod: https://github.com/ifarbod/premake-core/tree/android-fixes-stage-1

src/base/_foundation.lua
```diff
@@ -34,6 +34,7 @@
 	premake.CSHARP      = "C#"
 	premake.GCC         = "gcc"
 	premake.HAIKU       = "haiku"
+    premake.ANDROID     = "android"
```
tritao: Drive by review, but there's some minor indentation issues here

ifarbod: oops, will fix it now

modules/android/vsandroid_vcxproj.lua
ifarbod: @WorldofBay I like the `iif` approach more :)

ifarbod:
@samsinsane Thanks for the review, will squash everything into one now.

modules/android/vsandroid_vcxproj.lua
samsinsane: > is the Off entry even necessary here as it matches the default?

Yes, if a user explicitly specifies `Off` it should turn it off. Today the default is off, but tomorrow the default could easily be on. The user knows what they need, we shouldn't try to get all fancy here.

> also @samsinsane what about `p.ON` instead of `"On"`?

Not sure what you're referencing?

> with `iif` it then boils down to that:

The snippet doesn't work with two of the four possible values for `exceptionhandling`, and same as the first point, the user knows what they want.

modules/vstudio/vs2010_vcxproj.lua
ratzlaff: Huh, I did not know that github supported this. Nevermind then, its doing what it is supposed to!

PR #1266, opened 2019-04-05
Implement the 'visibility' API for Android projects

ifarbod:
As discussed in https://github.com/premake/premake-core/issues/1251#issuecomment-479895565

### DIFF
--- a/modules/android/vsandroid_vcxproj.lua
+++ b/modules/android/vsandroid_vcxproj.lua
@@ -338,6 +338,11 @@
 				end
 			end
 
+			-- -fvisibility=<>
+			if cfg.visibility ~= nil then
+				table.insert(opts, p.tools.gcc.cxxflags.visibility[cfg.visibility])
+			end
+
 			if #opts > 0 then
 				opts = table.concat(opts, " ")
 				vc2010.element("AdditionalOptions", condition, '%s %%(AdditionalOptions)', opts)


modules/android/vsandroid_vcxproj.lua
```diff
@@ -338,6 +338,17 @@
...
 
+			-- -fvisibility=<>
+			if cfg.visibility ~= nil then
+				local visibility_map = {
+					["Default"] 	= "default",
+					["Hidden"] 		= "hidden",
+					["Internal"] 	= "internal",
+					["Protected"] 	= "protected",
+				}
+				table.insert(opts, '-fvisibility=' .. visibility_map[cfg.visibility])
```
ratzlaff: This mapping exists in the [gcc tool](https://github.com/premake/premake-core/blob/master/src/tools/gcc.lua#L193). Does it need to be duplicated here?

ifarbod: Can I access it like this? `gcc.cxxflags.visibility` ?

ifarbod: I just tried it and I'm getting an error: `Error: [string "android/vsandroid_vcxproj.lua"]:268: attempt to index a nil value (global 'gcc')`

ifarbod: I think it's better now https://github.com/ifarbod/premake-core/commit/044cddd81a00e65a82e4eada63100b3ff799e60c

ifarbod:
@samsinsane There's the "Squash and merge" option on the dropdown next to the Merge button but sure, I'll squash them anyway.

PR #1271, opened 2019-04-24
Fix xcodebuildresources with paths that contain special characters

sp-jordi-vilalta:
The resource paths were used as patterns, which interpreted some special characters. This PR changes it to use a plain search.


ratzlaff:
Would it be possible to make a test for this specific case? This looks like something that could easily regress.

sp-jordi-vilalta:
Sure, I added a test.

ratzlaff:
Looks good to me. I know that @samsinsane will request you to squash your commits into one before he accepts it.

sp-jordi-vilalta:
Done ;)

samsinsane:
I don't really like this change, it removes functionality from an awful API to make it somewhat more awful but less awful for the simple case. You've made it easier to add paths with special characters, but you've made it impossible to only add the PNG files from a specific folder.

If the API kind was changed to be the same as `files`, it would allow for both `file-thing.ext` and `path/**.png`. Thoughts? Knowing the XCode generator, I doubt it could be this simple.

starkos:
I take it back: while this PR is technically correct (we should not be doing Lua pattern checks there), "fixing" it to support Premake patterns supersedes this change. Closing this one in favor of #1356.

@sp-jordi-vilalta: thanks for taking the time to submit this fix and kicking off the conversation! Sorry this one didn't get merged; it was the right idea.

PR #1275, opened 2019-05-01
Fixed various issues with escaping in CodeLite generator

samsinsane:


modules/codelite/codelite_project.lua
```diff
@@ -207,7 +207,10 @@
 			_x(4, '<IncludePath Value="%s"/>', project.getrelative(cfg.project, includedir))
 		end
 		for _, define in ipairs(cfg.defines) do
-			_x(4, '<Preprocessor Value="%s"/>', p.esc(define))
+			-- All characters encoded for XML (& -> &amp;) need to be escaped with backslash,
+			-- same applies to space. This was correct for CodeLite 12 and took a lot of trial
+			-- and error to verify.
+			_p(4, '<Preprocessor Value="%s"/>', p.esc(define):gsub('&', '\\&'):gsub(' ', '\\ '))
```
tdesveauxPKFX: Is it intended on the CodeLite side to have different escaping depending on the field in their project file?
This feel more like an issue to me but I maybe I'm wrong.

modules/codelite/codelite.lua
```diff
@@ -26,10 +26,19 @@
...
+		result = result:gsub('<', '&lt;')
+		result = result:gsub('>', '&gt;')
+		return result
+	end
+
 	function codelite.esc(value)
 		local result = value:gsub('&', '&amp;')
 		result = result:gsub('<', '&lt;')
 		result = result:gsub('>', '&gt;')
+		result = result:gsub('"', '&quot;')
```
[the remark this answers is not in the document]
samsinsane: I wanted to err on the side of caution with this, the escaping seems to be very fragile and inconsistent in CodeLite and I wanted to avoid future issues. For example, not escaping quotes results in the cURL project to silently fail to load in the Premake workspace - I believe there's another project that fails too.

modules/codelite/codelite_project.lua
samsinsane: Honestly, I have no idea. This one field seemed to have serious issues with everything, and caused the cURL project to not load. It seems like CodeLite converts the escaped values back, and passes the value to the compiler. If you don't use `\` on the `"`, strings come through without quotes and are treated as identifiers and builds fail. Having `\"` instead seems to cause `&`, `<`, `>` and ` ` to be treated like normal on the command line, so they also needed to be escaped to prevent builds failing, or random files being created on disk.

I'll see how these magic characters are handled in `gmake2` since both Make and CodeLite just pass params straight to the compiler. I'll then adjust this accordingly - last night I just kept iterating and didn't really consider what the other generators did to handle this craziness.

samsinsane: Looks like `gmake2` will only escape `"` and ` ` with a `\`. Builds fail if you have `&`, `<`, or `>` in `defines` without escaping them, I guess I could change the `"` escape to `\&quot;` which is what it used to be and then add ` ` -> `\ `. I'll see how that works out.

samsinsane: Fixed.

modules/codelite/codelite.lua
```diff
@@ -26,10 +26,20 @@
...
+		result = result:gsub('>', '&gt;')
+		return result
+	end
+
 	function codelite.esc(value)
 		local result = value:gsub('&', '&amp;')
 		result = result:gsub('<', '&lt;')
 		result = result:gsub('>', '&gt;')
+		result = result:gsub('"', '\\&quot;')
+		result = result:gsub(' ', '\\ ')
```
samsinsane: This is actually isolated entirely to the defines. Leaving this here breaks the debugger selection and the command line arguments.

alex-rass-88:
What about merging? CodeLite generator broken.

samsinsane:
@tdesveauxPKFX I made more changes since you last reviewed, but I'm going to merge this since as @alex-rass-88 has pointed out the generator is broken without these fixes.

PR #1278, opened 2019-05-01
Packaging script improvement

tdesveauxPKFX:
Finally managed to have some time to do this.

Changes are mostly to allow creating release packages without MinGW.

Also added a Bootstrap.bat that setup the environment for Visual Studio and run Bootstrap.mak.

Once this is merge, I will start the process for a new release.

### DIFF
--- a/Bootstrap.bat
+++ b/Bootstrap.bat
@@ -22,6 +22,9 @@ IF "%vsversion%" == "vs2010" (
 ) ELSE IF "%vsversion%" == "vs2017" (
 	CALL :VsWhereVisualBootstrap "%vsversion%" "15.0" "16.0"
 
+) ELSE IF "%vsversion%" == "vs2019" (
+	CALL :VsWhereVisualBootstrap "%vsversion%" "16.0" "17.0"
+
 ) ELSE (
 	ECHO Unrecognized Visual Studio version %vsversion%
 	EXIT /B 2
--- a/Bootstrap.bat
+++ b/Bootstrap.bat
@@ -1,0 +1,101 @@
+@ECHO OFF
+SETLOCAL
+SETLOCAL ENABLEDELAYEDEXPANSION
+
+SET vsversion=%1
+IF "%vsversion%" == "" (
+	SET vsversion=vs2015
+)
+
+IF "%vsversion%" == "vs2010" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "100"
+
+) ELSE IF "%vsversion%" == "vs2012" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "110"
+
+) ELSE IF "%vsversion%" == "vs2013" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "120"
+
+) ELSE IF "%vsversion%" == "vs2015" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "140"
+
+) ELSE IF "%vsversion%" == "vs2017" (
+	CALL :VsWhereVisualBootstrap "%vsversion%" "15.0" "16.0"
+
+) ELSE (
+	ECHO Unrecognized Visual Studio version %vsversion%
+	EXIT /B 2
+)
+
+REM On error, pause to allow user to notice it if script was launched through explorer
+IF %ERRORLEVEL% NEQ 0 (
+	PAUSE
+)
+
+EXIT /B %ERRORLEVEL%
+
+REM ===========================================================================
+
+REM Utils
+
+REM ===========================================================================
+
+REM %1: PremakeVsVersion -> ex: vs2015
+REM %2: VsVersion envvar -> ex: 140
+:LegacyVisualBootstrap
+
+SET "VsVersion_NoPoint=%~2"
+SET "VsEnvVar=VS%VsVersion_NoPoint%COMNTOOLS"
+SET "VsPath=!%VsEnvVar%!"
+
+IF NOT EXIST "%VsPath%vsdevcmd.bat" (
+	ECHO Could not find vsdevcmd.bat to setup Visual Studio environment
+	EXIT /B 2
+)
+
+CALL "%VsPath%vsdevcmd.bat" && nmake MSDEV="%~1" -f Bootstrap.mak windows
+EXIT /B %ERRORLEVEL%
+
+REM :LegacyVisualBootstrap
+
+
+REM ===========================================================================
+
+REM %1: PremakeVsVersion -> ex: vs2010
+REM %2: VisualStudio-style VSversionMin -> ex: 15.0
+REM %3: VisualStudio-style VSversionMax -> ex: 16.0
+:VsWhereVisualBootstrap
+SET "PremakeVsVersion=%~1"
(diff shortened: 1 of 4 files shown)


Bootstrap.bat
```diff
@@ -0,0 +1,101 @@
...
+) ELSE IF "%vsversion%" == "vs2013" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "120"
+
+) ELSE IF "%vsversion%" == "vs2015" (
+	CALL :LegacyVisualBootstrap "%vsversion%" "140"
+
+) ELSE IF "%vsversion%" == "vs2017" (
+	CALL :VsWhereVisualBootstrap "%vsversion%" "15.0" "16.0"
+
+) ELSE (
```
samsinsane: This should work with VS2019 too.

Bootstrap.bat
```diff
@@ -0,0 +1,101 @@
+@ECHO OFF
+SETLOCAL
+SETLOCAL ENABLEDELAYEDEXPANSION
+
+SET vsversion=%1
+IF "%vsversion%" == "" (
+	SET vsversion=vs2015
```
samsinsane: Perhaps this should default to vs2019? Or nothing at all?

scripts/package.lua
```diff
@@ -40,7 +40,17 @@
 -- Make sure I've got what I've need to be happy.
 --
 
-	local required = { "git", "make", "gcc", "premake5", "zip" }
+	local required = { "git" }
+
+	if not os.ishost("windows") then
+		table.insert(required, "make")
+		table.insert(required, "cc")
```
samsinsane: The previous list had `gcc` but this is `cc`, is this correct?

scripts/package.lua
```diff
@@ -88,79 +98,85 @@
...
-		error("unable to checkout branch " .. branch, 0)
-	end
+--
+-- Bootstrap Premake in the newly cloned repository
+--
 
-	z = os.executef("git submodule update --init")
+	print("Bootstraping Premake...")
+	if os.ishost("windows") then
+		z = execQuiet("Bootstrap.bat")
```
samsinsane: Perhaps a prompt for the VS version could be done here?

scripts/package.lua
```diff
@@ -88,79 +98,85 @@
...
+	else
+		local os_map = {
+			linux = "linux",
+			macosx = "osx",
+		}
+		z = execQuiet("make -j -f Bootstrap.mak %s", os_map[os.host()])
```
samsinsane: Should probably have a local for `os_map[os.host()]`, running this on BSD should fail more gracefully.

Additionally, perhaps the Bootstrap labels (Is that the correct terminology?) should match the Premake internal names? Having `os.host()` passed straight into _our_ Bootstrap.mak makes sense to me, thoughts? (Another PR for sure!)

scripts/package.lua
```diff
@@ -88,79 +98,85 @@
...
-	execQuiet("premake5 /to=build/gmake.bsd /os=bsd gmake")
+
+	genProjects("--to=build/vs2005 vs2005")
+	genProjects("--to=build/vs2008 vs2008")
+	genProjects("--to=build/vs2010 vs2010")
+	genProjects("--to=build/vs2012 vs2012")
+	genProjects("--to=build/vs2013 vs2013")
+	genProjects("--to=build/vs2015 vs2015")
+	genProjects("--to=build/vs2017 vs2017")
+	genProjects("--to=build/vs2019 vs2019")
```
samsinsane: Given this is run with Premake, perhaps we could iterate the actions instead of listing all the VS versions?

scripts/package.lua
```diff
@@ -88,79 +98,85 @@
...
+	genProjects("--to=build/gmake.windows --os=windows gmake")
+	genProjects("--to=build/gmake.unix --os=linux gmake")
+	genProjects("--to=build/gmake.macosx --os=macosx gmake")
+	genProjects("--to=build/gmake.bsd --os=bsd gmake")
```
samsinsane: Similar to above, but perhaps a local OS list to generate the makefiles for these four OS'? I guess the `linux` -> `unix` part would require more than the others would.

Bootstrap.bat
tdesveauxPKFX: I added a default so that the bat can be executed through the file explorer.
I set vs2015 as it's a sort of default in my head. I can set vs2019 if you find it more appropriate.

scripts/package.lua
tdesveauxPKFX: On both Linux and macOS, cc exists and is a symbolic link to gcc and clang respectively.

scripts/package.lua
tdesveauxPKFX: I will add an optional argument. I'd like this to be the base for an automated process so a prompt  would go against that.

Bootstrap.bat
samsinsane: My thoughts behind this is that anything that isn't latest would just appear as arbitrary. Having this prompt for a version would probably be the best option for those running through file explorer.

scripts/package.lua
```diff
@@ -88,79 +129,101 @@
...
-	execQuiet("premake5 /to=build/vs2019 vs2019")
-	execQuiet("premake5 /to=build/gmake.windows /os=windows gmake")
-	execQuiet("premake5 /to=build/gmake.unix /os=linux gmake")
-	execQuiet("premake5 /to=build/gmake.macosx /os=macosx gmake")
-	execQuiet("premake5 /to=build/gmake.bsd /os=bsd gmake")
+
+	local ignoreActions = {
+		"clean",
+		"embed",
+		"gmake2",
```
samsinsane: Maybe we should use `gmake2` instead of `gmake`?

scripts/package.lua
```diff
@@ -88,79 +129,101 @@
...
+		"package",
+		"test",
+
+		"codelite",
+		"xcode4",
```
samsinsane: After this release, we should consider bundling these two project types too. I would imagine that `codelite` would be similar to `gmake`.

Bootstrap.mak
```diff
@@ -61,7 +61,7 @@ mingw: $(SRC)
 	./build/bootstrap/premake_bootstrap --os=windows --to=build/bootstrap gmake
 	$(MAKE) -C build/bootstrap config=$(CONFIG)_$(PLATFORM)
 
-osx: $(SRC)
+macosx: $(SRC)
```
WorldofBay: For this you need to edit the travis.yml to pass "macosx" instead of "osx"

samsinsane:
I'm going to approve this, but you'll need to fix up the macOS builds - maybe for macOS using `macos`, `macosx` and `osx` should trigger the macOS job in `Bootstrap.mak`? Not the end of the world to support multiple names, assuming it works.

tdesveauxPKFX:
Bootstrapping should be fixed now.

If you think it's acceptable, I will merge once CI approve and do the following in another PR (after publishing a release):
- Bootstrap.mak default to latest Visual Studio present
- Package.lua sources use gmake2 action instead of gmake
- Package.lua sources include Premake projects for codelite and xcode

PR #1285, opened 2019-05-08
Use correct values for MinimumVisualStudioVersion in Android projects

ifarbod:


ifarbod:
I tried something like this but this is what I get:

```lua
local function preparePropertyGroup()
		system "android"
		local cfg = test.getconfig(prj, "Debug", platform)
		vc2010.propertyGroup(cfg)
	end

	function suite.minVisualStudioVersion()
		preparePropertyGroup()
		test.capture [[
	<MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
	<ApplicationType>Android</ApplicationType>
	<ApplicationTypeRevision>2.0</ApplicationTypeRevision>]]
	end
```
```
[  FAILED  ] test_android_project.minVisualStudioVersion
...make-core/modules/android/tests/test_android_project.lua:31: (1) expected:
        <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
...but was:
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Android'">
fulltext:
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Android'">

stack traceback:
        ...make-core/modules/android/tests/test_android_project.lua:31: in function 'testFunction'

 1 FAILED TESTS```

ratzlaff:
After your call to `preparePropertyGroup()` you probably want to call `android.androidApplicationType(cfg)` - so preparePropertyGroup may need to return that?

the output after `fulltext:` is what was captured (normally written to the file) for the unit test as a result of calling those functions in that order.

samsinsane:
You're also probably going to need to set the action with `p.action.set("vs2015")`, or whichever action you're intending to test.

ifarbod:
@ratzlaff Thanks, I've fixed it.
@samsinsane I thought `suite.setup()` already did that, so I did nothing for that.

samsinsane:
@ifarbod Sorry, I wasn't overly clear there. Testing `vs2015` is great, but it would be good to test all the actions that your code supports. If someone rearranged the if statements, which would result in `14.0` being output always, the tests should catch that.

ifarbod:
All right then, hope this is good enough.

samsinsane:
Perfect, thanks!

PR #1289, opened 2019-05-18
Haiku fixes

mmuman:
This fixes bootstrapping on Haiku.
I could probably squash this a bit.

Still some tests in error though.

Also the makefile in the tarball seems to have [hardcoded libs](https://github.com/haikuports/haikuports/blob/master/dev-util/premake/patches/premake-5.0.0%7Ealpha13.patchset#L54) and stuff, not sure how to fix this.

### DIFF
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -103,6 +103,16 @@ solaris: $(SRC)
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake2
 	$(MAKE) -C build/bootstrap -j`getconf NPROCESSORS_ONLN` config=$(CONFIG)
 
+haiku: $(SRC)
+	$(SILENT) rm -rf ./bin
+	$(SILENT) rm -rf ./build
+	$(SILENT) rm -rf ./obj
+	mkdir -p build/bootstrap
+	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_USE_POSIX -DLUA_USE_DLOPEN -D_BSD_SOURCE -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lbsd
+	./build/bootstrap/premake_bootstrap embed
+	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake2
+	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN` config=$(CONFIG)
+
 windows-base: $(SRC)
 	$(SILENT) if exist .\bin rmdir /s /q .\bin
 	$(SILENT) if exist .\build rmdir /s /q .\build
--- a/premake5.lua
+++ b/premake5.lua
@@ -197,6 +197,10 @@
 			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
 			links       { "m" }
 
+		filter "system:haiku"
+			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN", "_BSD_SOURCE" }
+			links       { "network", "bsd" }
+
 
 	-- optional 3rd party libraries
 	group "contrib"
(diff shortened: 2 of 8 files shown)


mmuman:
Any comment?

ratzlaff:
so, I don't know anything about haiku, just what I have read on wikipedia recently

Why does premake need to link against networking?

mmuman:
> Why does premake need to link against 'network'?

Probably because of cURL I suppose.

Bootstrap.mak
```diff
@@ -103,6 +103,16 @@ solaris: $(SRC)
 	./build/bootstrap/premake_bootstrap --to=build/bootstrap gmake2
 	$(MAKE) -C build/bootstrap -j`getconf NPROCESSORS_ONLN` config=$(CONFIG)
 
+haiku: $(SRC)
+	$(SILENT) rm -rf ./bin
+	$(SILENT) rm -rf ./build
+	$(SILENT) rm -rf ./obj
+	mkdir -p build/bootstrap
+	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_USE_POSIX -DLUA_USE_DLOPEN -D_BSD_SOURCE -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $? -lnetwork -lbsd
```
ratzlaff: what symbols does -lnetwork provide that premake_bootstrap requires?

premake5.lua
```diff
@@ -197,6 +197,10 @@
 			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN" }
 			links       { "m" }
 
+		filter "system:haiku"
+			defines     { "LUA_USE_POSIX", "LUA_USE_DLOPEN", "_BSD_SOURCE" }
+			links       { "network", "bsd" }
```
ratzlaff: links { "network" } probably needs to go in the contrib/curl/premake5.lua under a haiku filter

Bootstrap.mak
mmuman: Hmm, none it seems. I just used the same libs as for the rest, but it seems you don't use cURL for bootstrap so it works without. I'll make a fixup.

mmuman:
bsd is for getpass() btw.

premake5.lua
mmuman: Hmm Solaris links with libsocket… besides, it seems you're building cURL as a static lib anyway.

ratzlaff: Yes, but if you build premake without those contribs, or if cURL ever gets made into a binmodule, this is not necessary for general premake usage.

mmuman: But for now cURL is used as a static lib, so you must link it here, since you can't link static libs to anything (it's just an archive).

ratzlaff: That is a good point, I had forgotten that detail about static libraries.

How about putting `links { "network" }` inside a `if not _OPTIONS["no-curl"] then ... end` check like on line 182?

mmuman: Wouldn't luasocket also need it?

src/host/premake.h
```diff
@@ -43,7 +43,7 @@
 #define PLATFORM_STRING   "windows"
 #endif
 
-#define PLATFORM_POSIX  (PLATFORM_LINUX || PLATFORM_BSD || PLATFORM_MACOSX || PLATFORM_SOLARIS)
+#define PLATFORM_POSIX  (PLATFORM_LINUX || PLATFORM_BSD || PLATFORM_MACOSX || PLATFORM_SOLARIS || PLATFORM_HAIKU)
```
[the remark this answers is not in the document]
mmuman: IIRC I wanted to avoid adding a case somewhere while the POSIX implementation worked fine, I'd have to try without again…

mmuman:
> It would be great if you could squish this down to a single commit, it's only +20/-6.

Ok I'll squash all this when I get some time :-)

mmuman:
Just rebased and squashed, builds here.

mmuman:
Oh right, I think we're set.

PR #1294, opened 2019-05-28
Fix CodeLite generator.

alex-rass-88:
CodeLite generator now working.


samsinsane:
I'm going to close this since it looks like a partial copy of my PR #1275. Duplicating PRs like this doesn't help us get things fixed faster.

PR #1296, opened 2019-05-30
Fix source packaging

tdesveauxPKFX:


starkos:
I don't think this will do it, you need to run `premake embed` after you check out the sources to generate `scripts.c`. Unless I'm missing something?

You also probably didn't intend to delete `.gitattributes`, I assume…

starkos:
I stand corrected, source package is working fine now. But `.gitattributes` needs to be restored.

tdesveauxPKFX:
Sorry about the lack of context.

> I don't think this will do it, you need to run premake embed after you check out the sources to generate scripts.c. Unless I'm missing something?

embed was already called before the package creation here:
https://github.com/premake/premake-core/blob/e4d8121f9cda079153aec273758f7bdfa342cf43/scripts/package.lua#L155-L164

When I changed this script before, I removed the dependency on zip and used `git archive` instead but files needs to be tracked to be included in the package and I forgot to add `scripts.c`.

> You also probably didn't intend to delete .gitattributes, I assume…

It was intended. I added gitattributes to ignore some file on `git archive`.
I changed the way to ignore files in this PR as github also use git archive when you choose to download the repo instead of cloning it.
Also, it makes more sense to centralize the information about which file is ignored/added to the package.

As .gitattributes would be empty after this, I simply deleted it.

PR #1298, opened 2019-06-02
Update D module to VS2019 and latest D compiler options.

TurkeyMan:
Been meaning to get to this for ages.


TurkeyMan:
Yeah I dunno, I opened an issue about that a while ago to start discussion. The toolset issue is kinda problematic; premake is kinda incompatible with mixed-language projects for that reason.
Maybe you could filter it by language or something, but that feels really problematic too.
Could use parallel API's for different language toolsets, but that feels crappy as well. Not sure.

Issue #1299, opened 2019-06-05
Error: [string "src/base/fileconfig.lua"]:54: attempt to index a nil value (field 'configs')


alicekdev:
Using:
OS: Linux Mint
IDE: CodeLite

Lua file: Premake.5.lua (test file)
```
workspace "TestWks"
   architecture "x64"

   configurations { "Debug", "Release", "Dist" }

   configuration "Debug"
      defines { "DEBUG", "_DEBUG" }
      symbols "on"
      
   configuration "Release"
      defines { "NDEBUG" }
      optimize "on"
   
   startproject "wks.name"

   outputdir = "%{cfg.buildcfg}-%{cfg.system}-%{cfg.architecture}"

project "TestProj"
   location "TestProj"
   kind "SharedLib"
   language "C++"
   
   targetdir ("bin/" .. outputdir .. "/%{prj.name}")
   objdir ("bin-int/" .. outputdir .. "/%{prj.name}")
   
   pchheader "pch.h"
   pchsource "src/pch.cpp"

   files { "%{prj.name}/src/**.h", "%{prj.name}/src/**.cpp" }
```

Application file: src/base/fileconfig.lua
Line: 54
Link: [file](https://github.com/premake/premake-core/blob/master/src/base/fileconfig.lua#L54)

Completely not got any idea why this is happening, i would really like to to get this working on Linux using CodeLite IDE.

Hope someone/anyone can help?

samsinsane:
I'm not sure you can use tokens in `files`, there's two solutions here:
1. Change `%{prj.name}` to `TestProj`
2. Move the project into another `premake5.lua` in the `TestProj` folder and use `include "TestProj"`:
```lua
-- premake5.lua
workspace "TestWks"
   architecture "x64" -- You could use `platforms { "x64" }` instead
   startproject "wks.name" -- indentation doesn't undo `configuration`/`filter`, so you'll get a warning if you call this after a `configuration`/`filter`
   configurations { "Debug", "Release", "Dist" }

   configuration "Debug" -- You should probably move to `filter { "configurations:Debug" }` instead
      defines { "DEBUG", "_DEBUG" }
      symbols "on"
      
   configuration "Release" -- You should probably move to `filter { "configurations:Release" }` instead
      defines { "NDEBUG" }
      optimize "on"
   
   outputdir = "%{cfg.buildcfg}-%{cfg.system}-%{cfg.architecture}"

  include "TestProj"

-- TestProj.lua
project "TestProj"
[...]

PR #1301, opened 2019-06-10
Fix related to #1299

tdesveauxPKFX:
I'm not 100% sure about the tree fix. Maybe we still want `onleaf` called in this case and we should patch the `onleaf` functions instead.

### DIFF
--- a/src/base/fileconfig.lua
+++ b/src/base/fileconfig.lua
@@ -152,7 +152,11 @@
 --
 
 	function fileconfig.getconfig(fcfg, cfg)
-		return fcfg.configs[cfg]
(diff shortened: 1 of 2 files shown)


samsinsane:
To my understanding, this problem was only with the codelite generator so wouldn't that indicate an issue with the generator rather than `fileconfig` or `tree`?

starkos:
Since this one has stalled, I'm going to go ahead and close this one. I also think the tree fix should not be included: calling `onnode()` for an empty node seems valid to me. And there are no tests to cover the original bug. I'll see if I can submit a follow-up PR to address it, but closing this for now since it has gone stale.

PR #1317, opened 2019-07-23
deterministic "latest" SDK for vs2017

jeaiii:
using idea from here:
https://developercommunity.visualstudio.com/content/problem/140294/windowstargetplatformversion-makes-it-impossible-t.html

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -2760,20 +2760,6 @@
 		return m.conditionFromConfigText(vstudio.projectConfig(cfg))
 	end
 
---
--- Get the latest installed SDK 10 version from the registry.
---
-
-	function m.latestSDK10Version()
-		local arch = iif(os.is64bit(), "\\WOW6432Node\\", "\\")
-		local version = os.getWindowsRegistry("HKLM:SOFTWARE" .. arch .."Microsoft\\Microsoft SDKs\\Windows\\v10.0\\ProductVersion")
-		if version ~= nil then
-			return version .. ".0"
-		else
-			return nil
-		end
-	end
-
 
 --
 -- Output an individual project XML element, with an optional configuration
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -2519,7 +2519,7 @@
 				if _ACTION >= "vs2019" then
 					min = "10.0"
 				else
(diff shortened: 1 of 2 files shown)


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2519,7 +2519,7 @@
 				if _ACTION >= "vs2019" then
 					min = "10.0"
 				else
-					min = iif(_ACTION == "vs2017", m.latestSDK10Version(), nil)
+					min = nil
```
samsinsane: I would prefer if you just replaced `m.latestSDK10Version()` with `"$(LatestTargetPlatformVersion)"`.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2528,12 +2528,27 @@
 
 	end
 
+	function m.emitPlatformVersion(optSdkVersion)
```
samsinsane: This entire function can be reverted with the above change as it would just be a `nil` check.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2528,12 +2528,27 @@
...
+	end
 
 	function m.targetPlatformVersionGlobal(prj)
 		local min = m.targetPlatformVersion(prj)
-		if min ~= nil then
-			m.element("WindowsTargetPlatformVersion", nil, min)
+		
+		if _ACTION == "vs2017" then
+			-- This emits a property that when used causes vs2017 to detect the latest windows sdk version.  Always emit this in case we have a config dependent reference to it.
+			m.element("LatestTargetPlatformVersion", nil, "$([Microsoft.Build.Utilities.ToolLocationHelper]::GetLatestSDKTargetPlatformVersion('Windows', '10.0'))")
```
samsinsane: What versions of VS2017 have you tested?

modules/vstudio/vs2010_vcxproj.lua
jeaiii: Yes, I think it can be a much smaller change. I took made the change as is from our local version. I didn't want to make changes to it at the same time.

modules/vstudio/vs2010_vcxproj.lua
jeaiii: Pretty much all of them. It has been in use here for about a year.

starkos:
I've put together a cleaned up alternative PR for this one: #1424. I'm going to go ahead and close this one in favor of that new one; will reopen if #1424 can't be approved.

PR #1323, opened 2019-08-20
Update copyright year in support files

starkos:
Catch up our copyrights before the year is over and we have to do it again.


README.md
```diff
@@ -63,7 +63,7 @@ Support this project with your organization. Your logo will show up here with a
 
 ### Copyright & License
 
-Copyright &copy; 2002-2016 by Jason Perkins
+Copyright &copy; 2002-2019 by Jason Perkins
```
samsinsane: This is a bit of nitpick but the `LICENSE.txt` says `2003-2019` and this says `2002-2019`, that's weird.

starkos: I started coding in 2002, first release in 2003, but you're right that's weird. I'll make them both 2003.

Wow that was a long time ago.

samsinsane: > Wow that was a long time ago.

Over 16 years ago - I'm having flashbacks to using Windows ME and rebooting all the time so that I could open applications. That's enough remembering for me!

PR #1325, opened 2019-08-21
Improve the README

starkos:
You can [see the proposed new version here](https://github.com/starkos/premake-core/blob/update-readme/README.md).


starkos:
Thanks for catching that, I'll get it fixed.

PR #1332, opened 2019-09-12
Account for filename collisions on systems with case-insensitive file…

### LINKED ISSUE #1241: Windows - gmake2 case sensitivity
I am using latest Premake 5.0.0-dev on Windows with gmake2 (i am using the arm-none-eabi-* toolchain) builder.

There is a problem with two files of same name but unmatched cases, f.e.: _a/Rtc.cpp_ and _b/rtc.c_.
It is resolved to _Rtc.o_ and _rtc.o_ which is correct on case-sensitive platforms like Linux, but on Windows version of make it results to overriding one file with another and building only the latest with the following warning message:
> WAS_Tag.make:1094: warning: overriding commands for target `obj/Debug/Rtc.o'
> WAS_Tag.make:863: warning: ignoring old commands for target `obj/Debug/Rtc.o'

I have come [with this quick'n'dirty fix](https://github.com/kaldap/premake-core/commit/3ecf10965cf22286058249b18b73f33c9b4e31d2), but it probably is not the best solution.

ratzlaff:
**What does this PR do?**
Continues the work done in PR #1191 by accounting for case-insensitive filesystems

when computing sequence numbers for a compiled object filename, use the lowercase version of the base file name

closes #1241

### DIFF
--- a/src/base/oven.lua
+++ b/src/base/oven.lua
@@ -695,23 +695,40 @@
 -- conflicting object file names - hello1.o
 
 	function oven.uniqueSequence(f, cfg, seq, bases)
-		while true do
+		local good_sequence = true
+		repeat
+			-- be optimistic
+			good_sequence = true
 			f.sequence = seq[cfg] or 0
 			seq[cfg] = f.sequence + 1
 
-			if seq[cfg] == 1 then
+			if f.sequence == 0 then
+				-- first time seeing this objname
 				break
 			end
 
-			if not bases[f.objname] then
-				bases[f.objname] = {}
+			-- getting here has changed our sequence number, but this new "basename"
+			-- may still collide with files that actually end with this "sequence number"
+			-- so we have to check the bases table now
+
+			-- objname changes with the sequence number on every loop
+			local lowerobj = f.objname:lower()
+			if not bases[lowerobj] then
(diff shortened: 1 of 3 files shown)


src/base/oven.lua
```diff
@@ -695,23 +695,40 @@
 -- conflicting object file names - hello1.o
 
 	function oven.uniqueSequence(f, cfg, seq, bases)
-		while true do
+		local good_sequence = true
+		repeat
```
samsinsane: Not the end of the world or anything, but what was the reason for changing this?

codecat:
Apologies if this gets asked a lot, but is a new pre-release expected that contains this fix? I just ran into this exact problem myself on the latest pre-release version. No worries if a pre-release is not in sight, I can build it myself as well! 👍 

Just wanted to give a ping that someone randomly hit into this issue.. 😅

starkos:
Top of my Premake to-do list right now. I had planning to punch through some PRs and low-hanging issues and get the release out when I got back the States a few weeks ago, but as everything has gone sort of…sideways since then I haven't been able to make the progress I'd hoped. If I continues to drag on I'll abandon the fixes and just make the release.

PR #1345, opened 2019-10-10
New 'listWindowsRegistry' OS API method for enumerating content of single registry subkey

kaldap:
**What does this PR do?**

This PR is extending the 'os' API by adding new method called 'listWindowsRegistry' which returns table of all subkeys and values in the specified path. Nil is returned in the case of failure.

No breaking changes, just extended API.

* Useful in some cases, f.e. automatic selection of latest GCC toolchain on Windows.
* Only REG_SZ and REG_DWORD values are covered in test suite - did not find any reliable values of other types. But I did manual testing of all value types, so it should work properly.

Yes I did with single exception. Method name is in camel-case. I did it on purpose to keep it consistent with existing 'getWindowsRegistry' method.

### DIFF
--- a/src/host/os_getWindowsRegistry.c
+++ b/src/host/os_getWindowsRegistry.c
@@ -15,7 +15,7 @@ typedef struct RegKeyInfo
 	char * value;
 } RegKeyInfo;
 
-static HKEY get_key(const char **path)
+HKEY getRegistryKey(const char **path)
 {
 	if (_strnicmp(*path, "HKCU:", 5) == 0) {
 		*path += 5;
@@ -42,7 +42,7 @@ static HKEY get_key(const char **path)
 	return NULL;
 }
 
-static HKEY get_subkey(HKEY key, const char **path)
+static HKEY getSubkey(HKEY key, const char **path)
 {
 	HKEY subkey;
 	size_t length;
@@ -84,7 +84,7 @@ static HKEY get_subkey(HKEY key, const char **path)
 	return subkey;
 }
 
-static char * get_value(HKEY key, const char * name)
+static char * getValue(HKEY key, const char * name)
 {
 	DWORD length;
 	char * value;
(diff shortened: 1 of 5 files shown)


kaldap:
Found a registry value for a DWORD test which should be around there for some time already, so hopefully should make the test passing on all supported Windows versions.

What do you mean by "I’d personally prefer getWindowsRegistry to listWindowsRegistry"? Merge those two methods into single one with different behaviour for keys/values?

src/host/os_getWindowsRegistry.c
```diff
@@ -15,7 +15,7 @@ typedef struct RegKeyInfo
 	char * value;
 } RegKeyInfo;
 
-static HKEY get_key(const char **path)
+HKEY get_registry_key(const char **path)
```
starkos: Nitpick, but for consistency can you make this `getRegistryKey` (since you're changing anyway)?

src/host/os_listWindowsRegistry.c
```diff
@@ -0,0 +1,252 @@
...
+	const char * name;
+	const char * value;
+	DWORD valueSize;
+	DWORD type;
+} RegNodeInfo;
+
+typedef void (*ListCallback)(const RegNodeInfo * info, void * user);
+extern HKEY get_registry_key(const char** path);
+
+static const char* get_type_string(DWORD type)
```
starkos: Can you make this `getTypeString`?

src/host/os_listWindowsRegistry.c
```diff
@@ -0,0 +1,252 @@
...
+		case REG_MULTI_SZ:                   return "REG_MULTI_SZ";
+		case REG_RESOURCE_LIST:              return "REG_RESOURCE_LIST";
+		case REG_FULL_RESOURCE_DESCRIPTOR:   return "REG_FULL_RESOURCE_DESCRIPTOR";
+		case REG_RESOURCE_REQUIREMENTS_LIST: return "REG_RESOURCE_REQUIREMENTS_LIST";
+		case REG_QWORD:                      return "REG_QWORD";	
+		default:                             return NULL;
+	}
+}
+
+static HKEY open_key(const char *path)
```
starkos: …and `openKey`? You get the idea…

src/host/os_listWindowsRegistry.c
```diff
@@ -0,0 +1,252 @@
...
+			break;
+		}
+	}
+	lua_setfield(L, -2, "value");
+
+	// Complete the value subtable
+	lua_setfield(L, -2, info->name);
+}
+
+int os_listWindowsRegistry(lua_State* L)
```
starkos: Could this be `os.getWindowsRegistry` instead of `os.listWindowsRegistry`?

All nitpicks I know…

kaldap: I am afraid it could not, since it is allowed to have subkey and value of the same name inside a single key. In this particular case, it is not possible to choose the correct operation just from the path. 

What about calling it 'os.getWindowsRegistryKey'? Or having a second optional boolean argument to select between key and value...

starkos:
Looking good. If you could rebase against the latest master branch (and if you could squash the commits too, that would be a bonus), this looks ready to merge.

starkos:
Merged. Thanks for this!

Issue #1114, opened 2018-06-07
premake5 test run specific tests


TurkeyMan:
`premake5 test [testnames...]` should allow to run specific test(/s).

tdesveauxPKFX:
There is the `test-only` option. But I think it can only take one test at a time.

samsinsane:
`test-only` will take a suite or a specific test, and it doesn't do wildcard matching. Extending `test-only` to take a comma delimited list and probably wildcard matching would be pretty cool for module testing, amongst other things.

tvandijck:
closing, `--test-only=` works fine, it could be extended. Feel free to do so.

starkos:
Both? The `--test-only` flag isn't really needed, and it would be handy to be able to specify a select set of tests.

PR #1352, opened 2019-10-13
Improved test runner usage

samsinsane:
**What does this PR do?**

This PR extends the test runner to allow for some of the discussed items in #1114; wildcard support and specifying tests without using `--test-only`.

This shouldn't break the old functionality, but does extend it as mentioned above.

I wrote this several months before the commit date suggests, so I'm not sure where I got up to with it but it does seem to work. Some of the changes might be for the comma separated list, but I couldn't see anything to suggest I added support for that and it didn't work when tested.

- [x] Follow our [coding conventions]([https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions))

### DIFF
--- a/modules/self-test/self-test.lua
+++ b/modules/self-test/self-test.lua
@@ -41,12 +41,27 @@
 		m.loadTestsFromManifests()
 		m.detectDuplicateTests = false
 
-		local test, err = m.getTestWithIdentifier(_OPTIONS["test-only"])
-		if err then
-			error(err, 0)
+		local tests = {}
+		local isAction = true
+		for i, arg in ipairs(_ARGS) do
+			local _tests, err = m.getTestsWithIdentifier(arg)
+			if err then
+				error(err, 0)
+			end
+			
+			tests = table.join(tests, _tests)
+		end
+		
+		if #tests == 0 or _OPTIONS["test-only"] ~= nil then
+			local _tests, err = m.getTestsWithIdentifier(_OPTIONS["test-only"])
+			if err then
+				error(err, 0)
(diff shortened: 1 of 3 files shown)


modules/self-test/self-test.lua
```diff
@@ -41,12 +41,35 @@
...
 		m.detectDuplicateTests = false
 
-		local test, err = m.getTestWithIdentifier(_OPTIONS["test-only"])
-		if err then
-			error(err, 0)
+		local tests = {}
+		local isAction = true
+		for i, arg in ipairs(_ARGV) do
+			-- Skip any _OPTIONS
+			if not arg:startswith("/") and not arg:startswith("--") then
```
starkos: Could this use [_ARGS](https://github.com/premake/premake-core/wiki/_ARGS) instead?

starkos:
Quick question inline…

modules/self-test/self-test.lua
samsinsane: I don't see why not, I wasn't aware of `_ARGS` so that's cool. The docs aren't clear about what it contains, does it exclude `_OPTIONS`?

samsinsane: It does exclude the options, I'll update the PR to use `_ARGS` give me a minute 😄

samsinsane: The PR has been updated to use `_ARGS`, much better! Thanks for the tip @starkos!

PR #1353, opened 2019-10-16
Fix premake4.lua bootstrap build script

fountainment:
This PR add the missing include dir and fix '"luashim.h" not found' error when using premake4 bootstrap.


starkos:
Thanks for the contribution!

Issue #1310, opened 2019-07-08
Vertical-lines are replaced by slashes in the xcode 4 buildcommands.


tempura-sukiyaki:
Vertical-lines(`|`) are replaced with slashes(`/`) in xcode4 `buildcommands` (`prebuildcommands`, `prelinkcommands`, `postbuildcommands`).
Because of this, the following `postbuildcommands` will be converted unintentionally.

```lua: premake5.lua
postbuildcommands {
  'echo "abc" | grep "a"',
}
```

```bash: project.pbxproj
set -e
echo "abc" / grep "a"
```

Is there any way to avoid it?

samsinsane:
That's quite strange, does this only impact xcode4? I had a quick glance and couldn't see anything that would cause this. My initial reaction was that it's part of the deferred join system but it would impact all the actions not just xcode4.

WorldofBay:
it does impact vs2017 too, didn't test with more.

i searched for | in the code but there is no replacement whatsoever - only additional | like in the configuration formatting.

the following happens:
multiple | are handled the same as single |
| at the start or end of the string are cut without replacement
| anywhere else is replaced by a single /

i have no clue where it happens, could it be that path.normalize() causes this?

samsinsane:
@WorldofBay it's part of the deferred join system - [here](https://github.com/premake/premake-core/blob/94f6d63d4ce8341671cf03ab9a58bf2e4001faeb/src/host/path_join.c#L11) is the pipe, and [this](https://github.com/premake/premake-core/blob/94f6d63d4ce8341671cf03ab9a58bf2e4001faeb/src/host/path_join.c#L146-L151) will return `true` when the path contains a `|` and then [this](https://github.com/premake/premake-core/blob/94f6d63d4ce8341671cf03ab9a58bf2e4001faeb/src/host/path_join.c#L154-L191) will resolve the deferred join which will result in the path being `/`.

However, I think the real issue is that we're still treating these commands as paths when they might not contain a single path, and we're just destroying the actual commands.

WorldofBay:
why is `|` considered a path?

samsinsane:
> why is `|` considered a path?

It's not, as I said above, it's part of the deferred join system. This isn't really the problem though, it's just part of what causes the `|` to become `/`. As I stated above, the real problem is that we treat the commands as paths.

starkos:
Simple workaround would be to replace the pipe symbol in the path processor with something more obscure.

PR #1384, opened 2020-01-05
Improve issue #1310

tempura-sukiyaki:
**What does this PR do?**

-  If there is a whitespace immediately before and/or after `|`, it is not considered as a "deferred join delimiter".

-  In most cases, directory names do not have leading and/or trailing whitespace.
  In most cases, whitespace is added before and after "pipe" and "or".
  In such cases, applying this PR eliminates issue #1310.

- if directory name has a leading and/or trailing whietspace, "deferred join delimiter" will not work properly.
  However, all existing tests passed without change.

- This PR does not completely resolve issue #1310.
  So I don't write `close` comment.

- [x] Follow our [coding conventions]([https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions))

### DIFF
--- a/src/host/path_join.c
+++ b/src/host/path_join.c
@@ -8,7 +8,7 @@
 #include <string.h>
 #include "path_isabsolute.h"
 
-#define DEFERRED_JOIN_DELIMITER '|'
+#define DEFERRED_JOIN_DELIMITER '\a'
 
 char* path_join_single(char* buffer, char* ptr, const char* part, int allowDeferredJoin)
 {
--- a/src/host/path_join.c
+++ b/src/host/path_join.c
@@ -137,20 +137,9 @@ int path_deferred_join(lua_State* L)
 }
 
 
-static char *find_next_deferred_join_delimiter(const char *buffer, int offset)
-{
-	char *ptr = strchr(buffer + offset, DEFERRED_JOIN_DELIMITER);
-	while (ptr != NULL && (*(ptr + 1) == ' ' || (ptr != buffer && *(ptr - 1) == ' ')))
-	{
-		ptr = strchr(ptr + 1, DEFERRED_JOIN_DELIMITER);
-	}
-	return ptr;
-}
-
-
 int do_path_has_deferred_join(const char* path)
 {
-	return (find_next_deferred_join_delimiter(path, 0) != NULL);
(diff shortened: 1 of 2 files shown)


starkos:
Would it help to change the delimiter character? We could use a non-printable Unicode character for instance, which would never appear in a typical path? That way we wouldn't have to try to figure out if the delimiter was "real" or if it were part of the path?

tempura-sukiyaki:
If the "deferred join delimiter" can be replaced by another Unicode symbol, it seems safer.
However, I don't know the extent of the impact of doing so.
I think projects that already use `|` as the "deferred join delimiter" will not work.

samsinsane:
Lua doesn't support Unicode, this could be problematic when the paths are being manipulated in Lua?

Regardless, I feel that the problem isn't really that we use `|` for the delimiter, but that we treat `postbuildcommands` and friends as paths when they're not - only part of some of the elements will be a path sometimes, but we're assuming all of the elements are always entirely paths. I think the only solution here is to have users specify that they want the path manipulation through the `%{}` system.

starkos:
> I feel that the problem isn't really that we use | for the delimiter, but that we treat postbuildcommands and friends as paths when they're not

Agree 100%.

But to help @tempura-sukiyaki out, I do think it should be safe to switch the deferred join delimiter to a non-printable character like `"\x07"` (bell) or `"\x08"` (backspace). That will make it leave pipe symbols alone, and save the trouble of trying work around them.

Could you give that a try? If it doesn't work out, let me know here and I'll approve this PR for you.

Longer term, I would definitely like to address the root causes that made deferred joins seem necessary and see if we can find better solutions.

tempura-sukiyaki:
Thank you for your comment.

+ Replace `DEFFERD_JOIN_DELIMITER` in "src/host/path_join.c" from `|` to `\a` (bell).
+ Modify the test in "test/base/test_path.lua" accordingly.

Could you confirm my understanding is correct?

starkos:
Yes, that's the approach I would try.

tempura-sukiyaki:
I tried doing it. 🙂

starkos:
Did that help the problems you were seeing? Hope so! If you could you squash your changes and rebase them against the current master branch I'd be happy to approve and merge.

Issue #1400, opened 2020-02-11
Custom Build Steps


redorav:
**What problem will this solve?**
I'm trying to add Custom Build Steps to my build, I can now see how they would be done in the vcxproj (via the Visual Studio interface) but now I would like to add them to premake.

**What might be a solution?**
I'm trying to find out whether premake supports this but it seems like it doesn't as I've not been able to find any mention of it in the code, so perhaps I could give it a stab.

**What other alternatives have you already considered?**
I've considered build rules and pre/post events, but they all follow the "input -> output" model, whereas I'm trying to run something only once if certain dependencies have been modified with respect to a single output file. The custom built step works at a project, not file, level. Here's the documentation for the feature. https://devblogs.microsoft.com/visualstudio/custom-build-steps-tools-and-events/ and https://docs.microsoft.com/en-us/cpp/build/how-to-add-a-custom-build-step-to-msbuild-projects?view=vs-2019

Where do you think I could start and what should the interface be? I can see how the properties would fit in MSBuild, would I find issues trying to do the same thing in XCode or gmake2?

In Visual Studio there are several properties relating to this

- command line
- description
- outputs
- dependencies (other than output)

These are the ones I'm mostly interested in. Does something like this make sense?

project("MyProject")
    buildstepcommand("mytool.exe")
    buildstepdescription("executes tool")
    buildstepoutput("file.ext")
    buildstepdependencies { "a.file", "b.file" }

samsinsane:
There's [this](https://github.com/premake/premake-core/wiki/Custom-Build-Commands#custom-build-commands) using [`buildcommands`](https://github.com/premake/premake-core/wiki/buildcommands), [`buildmessage`](https://github.com/premake/premake-core/wiki/buildmessage), [`buildoutputs`](https://github.com/premake/premake-core/wiki/buildoutputs), and [`buildinputs`](https://github.com/premake/premake-core/wiki/buildinputs) - is this what you're looking for?

redorav:
The problem I have with those is that build commands are run once per build input, whereas I want to run the step only once, but based on multiple dependencies which could cause that one process to run. Does what I'm looking for make sense?

redorav:
Just a few more comments on this, it looks like in Visual Studio there are several ways to do custom build actions:

1) **Custom build tool:** This is used on a per-file basis. Essentially maps to the "Custom Build Tool" in Visual Studio.
2) **Rules:** Another mechanism for doing the above. Premake also calls them rules
[...]

Issue #1013, opened 2018-03-05
OSX: Attempt to index a nil value (global 'cfg')


redorav:
I have the following code that works on Windows:

```lua
workspace("Project")
	configurations { "Debug", "Release" }
	location ("Build/".._ACTION)
	architecture("x64")
	
	configuration "Debug"
		defines { "DEBUG", "_DEBUG" }
		symbols "on"

	configuration "Release"
		defines { "NDEBUG" }
		optimize "on"
	
project ("Project")
	kind("StaticLib")
	language("C++")
	
	files { "Source/lib/**.cpp", "Source/lib/**.hpp", "Source/lib/**.h" }
	includedirs { "Source/lib/include" }
	
	filter{"configurations:"} -- Removing this on Mac causes it to fail but works in Windows
	
	postbuildcommands 
	{
		"{copy} %{cfg.buildcfg.abspath} ../../Binaries/".."Project"..".".._ACTION..".%{cfg.buildcfg:lower()}".."%{cfg.buildtarget.extension}*"
	}
```

Somehow on the Mac cfg becomes nil if I don't do an explicit filter("configurations:") with basically no filter applied to it. On Windows it works, but on Mac it fails.

samsinsane:
What action were you using on macOS? `xcode4`?

redorav:
Yes, xcode4. Should I be using something else?

samsinsane:
No no, just wanted to verify that it was indeed `xcode4` and not `gmake` or `codelite`, or even the `vs*` actions.

samsinsane:
[This](https://github.com/premake/premake-core/blob/0eda22c602567e8b2efb9f1ba104fa766d5adbd4/modules/xcode/xcode_common.lua#L758-L793) will be the issue, it specifically handles the project level configurations, whereas VS [doesn't](https://github.com/premake/premake-core/blob/0eda22c602567e8b2efb9f1ba104fa766d5adbd4/modules/vstudio/vs2010_vcxproj.lua#L303-L307). I guess the fix would be to drop the optimization for project level items. @starkos thoughts?

starkos:
It seems like the Xcode exporter is trying to do the right thing here. I'm not clear on why `cfg` is becoming nil.

samsinsane:
[...]

PR #1403, opened 2020-02-17
Add build steps to Visual Studio

redorav:
**What does this PR do?**

Adds build steps to Visual Studio through the buildcommands interface. (**edit:** See issue #1400 for more information and discussion).

This adds missing functionality, to be able to have build steps with custom inputs and outputs

### DIFF
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -55,6 +55,7 @@ return {
 	"vc2010/test_assembly_refs.lua",
 	"vc2010/test_build_events.lua",
 	"vc2010/test_build_log.lua",
+	"vc2010/test_build_steps.lua",
 	"vc2010/test_character_set.lua",
 	"vc2010/test_compile_settings.lua",
 	"vc2010/test_config_props.lua",
--- a/modules/vstudio/tests/vc2010/test_build_steps.lua
+++ b/modules/vstudio/tests/vc2010/test_build_steps.lua
@@ -1,0 +1,92 @@
+--
+-- tests/actions/vstudio/vc2010/test_compile_settings.lua
+-- Validate compiler settings in Visual Studio 2010 C/C++ projects.
+-- Copyright (c) 2011-2020 Jason Perkins and the Premake project
+--
+
+	local p = premake
+	local suite = test.declare("vstudio_vs2010_build_steps")
+	local vc2010 = p.vstudio.vc2010
+	local project = p.project
+
+
+--
+-- Setup
+--
+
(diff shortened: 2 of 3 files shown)


starkos:
I haven't tried, but I suspect that this encounter the same bug I just fixed in #1401. This should be allowed:

```lua
project "MyProject"
   buildcommands { "cmd.exe %{cfg.buildcfg.abspath}" }
```

If you're trying to process that command at the project level, it will fail because `cfg` will not be set. See #1013 for more discussion.

redorav:
I just tested what you said and you're absolutely right, it works within the files filter but not at project scope. Why is it that project level commands cannot access configs? Is this something I can fix?

starkos:
I'd be curious what `%{cfg.targetdir}` emits in that situation.

The relevant condundrum is: Premake allows you to specify settings in a "higher" scope (global, workspace, project) that get inherited by lower or narrower scopes (configuration, file). So when I use script like…

```lua
project "MyProject"
   buildcommands { "cmd.exe %{cfg.targetdir}" }
```

…my _intention_ is that a separate command would be created for _each configuration_ (think "Debug", "Release") in my project. I would expect that my project's debug configuration to use `cmd.exe bin/Debug`, for example, and release to use `cmd.exe bin/Release`.

I'm not sure what the right answer would be here. A new set of APIs might the easiest, but I'm open to more general-purpose suggestions.

redorav:
As far as I can tell, this does the right thing in the case you describe. I have come up with the following:

configuration("Debug")
	targetdir ("MyTargetDir1")
		
configuration("Release")
	targetdir ("MyTargetDir2")

configuration {} -- Reset configurations so it applies to all

buildcommands
{
	"%{cfg.targetdir}"
}

And this expands the build command to ../../MyTargetDir1 only for Debug, and only MyTargetDir2 respectively on the Release configuration. It also respect configuration filters, so if I was to not have th configuration {} line, it only applies to Release.

starkos:
Thanks for doing that research, I think this looks workable. If you could add some unit tests and squash the commits I think it should be good to approve.

Thanks for contribution!

redorav:
I've added the unit tests and tried rebasing/squashing, but I'm making a mess like I did last time I tried doing the process. I think I conceptually understand what I'm trying to do, I just don't get how to do it with git...

redorav:
I've kind of done it in a brute force way by rewinding the history and re-pushing the commits in one go, I need to learn how to properly do it.

redorav:
Thanks for getting it in so quick, I love how premake keeps growing and maturing

PR #1404, opened 2020-02-21
Place all generated files in separate GENERATED list, so they all can…

### LINKED ISSUE #1380: gmake2: buildoutputs for custom rules not cleaned on make clean
With 5.0.0-dev

Mikhael-Danilov:
… be cleaned by clean target.

**What does this PR do?**

All files generated by regular or custom rules for gmake2/cpp projects will be placed in separate GENERATED list, and will be deleted on `make clean` invocation.

All files generated by custom rules for gmake2/cpp will be deleted on `make clean` invocation

Probably no. Yes, previously buildoutputs of custom rules gmake2/cpp was persisted after `make clean` now they will be deleted.

I use tolua for lua-c++ bindings, with this change generated files will be properly cleaned on `make clean` step.

closes #1380

- [v] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [v] Add unit tests showing fix or feature works; all tests pass
- [v] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [v] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [v] Minimize the number of commits

### DIFF
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -250,6 +250,11 @@
 		table.insert(fileset, filename)
 		cfg._gmake.filesets[kind] = fileset
 
+		local generatedKind = "GENERATED"
+		local generatedFileset = cfg._gmake.filesets[generatedKind] or {}
+		table.insert(generatedFileset, filename)
+		cfg._gmake.filesets[generatedKind] = generatedFileset
+
 		-- recursively setup rules.
 		cpp.addRuleFile(cfg, node)
 	end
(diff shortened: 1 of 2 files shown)


starkos:
Looking at the tests, it would appear that both `OBJECTS` and `GENERATED` are always the same. We already delete all of the object files on `make clean`, so it isn't obvious why this change is needed. 

Presumably you intended this code to also handle the output of custom build rules. If so, could you add a test to show that? I'd like to see a case where `OBJECTS` and `GENERATED` end up with different contents, otherwise it would be simpler to just use the list that's already there.

Mikhael-Danilov:
Thank you for feedback. 

I added `customBuildCommand_generatedCpp` test to show case I'm trying to handle: custom rule was used to generate C++ source and header files, said files will populate  `CUSTOM`, `OBJECTS` and `SOURCES` so `GENERATED` will be required to keep all their traces.

PR #1372, opened 2019-12-03
A draft of out of source build support using premake5

KOLANICH:

PR #1409, opened 2020-03-08
Use "to" option to set targetdir

starkos:
**What does this PR do?**

Extends the existing "to" option (i.e. `premake5 --to=build gmake`) to also change `targetdir`, so the `bin` folder and Premake binaries end up in the same folder as the project files that were used to build them.

Thanks to @KOLANICH for the original PR (#1372). This one has been rebased against master and squashed, the CI builds fixed, and this description added.

Previously, `premake5 --to=build gmake` would only change the location of the generated project files. When you ran `make` against those files, the libraries and object files would get placed in the `build` folder, but the `bin` folder would still be out next to the `Premake5.lua` script in the parent folder.

With this change, the `bin` folder will also get placed in `build`, make it easier to find.

The [Building Premake](https://github.com/premake/premake-core/wiki/Building-Premake) wiki page will need to be updated to reflect this change. I have it on my list to update once this lands.

- [ ] ~~Add unit tests showing fix or feature works; all tests pass~~ (n/a)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,5 +26,5 @@ matrix:
 
 script:
  - make -f Bootstrap.mak $TRAVIS_OS_NAME CONFIG=${CONFIGURATION}
- - bin/${CONFIGURATION}/premake5 test
+ - build/bootstrap/bin/${CONFIGURATION}/premake5 test
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,4 +16,4 @@ build_script:
   - cmd: nmake -f Bootstrap.mak MSDEV=vs2015 windows PLATFORM=%PLATFORM% CONFIG=%CONFIGURATION%
 
 test_script:
-  - cmd: bin\%CONFIGURATION%\premake5 test
+  - cmd: build\bootstrap\bin\%CONFIGURATION%\premake5 test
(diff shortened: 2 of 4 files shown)


starkos:
Hmm…apparently I need to do more research to fix the CI builds.

samsinsane:
@starkos this PR only changed the Premake binary `targetdir`, you'll need to change [luasocket](https://github.com/premake/premake-core/blob/6ca121993b786bc51ca09e7e722ce7ba8aac35a5/binmodules/luasocket/premake5.lua#L47-L51) and [example](https://github.com/premake/premake-core/blob/6ca121993b786bc51ca09e7e722ce7ba8aac35a5/binmodules/example/premake5.lua#L24-L28) too.

starkos:
I've dug into this further, and it isn't as simple as #1372 suggests. There are support libraries getting placed in the `build/` folder, and the `binmodules/example` and `binmodules/luasocket` will also have to be moved. Then the commands listed in `.travis.yml` and `appveyor.yml` all need to be able to complete successfully.

I gave this an honest effort but I don't consider this a high enough priority for the amount of time I'm putting into it. If someone else wants to run with it I'd welcome the addition.

PR #1412, opened 2020-03-23
fix issues with codelite (fixes #1411 and #1299)

### LINKED ISSUE #1411: CodeLite Generation Broken, Breaks with Multiple Platforms
Hello! I'm running premake5 5.0.0-alpha14

When I generate a CodeLite project with premake that has multiple platforms, generation seems to break. Specifically, if I use something like "links" to link another project. If I rebuild the workspace, everything seems to build fine, but if I build the specific project that calls "links" it gives me an error like "No rule to make target '".build-x86_64', needed by 'bin/x86_64/Debug/test_exe'"

This seems to be an old issue as I've come across something similar here: https://github.com/premake/premake-codelite/issues/17

In testing, it seems to be caused by the OR character '|' in the generated .project files.

image

I can induce the problem if I use configurations only with no platforms, and I use the '|' character.

`configurations{ "Debug|32","Release|32", "Debug|64", "Release|64" }`

If I don't use |, there are no issues.

I don't know why. But it seems like a straightforward solution is to not use |

As an extra reference, here's an example premake file (and repo) that reproduces the issue https://github.com/Spirrwell/premake-codelite-bug/blob/master/premake5.lua

arthurdead:
This fixes codelite generation if theres no files as reported here: #1299
And if theres multiple platforms as reported here: #1411

Also removes useless root folder from projects
Before:
2MVgHfb

After:

### DIFF
--- a/modules/codelite/codelite.lua
+++ b/modules/codelite/codelite.lua
@@ -21,7 +21,7 @@
 	function codelite.cfgname(cfg)
 		local cfgname = cfg.buildcfg
 		if codelite.workspace.multiplePlatforms then
-			cfgname = string.format("%s|%s", cfg.platform, cfg.buildcfg)
+			cfgname = string.format("%s-%s", cfg.platform, cfg.buildcfg)
 		end
 		return cfgname
 	end
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -385,12 +385,16 @@
 		_p(3, '</AdditionalRules>')
 	end
 
-	function m.isCpp11(cfg)
-		return (cfg.cppdialect == 'gnu++11') or (cfg.cppdialect == 'C++11') or (cfg.cppdialect == 'gnu++0x') or (cfg.cppdialect == 'C++0x')
+	function m.isCpp17(cfg)
+		return (cfg.cppdialect == 'gnu++17') or (cfg.cppdialect == 'C++17') or (cfg.cppdialect == 'gnu++1z') or (cfg.cppdialect == 'C++1z')
 	end
 
 	function m.isCpp14(cfg)
-		return (cfg.cppdialect == 'gnu++14') or (cfg.cppdialect == 'C++14') or (cfg.cppdialect == 'gnu++1y') or (cfg.cppdialect == 'C++1y')
+		return m.isCpp17(cfg) or (cfg.cppdialect == 'gnu++14') or (cfg.cppdialect == 'C++14') or (cfg.cppdialect == 'gnu++1y') or (cfg.cppdialect == 'C++1y')
+	end
+
+	function m.isCpp11(cfg)
+		return m.isCpp14(cfg) or (cfg.cppdialect == 'gnu++11') or (cfg.cppdialect == 'C++11') or (cfg.cppdialect == 'gnu++0x') or (cfg.cppdialect == 'C++0x')
 	end
 
 	function m.completion(cfg)
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -117,10 +117,16 @@
 		tree.traverse(tr, {
 			-- folders are handled at the internal nodes
 			onbranchenter = function(node, depth)
-				_p(depth, '<VirtualDirectory Name="%s">', node.name)
+				if(depth > 0)
+				then
+					_p(depth, '<VirtualDirectory Name="%s">', node.name)
+				end
 			end,
 			onbranchexit = function(node, depth)
-				_p(depth, '</VirtualDirectory>')
+				if(depth > 0)
+				then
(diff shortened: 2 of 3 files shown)


src/base/fileconfig.lua
```diff
@@ -152,7 +152,11 @@
 --
 
 	function fileconfig.getconfig(fcfg, cfg)
-		return fcfg.configs[cfg]
+		if fcfg.configs ~= nil then
```
samsinsane: @starkos this is the second time I've seen this change be suggested, what is the preferred way to fix the issue of this function indexing `nil`?

modules/codelite/codelite.lua
```diff
@@ -21,7 +21,7 @@
 	function codelite.cfgname(cfg)
 		local cfgname = cfg.buildcfg
 		if codelite.workspace.multiplePlatforms then
-			cfgname = string.format("%s|%s", cfg.platform, cfg.buildcfg)
+			cfgname = string.format("%s-%s", cfg.platform, cfg.buildcfg)
```
samsinsane: Why `-`? The comment in the linked issue indicates that `_` worked, and that would be consistent with how `gmake`/`gmake2` handles this.

modules/codelite/codelite_project.lua
```diff
@@ -117,10 +117,16 @@
 		tree.traverse(tr, {
 			-- folders are handled at the internal nodes
 			onbranchenter = function(node, depth)
-				_p(depth, '<VirtualDirectory Name="%s">', node.name)
+				if(depth > 0)
+				then
```
samsinsane: This is inconsistent with the rest of the code base. `if depth > 0 then` would be consistent with the rest of the code base.

modules/codelite/codelite_project.lua
```diff
@@ -117,10 +117,16 @@
...
+					_p(depth, '<VirtualDirectory Name="%s">', node.name)
+				end
 			end,
 			onbranchexit = function(node, depth)
-				_p(depth, '</VirtualDirectory>')
+				if(depth > 0)
+				then
```
samsinsane: Same as above.

modules/codelite/codelite_project.lua
```diff
@@ -379,12 +385,16 @@
...
 
-	function m.isCpp11(cfg)
-		return (cfg.cppdialect == 'gnu++11') or (cfg.cppdialect == 'C++11') or (cfg.cppdialect == 'gnu++0x') or (cfg.cppdialect == 'C++0x')
+	function m.isCpp17(cfg)
+		return (cfg.cppdialect == 'gnu++17') or (cfg.cppdialect == 'C++17') or (cfg.cppdialect == 'gnu++1z') or (cfg.cppdialect == 'C++1z')
 	end
 
 	function m.isCpp14(cfg)
-		return (cfg.cppdialect == 'gnu++14') or (cfg.cppdialect == 'C++14') or (cfg.cppdialect == 'gnu++1y') or (cfg.cppdialect == 'C++1y')
+		return m.isCpp17(cfg) or (cfg.cppdialect == 'gnu++14') or (cfg.cppdialect == 'C++14') or (cfg.cppdialect == 'gnu++1y') or (cfg.cppdialect == 'C++1y')
```
samsinsane: I find this to be a very strange change. Why does C++17 need to be viewed as C++14 and C++11?

modules/codelite/codelite.lua
arthurdead: i just didn't like seeing x86_64_Debug i thought x86_84-Debug was more pleasing

modules/codelite/codelite_project.lua
arthurdead: well c++17 has all the features of both c++14 and c++11 and c++14 has all the features of c++11 this is for code completion so before if i had c++17 the c++14 code completion would never be enabled even though it has all the features that was my logic anyway

samsinsane: Just trying to understand this a bit more:
- What happens to the code completion for C++17 features? You've added `isCpp17` but that doesn't get used to tell CodeLite you're using C++17.
- Without your changes, if you were using C++14, does that mean _only_ features added in C++14 have code completion? This is kind of why I find this change so strange.

src/base/fileconfig.lua
starkos: @arthurdead Can you provide a minimal example of the setup that caused you to need this fix? I'm not sure how to reproduce the problem.

starkos:
I'm going to close this PR, as it breaks a couple of our submission guidelines (tries to fix multiple issues, no unit tests), and the C++ standard changes appear to require more discussion before they can be merged.

Appreciate the work you're putting into fixing Codelite though! I opened #1419 to address the platform configuration names issue. Happy to consider the other fixes as well, but they will need to be submitted separately, so we  can approve them while the C++ standard stuff gets sorted out.

Issue #1438, opened 2020-04-29
Disable XP deprecation warning when compiling with msvc v141_xp toolset


withmorten:
Hi there - I often compile projects with the v141_xp toolset in VS2019 - but doing so emits the following warning:

`MSB8051: Support for targeting Windows XP is deprecated and will not be present in future releases of Visual Studio.`

This can be disabled by adding `<XPDeprecationWarning>false</XPDeprecationWarning>` within a PropertyGroup of the Project, but I haven't found a way to do so with a premake.lua script.

So then, I just now decided to just fix it rather hackily in the premake source itself, by simply adding this line to *any* vc project ... (my commit in my fork here):

https://github.com/withmorten/premake-core/commit/aedeb3273cb070b761d8bb22b6e3b4678588ab75

Now that I know it's possible, is there any better way this could be solved? I personally doubt anybody choosing this specific toolset wants to see the warning, but maybe it should be only enabled with a specific setting, but I honestly don't know how one would implement it in premake.

Any ideas?

Thanks for the great tool,
Morten

starkos:
I would be good with always adding it by default. Could you submit a PR to add it?

withmorten:
Of course!

starkos:
Closed by #1439.

PR #1439, opened 2020-04-29
Disable XP deprecation warning for Visual Studio projects

withmorten:
Related issue: https://github.com/premake/premake-core/issues/1438

### DIFF
--- a/modules/vstudio/tests/vc2010/test_globals.lua
+++ b/modules/vstudio/tests/vc2010/test_globals.lua
@@ -290,6 +290,42 @@ end
 	end
 
 
+	function suite.canSetXPDeprecationWarningToFalse_withV141XP()
+		toolset "v141_xp"
+		prepare()
+		test.capture [[
+<PropertyGroup Label="Globals">
+	<ProjectGuid>{42B5DBC6-AE1F-903D-F75D-41E363076E92}</ProjectGuid>
+	<Keyword>Win32Proj</Keyword>
+	<RootNamespace>MyProject</RootNamespace>
+</PropertyGroup>
+<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Globals">
+	<XPDeprecationWarning>false</XPDeprecationWarning>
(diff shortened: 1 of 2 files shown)


withmorten:
Looks like all the tests are failing (should have thought of that) - perhaps it at least should only be added to Windows projects (and I should maybe move it to the bottom of the globals list)?

samsinsane:
You should definitely ensure it only happens on Windows (`os.target() == p.WINDOWS`) but also you should ensure that it only happens for VS2019 (your issue mentions that it only started in VS2019), if this continues in VS2021 it can be updated to include that but we shouldn't have this happening when XP support is completely gone. Additionally, you should move the entry to the bottom of the group, you're going to break a lot of unit tests by putting it at the top. Once you've done this, I highly recommend you write unit tests to ensure that it doesn't happen in VS2017 or when `system "android"`, and that it does happen in VS2019.

samsinsane:
@withmorten thanks for the additional context on that!

withmorten:
The final result then looks like this, instead of adding it to the Globals, it gets added to the Configuration specific parts:

      <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Globals">
        <XPDeprecationWarning>false</XPDeprecationWarning>
      </PropertyGroup>
      <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Globals">
        <XPDeprecationWarning>false</XPDeprecationWarning>
      </PropertyGroup>
      <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='ReleaseNO|Win32'" Label="Globals">
        <XPDeprecationWarning>false</XPDeprecationWarning>
      </PropertyGroup>`

That's okay?

samsinsane:
> That's okay?

Seems good to me :+1:

withmorten:
Okay, now it only sets the warning to false when specifically choosing v141_xp, and only per config!

I hope the way I added the tests (one for global, one for specific config "Release" only) is okay!

PR #1446, opened 2020-05-13
Do not add <Link> for .NET resources or the resource will not be properly embedded.

rhuvendiek:
**What does this PR do?**

During generation of C# project I did notice resx files did not get properly embedded into the built assembly.

EmbedResource actions or VS won't generate <Link>"file"</Link> tags anymore.

There might be cases if the link is needed. Obviously I do not know the indepth internals of Visual Studio :(

### DIFF
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -156,7 +156,7 @@
 
 			if #contents > 0 or external then
 				_p(2,'<%s%s Include="%s">', info.action, condition, fname)
-				if external then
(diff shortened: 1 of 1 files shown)


starkos:
Hmm…it's been quite a while since I added support for external files to C# projects. At the time, adding `<link/>` was the only way to get Visual Studio to see a file that lived outside of the project folder. If you double-click the file within Solution Explorer, does it open? If so, then I'm good with merging.

rhuvendiek:
Tested with VS2017: Yes the filesopen just fine.
Also from my observation by poking into the csproj VS won't generate those <link> tag itself

Issue #1461, opened 2020-06-03
Clang in Visual Studio 2019


nickclark2016:
**What problem will this solve?**
Currently, using clang in Visual Studio can be a hassle with Premake.  The toolset specification to use Clang with VS projects is not intuitive to determine.  I suggest that the `clang` toolset value should alias to the correct toolset to use for Visual Studio.

**What might be a solution?**
Currently, an llvm-vs* option exists in Premake.  This works for VS2017 with the LLVM Compiler Toolchain Extension, but the Clang toolchain is suppoted out of the box with VS 2019.  Thus, the developers of the extension have recommended to use that rather than the extension, which does not work in VS 2019.  In order to effectively unify setting the toolset to `clang` across modules, I would suggest the following:

| Module | Toolset Value | Value in Files |
|----------|----------------|---------------|
| VS 2017 | llvm-vs2017 | LLVM-vs2017 |
| VS 2019 | clang | ClangCL |

However, this proposed solution could break any code that is dependent upon clang setting a certain value in Visual Studio's project files.

**What other alternatives have you already considered?**
Currently, I get this working by adding `msc-ClangCL` as the toolset, which aliases to Visual Studio's LLVM (clang-cl) toolset option.  However, this is not desirable as I would prefer to be able to set the toolset across the all modules in my project as `clang`.

**Anything else we should know?**
By doing something like this, it would reduce the amount of code needed to set the toolset for cross-platform projects which use different modules, e.g. a project that uses VS on Windows and gmake/gmake2 on Linux.  If this is something worthwhile, I have free time to develop a solution for this.

starkos:
It sounds like you have far more experience with this than I do. If you'd like to run with it, I'd be happy to assist with getting the changes approved and merged. I'm not super concerned with breaking compatibility for a good cause so long as we do the best we can to minimize the pain.

nickclark2016:
Sounds good.  I've verified a build locally on my Linux machine and verified functionality there.  I'll test a build out on Windows a bit later today and verify functionality, then I'll submit a PR.

greenkalx:
There's a problem with VS2019 where it expects 'ClangCL' with mixed case.
Premake lowercases this in src/_premake_init.lua.
I removed this lowercasing and rebuilt premake.
Without this fix, the General project properties do not open in VS.

Also, I wanted to use latest LLVM instead of the one included with VS2019, which was LLVM 8, but may be LLVM 10 now with VS2019 updates?
[...]

PR #1462, opened 2020-06-07
Added clang as a valid option for projects generated with vs2019

nickclark2016:
**What does this PR do?**

This PR unifies the functionality of the `clang` toolset across gmake/gmake2 and VS2019+ actions.  Formerly, `clang` would not resolve to a valid toolset, so it would be set to the default toolset of the Visual Studio version (e.g. `v142` for VS2019).  This allows the vs2010 project generator to resolve "clang" to the correct platform toolset in the project file.

This break any code that was relying on the value `clang` to resolve to `v142` in VS2019.  Instead, `clang` now resolves to `ClangCL` in the `PlatformToolset` of the project file.  Issue #1461 describes the full affects of the changes.  This should only affect projects generated with the `vs2019` action.

### DIFF
--- a/modules/vstudio/vs2019.lua
+++ b/modules/vstudio/vs2019.lua
@@ -28,7 +28,7 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None", "Utility", "SharedItems" },
 		valid_languages = { "C", "C++", "C#", "F#" },
 		valid_tools     = {
-			cc     = { "msc"   },
+			cc     = { "msc", "clang" },
 			dotnet = { "msnet" },
 		},
 
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -96,4 +96,7 @@ return {
 
 	-- Visual Studio 2013+ C/C++ Shared Items projects
 	"vc2013/test_vcxitems.lua",
+
+	-- Visual Studio 2019+ C/C++ Clang Projects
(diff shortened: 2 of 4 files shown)


modules/vstudio/vs2019.lua
```diff
@@ -28,8 +28,8 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None", "Utility", "SharedItems" },
 		valid_languages = { "C", "C++", "C#", "F#" },
 		valid_tools     = {
-			cc     = { "msc"   },
-			dotnet = { "msnet" },
+			cc     = { "msc", "clang" },
+			dotnet = { "msnet"        },
```
samsinsane, line 32: I don't believe we do this kind of thing anywhere else, can you revert this?

nickclark2016: Yup. I'll have those changes out for review tomorrow evening (eastern).

samsinsane: Sorry, I think there might have been a misunderstanding here. I was just referring to the whitespace in the `dotnet` line. It looks like you've reverted the `cc` change too?

nickclark2016: Yeah, I think I misunderstood.  It looked like the code was trying to keep the brackets in line, so that's what I was doing there.  I'll just put back the clang and do nothing with bracket alignment, if that works.

samsinsane: I might be wrong on the spacing, but I don't recall seeing that anywhere else. Thanks for that, and sorry about the confusion here.

PR #1468, opened 2020-06-12
macOS deployment target support for gcc and clang

noresources:
Use the value of systemversion to set the Apple-specific gcc/clang option -mmacosx-version-min=<version>, equivalent of the Xcode setting MACOSX_DEPLOYMENT_TARGET
add tests for gcc and clang

add -mmacosx-version-min= build flag on macOS when the project defines a systemversion.

This flag will be added in all actions that use the gcc/clang getcflags and getcxxflags method. The Xcode action doesn't. It already has a similar behavior but use a Xcode setting insteaod.

Tested on a real project on Linux (no changes) and macOS .

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -75,7 +75,7 @@
 		local cflags = config.mapFlags(cfg, clang.cflags)
 
 		local flags = table.join(shared, cflags)
-		flags = table.join(flags, clang.getwarnings(cfg))
+		flags = table.join(flags, clang.getwarnings(cfg), clang.getsystemversionflags(cfg))
 
 		return flags
 	end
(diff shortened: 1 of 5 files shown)


src/tools/clang.lua
```diff
@@ -84,6 +84,23 @@
...
+--
+-- Returns C/C++ system version related build flags
+--
+
+	function clang.getsystemversionflags(cfg)
+		local flags = {}
+
+		if cfg.system == p.MACOSX or cfg.system == p.IOS then
+			local minVersion = p.project.systemversion(cfg)
+			if (type (minVersion) == "string") and (string.match(minVersion, "^%d+%.%d+") ~= nil) then
```
samsinsane, line 96: The type will always be string or nil, so `minVersion ~= nil` is how we normally test this kind of thing. Additionally, we don't validate the version string in xcode [here](https://github.com/premake/premake-core/blob/2c42771f3f66b3efb9c64fdf9a0f69b02496c62f/modules/xcode/xcode_common.lua#L1206-L1209), doing validation here would be inconsistent.

noresources: 1. Okay for the "~= nil" test. 
2. So, shall I remove the validation or shall I add it to Xcode ? If validation is kept, it may be useful to add a generic macox system version validation function somewhere in the core.

starkos:
It would probably be better if you validate at the point where the value is set, to make it easier for the script author to fix it? [Here's an example of how to do it](https://github.com/premake/premake-core/blob/master/src/_premake_init.lua#L1237-L1257).

noresources:
> It would probably be better if you validate at the point where the value is set, to make it easier for the script author to fix it? [Here's an example of how to do it](https://github.com/premake/premake-core/blob/master/src/_premake_init.lua#L1237-L1257).

The allowed values depends on project configuration (mainly os target). As I understand it, the "allowed" function only know the input value.
If there is no systemversion validation in all other actions, maybe it is better to remove it here for now.

samsinsane:
Personally, I'm against validating version numbers - to me, it's just Premake getting in the way of the user when they know exactly what the version needs to be.

noresources:
Validation has been removed in the new commit

samsinsane:
PR looks good to me, but I'll let @starkos chime in on the validation before this is merged.

PR #1473, opened 2020-06-18
macOS os.getversion() improvements

noresources:
**What does this PR do?**
* Read macOS system version from system property list instead of reading kernel version.
* Add description for latest versions

* Same behavior for previously supported OS versions
* Correct version number for any new version of macOS instead of 0.0
* Maybe a bit slower

* I have kept the previous method as a fallback if the property list is removed in future releases.
* The big switch over kernel version has been replaced by `info->minorversion = kern_major - 4;`. The relation is correct for all previously supported versions and probably for future ones.
* The description will remain "Mac OS X" for new versions until a manual update of the code.

There is simpler methods to retrieve the OS version but they use [private API](https://opensource.apple.com/source/DarwinTools/DarwinTools-1/sw_vers.c.auto.html) or Objective-C code.

### DIFF
--- a/src/host/os_getversion.c
+++ b/src/host/os_getversion.c
@@ -108,6 +108,7 @@ int getversion(struct OsVersionInfo* info)
 
 #elif defined(PLATFORM_MACOSX)
 
+#include <CoreFoundation/CoreFoundation.h>
 #include <sys/param.h>
 #include <sys/sysctl.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


samsinsane:
> There is simpler methods to retrieve the OS version but they use private API or Objective-C code.

Is there a simpler Objective-C version that uses public APIs? How much simpler?

@starkos would you object to having a simpler implementation written in a different language? The code is already platform specific, being written in the "platform language" isn't going to make that much of a difference.

noresources:
@samsinsane Objective-C code relies one Cocoa AppKit framework. It is not simpler, just use standard APIs.
[Here is an implementation example](https://github.com/openjdk/jdk/blob/master/src/java.base/macosx/native/libjava/java_props_macosx.c#L237)

starkos:
I would not object to using Objective C in the macOS build.

noresources:
I have made some exeperiments with the Objective-C API and the sw_vers command line tool
1. If I modify the SystemVersion.plist values, both sw_vers and the Objective-C API returns the modified values immediatly. So, the current pull request does exactly what Apple is doing under the hood
2. The Objective-C API is only available on OSX 10.10+ (with a deprecated alternative)
3. To compile Objective-C, Bottstrap.mak and premake4/5.lua have to be changed.

starkos:
Thanks for doing that research. @samsinsane I'm okay with the current, simpler plist approach, if you have no objections?

PR #1477, opened 2020-07-07
Add support for .swift file for xcode

sp-rafael-lecina:
**What does this PR do?**
 - Set swift files as source code
 - Adds a ``swiftversion`` to be able to configure Swift Language Version

**How does this PR change Premake's behavior?**
 - Adds ``.swift`` on the file extension lists with proper values
 - Adds a DSL api to set the SWIFT_VERSION setting

Are there any breaking changes? Will any existing behavior change?
 - No

**Anything else we should know?**

Add any other context about your changes here.
 - closes #667

**Did you check all the boxes?**

- [x] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [x] Add unit tests showing fix or feature works; all tests pass
- [x] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] Minimize the number of commits

PR #1491, opened 2020-07-30
Use single value for settings on xcode

sp-rafael-lecina:
**What does this PR do?**
This PR includes the changes of https://github.com/premake/premake-core/pull/1477 . I don't know how I could split those PRs to include tests for both of them

This pull request enables single setting formatting fox xcode projects.
If you use cocoapods and it finds SWIFT_VERSION on a xcode setting formatted as a list cocoapods is not able to read the swift version and fails

It changes how xcode settings are formatted on the pbxproj file. Before this they were formatted as single element lists
```
SWIFT_VERSION = (
                    4.0,
                );
```
After this it will be generated as
```
SWIFT_VERSION = 4.0;
```

I checked on our own projects and this syntax works on Xcode

### DIFF
--- a/modules/xcode/tests/test_xcode_project.lua
+++ b/modules/xcode/tests/test_xcode_project.lua
@@ -61,6 +61,17 @@
 		]]
 	end
 
+	function suite.PBXBuildFile_ListsSwiftSources()
+		files { "source.swift", "Info.plist" }
+		prepare()
+		xcode.PBXBuildFile(tr)
+		test.capture [[
+/* Begin PBXBuildFile section */
+		4616E7383FD8A3AA79D7A578 /* source.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B2BDAE0539CBF12983B9120 /* source.swift */; };
+/* End PBXBuildFile section */
+		]]
+	end
+
 	function suite.PBXBuildFile_ListsResourceFilesOnlyOnceWithGroupID()
 		files { "English.lproj/MainMenu.xib", "French.lproj/MainMenu.xib" }
 		prepare()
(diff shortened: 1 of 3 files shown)


sp-rafael-lecina:
I'm not quite familiar with the codebase and the current hacks, but when you create an Xcode project and use a single value on a setting Xcode creates it using this format. This is a block a project created from Xcodeproject where I just added ``-ffast-math`` to OTHER_CFLAGS
Captura de pantalla 2020-08-05 a las 9 43 34
This means that the current projects created using premake for Xcode are abusing arguments lists for single value arguments

What would be a proper solution if you consider this a wrong approach? This PR was mainly uncomment a couple of lines, so somebody at some point already considererd this as an option

We found we needed this SWIFT_VERSION to not be on a list for cocoapods to be able to parse the version when upgrading to the latest Facebook SDK on an quite complex objective-C. This is something I think will be more common in the near future and a problem for people using cocoapods

samsinsane:
I can't really put it any simpler than: Using just your changes in #1477, `SWIFT_VERSION` doesn't come out as a list for me.

It would have been a list because you started off using `xcodebuildsettings` and that API forces all the values to be lists. The correct solution is exactly what you did, add a proper API to set this setting and avoid using `xcodebuildsettings`.

sp-rafael-lecina:
@samsinsane You are completely right. I added the tests to the other PR as they are here and they pass
I started by adding this ``SWIFT_VERSION`` to ``xcodebuildsettings``, made it work by adding the changes of this PR and then tried to add the ``swiftversion`` api because I thought it was cleaner but didn't notice this changes were not needed anymore. I should have added the tests from the beginning

PR #1499, opened 2020-08-19
GitHub Actions: Artifacts

Timic3:
**What does this PR do?**

This pull request uploads binary files after pipeline is finished and can be downloaded afterwards.

No.

This puts all files from bin/[release,debug] folder to a Zip file and uploads it to Actions server.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -3,27 +3,43 @@ name: CI Workflow
 on: [push, pull_request]
 
 jobs:
-  linux-and-macos:
-    runs-on: ${{ matrix.os }}
+  linux:
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
         config: [debug, release]
-        include:
-          - os: ubuntu-latest
-            osname: linux
-          - os: macos-latest
-            osname: osx
+        platform: [x64]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Build
-      run: make -f Bootstrap.mak ${{ matrix.osname }} CONFIG=${{ matrix.config }}
+      run: make -f Bootstrap.mak linux CONFIG=${{ matrix.config }}
     - name: Test
       run: bin/${{ matrix.config }}/premake5 test
-    - uses: actions/upload-artifact@v2
+    - name: Upload Artifacts
+      if: matrix.config == 'release'
+      uses: actions/upload-artifact@v2
       with:
-        name: ${{ matrix.osname }}-${{ matrix.config }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       run: make -f Bootstrap.mak ${{ matrix.osname }} CONFIG=${{ matrix.config }}
     - name: Test
       run: bin/${{ matrix.config }}/premake5 test
+    - uses: actions/upload-artifact@v2
```
samsinsane, line 24: It would be nice if this step had a better name, such as "Upload Artifacts". Same for the Windows step.

.github/workflows/ci-workflow.yml
```diff
@@ -21,6 +21,10 @@ jobs:
...
+      with:
+        name: ${{ matrix.osname }}-${{ matrix.config }}
```
samsinsane, line 26: Since macOS will be transitioning to ARM64 soon, it's probably worth including the platform here, adding the values into the `include` sections of `matrix`.

samsinsane (changes requested):
Thanks for this! I was looking at doing this a couple weeks ago and completely forgot haha!

Just a few changes. Also, I don't think it's necessary to upload the debug builds, but happy for this to be reviewed at a later date.

Timic3:
Thanks for the review, regarding the debug builds, I totally agree, they are also rather big (twice the size), so it would be best to remove them (they can of course be added back, if there is a need for them). With that said, I guess there is no need for `matrix.config` in the artifact names then? Or should I leave them there?

samsinsane:
That's a good question, I hadn't really thought about it - I don't think it really matters either way, I'm happy with whatever you decide.

.github/workflows/ci-workflow.yml
Timic3: Should I add platform for Linux as well? If not, it would either require an if statement, to check if the OS is MacOS or separate the job (linux-and-macos) into two.

Also wouldn't macosx look better in the filename instead of osx?

samsinsane: Yeah, I think adding platform Linux is a good idea - I imagine if ARM64 desktops and laptops become the norm, people will want ARM64 builds for Linux, and possibly Windows too.

Yeah, `macosx` would look better, it's also the name we use for the release assets - I must have just copied the Travis-CI name, oops!

Timic3: I had to separate the jobs due to `matrix.include` not supporting arrays (therefore we can't spawn more jobs for x64 and ARM64).

samsinsane:
Thanks for this!

PR #1515, opened 2020-09-11
ConformanceMode flag implemented for >= vs2017

### LINKED ISSUE #1329: Is there a way to specify Visual Studio specific project properties?
I want to specify "Debug Information Format", "Conformance Mode", "Disable Language Extensions" properties and may be some other. Generally I assume it should be possible to specify all VS project options in premake script. I can do this with buildoptions, like ```buildoptions { "/permissive-" }``` but then I don't see the value in the property pages. Is there a way to specify these options? Is there a list of all VS project options premake supports?
May be it's related to #1306

cos-public:
**What does this PR do?**

Provides support for the ConformanceMode flag for Visual Studio projects. Closes #1329.

Supports additional configuration flag.

### DIFF
--- a/modules/vstudio/tests/vc2010/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2010/test_compile_settings.lua
@@ -428,6 +428,34 @@
 		]]
 	end
 
+--
+-- If ConformanceMode flag is set, add <ConformanceMode> element (supported from VS2017)
+--
+
+	function suite.onConformanceMode_VS2015()
+		flags "ConformanceMode"
+		p.action.set("vs2015")
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<Optimization>Disabled</Optimization>
+		]]
+	end
+
+	function suite.onConformanceMode_VS2017()
+		flags "ConformanceMode"
+		p.action.set("vs2017")
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<Optimization>Disabled</Optimization>
+	<ConformanceMode>true</ConformanceMode>
+		]]
+	end
 
 --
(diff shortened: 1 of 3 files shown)


modules/vstudio/tests/vc2010/test_compile_settings.lua
```diff
@@ -428,6 +428,34 @@
 		]]
 	end
 
+--
+-- If ConformanceMode flag is set, add <ConformanceMode> element (supported from VS2017)
+--
+
+	function suite.onConformanceMode_VS2015()
```
samsinsane, line 435: This test doesn't confirm that `<ConformanceMode>` isn't output, you're just checking that everything up to the conformance mode element is output. You need to check that the element that would come after it is there, whatever that happens to be.

src/_premake_init.lua
```diff
@@ -468,6 +468,7 @@
 		kind  = "list:string",
 		allowed = {
 			"Component",           -- DEPRECATED
+			"ConformanceMode",
```
samsinsane, line 471: We don't add to `flags` anymore, you should create a new API called `conformancemode` with `kind = boolean`.

samsinsane (changes requested):

modules/vstudio/tests/vc2010/test_compile_settings.lua
cos-public: Could you suggest a correct way to check for absense of a flag? Relying on a default project properties is not reliable probably.

samsinsane: Like I said, you can just check for the element that is output after it. Otherwise, you can use `test.excludes("<ConformanceMode>true</ConformanceMode>", p.captured())`.

cos-public: `test.excludes()` is better here I think. Thank you.

samsinsane (changes requested):
Looks like you've got some XCode changes included in here, can you rebase those out or something?

cos-public:
I've rebased my branch, so the diff looks correct now. One of the tests is failing bacause of HTTP 502 error, which is irrelevant probably.

samsinsane (approved):

Issue #1520, opened 2020-09-15
toolset "clang" changes the formatting of links from X.lib to -lX for vs2019


nothke:
**What seems to be the problem?**
As the title says, putting `toolset "clang"` changes the form of `LinkName.lib` to `-lLinkName`, resulting in linking errors. Tested only with VS 2019.

**What did you expect to happen?**
The links should have .lib suffix, as they do without `toolset "clang"`

**How can we reproduce this?**
Sample:

``` lua
workspace "premake-test"
   configurations { "Debug", "Release" }

project "premake-test"
	kind "ConsoleApp"
	language "C++"
	architecture "x64"
	cppdialect "C++17"
	toolset "clang" -- comment this line for correct links
	targetdir "bin/%{cfg.buildcfg}"
	defines{"WIN32", "_CONSOLE"}

	files { "src/**.h", "src/**.cpp" }

	postbuildcommands {
		"{COPY} deps/sdl/lib/x64/SDL2.dll %{cfg.targetdir}"
	}

	includedirs { "deps/include" }

	libdirs{ 
		"deps/sdl/lib/x64",
	}

	links{
		"SDL2", -- becomes SDL2.lib with no toolset, but -lSDL2 with toolset "clang"
		"SDL2main"
	}

	filter "configurations:Debug"
		defines{"_DEBUG"}
		symbols "On"

	filter "configurations:Release"
		defines {"NDEBUG"}
		optimize "On"
```

This premake produces this in .vcxproj:
``` xml
<AdditionalDependencies>-lSDL2;-lSDL2main;%(AdditionalDependencies)</AdditionalDependencies>
```

but without toolset line, it produces this, the correct form:
``` xml
<AdditionalDependencies>SDL2.lib;SDL2main.lib;%(AdditionalDependencies)</AdditionalDependencies>
```

**What version of Premake are you using?**
Latest dev, built today.

samsinsane:
Thanks for the report! If you need a temporary workaround, you can override [this function](https://github.com/premake/premake-core/blob/15a07510ab689468c40d55215078e172e029de3b/modules/vstudio/vs2010_vcxproj.lua#L1394-L1410) using:
```lua
premake.override(premake.vstudio.vc2010, "additionalDependencies", function(oldfn, cfg, explicit)
  if cfg.system ~= premake.WINDOWS then
    oldfn(cfg, explicit)
  else
    local links = premake.vstudio.getLinks(cfg, explicit)
    if #links > 0 then
[...]

Issue #1507, opened 2020-08-30
toolset("clang") doesn't do anything


imgeself:
**What seems to be the problem?**
I am trying to use llvm toolset on windows. But when I set toolset to `clang`, premake generates projects with vs2019 toolset.

**What did you expect to happen?**
It generates visual studio project file that `PlatformToolset` set to `v142`. It should set to `ClangCL`

**What have you tried so far?**
I downloaded premake source code and modified [premake config file](https://github.com/premake/premake-core/blob/master/premake5.lua) and set the Premake5 project toolset to `clang`. It works there. But when I try on some other projects that using premake such as [Hazel](https://github.com/TheCherno/Hazel). It doesn't work there either.

**How can we reproduce this?**
I can reproduce this easily. Just create really basic workspace and project. Set toolset to clang.
```
workspace "Solution"
    configurations { "Debug", "Release" }
    startproject "Project"

project "Project"
    kind "ConsoleApp"
    toolset "clang"

```

**What version of Premake are you using?**
I tried alpha 15 and alpha 14.

samsinsane:
> But when I set toolset to "clang", premake generates projects with vs2019 toolset.

Please elaborate.

> I downloaded premake source code and modified some of the premake config files. When I set toolset to "clang" on those files it works.

Please elaborate.

> I tried on some other projects that using premake such as Hazel. It doesn't work there either.

They provide a pre-built version of Premake, you need to go talk to them about updating their version of Premake if you're having issues with their project.

> I can reproduce this easily.

Then provide a small sample as requested.

> I tried multiple versions

Then provide the version numbers as requested.

The bug template is there to be filled in, not ignored and mostly deleted. You only delay your own help by not doing things as requested.

imgeself:
I don't know how to elaborate more because the problem is very clear. I updated the bug report and hoping that it is more clear and understandable now.

> They provide a pre-built version of Premake, you need to go talk to them about updating their version of Premake if you're having issues with their project.

Hazel just a example, I am having this issue on my projects as well. I was trying to say that it happens other projects too.

nickclark2016:
@imgeself I merged in code a while back to do just what you are requesting (clang option for VS2019).  I've got a custom build of premake that I use with this added.  It's PR #1462.  @samsinsane would it be worthwhile to put out a new alpha release?

imgeself:
[...]

PR #1522, opened 2020-09-21
Fixed external library linking for Clang in VS2019+

nickclark2016:
**What does this PR do?**

Fixes problems referenced in #1520 and #1507 where the clang toolset produces a bad external library path when used with VS2019.  Current behavior deferred the library generation to gcc's implementation, such that external libraries received the string "-lName" (e.g. "-lSDL") instead of "name.lib" (e.g. "SDL.lib").

Changes the clang tool's `getlinks` method to defer to MSC when the action is VS2019 or newer.

N/A

### DIFF
--- a/modules/vstudio/tests/vc2019/test_link.lua
+++ b/modules/vstudio/tests/vc2019/test_link.lua
@@ -1,0 +1,43 @@
+--
+-- tests/actions/vstudio/vc2010/test_compile_settings.lua
+-- Validate compiler settings in Visual Studio 2019 C/C++ projects.
+-- Copyright (c) 2011-2020 Jason Perkins and the Premake project
+--
+
+local p = premake
+local suite = test.declare("vstudio_vs2019_compile_settings")
+local vc2010 = p.vstudio.vc2010
+local project = p.project
+
+--
+-- Setup
+--
+
+	local wks, prj
+
+	function suite.setup()
+		p.action.set("vs2019")
+		wks, prj = test.createWorkspace()
+	end
+
+	local function prepare(platform)
+		local cfg = test.getconfig(prj, "Debug", platform)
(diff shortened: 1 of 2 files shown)


src/tools/clang.lua
```diff
@@ -282,7 +283,12 @@
 --
 
 	function clang.getlinks(cfg, systemonly, nogroups)
-		return gcc.getlinks(cfg, systemonly, nogroups)
+		-- For visual studio projects, the links should match the MSC convention
```
samsinsane, line 286: This is wrong, it should only apply to Windows projects, not all projects. Furthermore, this change should be made in the VS module.

samsinsane (changes requested):
I appreciate the effort, but VS2019 supports more than just Windows with Clang.

src/tools/clang.lua
nickclark2016: I can easily move the change to the VS module.  Just so I go the agreed course of action, would you say:

* Changes made in the vs2019 module
* Changes only affect the clang and msc toolsets for windows projects

nickclark2016:
Not sure why the test is failing for macosx release on the response code test, especially since what I did *shouldn't* have affected that

samsinsane (approved):

nickclark2016:
Huh, wonder what the issue with the last run was.  Glad to see the release macosx build is working now, though.

samsinsane:
The HTTP unit tests connect to an external service, it's likely it was just a weird issue within the Azure network - I get this a lot at work with Azure Pipelines.

nickclark2016:
> 
> 
> The HTTP unit tests connect to an external service, it's likely it was just a weird issue within the Azure network - I get this a lot at work with Azure Pipelines.

Makes sense. Thanks for closing this out!

Issue #1518, opened 2020-09-12
Adding "Embed Without Signing" and "Embed & Sign" options for frameworks to xcode


thomashope:
I am using premake to generate an xcode project and have almost everything working, except I can't figure out how to set third party frameworks to be embedded into the final application bundle. Reading the wiki I found `copylocal` which appears to describe what I want to do, but is only for C# projects.

**Possible Workarounds**

1. Emulate "Embed Without Signing" by adding a `postbuildcommand` to copy the framework into `MyApp.app/Contents/Frameworks` (I haven't tried this).
2. Emulate "Embed & Sign" by adding a `postbuildcommand` to copy the framework and then sign it with `codesign`? (Also haven't tried this).
3. Open Xcode and change the drop down under _general > Frameworks, Libraries, and Embedded Content_, this is what i'm currently doing...

My current premake5.lua for reference: https://gist.github.com/thomashope/9616970abefa3e3ff55fb7d7316aa907

If there is already support for doing this and I just haven't found it then great! Otherwise I can start putting together a pull request to add this feature if someone can give me guidance on how best to integrate it.

**Adding to Premake**

If i'm going to work on a pull request my current plan for this would be to have `copylocal`, when used as in my gist, set the framework to "Embed & Sign" as the (untested) workaround for that seems like more of a PITA than the (also untested) workaround for "Embed Without Signing". I've been poking around the premake source and generated xcode projects and think I have a fair idea of how to do this, but will take some time anyway.

Suggestions welcome :) Thanks for all your work so far!

samsinsane:
The XCode generator is pretty awful to override using [`premake.override`](https://github.com/premake/premake-core/wiki/Overrides-and-Call-Arrays) but you can always try to do that as a fourth option.

I'm not 100% certain if reusing `copylocal` is the right option, if you need to be able to swap between signing and not signing, `copylocal` doesn't provide any flexibility there. @starkos any ideas on how this could or should be handled?

nickgravelyn:
It's not terribly complicated in the project file but it's mostly about figuring out how to cleanly specify it in Premake scripts. I've looked at this a couple of times but I found it really tricky to do with the current way that specifying libraries to link. I felt like it should actually be part of `links` but currently `links` just adds linker flags which isn't quite right for how you'd normally setup an embedded library for linking. Additionally you wouldn't have options for whether you just want it to copy vs copy and sign. It might just end up needing to be a custom `xcodeembed` function or similar.

[...]

PR #1524, opened 2020-09-26
Xcode embed libraries

nickgravelyn:
**What does this PR do?**

Provides a mechanism to specify `dylib`s and `framework`s to use Xcode's "Embed Libraries" functionality to optionally or embed & sign libraries. Trying to address #1518.

1. Xcode projects now support specifying `.dylib` extensions in `links`. These can be relative or absolute paths but the libraries end up getting treated like `.framework`s. That is, they are added to the Frameworks build phase and show up in the Frameworks group rather than just being added as additional linker flags. This was a prerequisite for `dylib`s to get embedded.
2. A new `xcodeembedlibraries` command was added that allows a project to specify the list of libraries to embed in a map, mapping the name of the library (not the path) to a string of either `"embed"` or `"embed-and-sign"`. 

While this functionally achieves the goals as far as I can tell, I'm not married to the implementation or interface so I'm open to suggestions there. The `xcodeembedlibraries` feels weird particularly because strings aren't great for this. I thought about just having the right hand side be true/false but that felt too opaque as to what it was doing. I also considered separate commands as discussed in #1518 but ultimately it felt nicer to have a single command since it's controlling a single list in Xcode.

I've written automated tests and verified this works on my own project (which embeds and signs a local `dylib`) but it probably needs more testing around the edges.

I left this as two commits so that it's easier to review the two changes described above. I'm happy to rebase/squash it later if desired for merging.

### DIFF
--- a/modules/xcode/_preload.lua
+++ b/modules/xcode/_preload.lua
@@ -46,6 +46,12 @@
 		}
 	}
 
+	p.api.register {
+		name = "xcodeembedlibraries",
+		scope = "config",
+		kind = "key-string",
+	}
+
 --
 -- Register the Xcode exporters.
 --
--- a/modules/xcode/tests/test_xcode_dependencies.lua
+++ b/modules/xcode/tests/test_xcode_dependencies.lua
@@ -69,6 +69,44 @@
 		]]
 	end
 
+	function suite.PBXBuildFile_ListsDependencyTargets_OnSharedLibWhenEmbedding()
+		kind "SharedLib"
+
+		project "MyProject"
+		xcodeembedlibraries
+		{
+			["MyProject2"] = "embed"
+		}
+
+		prepare()
+		xcode.PBXBuildFile(tr)
+		test.capture [[
+/* Begin PBXBuildFile section */
+		1BC538B0FA67D422AF49D6F0 /* libMyProject2-d.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 107168B810144BEA4A68FEF8 /* libMyProject2-d.dylib */; };
+		6514841E8D4F3CD074EACA5E /* libMyProject2-d.dylib in Embed Libraries */ = {isa = PBXBuildFile; fileRef = 107168B810144BEA4A68FEF8 /* libMyProject2-d.dylib */; };
+/* End PBXBuildFile section */
+		]]
+	end
+
+	function suite.PBXBuildFile_ListsDependencyTargets_OnSharedLibWhenEmbeddingAndSigning()
+		kind "SharedLib"
+
+		project "MyProject"
+		xcodeembedlibraries
+		{
+			["MyProject2"] = "embed-and-sign"
+		}
+
+		prepare()
+		xcode.PBXBuildFile(tr)
+		test.capture [[
+/* Begin PBXBuildFile section */
+		1BC538B0FA67D422AF49D6F0 /* libMyProject2-d.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 107168B810144BEA4A68FEF8 /* libMyProject2-d.dylib */; };
+		6514841E8D4F3CD074EACA5E /* libMyProject2-d.dylib in Embed Libraries */ = {isa = PBXBuildFile; fileRef = 107168B810144BEA4A68FEF8 /* libMyProject2-d.dylib */; settings = {ATTRIBUTES = (CodeSignOnCopy, ); }; };
+/* End PBXBuildFile section */
+		]]
+	end
+
 
 ---------------------------------------------------------------------------
 -- PBXContainerItemProxy tests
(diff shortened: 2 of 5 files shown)


thomashope:
Hi, nice work on the PR.

Read through the changes, the first commit looks good to me. Wasn't sure about the second one so I recreated your changes on my end and found a few issues in the process:

1) Embedded items should have have an additional entry in the `PBXBuildFile section`, even if they are not code signed. 

2) Frameworks should have `RemoveHeadersOnCopy` attribute set like so `		0DAEAB24FB163DD698103164 /* SDL2.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 7D41A17F89CC53F104B11FBF /* SDL2.framework */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };`. This attribute seems to do pretty much as described, if you look inside framework copied into the built .app you will see the Headers folder has been removed.

3) Pretty sure we need to add `LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks";` so that the executable finds and uses the correct framework placed inside the app bundle. I am guessing what happened on your end was the files appeared to be copied to the correct location but the executable selected you system installed libraries? You should be able to check that by moving/renaming your system libraries and see if your app crashes.

I have done all this testing using .frameworks because that's what I need it for. You can see that reflected in my code where it says 'Embed Frameworks' wheras yours says 'Embed Libraries'. Would be interested to know what xcode generates when you mix the two and if there are any additional cases to handle.

Are you able to you share your project folder structure / premake5.lua using .dylibs so I can test it out on my end?

[Here's the link](https://github.com/thomashope/premake-core/commit/5b570f32ae4cd1f6a37bb76a12307d99ec5ade73) to my commit sans tests if you want to take a look. I went for having separate `xcodeembedwithoutsigning` and `xcodeembedandsign` just to see how it would pan out.

Bearing in mind this is all stuff i've been figuring out from my own experimentation/looking at your code/and fiddling with the xcode project, so I may be wrong!

nickgravelyn:
Hey @thomashope. Thanks for taking a peek.

1. Oh yes good catch. I'll fix that up. I didn't have a use case for not signing so I didn't catch that it was missing.
2. That's a good point and seems reasonable. I don't use frameworks much so I forget about the embedded headers. IIRC Xcode lets you specify whether or not to remove headers but it seems reasonable to default to doing so and someone can come patch in a way to disable that if there's some reason they want to bloat their app bundle with a bundle of header files.
3. I don't think that should be done automatically. Sometimes libraries will already be built with that path into them, but for others I think the project should specify it themselves using `xcodebuildsettings { ["LD_RUNPATH_SEARCH_PATHS"] = "@executable_path/../Frameworks" }`.

I don't think there are any concerns with mixing the two. Xcode just names it based on what you add first, I believe. They're all linked and copied to the same place in the end.

nickgravelyn:
Ok I've addressed the first two points (did a rebase and kept it at two commits). You can see the test cases here: https://github.com/premake/premake-core/pull/1524/files#diff-06a55edcdb18495d6fe7e42f832b512dR122-R153

thomashope:
So I built your code and ran it on my project, and it works!

For reference here is what I had to do to my premake5.lua to get local frameworks correctly embedded into my app:

```lua
links {
    "third_party/sdl2/macos/SDL2.framework",
}

sysincludedirs {
    "third_party/sdl2/macos/SDL2.framework/Headers",
}

frameworkdirs {
    "third_party/sdl2/macos/",
}

xcodeembedlibraries {
    ["SDL2.framework"] = "embed-and-sign"
}

xcodebuildsettings {
    ["LD_RUNPATH_SEARCH_PATHS"] = "$(inherited) @executable_path/../Frameworks",
}
```

It's a fair bit of config for one framework and I think the user experience here can be improved, but this PR is a step in the right direction so it's a 👍 from me.

nickgravelyn:
That’s great to hear. I realized I need to do some more testing for embedding projects as well as loose file. Once that’s done I think the big question will just be about the interface to make sure it seems reasonable.

nickgravelyn:
Ok I've fixed things up so project references can also be marked to embed or embed & sign. Instead of using the file name it's the name of the project. This felt better since it matches with what users specify in `links`. I wrote some automated tests in the dependency test file but I also did a quick manual test and it worked in Xcode.

At this point things are functional for relative `dylib`/`framework` references as well as `SharedLib` project references. I'd be fine with this `xcodeembedlibraries` interface but I'm also happy to adjust it if someone has a more user-friendly idea of how we can expose these options.

starkos:
The not a fan of the table syntax for this one. I'm not a fan of the table syntax in general and use it only under duress, but this one really doesn't need it. I would much prefer that it use separate APIs instead.

```lua
links { "SDL" }
embed { "SDL" }
embedAndSign { "SDL" }
```

nickgravelyn:
I deleted my fork at some point so to continue this PR I had to re-fork and make a new PR. I've applied the API changes requested in #1619. Going to close this one out.

PR #1525, opened 2020-09-29
[Codelite] Only active (i.e. unique) configuration should be selected.

Jarod42:
**What does this PR do?**

Fix active configuration under codelite

Only codelite generator is changed.

To reproduce the issue:

- create a project with 3 configurations in premake (Debug, Profile, Release)
- generate for codelite
  - Here all 3 configurations are the active one :/
  - codelite handles the conflict by selecting the first active (so Debug in that case)

- open codelite
- switch to Release (so now Debug is not active, Release is (still) active, Profile is untouched (so active))
  - Profile is selected

### DIFF
--- a/modules/codelite/codelite_workspace.lua
+++ b/modules/codelite/codelite_workspace.lua
@@ -82,10 +82,12 @@
 
 		-- for each workspace config
 		p.push('<BuildMatrix>')
+		local selected = "yes" -- only one configuration should be selected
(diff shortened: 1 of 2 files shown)


starkos (approved):
Hmm…looks like we have `defaultplatform()` to select an initially active platform, but no `defaultconfiguration()`. I think this is a reasonable (and better) default implementation. Thanks for the contribution!

PR #1534, opened 2020-10-12
Added Ability to Disable JMC in Visual Studio

### LINKED ISSUE #1526: Need support for Just My Code in VS2017
A project has some code that crashes unless `<SupportJustMyCode>false</SupportJustMyCode>` is added to the VS2017 project.  VS2017 15.8 and higher have JustMyCode debugging enabled by default, and I'm using VS2019.

The project uses a lambda to generate code that gets copied to virtual memory in another process in order to invoke `CreateRemoteThread`.  JustMyCode injects assembly code into the lambda to call functions outside the lambda itself, thus copying the lambda code to the other process results in crashing assembly code.  Repos are here: [mridgers/clink](https://github.com/mridgers/clink) and [chrisant996/clink](https://github.com/chrisant996/clink).

**What might be a solution?**
Add a config option for VS2017.

After reading premake5 docs, doing internet searches, and reading source code in the premake repo, it looks like there is no realistic way to make premake emit the XML tag other than to modify the premake tool itself.  I also researched if there's a way to disable JustMyCode support via a `#pragma`, but there isn't.

I see four alternatives:
1. Ask project contributors to manually edit the vcxproj files after premake generates them.  That's not preferable because the project relies on premake to define a c++ macro with the head commit hash, and it uses the hash to identify whether it needs to copy an updated binary into place.  So the normal workflow has been to invoke premake after every git pull, and before every release or final build.
2. Create a script that post-processes the premake output.  That's not preferable because it requires multiple steps.
3. Learn how to improve premake itself.  That's not preferable because of the rampup cost.  It looks like [this SO post](https://stackoverflow.com/questions/39130550/injecting-properties-into-vs-projects-with-premake5/39151157#39151157) has rough high level directions for adding a certain kind of setting.  With some time and research, that could be a starting point for figuring out how to identify what kind of setting SupportJustMyCode is, where it should go, how to make it a no-op for other toolsets, etc.
4. Switch to a different toolset abstraction tool.  That's least preferable, because of existing build tooling that relies on premake.

Are there other options that I didn't notice?

I'm trying to revive these projects into working order, and I don't want to get sidetracked into making improvements in another project first.  So the best alternative for me right now is option 2, using a script to post-process the premake output.

When a project needs to use a new VS option that premake doesn't natively support yet, there doesn't seem to be a low cost path forward for a consumer of premake.  One must become a contributor to premake, in order to become a consumer of premake.  That makes it difficult for a multi-author project to use premake in the meantime.  It could be beneficial to have some way to inject arbitrary XML tags into the premake output, so that premake consumers don't have to postpone their own project progress while waiting for premake to natively support new toolset features.

Thank you for reading this and considering the feedback!

nickclark2016:
**What does this PR do?**

This PR resolves #1526.  It adds a "jmc" option to projects in order to disable just my code debugging.

Usage:
```lua
project "SomeProject"
   justmycode "Off"
```

There is no change for preexisting projects, thus it is a non-breaking change.  By settings `justmycode` to `"Off"` in a VS2017 or newer project, it adds a flag to the project file to disable JMC.

N/A

### DIFF
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -97,6 +97,8 @@ return {
 	-- Visual Studio 2013+ C/C++ Shared Items projects
 	"vc2013/test_vcxitems.lua",
 
-	-- Visual Studio 2019+ C/C++ Clang Projects
+	-- Visual Studio 2019+ C/C++ Projects
+	"vc2019/test_compile_settings.lua",
+	"vc2019/test_link.lua",
 	"vc2019/test_toolset_settings.lua"
 }
--- a/modules/vstudio/tests/vc2019/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2019/test_compile_settings.lua
@@ -1,0 +1,41 @@
+--
+-- tests/actions/vstudio/vc2010/test_compile_settings.lua
+-- Validate compiler settings in Visual Studio 2019 C/C++ projects.
+-- Copyright (c) 2011-2020 Jason Perkins and the Premake project
+--
+
+	local p = premake
+	local suite = test.declare("vstudio_vs2019_compile_settings")
+	local vc2010 = p.vstudio.vc2010
+	local project = p.project
+
+--
+-- Setup
+--
+
+	local wks, prj
+
+	function suite.setup()
+		p.action.set("vs2019")
+		wks, prj = test.createWorkspace()
+	end
+
+	local function prepare(platform)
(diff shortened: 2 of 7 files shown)


starkos (approved):
Thanks for this, and all the help you've been providing lately—much appreciated!

nickclark2016:
Just to follow up on this, can we get a `justmycode` page in the wiki?

rellensb:
Shouldn't this be a boolean? It seems kind of weird to be able to set it to `Off` but not to `On`

PR #1535, opened 2020-10-17
Added vs-latest action

### LINKED ISSUE #1479: Default Generator
I automatically call Premake on several projects, and currently I have the generator hardcoded into the script.
Problem is I want to use a different generator per platform, or switch to a different generator, or use the latest Visual Studio generator.

**What might be a solution?**
two part:
1. option to set a system wide default generator that will be used when no generator is explicitly specified.
2. "vs-latest" generator option that automatically finds and uses the latest Vistual Studio version that's installed.

might be possible to do dirty hacks like using environment variables.

nickclark2016:
**What does this PR do?**

This adds a `vs-latest` action (closes #1479).  When `vs-latest` is invoked, it checks for the `vswhere` executable on this system and runs it to determine the latest VS version.  If no version is found, `vswhere` does not exist, or the operating system is not windows, it defaults to `vs2019`, the current version as of writing this PR.

There are no breaking changes to the API.  This only adds functionality as described above.

This makes passing the unit tests dependent upon the machine that the tests are run on.

### DIFF
--- a/modules/vstudio/vslatest.lua
+++ b/modules/vstudio/vslatest.lua
@@ -1,0 +1,58 @@
+--
+-- vslatest.lua
+-- Extend the existing exporters with support for Visual Studio 2019.
+-- Copyright (c) Jason Perkins and the Premake project
+--
+
+	local p = premake
+	local vstudio = p.vstudio
+	local latest = nil
+
+	function getLatestVisualStudioVersion()
+		-- Get the most recent version installed
+		if latest == nil then
+			local version = nil
+			if os.host() == 'windows' then
+				local ver, err = os.outputof('"C:\\Program Files (x86)\\Microsoft Visual Studio\\Installer\\vswhere.exe" -latest -property catalog_productLineVersion')
+				if err ~= nil and err == 0 then
+					version = 'vs' .. ver
+				end
+			end
+
+			if version == nil then
+				version = "vs2019"
+				print("No Visual Studio installation found using vswhere.exe. Defaulting to " .. _ACTION .. ".")
+			end
+
+			latest = p.action.get(version)
+		end
+	
+		return latest
+	end
+
+---
+-- Define the Visual Studio Latest export action.
+---
+
+	newaction {
+		-- Metadata for the command line and help system
+		
+		trigger	    = "vs-latest",
+		shortname   = "Visual Studio - Latest",
+		description = "Generate Visual Studio project files of latest installation",
+
+		-- Copy data from the latest vs action
+		
+		targetos         = getLatestVisualStudioVersion().targetos,
+		toolset          = getLatestVisualStudioVersion().toolset,
+		valid_kinds      = getLatestVisualStudioVersion().valid_kinds,
+		valid_tools      = getLatestVisualStudioVersion().valid_tools,
+		onWorkspace      = getLatestVisualStudioVersion().onWorkspace,
+		onProject        = getLatestVisualStudioVersion().onProject,
+		onRule           = getLatestVisualStudioVersion().onRule,
+		onCleanWorkspace = getLatestVisualStudioVersion().onCleanWorkspace,
+		onCleanProject   = getLatestVisualStudioVersion().onCleanProject,
+		onCleanTarget    = getLatestVisualStudioVersion().onCleanTarget,
+		pathVars         = getLatestVisualStudioVersion().pathVars,
+		vstudio          = getLatestVisualStudioVersion().vstudio,
+	}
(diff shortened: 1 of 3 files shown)


modules/vstudio/vslatest.lua
```diff
@@ -0,0 +1,58 @@
...
+---
+-- Define the Visual Studio Latest export action.
+---
+
+	newaction {
+		-- Metadata for the command line and help system
+		
+		trigger	    = "vs-latest",
+		shortname   = "Visual Studio - Latest",
+		description = "Generate Visual Studio project files of latest installation",
```
starkos, line 42: Nitpick, but maybe this would be clearer as something like "Generate project files for newest installed Visual Studio"? At the very least, should be "project files *for* latest" instead of "project files of latest". Otherwise looks okay to me.

starkos (changes requested):

modules/vstudio/vslatest.lua
nickclark2016: Fixed!

modules/vstudio/vslatest.lua
```diff
@@ -0,0 +1,58 @@
...
+	newaction {
+		-- Metadata for the command line and help system
+		
+		trigger	    = "vs-latest",
+		shortname   = "Visual Studio - Latest",
+		description = "Generate Visual Studio project files for latest installation",
+
+		-- Copy data from the latest vs action
+		
+		targetos         = getLatestVisualStudioVersion().targetos,
```
samsinsane, line 46: This executes on start-up, not just when this action is used. You can see in the checks, that on non-Windows builds Premake now emits `No Visual Studio installation found using vswhere.exe. Defaulting to test.`

modules/vstudio/vslatest.lua
```diff
@@ -0,0 +1,58 @@
...
+			if os.host() == 'windows' then
+				local ver, err = os.outputof('"C:\\Program Files (x86)\\Microsoft Visual Studio\\Installer\\vswhere.exe" -latest -property catalog_productLineVersion')
+				if err ~= nil and err == 0 then
+					version = 'vs' .. ver
+				end
+			end
+
+			if version == nil then
+				version = "vs2019"
+				print("No Visual Studio installation found using vswhere.exe. Defaulting to " .. _ACTION .. ".")
```
samsinsane, line 24: Was this supposed to use `version` instead of `_ACTION`?

samsinsane (changes requested):
For what it's worth, this new action is broken - you need to modify `_ACTION` otherwise various elements won't be output correctly depending on the VS version. What happens when VS2021 comes out? This action just stops working for old builds of Premake just because you installed the newest version of VS? That sounds pretty lame.

modules/vstudio/vslatest.lua
nickclark2016: Hmm... that's not a behavior we want.   Is there a mechanism that I can use to hold this off until the vs-latest action is invoked? Or would it just be best to only emit the error if `_ACTION == vs-latest`.

samsinsane: I don't think so, from my understanding this kind of action is tricky to pull off since you need to execute this kind of thing before the oven/bake process. Your last idea sounds like the best move, but overall, I'm not sold on this new action as I stated in the overall review comment.

starkos:
What about hooking something early in the `_premake_main` call array, before the user script is run, do this sort of thing?

```lua
if _ACTION == 'vslatest' then
   _ACTION = findLatestVersion()
end
```

Untested, might not work, but would be pretty transparent if it did.

nickclark2016:
Wow I just noticed this is still out.  Sorry I've been out of it for a bit.  Let me re-acclimate and I'll try to get this cleaned up.

nickclark2016:
> 
> 
> What about hooking something early in the `_premake_main` call array, before the user script is run, do this sort of thing?
> 
> ```lua
> if _ACTION == 'vslatest' then
>    _ACTION = findLatestVersion()
> end
> ```
> 
> Untested, might not work, but would be pretty transparent if it did.

So I tried something like this out by hooking into the base/action.lua initialization.  When I go to run the tests, it doesn't like this.  I think the vslatest action may be the best approach, if we still want to go with this.  I'll hold off on continuing until we figure out if we want to continue on the original path, addressing @samsinsane's comments.

starkos:
> So I tried something like this out by hooking into the base/action.lua initialization. When I go to run the tests, it doesn't like this

That's not really enough info to provide any useful suggestions. What did you try? What issues did you experience? It looks like `action.initialize()` is early enough in the initialization order for this to work; is that where you did it? Did you also set `_ACTION`?

starkos:
@nickclark2016 I'm cleaning up the PR list—are you still working on this one?

nickclark2016:
@starkos Nah, I've been way too busy lately, so my PRs have fallen by the wayside.  I'll go ahead and close.

PR #1538, opened 2020-10-22
Runpathdirs improvements

noresources:
**What does this PR do?**
Fix and improve run path search paths support on gmake actions. Add support for xcode4

Automatically added run path search paths from sibling library is now handled by exporters instead of toolset. The previous behavior may not be accurate for all build system.

Not really, only gmake 1&2 supported runpathdirs and the behavior did not change for these exporters.

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -299,10 +299,20 @@
 
 --
 -- Return a list of decorated rpaths
+-- 
+-- @param cfg
+--    The configuration to query.
+-- @param dirs
+--    List of absolute paths
+-- @param mode
+--    Output mode
+--    - "linker" (default) Linker rpath instructions
+--    - "path" List of path relative to configuration target directory
 --
 
-	function gcc.getrunpathdirs(cfg, dirs)
+	function gcc.getrunpathdirs(cfg, dirs, mode)
 		local result = {}
+		mode = iif (mode == nil, "linker", mode)
 
 		if not (table.contains(os.getSystemTags(cfg.system), "darwin")
 				or (cfg.system == p.LINUX)) then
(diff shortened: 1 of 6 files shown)


starkos:
I'm a little uncomfortable adding run paths to a project that weren't explicitly requested by the user (the sibling projects). This probably needs an API similar to `links()` which allows sibling projects to be specified along with file system paths? I'll let @samsinsane (and anyone else who wants to review) have a say, but I'd vote against this one as-is.

noresources:
Currently, runpathdirs of sibling project are already automatically added. This patch just move this benavior to gmake action instead of gcc/clang code to ensure other actions are not affected. Please consider this PR as a code cleanup rather than a new feature.

As a reminder, this automatic behavior was previously added to fix the fact that binaries and libs generated by gmake actions cannot be moved from where they were built (library installation paths are hardcoded in the library binary on Linux & macOS). 

By the way, my initial proposal (years ago) for this was to add a `linkmode` API to let user choose between default install names (hardcoded) and automatic with runpaths etc.
The current solution was preferred :)

src/tools/gcc.lua
```diff
@@ -299,46 +299,40 @@
...
 -- Return a list of decorated rpaths
+-- 
+-- @param cfg
+--    The configuration to query.
+-- @param dirs
+--    List of absolute paths
+-- @param mode
+--    Output mode
+--    - "linker" (default) Linker rpath instructions
+--    - "patn" List of path relative to configuration target directory
```
samsinsane, line 310:
```suggestion
--    - "path" List of path relative to configuration target directory
```

samsinsane (changes requested):

starkos (changes requested):
Thank you for clarifying, I see what you mean now. Would it be possible to rebase this one against the current master branch? And then I go ahead and approve it.

noresources:
The branch was rebased with a bit of commit history cleanup plus runpath support for xcode4

starkos (approved):
Thanks!

PR #1552, opened 2020-11-03
Fix MSC LTO, runtime, subsystem

JoelLinn:
**What does this PR do?**

- Link-time optimizations now sets correct cl and ld flags.
- Run-Time selection now adheres to `runtime` setting.
- Set subsystem flag for `kind=="WindowedApp"` 

- The different Warning flags should not cause any troubles in any consumers.
- All other changes do in fact change the behavior. They however fix something that was broken or didn't work before.

### DIFF
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -28,6 +28,15 @@
 -- Returns list of C compiler flags for a configuration.
 --
 
+	local function getRuntimeFlag(cfg, isstatic)
+		local rt = cfg.runtime
+		local flag = iif(isstatic, "/MT", "/MD")
+		if (rt == "Debug") or (rt == nil and config.isDebugBuild(cfg))  then
+			flag = flag .. "d"
+		end
+		return flag
+	end
+
 	msc.shared = {
 		clr = {
 			On = "/clr",
@@ -37,6 +46,7 @@
 		},
 		flags = {
 			FatalCompileWarnings = "/WX",
+			LinkTimeOptimization = "/GL",
 			MultiProcessorCompile = "/MP",
 			NoMinimalRebuild = "/Gm-",
 			OmitDefaultLibrary = "/Zl"
@@ -87,11 +97,11 @@
 		},
 		staticruntime = {
 			-- this option must always be emit (does it??)
-			_ = function(cfg) return iif(config.isDebugBuild(cfg), "/MDd", "/MD") end,
+			_ = function(cfg) return getRuntimeFlag(cfg, false) end,
 			-- runtime defaults to dynamic in VS
-			Default = function(cfg) return iif(config.isDebugBuild(cfg), "/MDd", "/MD") end,
-			On = function(cfg) return iif(config.isDebugBuild(cfg), "/MTd", "/MT") end,
-			Off = function(cfg) return iif(config.isDebugBuild(cfg), "/MDd", "/MD") end,
+			Default = function(cfg) return getRuntimeFlag(cfg, false) end,
+			On = function(cfg) return getRuntimeFlag(cfg, true) end,
+			Off = function(cfg) return getRuntimeFlag(cfg, false) end,
 		},
 		stringpooling = {
 			On = "/GF",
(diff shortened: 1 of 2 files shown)


src/tools/msc.lua
```diff
@@ -28,6 +28,15 @@
 -- Returns list of C compiler flags for a configuration.
 --
 
+	local function getruntimeflag(cfg, isstatic)
```
starkos, line 31: Please use camel-case (`getRuntimeFlag`, `isStatic`) for names.

src/tools/msc.lua
```diff
@@ -28,6 +28,15 @@
...
+		local rt = cfg.runtime
+		local flag = iif(isstatic, "/MT", "/MD")
+		if (rt == "Debug") or (rt == nil and config.isDebugBuild(cfg))  then
```
starkos, line 34: Should `(rt == "Debug")` go in `config.isDebugBuild()`? Seems like this should be consistent behavior across tools.

src/tools/msc.lua
```diff
@@ -368,15 +379,15 @@
 		local result = {}
 
 		for _, enable in ipairs(cfg.enablewarnings) do
-			table.insert(result, '/w1"' .. enable .. '"')
```
starkos, line 371: Why is this being changed?

tests/tools/test_msc.lua
```diff
@@ -87,7 +87,8 @@
 	function suite.ldflags_onLinkTimeOptimizations()
 		flags "LinkTimeOptimization"
 		prepare()
-		test.contains("/GL", msc.getldflags(cfg))
+		test.contains("/GL", msc.getcflags(cfg))
```
starkos, line 90: This test suite is for `ldflags`. `cflags` tests should go into their own suite.

tests/tools/test_msc.lua
```diff
@@ -237,11 +238,11 @@
 	end
 
 	function suite.cflags_onSpecificWarnings()
-		enablewarnings { "enable" }
-		disablewarnings { "disable" }
-		fatalwarnings { "fatal" }
+		enablewarnings { "1234" }
+		disablewarnings { "5678" }
+		fatalwarnings { "9753" }
```
starkos, line 243: Why is this being changed?

starkos (changes requested):

starkos:
FYI, if you had split these fixes into separate PRs (LTO, WindowedApp, runtime flags, warning changes) we could have approved the simpler changes. Because you submitted them all as one PR, they are all now going to be held up awaiting changes. The first checkbox on our PR template says "Focus on a single fix or feature" for a reason.

src/tools/msc.lua
JoelLinn: Depends on what the definition of `config.isDebugBuild()` is. You may have a configuration with debug symbols, no optimisations but using the runtime library. I would consider that a debug build. About the other way around, I can't tell.

src/tools/msc.lua
JoelLinn: If the goal is to follow VS it may indeed make sense to reject it, although the MSC docs don't mention quotes.
In my case, I have a custom action/generator which somewhere down the line breaks with quotes - granted, that is not an argument to have it changed in the core.

tests/tools/test_msc.lua
JoelLinn: At least to my understanding, the [MSC docs](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-160) state the warnings are identified by a four-digit number.

tests/tools/test_msc.lua
JoelLinn: I split up the tests

tests/tools/test_msc.lua
JoelLinn: Undid those change.

src/tools/msc.lua
JoelLinn: Undid those changes.

src/tools/msc.lua
JoelLinn: I am still unsure about this. The [description of `isDebugBuild`](https://github.com/premake/premake-core/blob/1aeb39dfcef8d5d9fdb893c95bd6d3c6ad150cb6/src/base/config.lua#L427-L431) hints it chooses the MS runtime version but as stated you may want to have symbols without using the Debug runtime. At least to me, the name of the function and what it does fit very well together.

starkos: Let's leave `isDebugBuild()` alone for now, until there's a more compelling reason to change it.

starkos:
@samsinsane - does this one look reasonable to you, now that some of the cruft has been trimmed out?

@JoelLinn - you'll need to rebase these changes against the latest master branch (okay if you want to wait for review)

samsinsane (approved):

samsinsane:
@starkos Yeah, overall it looks good.

starkos:
Cool. @JoelLinn If you can rebase and squash these changes I'll get it merged. Thanks for the contribution!

JoelLinn:
Done. I don't know what to squash though as there is only one commit...

starkos (approved):

PR #1554, opened 2020-11-04
clang Use `llvm-ar` linker when LTO flag is set.

JoelLinn:
**What does this PR do?**

Changes the static linker from `ar` to `llvm-ar` when the `clang` toolset is used and the `LinkTimeOptimization` flag is set.
Although many times `GNU ar` will also work, for some clang features like `-flto` (link-time optimisation, I only came across this one but could be more) `llvm-ar` is needed.

It should not break or change build-file output for existing projects (that may not have `llvm-ar` installed) since `llvm-ar` is only conditionally used and still defaults to `ar`.

Before this change, it was not possible to use the `LinkTimeOptimization` flag with the clang toolset unless using the "AR=llvm-ar" environment variable.

### DIFF
--- a/modules/gmake/tests/cpp/test_clang.lua
+++ b/modules/gmake/tests/cpp/test_clang.lua
@@ -44,3 +44,20 @@ ifeq ($(config),debug)
 		]]
 	end
 
+	function suite.usesCorrectCompilersAndLinkTimeOptimization()
+		flags { "LinkTimeOptimization" }
+		make.cppConfigs(prj)
+		test.capture [[
+ifeq ($(config),debug)
+  ifeq ($(origin CC), default)
+    CC = clang
+  endif
+  ifeq ($(origin CXX), default)
+    CXX = clang++
+  endif
+  ifeq ($(origin AR), default)
+    AR = llvm-ar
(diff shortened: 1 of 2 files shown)


samsinsane:
Please fill in the template, it's there for a reason.

JoelLinn:
> Please fill in the template, it's there for a reason.

please reopen

tritao:
It seems a bit inconsistent that the linker will be different depending on if LTO is enabled or not.

Is there any reason not to just always use `llvm-ar` when using the `clang` toolset?

JoelLinn:
> Is there any reason not to just always use `llvm-ar` when using the `clang` toolset?

* It may break existing projects since usually llvm packages are not required by clang-packages. (llvm-ar is not installed)
* Afaik (if that counts anything here) cmake also uses plain `ar` first in its makefiles unless certain features are enabled, only then `llvm-ar`, `llvm-ranlib` and friends are emitted. I never looked into when exactly they switched, just noticed it in the makefiles

tritao:
Makes sense, thanks.

samsinsane (approved):

PR #1555, opened 2020-11-05
Added API to disable fast up to date checks

### LINKED ISSUE #1443: Better API for generating PropertyDefinition in VS project files?
Is there better api for generating custom property definitions in vcxproj file? I found old answer on SO https://stackoverflow.com/a/39151157/6119618 and it seems too outdated and too complicated.

Could you, please, elaborate on how to add 
```XML
<PropertyGroup>
    <DisableFastUpToDateCheck>true</DisableFastUpToDateCheck>
</PropertyGroup> 
```
into my project?

nickclark2016:
**What does this PR do?**

Adds DisableFastUpToDateCheck API, which is only useful for Visual Studio projects.  When `fastuptodate` is `Disable`, it adds the relevant XML entry to the globals property group.  Closes #1443.

No changes to existing behavior.  Adds XML entries when `fastuptodate` is set to `Disable`.

API:

```lua
project "name"
    -- To Disable
    fastuptodate "Off"

    -- To Enable
    fastuptodate "On" -- does nothing, added for completeness
```


src/_premake_init.lua
```diff
@@ -430,6 +430,16 @@
 		kind = "string",
 	}
 
+	api.register {
+		name = "fastuptodate",
+		scope = "project",
+		kind = "string",
```
samsinsane, line 436: This should use `boolean`, we already have a concept of `On` and `Off`, it would be confusing to have a random set of APIs use `Enable`/`Disable`.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -133,13 +133,14 @@
...
 			m.windowsTargetPlatformVersion,
+			m.fastUpToDateCheck,
 		}
 	end
 
 	m.elements.globalsCondition = function(prj, cfg)
 		return {
 			m.windowsTargetPlatformVersion,
-			m.xpDeprecationWarning,
+            m.xpDeprecationWarning,
```
samsinsane, line 143: This should be reverted, this change is wrong and unnecessary.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2666,6 +2667,13 @@
 	end
 
 
+	function m.fastUpToDateCheck(prj)
+		if prj.fastuptodate == "Disable" then
+			m.element("DisableFastUpToDateCheck", nil, "true")
```
samsinsane, line 2672: Is this available from VS2010, or was it added in VS2019? If the latter, the above check should include `_ACTION >= "vs2019"`.

samsinsane (changes requested):

modules/vstudio/vs2010_vcxproj.lua
nickclark2016: Good catch, thought I had reverted this.

modules/vstudio/vs2010_vcxproj.lua
nickclark2016: Available since VS2010 I believe.  I've found some only MSDN forum posts about it pretty far back.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2666,6 +2667,13 @@
 	end
 
 
+	function m.fastUpToDateCheck(prj)
+		if prj.fastuptodate == false then
+			m.element("DisableFastUpToDateCheck", nil, "true")
```
samsinsane, line 2672: Having another look at this, this should probably emit any time that `fastuptodate` is set.
```lua
if prj.fastuptodate ~= nil then
  m.element("DisableFastUpToDateCheck", nil, iif(prj.fastuptodate, "false", "true"))
end
```

The idea here is that, if the user is specifying the value then Premake should honour that request regardless of what the default may be - the default might change.

samsinsane (dismissed):

modules/vstudio/vs2010_vcxproj.lua
starkos: Agree with @samsinsane, good catch. If you look at the other switches, "Default" always leaves the toolset's default behavior, while "On" and "Off" set it explicitly.

nickclark2016: I can definitely do this.  Sorry on the slow follow up, work + school has been insane recently.

starkos (changes requested):
I'll return the apology for the slow turnaround. Can you rebase this against the current main branch and I'll go ahead and merge?

nickclark2016:
I hit update. I think that rebases.  If not, feel free to let me know and I'll manually rebase.

starkos (approved):

starkos:
Thanks for the quick follow up!

PR #1569, opened 2020-12-15
Fix --help output when using mixed-case command line options

ratzlaff:
**What does this PR do?**

Prior to this PR, using `--help` would silently fail at iterating over the options table when `newoption` was used to add a trigger that contained capital letters. 

using `--help` should now properly display all user-defined options

### DIFF
--- a/src/base/option.lua
+++ b/src/base/option.lua
@@ -65,6 +65,7 @@
 --
 
 	m.list = {}
+	setmetatable(m.list, _OPTIONS_metatable)
 
 
 --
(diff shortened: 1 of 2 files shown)


starkos (commented):
I do see this, but I don't understand the changes entirely and want to take a look before approving. Not sure where/what that metatable is, and the unit test changes seems like a lot just to test a list iteration (why not put the values into a new list and check that instead of printing it out? Why create a whole new options table instead of using the same approach as what was there?)

starkos:
As above, this is way too heavy-handed for a simple fix. I'm closing it in favor of #1617.

ratzlaff:
No worries here, looks like that is better anyway =)

PR #1589, opened 2021-02-16
os.outputof: add a second argument to select which stream to output

noresources:
local o, e = os.outputof(cmd, streams)

    Where streams could be one of
    - "output" Only return standard output stream content
    - "error" Only return standard error stream content
    - "both" (default) Return both streams content

Allow to filter standard output or error stream content from os.outputof() return value.
Ignoring error output is important when the invoked command may return an exit code 0 (success) but output some content on the error stream which will pollute the text result.

The default behavior follow the previous one (report both streams)

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -456,11 +456,24 @@
 --
 -- Run a shell command and return the output.
 --
+-- @param cmd Command to execute
+-- @param streams Standard stream(s) to output
+-- 		Must be one of 
+--		- "botn" (default)
+--		- "output" Return standard output stream content only
+--		- "error" Return standard error stream content only
+--
 
-	function os.outputof(cmd)
+	function os.outputof(cmd, streams)
 		cmd = path.normalize(cmd)
+		local redirection = " 2>&1"
+		if streams == "output" then
+			redirection = " 2>/dev/null"
+		elseif streams == "error" then
+			redirection = " 2>&1 1>/dev/null"
+		end
 
-		local pipe = io.popen(cmd .. " 2>&1")
+		local pipe = io.popen(cmd .. redirection)
 		local result = pipe:read('*a')
 		local success, what, code = pipe:close()
 		if success then
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -466,11 +466,16 @@
 
 	function os.outputof(cmd, streams)
 		cmd = path.normalize(cmd)
-		local redirection = " 2>&1"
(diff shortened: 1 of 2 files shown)


src/base/os.lua
```diff
@@ -455,12 +455,25 @@
 
 --
 -- Run a shell command and return the output.
+--
+-- @param cmd Command to execute
+-- @param streams Standard stream(s) to output
+-- 		Must be one of 
+--		- "botn" (default)
```
samsinsane, line 462:
```suggestion
--		- "both" (default)
```

src/base/os.lua
```diff
@@ -455,12 +455,25 @@
...
 
-	function os.outputof(cmd)
+	function os.outputof(cmd, streams)
 		cmd = path.normalize(cmd)
+		local redirection = " 2>&1"
+		if streams == "output" then
+			redirection = " 2>/dev/null"
+		elseif streams == "error" then
+			redirection = " 2>&1 1>/dev/null"
+		end
```
samsinsane, line 474: Should this perhaps validate that `streams` is a valid value? I imagine an error would be better than it just falling back to `"both"`?

tests/base/test_os.lua
```diff
@@ -184,8 +184,26 @@
...
+			and os.isdir (_TESTS_DIR)
+		then
+			local ob, e = os.outputof ("ls " .. _TESTS_DIR .. "/base")
+			local oo, e = os.outputof ("ls " .. _TESTS_DIR .. "/base", "output")
+			test.isequal (oo, ob)
+			local s, e = string.find (oo, "test_os.lua")
+			test.istrue(s ~= nil)
+			
+			local o, e = os.outputof ("ls " .. cwd .. "/base", "error")
+			test.istrue(o == nil or #o == 0)
```
samsinsane, line 203: This test should probably validate that using `"error"` actually works, this just confirms that it doesn't get anything from the `"output"` stream.

samsinsane (changes requested):

tests/base/test_os.lua
noresources: When I wrote this patch, I had more relevant tests that works locally but failed in the github actions context. So I added this minimal test. Note that the master branch does not contain any test about `os.outputof` output (only error code)

src/base/os.lua
noresources: I have updated the PR in this way

Issue #1547, opened 2020-10-28
Move documentation into main repository


starkos:
**What problem will this solve?**

The GitHub wiki, where the documentation is currently stored, has been subject to spam and is currently locked down.

Ideally, the documentation would just be a collection of Markdown files in a `docs/` folder, [something like this](https://github.com/starkos/premake-next/tree/master/docs); that would allow us to update the docs right alongside the code and submit it all together in a single pull request. Downside is that we lose GitHub's wiki page search feature.

**What might be a solution?**

Set up an open source documentation website package like [Docusaurus](https://docusaurus.io) or [MkDocs](https://www.mkdocs.org/) to format and serve the documentation.

**What other alternatives have you already considered?**

That's pretty much it. Don't think re-opening the wiki is a viable option.

**Anything else we should know?**

Bonus points if we can serve documentation for different versions of Premake from the same site, i.e 5.x vs. 6.x.

*(You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to requests like these!)*

nickclark2016:
@starkos if you decide to keep with GitHub wiki, I'd be willing to help maintain it as new functionality is added.

samsinsane:
Surely there's a way to deploy to the GitHub wiki in the GitHub actions? Also, maybe deploying the docs to premake.github.io would be better. Assuming the pages are indexed by Google, you can just search `<term> site:premake.github.io` and get all of the pages that reference that term, instead of just the page that contains that term in the title.

starkos:
@samsinsane Hmm…an action that pushes all changes from a folder in the main repo into the wiki would solve the biggest issues, that's true. And it avoids adding another dependency to the project, that's a bonus. The search/navigation isn't nearly as nice as a dedicated package—for sake of comparison: Docusaurus generated docs for [Redux](https://redux.js.org/api/api-reference) and [Babel](https://babeljs.io/docs/en/). And I'm not sure how we'd juggle v5 vs. v6 docs if there are significant changes (which is looking pretty likely)? [Here's how Jest does it](https://jestjs.io/en/versions) (also Docusaurus).

@nickclark2016 I'm going to put a shout out for more reviewers/approvers shortly. Would be very happy to have the help.

redorav:
I actually like the search-as-you-type feature of the current wiki. The example docs you've put forward sure look nice though...

KyrietS:
[...]

PR #1602, opened 2021-03-23
Rework website home & download pages

starkos:
**What does this PR do?**

Reworks the placeholder home and download pages on the new Docusaurus-based website.

No changes to Premake itself.

I still need to rework the header and footer, and move some content from `docs/` to `pages/` such as the help and "who uses" pages. Tracking the work on #1547.

- [ ] ~Add unit tests showing fix or feature works; all tests pass~


website/src/pages/download.js
```diff
@@ -0,0 +1,106 @@
...
+			<Container>
+				<Row>
+					<Column>
+						<h3>Pre-Built Binaries</h3>
+						<p>Binaries simply need to be unpacked and placed somewhere on the system search path or any other convenient location.</p>
+						<p>
+							<DownloadLink arch="windows" /><br />
+							<DownloadLink arch="linux" /><br />
+							<DownloadLink arch="macos" /><br />
+						</p>
```
KyrietS, line 68: `<div>` cannot appear as a descendant of `<p>`. You will see this error when you go to download page and open console in your browser. It's not react error. Website is compiled correctly but accorting to W3C you cannot put *block elements* inside `<p>` and React is kind enough to report this in the console.

My suggestion here is to change this paragraph with `DownloadLink`s into list `<ul>`, `<li>` then add a class to `<ul>`, e.g. "download-links" and style it. (then you can remove "download-link" class from `DownloadLink`). Unlike `<p>`, lists can contain *divs* and other *block elements*.

* To remove list bullets use: `list-style: none;`
* You can also style each item in the list: `.download-links > li { /*...*/ }`

website/src/pages/download.js
```diff
@@ -0,0 +1,106 @@
...
+					<Column>
+						<h3>Build It Yourself</h3>
+						<p>
+							The source code package includes project files for all supported toolsets.
+							See <Link to="https://github.com/premake/premake-core/blob/master/BUILD.txt">BUILD.txt</Link> for
+							build instructions.
+						</p>
+						<p>
+							<DownloadLink arch="src" />
+						</p>
```
KyrietS, line 79: Same problem as above. I suggest using list with one item. Left padding is natural in lists. You don't have to create it artificially for every `DownloadLink`.

website/src/pages/download.js
```diff
@@ -0,0 +1,106 @@
...
+		filename = 'windows.zip';
+		label = 'Windows';
+		break;
+	}
+	return (
+		<div className="download-link">
+			<Link to={`https://github.com/premake/premake-core/releases/download/v${LATEST_VERSION}/premake-${LATEST_VERSION}-${filename}`}>
+				<b>{label}</b>
+			</Link>
+		</div>
```
KyrietS, line 37: I don't think that using `div` here and styling it in this way is a good idea. I suggest removing `div` so the  `Link` becomes the top element. To understand my point of view please look at my comments below.

website/src/pages/index.js
```diff
@@ -1,104 +1,79 @@
...
+			</Link>
+		</div>
+	</Container>;
+
+
+const Feature = ({ title, children }) =>
+	<div className="col col--4">
+		<h3>{title}</h3>
+		<div>{children}</div>
+	</div>;
```
KyrietS, line 33: I suggest using `<Column> </Column>` here. It's already imported but it's not used (yet).

KyrietS (changes requested):
> I still need to rework the header and footer, and move some content from docs/ to pages/ such as the help and "who uses" pages.

Just a reminder that you can create simple pages in Markdown. Docusaurus will take care of it. You don't have to use JS and React in every page.

starkos:
Good feedback, thanks! I was going back and forth on the layout of the download links and didn't clean things up properly; I've switched them to `<ul>` as suggested. I'm hoping/intending to loop back and associate icons with each one at some point (why I wrapped them in `<div>`).

Console errors are gone. I also backed out my formatting change to `docusaurus.config.js` to avoid conflicts with your commit.

KyrietS (approved):
Thanks for the changes you've made. It looks great now. Those `DownloadLink` lists still can be simplified and refactored but it's good enough to merge this commit in my opinion.

And I wouldn't worry about conflicts in `docusaurus.config.js`. I can always do rebase and force-push to my branch :-)

PR #1619, opened 2021-04-16
Xcode embed libraries

nickgravelyn:
This is a continuation of #1524. I apparently deleted my fork at some point so in order to address the PR feedback I had to re-fork and open a new PR. My apologies for making this harder to track as a result.

0e93130 and 4ed7424 are just taken directly from #1524. 626eedb is the API change requested by @starkos in #1524. The tests all passed for me locally and I was able to generate an Xcode project that did the right thing so the changes still appear to work for me.

### DIFF
--- a/modules/xcode/_preload.lua
+++ b/modules/xcode/_preload.lua
@@ -47,9 +47,15 @@
 	}
 
 	p.api.register {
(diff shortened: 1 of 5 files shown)


starkos (approved):
Thanks for making those changes!

PR #1627, opened 2021-05-17
Enable pchheader for Codelite.

### LINKED ISSUE #1621: Precompiled headers not working with codelite generator
When making a codelite project from the generator the precompiled headers are not being set in the project file, everything else is there.

Under the project settings -> compiler -> Pre Compiled Headers    Nothing is filled out for the header

I tried to build it for Visual Studio to make sure the configuration file was correct and sure enough it did set the header in that project so I know I have it set up correctly.  (note I don't have Visual Studio I was just looking at the output)

In the configuration file set pchheader "all.hpp" and I also made a source for and set that to pchsource "all.cpp"  Both are in the root directory of the project.

I'm using Premake5 on Arch Linux.

I attached my test solution I was using to see if premake was something I was going to use. 

[example.tar.gz](https://github.com/premake/premake-core/files/6345069/example.tar.gz)

Jarod42:
**What does this PR do?**

Add missing support of pchheader for Codelite generator:

close #1621

bug fixes of Codelite generator

I also added pch to my sample tests in https://github.com/Jarod42/premake-sample-projects
My check there is partially broken as even if Codelite actually failed to build, test is marked as successful :-(
but log show the issue.

### DIFF
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -188,43 +188,6 @@
 		}
 	end
 
-	local function getpch(cfg)
-		-- If there is no header, or if PCH has been disabled, I can early out
-		if not cfg.pchheader or cfg.flags.NoPCH then
-			return nil
-		end
-
-		-- Visual Studio requires the PCH header to be specified in the same way
-		-- it appears in the #include statements used in the source code; the PCH
-		-- source actual handles the compilation of the header. GCC compiles the
-		-- header file directly, and needs the file's actual file system path in
-		-- order to locate it.
-
-		-- To maximize the compatibility between the two approaches, see if I can
-		-- locate the specified PCH header on one of the include file search paths
-		-- and, if so, adjust the path automatically so the user doesn't have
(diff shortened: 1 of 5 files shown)


modules/codelite/codelite_project.lua
```diff
@@ -188,6 +188,42 @@
 		}
 	end
 
+	local function getpch(cfg)
+		-- If there is no header, or if PCH has been disabled, I can early out
```
Jarod42, line 192: That function is similar to the one from gmake/gmake2.
Not sure where common code should be factorized.

starkos: If the code is exactly the same, you could `require()` in the gmake module here and then just call that implementation directly? Or it could be moved to a helper method in `gcc.lua` perhaps?

Jarod42: Moved in gcc.

starkos (approved):

starkos:
Thanks for the contribution!

PR #1629, opened 2021-05-20
Added support for Premake on macOS universal binary

tempura-sukiyaki:
**What does this PR do?**

This change allows Premake to be built for Arm64/x86_64 universal binaries on macOS.

Premake runs on Arm64 on Macs using the Apple M1 processor.
There are no other changes.

Building universal binaries requires Xcode Command Line Tools version 12.2 or later.


samsinsane (changes requested):
The main issue with supporting a universal binary for macOS is that supporting the M1 requires macOS 11.0 minimum. Should we continue to build an x86_64 only binary with whatever version of macOS we target, but also build a universal binary with an 11.0 requirement?

tempura-sukiyaki:
It may be a matter of opinion whether Premake needs to support universal binaries right now or not.
As long as Rosetta 2 is supported, I don't think there is any need to rush.
However, I think it will have to be done eventually.

starkos:
ARM has definitely got momentum; we're probably going to have support it for all platforms soon. I don't want to require macOS 11.0 yet, so this will need to be a switch of some kind. 

What about a new command line flag like `--arm` or `--arch=arm` to control the binary type? Something not tied to macOS so we can support ARM Linux & Windows too?

tempura-sukiyaki:
I wrote the source code to try it out.
How does it look like this?

https://github.com/premake/premake-core/compare/master...tempura-sukiyaki:premake-arch

However, I do not have a Linux environment, so I have not been able to verify that this works correctly on Linux.

In Visual Studio, I think it might be better to ignore this option and always include all platforms.

https://github.com/premake/premake-core/compare/master...tempura-sukiyaki:premake-arch2

samsinsane (approved):

nickclark2016:
What's the status on this PR?

PR #1631, opened 2021-05-26
Sort `buildinputs` and `buildoutputs` on xcode4

tempura-sukiyaki:
**What does this PR do?**

Fixed a issue where xcodeproj would be modified even if the Premake script was not modified if multiple paths were specified for `buildinputs` and/or `buildoutputs`.

The above issue was caused by an indeterminate sort order of Lua's key-value tables, so that part has been fixed.

None.

### DIFF
--- a/modules/xcode/tests/test_xcode_project.lua
+++ b/modules/xcode/tests/test_xcode_project.lua
@@ -1332,15 +1332,15 @@
 			);
 			inputPaths = (
 				"file.a",
+				"file.3",
 				"file.1",
 				"file.2",
-				"file.3",
 			);
 			name = "Build \"file.a\"";
 			outputPaths = (
-				"file.4",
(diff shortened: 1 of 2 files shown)


samsinsane:
Do the other generators sort these too? We should try to be consistent.

Also, is there a case where a user wouldn't want the inputs to be sorted, but instead in the order they specified?

starkos (changes requested):
Full agreement that the order needs to be deterministic, but think the ideal option is to output them in the same order in which they appear in the source script. Simple fix: just replace `table.insertsorted()` with `table.insert()` and adjust the tests. If you agree and can make that change, we can get this one merged. Thanks for taking the time to tackle this!

samsinsane (approved):

starkos (approved):

PR #1654, opened 2021-06-22
xcode4: Fix missing link of sibling project with custom targetextension

noresources:
When a node category cannot be determined by extension,
use the configuration kind when available.

Add unit test to illustrate the case.

**What does this PR do?**
Project with a custom targetextension were not linked to other projects 
because the strategy used to determine Xocde node type was uniquely based
on node name extension.
This patch adds a "last chance" fallback based on attached configuration kind

Fix a edge case bug. Should not change the behavior for most common cases.

I kept the previous behavior as the main strategy to avoid breaking changes.

### DIFF
--- a/modules/xcode/xcode_common.lua
+++ b/modules/xcode/xcode_common.lua
@@ -46,6 +46,8 @@
 		}
 		if node.isResource then
 			return "Resources"
+		elseif node.cfg and (node.cfg.kind == p.SHAREDLIB or node.cfg.kind == p.STATICLIB) then
+			return "Frameworks"
 		end
 		return categories[path.getextension(node.name)]
 	end
--- a/modules/xcode/tests/test_xcode_dependencies.lua
+++ b/modules/xcode/tests/test_xcode_dependencies.lua
@@ -198,6 +198,24 @@
 		]]
 	end
 
+function suite.PBXFrameworksBuildPhase_ListsDependencies_OnSharedLibWithTargetExtension()
+		kind "SharedLib"
+		targetextension ".plugin"
+		prepare()
+		xcode.PBXFrameworksBuildPhase(tr)
+		test.capture [[
+/* Begin PBXFrameworksBuildPhase section */
+		9FDD37564328C0885DF98D96 /* Frameworks */ = {
+			isa = PBXFrameworksBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				6B7205267D294518F2973366 /* libMyProject2-d.plugin in Frameworks */,
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+		};
+/* End PBXFrameworksBuildPhase section */
+		]]
+	end
 ---------------------------------------------------------------------------
 -- PBXCopyFilesBuildPhaseForEmbedFrameworks tests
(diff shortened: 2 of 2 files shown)


modules/xcode/xcode_common.lua
```diff
@@ -47,7 +47,14 @@
 		if node.isResource then
 			return "Resources"
 		end
-		return categories[path.getextension(node.name)]
+		local category = categories[path.getextension(node.name)]
+		if (not category)
+			and node.cfg
+			and (node.cfg.kind == p.SHAREDLIB or node.cfg.kind == p.STATICLIB)
+		then
+			category = "Frameworks"
```
samsinsane, line 55: Perhaps this should be treated like an early return like `isResource`? Is there a situation where a static or shared library wouldn't be in the `Frameworks` category?
```lua
if node.isResource then
  return "Resources"
elseif node.cfg and (node.cfg.kind == p.SHAREDLIB or node.cfg.kind == p.STATICLIB) then
  return "Frameworks"
else
  return categories[path.getextension(node.name)]
end
```

modules/xcode/xcode_common.lua
```diff
@@ -47,7 +47,14 @@
 		if node.isResource then
 			return "Resources"
 		end
-		return categories[path.getextension(node.name)]
+		local category = categories[path.getextension(node.name)]
+		if (not category)
+			and node.cfg
+			and (node.cfg.kind == p.SHAREDLIB or node.cfg.kind == p.STATICLIB)
```
samsinsane, line 53: I don't think we split if statements over multiple lines like this anywhere else in the code base - this should probably just be on a single line for consistency.

samsinsane (dismissed):

modules/xcode/xcode_common.lua
noresources: Sorry, I have a low vision and I tend to write code with small line length.

modules/xcode/xcode_common.lua
noresources: As I write in the PR, I kept the extension-based test first to avoid possible breaking changes. However, I agree your suggestion is probably the correct one. If everyone is ok with the early return, I will change the code that way.

samsinsane: @starkos thoughts on using the early return instead of the fallback? I'm happy with either for what it's worth.

starkos: I agree on the early fallback.

modules/xcode/xcode_common.lua
starkos: I'm pretty sure I've used this style elsewhere in the past (though it may have been refactored away over time). I'm okay with the multiple lines in this case.

noresources:
Code updated with early return and a single-line "if"

starkos (approved):
This looks good to me. Would you mind rebasing it against or merging it with the latest main branch? It should be trivial merge, but a big line ending fix just landed which might cause issues.

noresources:
> This looks good to me. Would you mind rebasing it against or merging it with the latest main branch? It should be trivial merge, but a big line ending fix just landed which might cause issues.
Branch was rebased (twice) without issues

PR #1661, opened 2021-06-30
Add frameworkdirs support to gmake and gmake2 with gcc/clang toolsets…

noresources:
… on macOS systems

* Add new optional parameter to toolset.getincludedirs(dirs, sysdirs, frameworkdirs)
* Translate frameworkdirs to -F<path> build & linker flags
* Add tests

It should be easy to update CodeLite too but I can't really test it. 


src/tools/gcc.lua
```diff
@@ -284,12 +284,20 @@
...
+	function gcc.getincludedirs(cfg, dirs, sysdirs, frameworkdirs)
 		local result = {}
-		for _, dir in ipairs(dirs) do
+		for _, dir in ipairs(dirs or {}) do
 			dir = project.getrelative(cfg.project, dir)
 			table.insert(result, '-I' .. p.quoted(dir))
 		end
+
+		if (table.contains(os.getSystemTags(cfg.system), "darwin")
+			and cfg.frameworkdirs ~= nil and #cfg.frameworkdirs) then
```
samsinsane, line 295: This should be on one line. This isn't using the right variable, shouldn't be using `cfg.` for `frameworkdirs`. Also, this is not consistent with the other `dirs` blocks by checking the value of `cfg.frameworkdirs`/`frameworkdirs`.

src/tools/gcc.lua
```diff
@@ -284,12 +284,20 @@
...
+			for _, dir in ipairs(frameworkdirs or {}) do
+				table.insert(result, '-F' .. p.quoted(dir))
```
samsinsane, line 297: This isn't consistent with the other blocks, the others use `project.getrelative()` on `dir`, why doesn't this one?

src/tools/gcc.lua
```diff
@@ -284,12 +284,20 @@
 -- Decorate include file search paths for the GCC command line.
 --
 
-	function gcc.getincludedirs(cfg, dirs, sysdirs)
+	function gcc.getincludedirs(cfg, dirs, sysdirs, frameworkdirs)
 		local result = {}
-		for _, dir in ipairs(dirs) do
+		for _, dir in ipairs(dirs or {}) do
```
samsinsane, line 289: Is this change necessary? Is the `or {}` required on any of the blocks in this function?

src/tools/gcc.lua
```diff
@@ -481,6 +489,13 @@
 			table.insert(flags, '-L' .. p.quoted(dir))
 		end
 
+		if (table.contains(os.getSystemTags(cfg.system), "darwin")
+			and cfg.frameworkdirs ~= nil and #cfg.frameworkdirs) then
```
samsinsane, line 493: This should be on one line. Also, this is not consistent with the other `dirs` blocks by checking the value of `cfg.frameworkdirs`.

samsinsane (changes requested):

src/tools/gcc.lua
noresources: Ok for the "one-line".

The use of cfg. is a copy-paste-is-evil mistake (from `gcc.getLibraryDirectories(cfg)`), that I previously fixed and re-introduced while rebasing for this PR...

The test on frameworkdirs value is made to minimize changes in code. Otherwise it will
require to add `, {}` on all `toolset.getincludedirs()` calls. I suggest to keep it for now.

src/tools/gcc.lua
noresources: During tests I always used absolute path so i didn't catch my attention. I will made more tests to ensure relative paths are accepted by the -F option

src/tools/gcc.lua
noresources: No, ti was (probably) made due to an invalid call of this function in a test. Test will be fixed

src/tools/gcc.lua
noresources: Ok for the "one line". Here, `cfg.frameworkdirs` is always valid so I will remove the unnecessary tests

src/tools/gcc.lua
samsinsane: > The test on frameworkdirs value is made to minimize changes in code. Otherwise it will
require to add `, {}` on all `toolset.getincludedirs()` calls. I suggest to keep it for now.

Sorry to be pedantic, but isn't that why you have the `or {}` inside? Which is how the `sysdirs` loop works too. Checking if `frameworkdirs` is not `nil` isn't necessary because you're already doing `ipairs(frameworkdirs or {})`?

samsinsane (approved):

modules/gmake/gmake_cpp.lua
```diff
@@ -521,7 +521,7 @@ end
 
 
 	function make.includes(cfg, toolset)
-		local includes = toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs)
+		local includes = toolset.getincludedirs(cfg, cfg.includedirs, cfg.sysincludedirs, cfg.frameworkdirs)
```
Jarod42, line 524: Should probably be done for all exporters.

modules/gmake2/gmake2_cpp.lua
```diff
@@ -531,8 +531,8 @@
 			end
 		end
 
-		if fcfg.includedirs or fcfg.sysincludedirs then
-			local includes = toolset.getincludedirs(cfg, fcfg.includedirs, fcfg.sysincludedirs)
+		if fcfg.includedirs or fcfg.sysincludedirs or fcfg.frameworkdirs then
+			local includes = toolset.getincludedirs(cfg, fcfg.includedirs, fcfg.sysincludedirs, cfg.frameworkdirs)
```
Jarod42, line 535:
```suggestion
			local includes = toolset.getincludedirs(cfg, fcfg.includedirs, fcfg.sysincludedirs, fcfg.frameworkdirs)
```

src/tools/gcc.lua
```diff
@@ -284,12 +284,20 @@
...
-	function gcc.getincludedirs(cfg, dirs, sysdirs)
+	function gcc.getincludedirs(cfg, dirs, sysdirs, frameworkdirs)
 		local result = {}
 		for _, dir in ipairs(dirs) do
 			dir = project.getrelative(cfg.project, dir)
 			table.insert(result, '-I' .. p.quoted(dir))
 		end
+
+		if (table.contains(os.getSystemTags(cfg.system), "darwin") and frameworkdirs ~= nil) then
+			for _, dir in ipairs(frameworkdirs or {}) do
```
Jarod42, line 295: Below doesn't have `or {}`.

Jarod42 (changes requested):

modules/gmake/gmake_cpp.lua
noresources: This "miss" was intentional. I don't think resflags needs framework dirs.

src/tools/gcc.lua
noresources: `and frameworkdirs ~= nil` is probably the one to remove here

modules/gmake/gmake_cpp.lua
Jarod42: For resources, it might be OK, but I would pass `nil` explicitly.

For gmake2, only for "per-file" configuration has been done.
Missing for Codelite generator...

noresources: I am not against an explicit nil, however there is currently no `toolset.getincludedirs(cfg, inc, nil)` when sysdir is not mandatory.
About gmake2, Ok, I didn't understand your first message quite well. This will be fixed.
About CodeLite, I said in the PR message I am not using CodeLite, have no knowledge about it and I can't test it.

Jarod42: `sysdir` is missing only for resources flags (I wonder it is not a bug BTW).

You should add `frameworkdirs` for all non-resources flags call (for all generators).

It is the toolset which "manages" it. (as msvc which does nothing).
if generator doesn't use directly compiler flags, it iterates directly on `cfg.includedirs` or similar, and then indeed, it would require knowledge of the generator.

For test, I started https://github.com/Jarod42/premake-sample-projects to test with projects with multiple generators.
Nothing platform specific yet, as `frameworks` 

If you can provide sample test project (usable in github action) (a variant of project-01 which test includedirs), I can add it.

noresources:
Hello,
I finally had some spare time and motivation to run a macOS VM, install CodeLite and test frameworkdirs  support.
I also add the runpathdirs support to it (one line change).
Finally I rebased/cleanup commits on top of master.
Hope this will allow to close this PR.

modules/gmake/gmake_cpp.lua
starkos: @Jarod42 It is okay if contributors only want to target a single toolset. Not everyone has access to every toolset; allowing people to contribute where they can is more important. And please do not confuse things by asking people to provide you with tests for a separate repository; unit tests are all that we require at this time. Thanks!

starkos (approved):
Approving, but will let @samsinsane make the call on the merge since they've been more involved in the review.

starkos:
Oh shoot, I see @samsinsane already approved…merging…

modules/gmake/gmake_cpp.lua
Jarod42: @starkos: What bothers me is that, then projects are generator-specific without any errors/warnings for users when used with incompatible generators.

Sorry, I didn't want to confuse people with extra tests in another repository.

Notice that you have tests just for gcc, but not for gmake/gmake2.

starkos: > What bothers me is that, then projects are generator-specific without any errors/warnings for users when used with incompatible generators.

The alternative is that people don't contribute. We've been down that path before. I do agree that we need to do a better job of surfacing compatibility information, and the documentation seems an appropriate place to do that.

@Jarod42 If you have general comments on broad topics ("features should be implemented for all generators") please use the provided Discussions rather than cluttering up the PR review notes. I have no longer have any idea what change was requested by this thread, and if it has been done or not.

src/tools/gcc.lua
starkos: I believe this one has been resolved?

PR #1672, opened 2021-07-26
C# Symbol Premake -> VS Mapping Changes

### LINKED ISSUE #1670: Visual Studio C# Project - Symbols "Off" seting debugging information to "pdbonly"
I have the following filters set inside a **premake5** project declaration for a C# SharedLib project:

```lua
filter "configurations:Debug"
        symbols "Off"
        defines { "DEBUG", "TRACE" }

filter "configurations:Release"
        defines { "TRACE" }
        optimize "On"
```

This is part of the resulting vs2019 project file:
```xml
<PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
    <PlatformTarget>AnyCPU</PlatformTarget>
    <DebugType>pdbonly</DebugType>
    <Optimize>false</Optimize>
    <OutputPath>..\..\Assemblies\</OutputPath>
    <BaseIntermediateOutputPath>obj\Any CPU\Debug\</BaseIntermediateOutputPath>
    <IntermediateOutputPath>$(BaseIntermediateOutputPath)</IntermediateOutputPath>
    <DefineConstants>DEBUG;TRACE</DefineConstants>
    <ErrorReport>prompt</ErrorReport>
    <WarningLevel>4</WarningLevel>
  </PropertyGroup>
```
As you can see `<DebugType>` has been set to `pdbonly` instead of `none`.
I have tried with the other `symbols` values, bot none disables it completelly.

Even if its not a bug and its intentional, sometimes completelly disabling debugging information is needed, and there should be an option to do so.

nickclark2016:
**What does this PR do?**

This PR closes #1670.  It allows for `none` to be set for the debug symbol value of Visual Studio C# projects.

Previous Mapping

| Premake Symbol Value  | Visual Studio C# Project Value |
|-----------------------|--------------------------------|
| Off, Full, FastLink, Default | pdbonly |
| On | full |

New/Proposed Mapping

| Premake Symbol Value  | Visual Studio C# Project Value |
|-----------------------|--------------------------------|
| Off | none |
| On, FastLink | pdbonly |
| Full | full |
| Default | portable |

This changes the behavior of the `Off`, `On`, `Full`, and `Default` C# symbol outputs.

The rational behind the change of `Default` to map to `portable` is because VS2019 defaults to `portable` for the debug symbol output.

- [x] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes (Not applicable, no new API added)

### DIFF
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -278,11 +278,27 @@
 --
 
 	function dotnetbase.debugProps(cfg)
-		if cfg.symbols == p.ON then
-			_p(2,'<DebugSymbols>true</DebugSymbols>')
-			_p(2,'<DebugType>full</DebugType>')
+		if _ACTION >= "vs2019" then
+			if cfg.symbols == "Full" then
+				_p(2,'<DebugType>full</DebugType>')
+				_p(2,'<DebugSymbols>true</DebugSymbols>')
+			elseif cfg.symbols == p.OFF then
+				_p(2,'<DebugType>none</DebugType>')
+				_p(2,'<DebugSymbols>false</DebugSymbols>')
+			elseif cfg.symbols == p.ON or cfg.symbols == "FastLink" then
+				_p(2,'<DebugType>pdbonly</DebugType>')
+				_p(2,'<DebugSymbols>true</DebugSymbols>')
+			else
+				_p(2,'<DebugType>portable</DebugType>')
+				_p(2,'<DebugSymbols>true</DebugSymbols>')
+			end
 		else
-			_p(2,'<DebugType>pdbonly</DebugType>')
+			if cfg.symbols == p.ON then
+				_p(2,'<DebugSymbols>true</DebugSymbols>')
+				_p(2,'<DebugType>full</DebugType>')
+			else
+				_p(2,'<DebugType>pdbonly</DebugType>')
+			end
 		end
 		_p(2,'<Optimize>%s</Optimize>', iif(config.isOptimizedBuild(cfg), "true", "false"))
 	end
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -12,6 +12,7 @@ return {
 	"cs2005/test_compiler_props.lua",
 	"cs2005/test_no_warn.lua",
 	"cs2005/test_debug_props.lua",
+	"cs2005/test_debug_props_2019.lua",
 	"cs2005/test_files.lua",
 	"cs2005/test_icon.lua",
 	"cs2005/test_netcore.lua",
--- a/modules/vstudio/tests/cs2005/test_debug_props_2019.lua
+++ b/modules/vstudio/tests/cs2005/test_debug_props_2019.lua
@@ -1,0 +1,130 @@
+--
+-- tests/actions/vstudio/cs2005/test_debug_props_2019.lua
+-- Test debugging and optimization flags block of a Visual Studio 2019+ C# project.
+-- Copyright (c) 2012-2021 Jason Perkins and the Premake project
+--
+
+	local p = premake
+	local suite = test.declare("vstudio_cs2005_debug_props_2019")
+	local cs2005 = p.vstudio.cs2005
+	local dn2005 = p.vstudio.dotnetbase
+	local project = p.project
+
+
+--
+-- Setup and teardown
+--
+
+	local wks, prj
+
+	function suite.setup()
+		p.action.set("vs2019")
+		wks, prj = test.createWorkspace()
+	end
+
+	local function prepare()
+		local cfg = test.getconfig(prj, "Debug")
+		dn2005.debugProps(cfg)
+	end
+
+
+--
+-- Check the handling of the Symbols flag.
+--
+
+	function suite.debugSymbols_onNoSymbolsFlag()
+		prepare()
+		test.capture [[
+		<DebugType>portable</DebugType>
+		<DebugSymbols>true</DebugSymbols>
+		<Optimize>false</Optimize>
+		]]
+	end
+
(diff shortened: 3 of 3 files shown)


starkos:
Have you confirmed the existence of "portable" for older versions of Visual Studio and the various .NET project types? I'm pretty sure at one point it was only supported for .NET Core?

nickclark2016:
Good question.  From what I found about it, it appears to be fairly well supported at this point, with the following notable exceptions:
* Older VS debugger versions (before VS 2015 Update 2)
* .NET Framework versions 4.7.1 and earlier
* Some code analysis tools
* Some symbol servers
* Post compilation steps that consume the non-portable/Windows PDBs
* Some .NET decompilers
* Microsoft DIA

[Source](https://github.com/dotnet/core/blob/e6049bb60307d987e044c39a106e0d6cf98857a3/Documentation/diagnostics/portable_pdb.md)

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -278,11 +278,18 @@
...
-			_p(2,'<DebugSymbols>true</DebugSymbols>')
+		if cfg.symbols == "Full" then
 			_p(2,'<DebugType>full</DebugType>')
+			_p(2,'<DebugSymbols>true</DebugSymbols>')
+		elseif cfg.symbols == p.OFF then
+			_p(2,'<DebugType>none</DebugType>')
+			_p(2,'<DebugSymbols>false</DebugSymbols>')
+		elseif cfg.symbols == "Default" then
+			_p(2,'<DebugType>portable</DebugType>')
+			_p(2,'<DebugSymbols>true</DebugSymbols>')
```
samsinsane, line 289: I disagree with this. `Default` is supposed to be what the editors default is - not what the editors default is at the time of writing the code.

nickclark2016: Is there a good way to determine what that is?

samsinsane: Don't emit anything and let VS default it? To me, that's what this value is supposed to mean. I did see the comment from @starkos in the linked issue after I commented this though. I personally disagree with the approach of emitting the project templates by default in Premake - it's frustrating having some default baked in that you can't turn off.

Furthermore, this PR will create an inconsistency around what `symbols "Default"` means. Look at what `vcxproj` does, `Default` doesn't emit anything. Look at what the `msc` and `dotnet` tools do, both only support `On`, `Default` doesn't emit anything.

Perhaps a better way to set a "default" value would be using this system instead?
https://github.com/premake/premake-core/blob/6221a49b7eecd7b10b927567efbd9a7545ddf397/src/_premake_init.lua#L1824-L1894

This isn't a hill I'm going to die on or anything, so if you guys feel strongly about emitting `portable` I'm happy for that to happen.

starkos:
> Don't emit anything and let VS default it? To me, that's what this value is supposed to mean.

I've always interpreted it as "match what you'd get by default in a newly created project". For a command line tool that just means don't set the value and let it use its internal default. For IDEs, I've tried to match what you would get if ran `File > New Project...`, since that's what I think most people would consider the "default" behavior. This might differ from not setting the switch at all.

But I do agree with @samsinsane that using "portable" as a blanket fault smells wrong. IMHO, to identify the correct value you'd have to go back and run `File > New Project...` in each major version of the IDE and see what you get. Presumably the logic that is there today reflects what we discovered in the past, so my recommendation would be to see what VS2019 puts in a new project and copy that, making sure not to change the value for any older versions of VS.

My $0.02.

modules/vstudio/vs2005_dotnetbase.lua
starkos: > Don't emit anything and let VS default it? To me, that's what this value is supposed to mean.

I've always interpreted it as "match what you'd get by default in a newly created project". For a command line tool that just means don't set the value and let it use its internal default. For IDEs, I've tried to match what you would get if ran `File > New Project...`, since that's what I think most people would consider the "default" behavior. This might differ from not setting the switch at all.

But I do agree with @samsinsane that using "portable" as a blanket fault smells wrong. IMHO, to identify the correct value you'd have to go back and run `File > New Project...` in each major version of the IDE and see what you get. Presumably the logic that is there today reflects what we discovered in the past, so my recommendation would be to see what VS2019 puts in a new project and copy that, making sure not to change the value for any older versions of VS.

My $0.02. Also thanks and h/t @nickclark2016 for digging into this.

nickclark2016: Works for me.  I'll go back and pull out that so we emit nothing on default and update the tests accordingly.  Appreciate your inputs @samsinsane @starkos.

nickclark2016: So course of action here is to say if action >= VS2019, then use portable (the default), else continue to use pdbonly? That works for me.  I appreciate the input :)

nickclark2016: So just to follow up, I threw the old code behind an else block, and used the new code for VS2019 forward.

starkos (approved):
Thanks for that!

starkos:
@nickclark2016 Nitpick, but please continue to stage your PR branches on your own fork of premake-core, rather than pushing the branch to premake-core itself. That way our release scripts can correctly pull out your username from the auto-generated commit message (and it avoids clutter and old branches hanging around on premake-core). Thanks!

nickclark2016:
> 
> 
> @nickclark2016 Nitpick, but please continue to stage your PR branches on your own fork of premake-core, rather than pushing the branch to premake-core itself. That way our release scripts can correctly pull out your username from the auto-generated commit message (and it avoids clutter and old branches hanging around on premake-core). Thanks!

Of course. I didn't realize until after putting in the PR that I had my remote set to premake and not my fork.  I'll be more careful on this next time :)

PR #1700, opened 2021-08-27
Start of GNU Makefile exporter for 6.x

nickclark2016:
**What does this PR do?**

Starts implementation of the GNU Makefile exporter for the 6.x branch.

N/A

TODO Before Merge:
- [x] Unit Testing

Follow Up Needs
- Support for project types other than executables
- Support for linking external libraries
- Dependencies


starkos:
Trying to make some time to sit and review this…

nickclark2016:
Sounds good.  I'll be driving the next few days, so I'll get around to responses next weekend.

starkos (commented):
On the whole, I'd say this looks pretty good. There are minor nitpicks like…

```make
OBJDIR = bin-int/Premake/Debug  # why "bin-int" and not the previous default "obj"?
```

…but nothing that won't come out in the wash. So that's great, thanks for putting it together!

Before pushing on ahead here though, I really think it would be worth thinking through what the v6 makefiles should look like, while we have this opportunity to break big things. Should/could we be able to build Windows/Linux/macOS out of a single makefile (pretty sure CodeBlocks & CodeLite support this idea too)? How should/could per-file build settings be expressed? What about custom build rules? Would it be possible to _not_ inherit project settings into the build configurations, and instead output all of the project-level stuff in a leading block, and then only break out the values that actually differ in the build configuration if-else-end?

Because Makefiles are so open-ended, we have the opportunity to do things different/better. These bigger questions never really got properly addressed in v5; I want to make sure they are looked at for v6 before we dig in too deep to porting the previous implementations. Does that make sense?

modules/gmake/src/sln.lua
```diff
@@ -0,0 +1,251 @@
+---
+-- GNU Makefile solution exporter.
```
starkos, line 2: I realize you were just lifting from the existing VS exporter, but for makefiles these should be called "workspace" and "wks" rather than "solution" and "sln", which are VS specific terms. Same with the unit test names, etc.

modules/gmake/tests/sln/sln_help_rule_tests.lua
```diff
@@ -0,0 +1,89 @@
...
+help:
+	@echo "Usage: make [config=name] [target]"
+	@echo ""
+	@echo "Targets:"
+	@echo "	all [default]"
+	@echo "	clean"
+	@echo "	help [Prints this message]"
+	@echo "	MyProject"
+	@echo ""
+	@echo "For more information, see https://github.com/premake/premake-core/wiki"
```
starkos, line 87: This will need to get changed in both v5 & v6, should be https://premake.github.io now.

nickclark2016:
> On the whole, I'd say this looks pretty good. There are minor nitpicks like…
> 
> ```makefile
> OBJDIR = bin-int/Premake/Debug  # why "bin-int" and not the previous default "obj"?
> ```
> 
> …but nothing that won't come out in the wash. So that's great, thanks for putting it together!
> 
> Before pushing on ahead here though, I really think it would be worth thinking through what the v6 makefiles should look like, while we have this opportunity to break big things. Should/could we be able to build Windows/Linux/macOS out of a single makefile (pretty sure CodeBlocks & CodeLite support this idea too)? How should/could per-file build settings be expressed? What about custom build rules? Would it be possible to _not_ inherit project settings into the build configurations, and instead output all of the project-level stuff in a leading block, and then only break out the values that actually differ in the build configuration if-else-end?
> 
> Because Makefiles are so open-ended, we have the opportunity to do things different/better. These bigger questions never really got properly addressed in v5; I want to make sure they are looked at for v6 before we dig in too deep to porting the previous implementations. Does that make sense?

Yeah, the OBJDIR was out of habit for my personal projects. Will change to `obj`.  For printing the values specific to configurations separate from the project-wide settings, I'd argue that we should have that built into the DOM, where when we get configuration-specific values rather than configuration and project values.  Is this something that's doable?  I think that'd make sense for getting file-specific configuration values as well.  I haven't had the chance to look into file-specific values yet, but that's on my list to do :).  I'll look into fixing those other comments today and get it back out.  Thanks for taking a look!

modules/gmake/gmake.lua
```diff
@@ -0,0 +1,333 @@
...
+local path = require('path')
+local premake = require('premake')
+local State = require('state')
+
+local gmake = {}
+
+gmake.sln = doFile('./src/sln.lua', gmake)
+gmake.proj = doFile('./src/proj.lua', gmake)
+
+local _ARCHITECTURES = {
```
samsinsane, line 15: This shouldn't be `local`, doing this makes it impossible to extend from a module or script. Also, this isn't used anywhere.

samsinsane (commented):
I'm not sure how the gmake2 generator supported multiple jobs (`-j#`) properly compared to the gmake generator, but this rewrite needs to ensure it doesn't go backwards on supporting multiple jobs.

nickclark2016:
The `-j` flag is done on command line, not in a rule.

modules/gmake/gmake.lua
nickclark2016: Tbh, this was just copied from the vstudio module, but you're right, it's unused. Will remove.

samsinsane:
> The `-j` flag is done on command line, not in a rule, but I definitely agree we don't want to break multi-core/multi-job builds.

I know that, but there was _something_ about the original gmake generator that caused build failures if you used the `-j` flag which the gmake2 generator fixed. My only concern about another rewrite is reintroducing this and creating a reason to rewrite the generator again.

modules/gmake/tests/sln/sln_help_rule_tests.lua
nickclark2016: > I'm not sure how the gmake2 generator supported multiple jobs (`-j#`) properly compared to the gmake generator, but this rewrite needs to ensure it doesn't go backwards on supporting multiple jobs.



> > The `-j` flag is done on command line, not in a rule, but I definitely agree we don't want to break multi-core/multi-job builds.
> 
> I know that, but there was _something_ about the original gmake generator that caused build failures if you used the `-j` flag which the gmake2 generator fixed. My only concern about another rewrite is reintroducing this and creating a reason to rewrite the generator again.

Ohhh, following now :).  I'll keep an eye out for it and see if I can get it to break.

starkos: Yeah, I'm super NOT thrilled to be rewriting this stuff again, TBH. But between the two different modules and all the various authors it's kind of a mess. To @samsinsane's point: definitely work against the Gmake2 module as much as possible, if you weren't already, as that's been getting the bulk of the recent fixes.

starkos:
> For printing the values specific to configurations separate from the project-wide settings, I'd argue that we should have that built into the DOM, where when we get configuration-specific values rather than configuration and project values.

If you remove [the call to `withInheritance()` from `fetchProjectConfig()`](https://github.com/premake/premake-core/pull/1700/files#diff-0ee3ea57ff2c1ba4adc911a62160aa61d36379032731e464972cd6076a9f9c51R213), you will get just the configuration values, without the project-level settings. But you'll still need to add logic to the exporter to write out the project-level settings, in addition to iterating over the build configurations. This is all just theoretical at this point mind you, there may be good reasons NOT to do this that I'm not seeing.

And once more so it doesn't get lost in the conversation: IMHO we really need some R&D on things like custom build rules, per-file build settings, and multi-platform targeting before we push too much further on this, since they could have big impacts on how the files should be written.

nickclark2016:
> > For printing the values specific to configurations separate from the project-wide settings, I'd argue that we should have that built into the DOM, where when we get configuration-specific values rather than configuration and project values.
> 
> If you remove [the call to `withInheritance()` from `fetchProjectConfig()`](https://github.com/premake/premake-core/pull/1700/files#diff-0ee3ea57ff2c1ba4adc911a62160aa61d36379032731e464972cd6076a9f9c51R213), you will get just the configuration values, without the project-level settings. But you'll still need to add logic to the exporter to write out the project-level settings, in addition to iterating over the build configurations. This is all just theoretical at this point mind you, there may be good reasons NOT to do this that I'm not seeing.
> 
> And once more so it doesn't get lost in the conversation: IMHO we really need some R&D on things like custom build rules, per-file build settings, and multi-platform targeting before we push too much further on this, since they could have big impacts on how the files should be written.

TIL.  I'll update that in my project then.

KyrietS:
*#docsnazi here* 😄

How does this amount of code without touching single *.md page has:
- [x] Align documentation to your changes

Please correct me if I'm clinging unnecessarily but what is the strategy for keeping docs up-to-date with the implementation now? Premake 5.0 has such a large docs-debt now that it is no longer manageable. And I'd love to see improvement in that aspect for 6.x.

nickclark2016:
> _#docsnazi here_ 😄
> 
> How does this amount of code without touching single *.md page has:
> 
> * [x]  Align documentation to your changes
> 
> Please correct me if I'm clinging unnecessarily but what is the strategy for keeping docs up-to-date with the implementation now? Premake 5.0 has such a large docs-debt now that it is no longer manageable. And I'd love to see improvement in that aspect for 6.x.

I didn't add any API 😛

KyrietS:
> I didn't add any API 😛

Oh, okay... fair enough for now, but I'll see you tomorrow! 😄

modules/gmake/gmake.lua
samsinsane: > **EDIT**: I did find a usage. We use this to look up the architecture instead of assuming that the value in the premake api is the value that is passed directly to the makefile.

VS requires specific values for the architectures, we definitely shouldn't be bringing that concept to any generator that doesn't need it. I don't want to update Premake just to get access to a new architecture, I want to just add a value into an array and be done with it. Also, you can't write modules to add support for other systems if this is how things work.

Additionally, for VS, we should be doing `lookupTable[cfg.architecture] or cfg.architecture`. The "win32 hack" in Premake5 was a pain for me every single time I tried to extend the vstudio module to support a new platform/architecture. I understand _why_ the VS module has the table, but it shouldn't be `local` and we shouldn't be defaulting to `Win32` we should be defaulting to what the user said - Premake needs to get out of the way.

modules/gmake/gmake.lua
```diff
@@ -0,0 +1,338 @@
...
+	x86_64 = 'x86_64',
+	arm = 'ARM',
+	arm64 = 'ARM64'
+}
+
+
+---
+-- Table of all toolsets.
+---
+gmake.toolsets = {
```
Jarod42, line 26: Shouldn't those toolset specific stuff be moved in the appropriate toolset directly?

nickclark2016: Very much agreed, just waiting for the toolset API then I'll do it.

starkos:
> Premake 5.0 has such a large docs-debt now that it is no longer manageable. And I'd love to see improvement in that aspect for 6.x.

Agreed. I didn't document the earlier work because I wasn't yet sure if it was even going to work, and we didn't have Docusaurus yet. Now that it does, and we do, I'm making an effort to document everything new that I add, and regularly do a little bit of catch-up on the earlier stuff. 

There's no obligation at all, but you're welcome to contribute to the 6.x stuff as well if you like.

modules/gmake/gmake.lua
starkos: @Jarod42 We weren't really as clear as we could have been on the intention of this PR. We're just trying to bootstrap a 6.x makefile exporter up to the same level as the current 6.x VS exporter: a minimal implementation able to generate a barebones but working makefile for Premake itself. I'm also working an Xcode version of the same thing. Some things—like toolsets and everything related to them—will be placeholders until we have this baseline work in place.

modules/gmake/gmake.lua
starkos: @samsinsane Agreed on all points, but this was just intended as a placeholder to get things working—I had a comment to indicate that in the VS version that didn't get carried over. I believe (but don't yet know for sure) that these will end up in the eventual toolset abstraction, and will be extensible from there. Once we have baseline exporters for VS (done), Gmake and Xcode (in progress) that is on my shortlist to dig into.

PR #1701, opened 2021-08-28
add support for external warning levels in visual studio

### LINKED ISSUE #1692: Add support for MSVC external includes
premake has support for sysincludedirs for a long time. It works great for clang and gcc in my experience. However for visual studio the directories in sysincludedirs are threated as normal includes. This leads to warning generation in those includes. It's annoying if you want to use the FatalWarnings flag that premake provides.

**What might be a solution?**
Visual Studio has support for external includes since Visual Studio 2017 version 15.6 Preview 1: https://devblogs.microsoft.com/cppblog/broken-warnings-theory/

But premake has no support for it yet. Is it possible to add this feature? If you can point me to the right direction I can implement it myself.

You can work around the missing support by using:
`buildoptions {"/experimental:external  /external:W0"}`
`buildoptions {"/external:I <include path>"}`

The downside of this approach is that the result is not clearly visible in the project settings inside visual studio. It does not show up as  external includes:
external1
but you can see them in the additional options:
external2

StellaASchlotter:
**What does this PR do?**
Add support for external warnings in visual studio

sysincludedirs now works slightly different for visual studio.  Instead of adding the paths in sysincludedirs  "VC++ Directories" properties panel under "Include Directories" it will add it to "External Include Directories"

It was only tested on visual studio 2019.

closes #1692

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -2762,7 +2762,6 @@
 	end
 
 
-
 	function m.disableSpecificWarnings(cfg, condition)
 		if #cfg.disablewarnings > 0 then
 			local warnings = table.concat(cfg.disablewarnings, ";")
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -356,6 +356,7 @@
 		local calls = {
 			m.precompiledHeader,
 			m.warningLevel,
+			m.externalWarningLevel,
 			m.treatWarningAsError,
 			m.disableSpecificWarnings,
 			m.treatSpecificWarningsAsErrors,
@@ -2133,7 +2134,7 @@
 	function m.includePath(cfg)
 		local dirs = vstudio.path(cfg, cfg.sysincludedirs)
 		if #dirs > 0 then
-			m.element("IncludePath", nil, "%s;$(IncludePath)", table.concat(dirs, ";"))
+			m.element("ExternalIncludePath", nil, "%s;$(ExternalIncludePath)", table.concat(dirs, ";"))
 		end
 	end
 
@@ -2761,6 +2762,7 @@
 	end
 
 
+
 	function m.disableSpecificWarnings(cfg, condition)
 		if #cfg.disablewarnings > 0 then
 			local warnings = table.concat(cfg.disablewarnings, ";")
@@ -2815,6 +2817,11 @@
 		if cfg.warnings then
 			m.element("WarningLevel", condition, map[cfg.warnings] or "Level3")
 		end
+	end
+
+	function m.externalWarningLevel(cfg)
+		local map = { Off = "TurnOffAllWarnings", High = "Level4", Extra = "Level4", Everything = "EnableAllWarnings" }
+		m.element("ExternalWarningLevel", nil, map[cfg.externalwarnings] or "Level3")
 	end
 
 
--- a/website/docs/externalwarnings.md
+++ b/website/docs/externalwarnings.md
@@ -1,4 +1,4 @@
-Controls the number of warnings that are shown by the compiler. This setting only applies to visual studio 17 and newer. It will control the warnings shown for includes given by sysincludedirs
+Controls the number of warnings that are shown by the compiler. This setting only applies to Visual Studio 17 and newer. It will control the warnings shown for includes given by [sysincludedirs](sysincludedirs.md).
 
 ```lua
 externalwarnings "value"
--- a/website/docs/externalwarnings.md
+++ b/website/docs/externalwarnings.md
@@ -1,0 +1,34 @@
+Controls the number of warnings that are shown by the compiler. This setting only applies to visual studio 17 and newer. It will control the warnings shown for includes given by sysincludedirs
+
+```lua
+externalwarnings "value"
+```
+
+If no value is set for a configuration, the toolset's default warning level will be used.
+
+### Parameters ###
+
+`value` specifies the desired level of warning:
+
+| Value       | Description                                            |
+|-------------|--------------------------------------------------------|
+| Off         | Do not show any warning messages.                      |
+| Default     | Use the toolset's default warning level.               |
+| Extra       | Enable the toolset's maximum warning level.            |
+
+### Applies To ###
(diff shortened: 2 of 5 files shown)


nickclark2016:
Can we get some unit tests to show this working?

StellaASchlotter:
Working on it. Locally it works for me already. I know thats not enough to get it approved :)

Do you have any thoughts on the splt between the warning level and the external include dirs. For gcc and clang setting a include dir with `sysincludedirs` also sets the warnings to `Off`. This is different in visual studio now. I wonder if the default behavior for vs should also be warnings off?

nickclark2016:
I'd argue that we want the same default behavior across platforms, but I could see the argument for making it default to the exported platform.  Thoughts @starkos @samsinsane?

website/docs/externalwarnings.md
```diff
@@ -0,0 +1,34 @@
+Controls the number of warnings that are shown by the compiler. This setting only applies to visual studio 17 and newer. It will control the warnings shown for includes given by sysincludedirs
```
KyrietS, line 1:
```suggestion
Controls the number of warnings that are shown by the compiler. This setting only applies to Visual Studio 17 and newer. It will control the warnings shown for includes given by [sysincludedirs](sysincludedirs.md).
```

website/docs/externalwarnings.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Parameters ###
+
+`value` specifies the desired level of warning:
+
+| Value       | Description                                            |
+|-------------|--------------------------------------------------------|
+| Off         | Do not show any warning messages.                      |
+| Default     | Use the toolset's default warning level.               |
+| Extra       | Enable the toolset's maximum warning level.            |
```
KyrietS, line 17: What about `High` and `Everything` values?

website/docs/sysincludedirs.md
```diff
@@ -4,9 +4,12 @@ Specifies the system include file search paths.
 sysincludedirs { "paths" }
 ```
 
-For Visual Studio, these paths are placed in the "VC++ Directories" properties panel. For GCC and Clang, they are preceded with the `-isystem` flag, rather than `-I`. For toolsets which do not support the concept of system include directories, they are treated as a normal include directory.
+For Visual Studio, these paths are placed in the "VC++ Directories" properties panel under "External Include Directories". Note that unlike gcc and clang this does not control warning levels. See externalwarnings to control the warning level for sysincludes in visual studio.
```
KyrietS, line 7:
```suggestion
For Visual Studio, these paths are placed in the "VC++ Directories" properties panel under "External Include Directories". Note that unlike gcc and clang this does not control warning levels. See [externalwarnings](externalwarnings.md) to control the warning level for sysincludes in Visual Studio.
```

website/docs/sysincludedirs.md
```diff
@@ -4,9 +4,12 @@ Specifies the system include file search paths.
...
+
+For GCC and Clang, they are preceded with the `-isystem` flag, rather than `-I`. For toolsets which do not support the concept of system include directories, they are treated as a normal include directory.
 
 Include files located via a system include directory are treated as correct: no warnings will be shown for the contents of the file.
+Note that this is different for visual studio.
```
KyrietS, line 12:
```suggestion
Note that this is different for Visual Studio.
```

KyrietS (commented):
Hi, some comments on the documentation from me.
In general Visual Studio should always be capitalized. And it's good to link to other Premake's terms when you mention them.

starkos (changes requested):
Thanks for putting this together!

But uh…no, we're not changing the handling of all include paths for Visual Studio. Please roll that bit back and focus this PR on external warning levels only.

If you want to expose the ability to set "external" include paths, please introduce a new API (`externalincludedirs` would be an obvious name) to support that. It should go into a separate PR, instead of mixing multiple unrelated feature changes into a single PR.

If we can trim this down to just the external warnings switch, and address the comments on tests and documentation, we can go ahead and merge. Thanks again!

StellaASchlotter:
@starkos according to the documentation of Visual Studio. The external warning levels only apply for  the external include dirs.
The example in https://devblogs.microsoft.com/cppblog/broken-warnings-theory/ shows that the external warning level and the external include dir must be specified.  So those two features are tied together and are not unrelated in my opinion. I can still split this in multiple PR's.

Another point:
Suppose we introduce externalincludedirs as a seperate thing. For projects that should support both gcc/clang and VisualStudio, both `sysincludedirs` and `externalincludedirs` must be set. Then you need to be careful that you set those inside a filter. Otherwise you would end up with the sysincludedirs als normal includes + the externalincludedirs as external include in Visualstudio. I guess this could be pointed out in the documentation.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2752,6 +2753,7 @@
 	end
 
 
+
```
starkos, line 2756: Please avoid introducing spurious formatting changes; thanks!

website/docs/sysincludedirs.md
starkos: In addition to @KyrietS changes, it would be useful to know what "unlike gcc and clang this does not control warning levels" means. What do GCC and Clang do differently? I don't know what GCC & Clang do TBH but assuming they suppress warnings in system headers I might put the important bit first and say "To control warning levels for system includes on Visual Studio, see [externalwarnings]; other toolsets will suppress warnings and errors in system includes."

starkos (commented):
I don't have a project in front of me to exercise this, so help me out: are you saying that warnings & errors are NOT being suppressed in system includes by Visual Studio? I kind of thought that was the point of them?

Jarod42:
Should it be closed as superseded by #1754?

website/docs/sysincludedirs.md
Jarod42, line 7: `function msc.getincludedirs` doesn't reflect the External include Directories

StellaASchlotter:
Currently I can't continue working on this. Another solution for externalincludedirs is being developed in https://github.com/premake/premake-core/pull/1754. If you merge that you can close this PR.

starkos:
Based on the comments, closing this one in favor of #1754.

PR #1702, opened 2021-08-28
[Docs] Add a cruel script to dump all the values of tables

Pirulax:
**What does this PR do?**
Adds a cruel script to dump the content of tables available inside tokens

It might help people, as the docs aren't complete.


starkos (commented):
Thank you for taking the time to put this together! I'm going to vote "no" on this one though. Neither myself nor Google have any idea what "a cruel script" is, there's no explanation of what this is or why you'd want it, and it is just documenting a hack, IMHO.

An action to dump Premake's view of things would be useful, but this isn't quite that.

PR #1710, opened 2021-09-07
Add support for SSE 4.2.

ActuallyaDeviloper:
**What does this PR do?**

This adds support for SSE 4.2 in Premake. The change is quite simple. The lack of SSE 4.2 support so far was probably just an oversight.

There are no behaviour changes for already written and currently working Premake build scripts.

I did not add new tests, since there are none for global configurations themselves and I just added new table entries and really did not change any actual logic.


modules/d/tools/ldc.lua
```diff
@@ -116,6 +116,7 @@
 			SSE3 = "-mattr=+sse3",
 			SSSE3 = "-mattr=+ssse3",
 			["SSE4.1"] = "-mattr=+sse4.1",
+			["SSE4.2"] = "-mattr=+sse4.2",
```
nickclark2016, line 119: Just because I'm not as familiar with D, is SSE 4.2 supported by LDC?

nickclark2016 (changes requested):
Please add a unit test for the value.

modules/d/tools/ldc.lua
ActuallyaDeviloper: I am not able to find documentation for that but at gcc.godbolt.org the LDC compiler accepts it.

ActuallyaDeviloper:
Thank your for your speedy response.

Do you really think a test is necessary? There are also no tests that I can find for SSE 3, SSSE 3 or SSE 4.1 and I think that is totally fine that way.

If you insist that this must be tested, I am actually not sure how to. The only tests for SSE-like kind of things seem to be for specific toolsets. Would I need to make a new test for each of them?

nickclark2016:
I'll defer to @starkos or @samsinsane for final decision on a unit test, but it's part of our checklist.

modules/d/tools/ldc.lua
nickclark2016: Good enough for me.

website/docs/vectorextensions.md
```diff
@@ -13,14 +13,15 @@ If no value is set for a configuration, the toolset's default vector extension s
...
-| AVX2        | Use Advanced Vector Extensions 2.                      |
 | IA32        | Use Intel Architecture 32-bit                          |
 | SSE         | Use the basic SSE instruction set.                     |
 | SSE2        | Use the SSE2 instruction set.                          |
 | SSE3        | Use the SSE3 instruction set.                          |
 | SSSE3       | Use the SSSE3 instruction set.                         |
 | SSE4.1      | Use the SSE4.1 instruction set.                        |
+| SSE4.2      | Use the SSE4.2 instruction set.                        |
+| AVX         | Use Advanced Vector Extensions.                        |
+| AVX2        | Use Advanced Vector Extensions 2.                      |
```
samsinsane, line 24: Reordering this to be based on release date instead of alphabetical should be done as a different PR - this PR is adding SSE4.2 and should only add SSE4.2.

samsinsane (changes requested):
Regarding the unit tests, I guess it would depend if SSE4.2 is only supported by a subset of VS. It looks like that's how it's been tested before, I'm not sure if explicitly testing each possible value is worth the effort but given how we map the values for VS perhaps we should ensure we are mapping correctly? I think this can be merged without any unit tests though.

As for _how_ to do the tests, just as they are already:
https://github.com/premake/premake-core/blob/65deb619f8d5579487def157d7c369e5c6d18864/modules/vstudio/tests/vc2010/test_vectorextensions.lua#L33-L93
https://github.com/premake/premake-core/blob/65deb619f8d5579487def157d7c369e5c6d18864/tests/tools/test_msc.lua#L298-L320
https://github.com/premake/premake-core/blob/65deb619f8d5579487def157d7c369e5c6d18864/tests/tools/test_gcc.lua#L127-L149

ActuallyaDeviloper:
You are right, there are no changes regarding the VS version because VS (unfortunately) can't directly target SSE3 - SSE4.2 in any version with it's code generation controlled by these compiler switches.

I have made the AVX entry the second one again even though I am sceptical of the existing order in regards to if there is any meaning to it.

website/docs/vectorextensions.md
samsinsane: > this PR is adding SSE4.2 and should only add SSE4.2.

samsinsane (approved):

starkos (approved):

ActuallyaDeviloper:
Can we maybe make progress with this?

@nickclark2016 Apparently both @starkos and @samsinsane consider the level of preexisting testing sufficient. If I am not mistaken, your counter argument was only due to the formal requirement of fulfilling the checklist. Can you accept this patch set?

nickclark2016:
If tests are added for MSC (see `test_msc.lua`) and GCC (see `test_gcc.lua`), I'm happy to approve.

ActuallyaDeviloper:
@nickclark2016 Can we make progress with this now?

PR #1712, opened 2021-09-07
Add OpenMP support for VisualStudio

### LINKED ISSUE #1312: Enabling OpenMP in vs2017
I've seen from other projects that OpenMP can be enabled simply:
	`configuration{ "vs*",  "release" }
	buildoptions { "/openmp" }	`	
In my project this doesn't seem to work as "/openmp" goes to AdditionalOptions instead of OpenMPSupport. I know that the modules folder have the answer but I got lost and I need some help.
Thank you

T-rvw:
**What does this PR do?**

When I tried to use Premake5 to set VisualStudio project property -> C++ -> Language -> Open MP Support, I didn't find any solution. I learned that [issue 1312](https://github.com/premake/premake-core/issues/1312) reported in the past and it also didn't get answered. So this PR add support for this compile option.

Closes #1312 

It will add `<OpenMPSupport>true</OpenMPSupport>` node inside `<ClCompile>...</ClCompile>` group

No.

- [Yes] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [Yes] Add unit tests showing fix or feature works; all tests pass
- [Yes] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [Yes] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [Yes] Minimize the number of commits
- [Yes] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

### DIFF
--- a/modules/vstudio/tests/vc2019/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2019/test_compile_settings.lua
@@ -39,3 +39,30 @@
 	<SupportJustMyCode>false</SupportJustMyCode>
 		]]
 	end
+
+--
+-- Check ClCompile for OpenMPSupport
+--
+	function suite.openmpOn()
+		openmp "On"
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<Optimization>Disabled</Optimization>
+	<OpenMPSupport>true</OpenMPSupport>
+		]]
+	end
+
+	function suite.openmpOff()
+		openmp "Off"
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<Optimization>Disabled</Optimization>
+	<OpenMPSupport>false</OpenMPSupport>
+		]]
+	end
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -370,6 +370,7 @@
 			m.functionLevelLinking,
 			m.intrinsicFunctions,
 			m.justMyCodeDebugging,
+			m.supportOpenMP,
 			m.minimalRebuild,
 			m.omitFramePointers,
 			m.stringPooling,
@@ -2160,6 +2161,14 @@
 
 		if _ACTION >= "vs2017" and jmc == "Off" then
 			m.element("SupportJustMyCode", nil, "false")
+		end
+	end
+
+	function m.supportOpenMP(cfg)
+		if cfg.openmp == "On" then
+			m.element("OpenMPSupport", nil, "true")
+		elseif cfg.openmp == "Off" then
+			m.element("OpenMPSupport", nil, "false")
 		end
 	end
 
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -118,6 +118,10 @@
 		},
 		justmycode = {
 			Off = "false"
+		},
+		openmp = {
+			On = "/openmp",
+			Off = "/openmp-"
 		}
 
 	}
--- a/website/docs/openmp.md
+++ b/website/docs/openmp.md
@@ -1,0 +1,34 @@
+Enable or disable [OpenMP](https://en.wikipedia.org/wiki/OpenMP).
+
+```lua
+openmp "value"
+```
+If no value is set for a configuration, the toolset's default OpenMP option (usually Off) will be performed.
+
+### Parameters ###
+
+`value` is one of:
+
+| Value   | Description                                       |
+|---------|---------------------------------------------------|
+| On      | Turn on OpenMP.                                   |
+| Off     | Turn off OpenMP.                                  |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0-beta1 or later. Currently only implemented for Visual Studio 2010+. As a workaround for other toolsets, you can use [buildoptions](buildoptions.md) like this:
+
+```lua
+filter "toolset:not msc*"
+	buildoptions "-fopenmp"
+```
+
+## Examples ##
+
+```lua
+openmp "On"
+```
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1449,6 +1449,16 @@
 		}
 	}
 
+	api.register {
+		name = "openmp",
+		scope = "project",
+		kind = "string",
+		allowed = {
+			"On",
+			"Off"
+		}
+	}
+
 -----------------------------------------------------------------------------
 --
 -- Field name aliases for backward compatibility
--- a/website/docs/optimize.md
+++ b/website/docs/optimize.md
@@ -4,7 +4,7 @@ The **optimize** function specifies the level and type of optimization used whil
 optimize "value"
 ```
 
-If no value is set for a configuration, the toolsets default optimization (usually none) will be performed.
+If no value is set for a configuration, the toolset's default optimization (usually none) will be performed.
 
 ### Parameters ###
 
--- a/website/docs/strictaliasing.md
+++ b/website/docs/strictaliasing.md
@@ -4,7 +4,7 @@ Sets the level of allowed pointer aliasing.
 strictaliasing "value"
 ```
 
-If no value is set for a configuration, the toolsets settings will be used.
+If no value is set for a configuration, the toolset's settings will be used.
 
 ### Parameters ###
 
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -176,6 +176,7 @@ module.exports = {
 						'nuget',
 						'nugetsource',
 						'objdir',
+						'openmp',
 						'optimize',
 						'pchheader',
 						'pchsource',


nickclark2016 (changes requested):
Please add unit tests to verify functionality.

T-rvw:
> Please add unit tests to verify functionality.

Thanks for your quick reply. I added a test function to check it.

src/tools/msc.lua
```diff
@@ -118,6 +118,10 @@
 		},
 		justmycode = {
 			Off = "false"
```
T-rvw, line 120: I wrote `justmycode "On"` and got an error in the past... I think we should always apply `On` and `Off` as a whole to users.

nickclark2016: Correct.  I have that on my backlog, or feel free to add that in yourself in a separate PR :)

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2163,6 +2164,12 @@
 		end
 	end
 
+	function m.supportOpenMP(cfg)
+		if cfg.openmp == "On" then
```
T-rvw, line 2168: I think it is unnecessary to check Visual Studio version here.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2163,6 +2164,12 @@
...
+			m.element("OpenMPSupport", nil, "true")
+		end
```
nickclark2016, line 2170: Let's output false here if "openmp" is set to false.  This way, we aren't relying on the exporter's default.

modules/vstudio/tests/vc2019/test_compile_settings.lua
```diff
@@ -38,4 +38,19 @@
...
 	<SupportJustMyCode>false</SupportJustMyCode>
 		]]
-	end
\ No newline at end of file
+	end
+
+--
+-- Check ClCompile for OpenMPSupport
+--
+	function suite.SupportOpenMP()
```
nickclark2016, line 46: Add a test case for off

nickclark2016 (changes requested):

modules/vstudio/vs2010_vcxproj.lua
nickclark2016: I know there's OpenMP support back at least as far as 2015, so I'd be happy with no version check.

src/tools/msc.lua
nickclark2016: Appreciate you pulling that off my backlog then

modules/vstudio/vs2010_vcxproj.lua
T-rvw: I learned that VS2010 started to support OpenMP and premake only supports VS2013 and above versions. So it's OK not to check it.

nickclark2016 (approved):
I'm happy with the PR. Recommend squash and merge strategy to get these commits to a single commit for merge.  @T-rvw, for future reference, please amend commits with changes or squash before pushing so we have minimal commits for merging back into the repo.  Looks good 👍

samsinsane (approved):

Jarod42:
Shoudn't `-fopenmp` be added to gcc/clang toolset too.

The other way would be to search for `/openmp` in buildoption to place it in correct place in visual project.

T-rvw:
> Shoudn't `-fopenmp` be added to gcc/clang toolset too.
> 
> The other way would be to search for `/openmp` in buildoptions to place it in correct place in visual project (as from premake point of view, it is a buildoption like another). (I know that Codelite/CodeBlock doesn't hardcode compiler options and extract them as checkboxes from "commandline")

Thanks for providing information about `-fopenmp` in gcc/clang toolset. Do you think I should support `-fopenmp` for gcc/clang in this PR or a separated one? I'm willing to support OpenMP not only for VisualStudio.

If we use "The other way" to support OpenMP, it looks that we need to translate `/openmp` to `-fopenmp` for gcc/clang user. It is a little confusing to write `/openmp` from their perspective. Current way has a consistent usage for user although we need to add a new api.

Jarod42:
I think that that PR should fully (so each toolset) support Open MP.

Not sure I would want each custom visual studio node to be a premake option (which IMO then need support for other toolsets)).
So, my alternative solution is visual studio specific:
instead of having special premake option, user just do:
```
filter "toolset:msc*"
    buildoptions "/openmp"
```

and on premake side, visualStudio module parses the buildoptions to detect if `/openmp` is present, and if so, instead of putting it in extra build command flag, set the information in the node as you did.

as currently `buildoptions "/openmp"` has just *"glitch"* issues (it is in the command line, but the dedicated widget don't show its true value).

samsinsane:
> The other way would be to search for `/openmp` in buildoptions to place it in correct place in visual project (as from premake point of view, it is a buildoption like another). (I know that Codelite/CodeBlock doesn't hardcode compiler options and extract them as checkboxes from "commandline")

If only Premake had some kind of API system to let the user specify something as simple as `openmp "On"`.

Don't over-complicate a problem that is already solved.

Jarod42:
@samsinsane: I would like that Premake is less visual studio specific/centric and more portable.
My concern is not to over-complicate a problem **partially** solved, but having a proper solution/API.

Visual Studio has more than 100 options (with "property entry"). gcc/clang also have specific flags (not necessary overlapping). I'm not sure Premake wants to have equivalent options for each field.

I don't know if OpenMP is generic enough to be part of Premake (toolset) API (it might be).
and if it is, then it should be done for all toolsets (and it is my main concern).

starkos:
> I think that that PR should fully (so each toolset) support Open MP.

@Jarod42 Once again, we have no objection to toolset-specific contributions, and encourage people to improve Premake where they are able. 

> visualStudio module parses the buildoptions to detect if /openmp is present, and if so, instead of putting it in extra build command flag, set the information in the node as you did.

@Jarod42 No. We are not doing this and contributions that do will be rejected. Please use the API system, and do not try to invent your own approach to an already solved problem.

starkos (changes requested):
Two changes please:

1. A unit test would be very much appreciated, you can just copy any of the existing tests.
2. A new documentation page for `openmp()`. Again, you can copy any of the existing pages in `website/docs`, and there are instructions on adding a new page in `website/README.md`. Please note that this feature is currently Visual Studio only.

Thanks! (And sorry for the conflicting feedback; trying to juggle too much!)

T-rvw:
> Two changes please:
> 1. A unit test would be very much appreciated, you can just copy any of the existing tests.
> 2. A new documentation page for `openmp()`. Again, you can copy any of the existing pages in `website/docs`, and there are instructions on adding a new page in `website/README.md`. Please note that this feature is currently Visual Studio only.
> 

1. Unit test was added in `tests/vc2019/test_compile_settings.lua`. Do you mean that I should create a new lua file? If so, which folder should be suitable to place new unit test?

2. Thanks! I'm going to create document page.

website/docs/openmp.md
```diff
@@ -0,0 +1,29 @@
...
+| On      | Turn on OpenMP.                                   |
+| Off     | Turn off OpenMP.                                  |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Require a version 5.0 alpha 16 or later. Currently only implemented for Visual Studio 2010+.
```
T-rvw, line 23: @starkos Please help me to check if the version is right.

website/docs/openmp.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Examples ##
+
+```lua
+openmp "On"
```
Jarod42, line 28: How about

```lua
filter "toolset:msc*"
  openmp "On"
filter "toolset:not msc*"
  buildoptions "-fopenmp"
```

Jarod42:
For information,
I add a test in my repo https://github.com/Jarod42/premake-sample-projects/tree/OpenMP
(@T-rvw : that repo is NOT approved by Premake).

(openMP defines under the hood `_OPENMP`, and I test that define presence).

Notice that old

```lua
filter "toolset:msc*"
  buildoptions "/openmp"
filter "toolset:not msc*"
  buildoptions "-fopenmp"
```
works (even if option is not displayed correctly inside visual studio).

Now we should have

```lua
filter "toolset:msc*"
  openmp "On"
filter "toolset:not msc*"
  buildoptions "-fopenmp"
```

website/docs/openmp.md
T-rvw: Do you have a try that `filter "toolset:not msc*"
  buildoptions "-fopenmp"` works in gcc/clang? `buildoptions "/openmp"` doesn't work in msc.

T-rvw:
@Jarod42 buildoptions "/openmp" for `msc` should not work. See [#1312](https://github.com/premake/premake-core/issues/1312). If `buildoptions "-fopenmp"` works for `not msc`, I am OK to write it into document as a workaround before `openmp` supports other toolsets.

Thanks!

website/docs/openmp.md
Jarod42: In my sample-project test (https://github.com/Jarod42/premake-sample-projects/tree/OpenMPOld),

I used:

```
filter "toolset:msc*"
		buildoptions "/openmp"
	filter "toolset:not msc*"
		buildoptions "-fopenmp"
```
and my tests pass for visual studio 2019 (I only test that version of visual) as well:

https://github.com/Jarod42/premake-sample-projects/runs/3608755488?check_suite_focus=true

I suggest to continue this discussion in my repo instead of here...

Jarod42:
> buildoptions "/openmp" for msc should not work. See #1312

I disagree on "not work". It indeed displays wrongly as the option is misplaced, but project would compile fine.

As when you do `buildoptions "/O2"`, and Optimization field is wrongly displayed, but you use `/O2` option anyway.
(for me, it would be to module visual to place the options to the correct place, but it is not the opinion of Premake team).

> If `buildoptions "-fopenmp"` works for `not msc`.

`buildoptions "-fopenmp"` pass my integration tests for gmake/gmake2/codelite (for gcc/linux).
I don't have tests with clang :-/ (and it seems with quick search that clang requires `-fopenmp=libomp` :-/ )
so filter should be `filter "toolset:gcc"`.

website/docs/openmp.md
T-rvw: See our discussions in https://github.com/Jarod42/premake-sample-projects/commit/65b513d671473923d3a5a964e162ce72bbb1acf3.

T-rvw:
@Jarod42 is nice to point out display issue existed in Visual Studio. That is what I don't know in previous discussions.

Let me rethink about this story, I would like to use `OpenMP` for my project in VisualStudio. So I searched `openmp` in Premake docs but got nothing. So I tried to type `/openmp` in build option and checked VisualStudio's `Open MP Support` display option because I didn't want to wait a build time to see if it works. Then I created pull request for Premake to support correct '/openmp' option and correct display.

My opinion is:

1. An extra API such as `openmp` will save Premake users' time. These people who don't know `/openmp`(`msc` as an example) build option can leverage Premake doc to use it more easily.
2. To translate `/openmp` to `<OpenMPSupport>true</OpenMPSupport>` for VisualStudio, it looks like a workaround for me. I think it must add proper comments about this VisualStudio display issue. Or it is hard to understand to do so. And we can't control if Mircosoft team will fix it or not. Then we maybe have efforts to remove dead codes if they fix it.
3. `openmp "On"` is more straightforward than using toolset filters to type different build options. It is a fact that we need to write filter for VS/gcc/clang, right? But I love to use Premake as my build tool because it is easier to use than CMake and other build tools. And writing filters are more like writing platform specified macros for cross-platform projects. I would like to write these things as less as possible.

website/docs/openmp.md
starkos: Marking this resolved, as conversation has moved.

website/docs/openmp.md
```diff
@@ -0,0 +1,34 @@
+Enable or disable [OpenMP](https://en.wikipedia.org/wiki/OpenMP).
+
+```lua
+openmp "value"
+```
+If no value is set for a configuration, the toolsets default OpenMP option (usually Off) will be performed.
```
nickclark2016, line 6: toolset's, not toolsets

website/docs/openmp.md
```diff
@@ -0,0 +1,34 @@
...
+| On      | Turn on OpenMP.                                   |
+| Off     | Turn off OpenMP.                                  |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Require a version 5.0 alpha 16 or later. Currently only implemented for Visual Studio 2010+. As a workaround for other toolsets, you can use [buildoptions](buildoptions.md) like this:
```
nickclark2016, line 23: Technically, a build of alpha 17 or later.  Alpha 16 does not have this included.

nickclark2016 (changes requested):

nickclark2016:
Just a few minor nits, I'll be happy once addressed.

website/docs/openmp.md
T-rvw: Also fixed this typo for other docs.

website/docs/openmp.md
T-rvw: Do you mean `Require a version 5.0 beta 1 or later.`?

nickclark2016: Yup!

website/docs/openmp.md
```diff
@@ -0,0 +1,34 @@
...
+| On      | Turn on OpenMP.                                   |
+| Off     | Turn off OpenMP.                                  |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Require a version 5.0 beta 1 or later. Currently only implemented for Visual Studio 2010+. As a workaround for other toolsets, you can use [buildoptions](buildoptions.md) like this:
```
starkos, line 23: So close! The only nitpick I can find is the phrasing here is a little confusing and doesn't match the rest of the documentation. Instead of "Require a version 5.0 beta 1 or later" make it "Premake 5.0-beta1 or later.". See [filter](https://premake.github.io/docs/filter/#availability) for an example. Thanks so much for putting this together and dealing with all of our change requests! :)

starkos (changes requested):

nickclark2016 (approved):

starkos (approved):
🎉  Merging…

PR #1723, opened 2021-09-28
Adds condition to the VS csproj ItemGroup when linking with a DLL

cicanci:
**What does this PR do?**

Consider the following `premake5.lua` script line:
```lua
links { "bin/%{cfg.platform}/%{cfg.buildcfg}/MyLib.Sharp.dll" }
```

If you have more than one optimizations it will only consider the first one found. So, for example, if you have `Debug` and `Release` optimizations, the only DLL linked will be the `Debug` one, resulting in the following VS csproj configuration:
```xml
<ItemGroup>
  <Reference Include="MyLib.Sharp">
  <HintPath>..\..\bin\x64\Debug\MyLib.Sharp.dll</HintPath>
  </Reference>
</ItemGroup>
```

The changes in this PR make it possible to apply the same condition used in the `<PropertyGroup>` to the `<ItemGroup>` when linking a DLL in a VS csproj, resulting in the following VS csproj configuration (considering the previous example). 

```xml
<ItemGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
  <Reference Include="MyLib.Sharp">
  <HintPath>..\..\bin\x64\Debug\MyLib.Sharp.dll</HintPath>
  </Reference>
</ItemGroup>
<ItemGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|x64' ">
  <Reference Include="MyLib.Sharp">
  <HintPath>..\..\bin\x64\Release\MyLib.Sharp.dll</HintPath>
  </Reference>
</ItemGroup>
```

After this PR, any VS csproj that links a DLL will have the condition applied for each optimization defined. For example, if only `Debug` is used, the VS csproj will change from this:
```xml
<ItemGroup>
  <Reference Include="MyLib.Sharp">
  <HintPath>..\..\bin\x64\Debug\MyLib.Sharp.dll</HintPath>
  </Reference>
</ItemGroup>
```

To this:
```xml
<ItemGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
  <Reference Include="MyLib.Sharp">
  <HintPath>..\..\bin\x64\Debug\MyLib.Sharp.dll</HintPath>
  </Reference>
</ItemGroup>
```

Not really, the changes are pretty straight forward.


nickclark2016:
Looks like we have tests failing here.  Please investigate why these tests are failing, resolve the failures, and then add unit testing for the new piece of functionality.

modules/vstudio/tests/cs2005/test_assembly_refs.lua
```diff
@@ -35,7 +35,9 @@
 	function suite.emptyGroup_onNoLinks()
 		prepare()
 		test.capture [[
-	<ItemGroup>
+	<ItemGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+	</ItemGroup>
+	<ItemGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
```
Jarod42, line 40: Wonder if tests should not have only one config by default, to avoid to duplicate each expected.

cicanci: Perhaps that would change other tests by limiting to one configuration? I understand it shows duplicated content on each `<ItemGroup>` but that is the expected behaviour with this PR.

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -328,29 +328,27 @@
 -- Write out the references item group.
 --
 
-	dotnetbase.elements.references = function(prj)
+	dotnetbase.elements.references = function(prj, cfg)
 		return {
 			dotnetbase.assemblyReferences,
 			dotnetbase.nuGetReferences,
```
Jarod42, line 334: Shouldn't we add `cfg` parameter to `dotnetbase.nuGetReferences` even if unused.

modules/vstudio/tests/cs2005/test_assembly_refs.lua
Jarod42: I meant,

- dedicated test(s) to see that multiple configurations works. (maybe with different and same content).
- previous tests with single configuration (So indeed all those tests will still be modified with the `Condition` part), but would avoid the duplicate nodes.

Not sure how it works neither, but, alternatively, could "Condition" be added only when content differs? (then previous test would works, and you can add a test with different content depending of the config).

Note: I am just a contributor, not a maintainer.

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -328,29 +328,27 @@
 -- Write out the references item group.
 --
 
-	dotnetbase.elements.references = function(prj)
+	dotnetbase.elements.references = function(prj, cfg)
```
starkos, line 331: If we're putting a configuration condition on this `ItemGroup`, then I think everything in it should target that configuration. I would…

1. Change this line to pass only `cfg`, not `prj`
2. Update your changes to `assemblyReferences` to accept `cfg` as the only argument
3. Update `nuGetRefefences` to accept only `cfg`, and replace this line `local cfg = p.project.getfirstconfig(prj)` with `local prj = cfg.project`

I did a quick test and I believe these are all the changes that are needed. That should bring this `ItemGroup` into alignment with the rest of the code. Does that make sense? Let me know if you have questions.

starkos (changes requested):

starkos:
Also, if you could squash your changes down to a single commit (you can force push to this branch to overwrite the previous commits) it will make it easier for us to merge it. Thanks!

starkos:
Merged, thanks!

PR #1728, opened 2021-10-04
Added action to Premake scripts to check for and generate missing documentation

samsinsane:
- Added auto-generated documentation files for APIs without docs
- Renamed editAndContinue.md to editandcontinue.md to match convention

**What does this PR do?**

Adds a helpful action for the Premake project to auto-generate new documentation.

Prevents CI builds from succeeding if documentation is missing.

Under certain conditions the git commands won't work, for example if a user hasn't allowed tags to be downloaded. This PR doesn't necessarily focus on a single fix/feature as the template required for missing documentation needs to be reviewed. Similarly, `editAndContinue` needed to be corrected to avoid a false-positive on the missing documentation.


premake5.lua
```diff
@@ -40,6 +40,102 @@
...
+						if stage == "alpha" and requiredversion == 17 then
+							stage = "beta"
+							requiredversion = 1
+						end
+						docfile:write("Premake 5.0.0 " .. stage .. " " .. requiredversion .. " or later.\n\n")
+					else
+						docfile:write("Premake 5.0.0 unknown or later.\n\n")
+					end
+					path.normalize = pathnormalize
+					docfile:write("### Examples ###\n\n")
```
Jarod42, line 127: Not sure about the (dummy) example section (not all existing doc contains example).
Especially for the ones when `allowed` is a `table`.

Commented dummy section "See also" might be useful for consistency.

Wonder if `register` should not have a "description" field (as the new option has) to.

samsinsane: > Not sure about the (dummy) example section (not all existing doc contains example). Especially for the ones when `allowed` is a `table`.
> 
> Commented dummy section "See also" might be useful for consistency.

I'm not sure which API I was using as the template, but I believe it was [this one](https://github.com/premake/premake-core/blob/master/website/docs/debugger.md). Which based on the [wiki commit](https://github.com/premake/premake-core/wiki/_compare/d407be92ee2f8cd0030beb177bbece5d5199f124) that added it, looks like it was just a bunch of auto-generated files. This is the template that we've done before, so I figured it was good to continue using it.

As for removing examples and adding "see also", I just don't quite understand why we would do that? In my opinion, not every API requires a "see also", but most APIs should probably have an example of their usage. Even if that example is just showing an allowed value being used. I don't think this should be part of the auto-gen docs though, I believe the docs writer needs to make a call on which value should be used as an example, with an appropriate description of how it's being used. For example, [`disablewarnings`](https://premake.github.io/docs/disablewarnings#examples) this shows how you can actually use it, not just calling the API blindly, as well as providing a description of what the example does. [`compileas`](https://premake.github.io/docs/compileas#examples) is another example of using the API, this time with a set of allowed values, however it doesn't contain a description of what the example is doing. I think a placeholder example is enough to get doc writers going, but perhaps the placeholder needs more guiding text?

> Wonder if `register` should not have a "description" field (as the new option has) to.

I'm not quite sure this is relevant to this PR? This should probably be it's own issue/discussion instead.

KyrietS:
Just remember that those auto-generated drafts won't appear on the sidebar but will still be searchable.

Sidebar auto-generation (which I've been talking about for months and it's still work-in-progress 😅) is based on folders. So eventually we would like to get something similar to [6.x docs arranged in folders](https://github.com/premake/premake-core/tree/6.x/website/docs). For now it is ok to have all markdown files in one folder, but I plan to change that at some point.

samsinsane:
> Just remember that those auto-generated drafts won't appear on the sidebar but will still be searchable.

Oh, that was not intentional at all - even though I modified the sidebar.js it didn't even occur to me to add these new files. Should I add these files? I think even just availability information being there is better than nothing. @KyrietS what are your thoughts?

premake5.lua
Jarod42: I meant generated files would requires someone to fix content anyway, so I proposed
```
<!-- 
Example:
...
See Also:
-->
```
So not visible as default, but allow the ones who fix content to respect the template.
but nvm.

KyrietS:
I think that these drafts should be listed in the sidebar so that users can see it and (hopefully) help us develop the documentation. I'm just wondering that maybe we should create a separate section in the sidebar for auto-generated pages to keep track of them 🤔. 

*(And personally, I wouldn't bother to automate the process of adding these pages to the sidebar)*

Also, I think it would be nice if each draft had some sort of encouragement for the reader to help with the documentation improvements. "This page was auto-generated. Feel free to help us improve the documentation by creating a pull request." or something like that.

samsinsane:
@KyrietS I can just remove the commented out section for the files that I added, and just leave it in the template generation?

KyrietS:
@samsinsane Okay, then please do it. Their fix will be for the next version of Docusaurus anyway.

PR #1749, opened 2021-11-06
Changed C++20 to emit C++20 instead of C++Latest in MSVC

### LINKED ISSUE #1746: cppdialect c++20 outputs c++latest on VS2019
I use cppdialect "C++20" currently, which resolves to `/std:c++latest` but in my version of vs2019 I want it to resolve to `/std:c++20`.

Imported from this comment: https://github.com/premake/premake-core/issues/971#issuecomment-962028454

nickclark2016:
**What does this PR do?**

Changes C++20 to emit `stdcpp20` instead of `stdcpplatest` in VS projects VS2017 or newer.  Adds `c11` and `c17` outputs to VS 2019 and newer projects.

Existing projects with the `cppdialect` set to `C++20` will now emit `stdcpp20` instead of `stdcpplatest` for VS 2017 and newer.

Closes #1746


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1477,9 +1477,15 @@
 			elseif (cfg.cppdialect == "C++17") then
 				m.element("LanguageStandard", nil, 'stdcpp17')
 			elseif (cfg.cppdialect == "C++20") then
-				m.element("LanguageStandard", nil, 'stdcpplatest')
+				m.element("LanguageStandard", nil, 'stdcpp20')
```
samsinsane, line 1480: According to the docs, this is only available in VS2019?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1477,9 +1477,15 @@
...
 			elseif (cfg.cppdialect == "C++latest") then
 				m.element("LanguageStandard", nil, 'stdcpplatest')
+			elseif _ACTION >= "vs2019" then
+				if (cfg.cdialect == "C11") then
+					m.element("LanguageStandard_C", nil, 'stdc11')
+				elseif (cfg.cdialect == "C17") then
+					m.element("LanguageStandard_C", nil, 'stdc17')
+				end
```
samsinsane, line 1488: These should be in their own function independent of `cppdialect` - this change currently means specifying both `cppdialect` and `cdialect` will result in only the C++ standard to be emit by Premake.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1539,7 +1545,7 @@
 			elseif (cfg.cppdialect == "C++17") then
 				table.insert(opts, "/std:c++17")
 			elseif (cfg.cppdialect == "C++20") then
-				table.insert(opts, "/std:c++latest")
+				table.insert(opts, "/std:c++20")
```
samsinsane, line 1548: Similar to the comment above, VS2015 likely doesn't support this flag.

src/_premake_init.lua
```diff
@@ -757,6 +757,7 @@
 			"C90",
 			"C99",
 			"C11",
+			"C17",
```
samsinsane, line 760: This should probably be done in a separate PR - I think it's far more likely `gmake2` users are going to want to use this than VS users. Similarly, there should be the `gnu17` option too.

samsinsane (changes requested):

src/_premake_init.lua
nickclark2016: I was just getting all the language flags for MSVC up to date in their mappings. I'd be happy to do gmake2 in a separate PR.

samsinsane: Yeah that's fair enough, I'll resolve this comment.

samsinsane (approved):

PR #1754, opened 2021-11-13
Improve external include & warning support

### LINKED ISSUE #1692: Add support for MSVC external includes

englercj:
**What does this PR do?**

Adds support for "external" headers to Visual Studio 2022. This adds three new APIs:

- `externalwarnings` - Controls warning levels for headers considered "external" (`/external:Wn`, `<ExternalWarningLevel>`)
- `externalanglebrackets` - When set to "On" treats all headers included with angle brackets as "external" (`/external:anglebrackets`, `<TreatAngleIncludeAsExternal>`)
- `externalincludedirs` - Include directory paths to treat as "external" (`/external:Ipath`, `<>`)

This is a different take on this support than #1701. Here instead of repurposing `sysincludedirs`, I added a new concept of `externalincludedirs`. I think these two things are fundamentally different as "external" in this context means "not part of my project" which doesn't necessarily mean "system". For example, I use these for some third-party non-system libraries, which I include using double quotes and not angle brackets.

MSVC Docs: https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170

Existing behavior should be unaffected, all new functionality is behind new APIs.

closes #1692


KyrietS:
Please add your new markdown pages to [website/sidebar.js](https://github.com/premake/premake-core/blob/master/website/sidebars.js)

src/tools/msc.lua
```diff
@@ -398,3 +407,13 @@
 
 		return result
 	end
+
+
+	function msc.getexternalincludedirs(cfg)
```
Jarod42, line 412: To be coherent with other toolset, that implementation should be included in `getincludedirs`
and just call `msc.getincludedirs(cfg, nil, nil, cfg.externalincludedirs, nil)`

nickclark2016:
Out of scope of this PR, but how exactly should this be handled by other build systems?  For build systems without this concept, should `externalincludedirs` be treated as `sysincludedirs`?  I think this question should be answered and the consequences understood before we approve/deny this.

Jarod42:
> I think these two things are fundamentally different as "external" in this context means "not part of my project" which doesn't necessarily mean "system".

Note that msvc does nothing for `sysincludedirs`, so it might just be a naming issue.
What would you place in "system"?
Aliasing might be an option if really you dislike the name.

StellaASchlotter:
I am the author of https://github.com/premake/premake-core/pull/1701. Unfortunately I can't continue working on premake for the next few months. However I want to point out that I think it is a bug in premake that sysincluddirs do not supress warnings in visual studio  but do in gcc and clang. I wanted to address this as part of my PR. 

My idea was to make sysincludedirs behave the same on all systems so I don't need multiple instructions for the same behavior.

But I am okay with closing my PR and going forward with this one. Any support for externalincludedirs is better than none.

englercj:
> What would you place in "system"?

I guess I'd separate include paths I expect to exist on the build system (Windows SDK or glibc++, for example). And external would be dependencies of my application that aren't from the system, for example sqlite, freetype, or other third-party libraries that are external to my app but not part of the system.

Maybe this distinction is moot or undesired, and that's fine I'm open to any solution that lets me tag file paths as external in MSVC so I can disable warnings when including them.

> My idea was to make sysincludedirs behave the same on all systems so I don't need multiple instructions for the same behavior.

If I change this PR to fall back to sysincludedirs on other build systems then you will get the behavior you expect of this new instruction silencing warnings on all platforms. Is that a reasonable solution?

src/tools/msc.lua
englercj: Will fix.

website/docs/externalincludedirs.md
```diff
@@ -0,0 +1,36 @@
...
+You can also use wildcards to match multiple directories. The * will match against a single directory, ** will recurse into subdirectories as well.
+
+```lua
+externalincludedirs { "../includes/**" }
+```
+
+### See Also ###
+
+* [externalanglebrackets](externalanglebrackets.md)
+* [externalwarnings](externalwarnings.md)
```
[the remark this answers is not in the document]
englercj: Will fix.

website/docs/externalwarnings.md
```diff
@@ -0,0 +1,37 @@
...
+### Examples ###
+
+```lua
+externalwarnings "Off"
+```
+
+### See Also ###
+
+* [externalanglebrackets](externalanglebrackets.md)
+* [externalincludedirs](externalincludedirs.md)
```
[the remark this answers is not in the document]
englercj: Will fix.

website/docs/externalwarnings.md
```diff
@@ -0,0 +1,37 @@
+Controls the number of warnings that are shown by the compiler for headers that are considered external. This setting only applies to Visual Studio 17 and newer.
```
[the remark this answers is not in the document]
englercj: I copy-pasted this from `warnings.md`, did you want me to change the word there too?

website/docs/externalincludedirs.md
```diff
@@ -0,0 +1,36 @@
+Specifies the include file search paths for the compiler, treating headers included from these paths as external. This setting only applies to Visual Studio 17 and newer. In other toolsets it is treated the same as [includedirs](includedirs.md).
```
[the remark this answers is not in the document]
englercj: Will fix.

website/docs/externalanglebrackets.md
```diff
@@ -0,0 +1,27 @@
+Treats all headers included by `#include <header>`, where the header file is enclosed in angle brackets (`< >`), as external headers. This setting only applies to Visual Studio 17 and newer.
```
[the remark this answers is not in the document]
englercj: Will fix.

nickclark2016:
I've had some time to dig into the implications of this as opposed to co-opting the existing API.  As it currently stands, I don't think I'd want to accept the PR as-is (however I appreciate the work 😃).  Below I've listed some of my research into the issue as well as what my take on the correct course of action is.

* `externalincludedirs` - This is fundamentally no different than `sysincludedirs`.  While this may not semantically be what you want, they have roughly the same meaning.  These both seem to mean any include directory other than includes from projects that the user owns.
* `externalanglebrackets` - This is something unique to Visual Studio as far as I can tell.  Since this has no analog to any other toolchain, should we make the API part of the visual studio module or should we ignore it in other toolchains.
* `externalwarnings` - This equates to the `-Wsystem-headers` flag in GCC/Clang.

What do I propose?

* I believe that we should coalesce the existing `sysincludedirs` and `externalincludedirs` to a single API called `externalincludedirs`.  I think this name works better, as it is the more all-encompassing of the two.  This will prevent the need for an exporter-specific API.  This would behave by writing `externalincludedirs` to the `externalincludedirs` in Visual Studio, but for GNU makefile it would write it to `-isystem`.
* For `externalwarnings`, I believe we should try to mirror as closely to the current `warnings` options as possible.  See below table for flag mapping.

| Premake Symbol Value | Visual Studio | GNU Makefile     |
|----------------------|---------------|------------------|
| Off                  | /external:W0  |                  |
| Default              | /external:W3  | -Wsystem-headers |
| High                 | /external:W4  | -Wsystem-headers |
| Extra                | /external:W4  | -Wsystem-headers |
| Everything           | /external:W4  | -Wsystem-headers |

* Finally, I believe that the `externalanglebrackets` API should be specific to Visual Studio.

This would have impacts on the API that already exists for `sysincludedirs`.  Instead of removing the API, I would have it deprecated in favor of the new `externalincludedirs` and removed in the full release.

Thoughts on this?

starkos:
+1 on this plan.

> This would have impacts on the API that already exists for sysincludedirs. Instead of removing the API, I would have it deprecated in favor of the new externalincludedirs and removed in the full release.

In favor of deprecating, but would probably not remove until v6 at this point, IMHO. Another option is to make `sysincludedirs` and alias for `externalincludedirs`?

englercj:
> I believe that we should coalesce the existing sysincludedirs and externalincludedirs to a single API called externalincludedirs

I think this makes a lot of sense, I'll get that done.

> For externalwarnings, I believe we should try to mirror as closely to the current warnings options as possible. See below table for flag mapping.

I can make this change, but the reason I had them different originally is because I didn't want the same words with different meanings. For example, `"Everything"` in the `warnings` API means `/Wall` but in the `externalwarnings` it would mean `/W4`. That may be surprising for some users so instead of having multiple `/W4`s I just cut that one out. I'll add it back in for now and let people review.

> Finally, I believe that the externalanglebrackets API should be specific to Visual Studio.

It is currently only used in vs project generation and ignored by other toolsets. Did you want me to do something different here?

englercj:
> In favor of deprecating, but would probably not remove until v6 at this point, IMHO. Another option is to make sysincludedirs and alias for externalincludedirs?

I don't have strong feelings on this, ahppy to do whatever people want here.

nickclark2016:
> > I believe that we should coalesce the existing sysincludedirs and externalincludedirs to a single API called externalincludedirs
> 
> I think this makes a lot of sense, I'll get that done.
> 
> > For externalwarnings, I believe we should try to mirror as closely to the current warnings options as possible. See below table for flag mapping.
> 
> I can make this change, but the reason I had them different originally is because I didn't want the same words with different meanings. For example, `"Everything"` in the `warnings` API means `/Wall` but in the `externalwarnings` it would mean `/W4`. That may be surprising for some users so instead of having multiple `/W4`s I just cut that one out. I'll add it back in for now and let people review.
> 
> I also don't like `"Default"` being `/external:W3`. The visual studio default is to inherit the project's warning level, which matches the behavior of older versions. I'd like to keep that behavior as the default.

For this, I think we should personally make the `warnings` and `externalwarnings` flags match as much as possible. To me, this would be the most straight forward.  It would be nice to have `inherit` as well, but I think the mismatch for `Default` would be confusing.

> > Finally, I believe that the externalanglebrackets API should be specific to Visual Studio.
> 
> It is currently only used in vs project generation and ignored by other toolsets. Did you want me to do something different here?

So that's one option. Another option would be to register the API on the vstudio module.  See this file: https://github.com/premake/premake-core/blob/master/modules/vstudio/_preload.lua

englercj:
OK I think I've addressed all the feedback at this point, aside from deprecating `sysincludedirs` and making it an alias of `externalincludedirs`. I wanted the naming debate of which one to keep settle first so I knew how to proceed.

src/tools/msc.lua
```diff
@@ -237,13 +246,21 @@
...
+	function msc.getincludedirs(cfg, dirs, sysdirs, extdirs, frameworkdirs)
 		local result = {}
 		dirs = table.join(dirs, sysdirs)
 		for _, dir in ipairs(dirs) do
 			dir = project.getrelative(cfg.project, dir)
 			table.insert(result, '-I' ..  p.quoted(dir))
 		end
+		for _, dir in ipairs(extdirs or {}) do
+			dir = project.getrelative(cfg.project, dir)
+			if _ACTION >= "vs2022" then
```
Jarod42, line 258: Logically, test should be against the toolset version, as other actions can use msc toolset.

website/docs/externalanglebrackets.md
```diff
@@ -0,0 +1,28 @@
...
+| Off     | Default. Headers are treated normally. |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0 or later.
+Visual Studio 2022 versiuon 17.0 or later.
```
Jarod42, line 23:
```suggestion
Visual Studio 2022 version 17.0 or later.
```
typo has been copy pasted...

website/docs/externalwarnings.md
```diff
@@ -0,0 +1,40 @@
...
+
+`value` specifies the desired level of warning:
+
+| Value       | Description                                            |
+|-------------|--------------------------------------------------------|
+| Off         | Do not show any warning messages.                      |
+| Default     | Use the toolset's default warning level.               |
+| Extra       | Enable the toolset's maximum warning level.            |
+| High        | Needs documentation                                    |
+| Everything  | Needs documentation                                    |
```
Jarod42, line 19: To document.

Jarod42:
My opinion about naming:
- `externalincludedirs` seems clearer than `sysincludedirs`.
- Having both, one as aliasing of the other is ok.
- `sysincludedirs` has historic advantage, but deprecating `sysincludirs` is ok for me.

If we keep only `sysincludedirs`, then the new `externalanglebrackets`, `externalwarnings` should be renamed IMO.

src/tools/msc.lua
englercj: I was thinking that as well. How do I check that here?

englercj:
Made changes to deprecate `sysincluddirs` and only have an implementation for `externalincludedirs`.

src/_premake_init.lua
```diff
@@ -1483,6 +1476,18 @@
 		}
 	}
 
+	api.register {   -- DEPRECATED 2021-11-16
```
Jarod42, line 1479: Should it be `api.alias()` instead?

website/docs/externalincludedirs.md
```diff
@@ -1,9 +1,13 @@
-Specifies the include file search paths for the compiler, treating headers included from these paths as external. In toolsets other than Visual Studio 2022 this is treated the same as [sysincludedirs](sysincludedirs.md).
+Specifies the include file search paths for the compiler, treating headers included from these paths as external.
 
 ```lua
 externalincludedirs { "paths" }
 ```
 
+For Visual Studio, these paths are placed in the "VC++ Directories" properties panel. For GCC and Clang, they are preceded with the `-isystem` flag, rather than `-I`. For toolsets which do not support the concept of external include directories, they are treated as a normal include directory.
+
+Include files located via an external include directory are treated as correct: no warnings will be shown for the contents of the file.
```
Jarod42, line 9:
```suggestion
Include files located via an external include directory are treated specially, see [externalwarnings](externalwarnings.md).
```

website/docs/sysincludedirs.md
```diff
@@ -36,5 +36,6 @@ sysincludedirs { "../includes/**" }
 
 ### See Also ###
```
Jarod42, line 37: Page should probably just state, alias of `[externalincludedirs](externalincludedirs.md)` and deprecated statut.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2144,9 +2149,21 @@
 
 
 	function m.includePath(cfg)
```
Jarod42, line 2151: Those 2 methods should probably be merged, as they both handle cfg.externalincludedirs.
or vs2022_vcxproj created?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2835,6 +2853,25 @@
 	end
 
 
+	function m.externalWarningLevel(cfg)
+		if _ACTION >= "vs2022" then
```
Jarod42, line 2857: So probably dedicated file for vs2022 stuff.

src/_premake_init.lua
englercj: I copied the pattern of `debuggerflavor` being deprecated in `vstudio/_preload.lua`. Let me know if I should do something different.

modules/vstudio/vs2010_vcxproj.lua
englercj: I combined the functions, moving this into a vcxproj file specifically for 2022 would be a different pattern than all the other vs-version specific stuff in this file which checks _ACTION.

englercj:
Are there additional changes people are interested in seeing here?

src/tools/msc.lua
Jarod42: I would say something like `cfg.toolset >= "msc-v143"`.

Jarod42: @starkos: from [toolset](https://premake.github.io/docs/toolset/), it seems we can use `"v143"` directly instead of `"msc-v143"`, I didn't see code to unify it afterward.
As `"vyyy" > "msc-xxx"`, that check might be problematic. which is the correct method?

starkos (approved):
This looks good to me now. I've updated the PR title to be more accurate, and approving for merge.

starkos:
@englercj Can you please rebase this against the latest main branch, and squash the commits?

Jarod42:
@starkos:
It seems you miss my comment:
> from toolset, it seems we can use "v143" directly instead of "msc-v143", I didn't see code to unify it afterward.
As "vyyy" > "msc-xxx", that check might be problematic. which is the correct method?

starkos:
> @starkos: It seems you miss my comment:
> > from toolset, it seems we can use "v143" directly instead of "msc-v143", I didn't see code to unify it afterward.
> > As "vyyy" > "msc-xxx", that check might be problematic. which is the correct method?

I did miss it, thanks for escalating. It [looks like the value will get normalized to "msc-v143"](https://github.com/premake/premake-core/blob/master/src/base/tools.lua#L32-L53).

src/tools/msc.lua
Jarod42: `cfg.toolset >= "msc-v143"` seems the way to go.

englercj: Changing this line to `if cfg.toolset >= "msc-v143" then` causes an error in the tests:

```
[  FAILED  ] 1 tests, listed below:
[  FAILED  ] tools_msc.cflags_onExternalIncludeDirs
[string "src/tools/msc.lua"]:187: attempt to compare string with nil
stack traceback:
        [string "src/tools/msc.lua"]:187: in field 'getincludedirs'
        ...s/engle/source/oss/premake-core/tests/tools/test_msc.lua:301: in function 'testFunction'
```

Seems like the `cfg.toolset` value is not always set.

englercj:
Hey @starkos, just wanted to ping about this comment: https://github.com/premake/premake-core/pull/1754#discussion_r758605807

I tried the method of comparing `cfg.toolset` but it is sometimes `nil`. Please advise.

Jarod42:
I suppose `p.action.set("vs2010")` doesn't set associated toolset.
You probably need to set toolset also afterward :-/

I hope it is only a Unit testing issue.

src/tools/msc.lua
Jarod42: `cfg.toolset and cfg.toolset >= "msc-v143"` here and set toolset or action in test_msc.lua does the job.
Actually, in UT of toolset, we don't set neither action nor toolset, that'is why you got issue.

englercj:
Rebased and ready for merge.

src/tools/msc.lua
englercj: Fixed.

englercj:
Fixed the toolset check and squashed the commits onto latest master. I think we're good to go.

Jarod42 (approved):

peter1745:
Can't wait for this to be merged in so vs2022 support isn't *meh* anymore

englercj:
Rebased atop latest master, ready for merge AFAIK.

starkos:
Done, thanks!

StellaASchlotter:
@englercj was there a reason why its only for visual studio 2022 and up? Externalincludes are supported in vs2019.

englercj:
@StellaASchlotter It was experimental in most versions of vs2019 and required the `/experimental:external` flag to be set in order to work. It was only non-experimental as of vs2019 16.10. I didn't deal with the complexity of checking when to set the experimental flag because vs2022 support was sufficient for my needs, and it is guaranteed to work there in all versions.

StellaASchlotter:
Ah thanks for this information. I was not aware of this. I will upgrade the VS version in my team then.

PR #1763, opened 2021-11-26
Fix `make clean` handling of generated files

### LINKED ISSUE #1650: Make clean is not working with generated files
I have a simple premake5 file that creates makefiles for a small project of mine. When I run premake5 gmake2 all makefiles are created successfully and I can call make to compile to an executable and run it.
The problem is when I try to call make clean, I get an error saying: `The directory name is invalid. and make: *** [clean] Error 267`

Build directory should have been cleaned

I've narrowed it down to one line in the makefile which should remove all the object files, but causes an error and stops any more files from being removed.

`$(SILENT) if exist $(subst /,\\,$(GENERATED)) rmdir /s /q $(subst /,\\,$(GENERATED))`

This is the line that is causing the error. Removing it fixes the problem but causes my bin/obj folder to remain.

Calling make clean gives this error.

premake5 (Premake Build Script Generator) 5.0.0-alpha16.

This issue is based on https://stackoverflow.com/questions/64414270/make-clean-is-not-working-when-using-premake5
though this problem occurs with me too.

starkos:
**What does this PR do?**

Replaces an incorrect `rmdir` with `del` to fix removal of generated files on `make clean`. Closes #1650.

No changes other than the fix.

Nope.

### DIFF
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -690,7 +690,7 @@
 		_p('\t$(SILENT) rm -rf $(OBJDIR)')
 		_p('else')
 		_p('\t$(SILENT) if exist $(subst /,\\\\,$(TARGET)) del $(subst /,\\\\,$(TARGET))')
-		_p('\t$(SILENT) if exist $(subst /,\\\\,$(GENERATED)) rmdir /s /q $(subst /,\\\\,$(GENERATED))')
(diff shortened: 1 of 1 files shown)


samsinsane (approved):
The `rm -rf` above on line 689, should be changed to `rm -f` to match the change here. This can still be merged, but I think this will create confusion around exactly what `$(GENERATED)` contains.

PR #1764, opened 2021-11-26
Always use `llvm-ar` for Clang

starkos:
**What does this PR do?**

Adjusts the Clang toolset adapter to always use its included `llvm-ar` instead of `ar`, which may or may not be installed. Fixes #1752.

No changes other than described.

Nope.


starkos:
Looks like `llvm-ar` isn't available on the CI servers, so that answers that question.

nickclark2016:
We *should* be able to do a check using `which` in the makefile to see if `llvm-ar` exists.  If it does, use `llvm-ar`, else we use `ar`.  This should probably only be emitted to the makefile if we are using an LLVM toolchain.

PR #1770, opened 2021-12-07
Add a workroud for localized dates on the website

### LINKED ISSUE #1748: Dates in docs aren't localised
Date formats in the docs aren't localised based on the users browser settings, and don't indicate they're hardcoded to the US locale.

I expected my locale to be used, it's literally less effort, just don't specify a locale. Alternatively, ISO 8601 would be fine.

Tried another browser to confirm that the locale support in my primary browser wasn't broken or wrong.

Change your language in your browser from "English" or "English (US)" to "English (Australia)" or "English (UK)".

N/A.

I couldn't find anything in the Docusaurus config file to indicate that we set this, so it looks like Docusaurus unnecessarily hardcodes a locale.

KyrietS:
**What does this PR do?**

Added script that replaces the date in *Last updated on* section. It uses standard JavaScript function [toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) to produce localized date.

I've created a demo based on this commit so please check if it works for you ([LINK](https://kyriets.github.io/premake-core/docs/Using-Premake)). But keep in mind that it is your browser that decides what format to use. For me, every date now looks like this:
image

Fixes #1748 

Website only.

Currently, Docusaurus will always try to put mm/dd/yyyy date on every page. It is a workaround as Docusaurus doesn't allow changing metadata of a page (yet).

I've also bumped Docusaurus version from `beta.6` to `beta.9`. This update did not require any changes.


KyrietS:
I can already see that it's half working ☹ Tricking React has never been easy 😅 I'll mark this as draft for now.

KyrietS:
Okay, it should be all fine now. I'm waiting for comments and opinions 🤠

website/static/js/localizeLastUpdatedDate.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function localizeTimeElement(element) {
+    if (element instanceof HTMLTimeElement && element.hasAttribute('datetime')) {
+        const iso8601 = element.getAttribute('datetime');
+        const date = new Date(iso8601);
+        element.textContent = date.toLocaleDateString(undefined, {
+            year: "numeric",
+            month: "2-digit",
+            day: "2-digit",
```
samsinsane, line 26: Why force `DD`, `MM`, and `YYYY`? My original issue was that the date format wasn't respecting the users locale and this kind of does the same thing.

KyrietS: Yes I wanted to force leading zeros in date format and still use locale=undefined. I think you're right, let the browser decide.

samsinsane (approved):

PR #1773, opened 2021-12-08
Add C++ module file extensions and "allmodulespublic" option

hannes-harnisch:
**What does this PR do?**

Visual Studio automatically detects `.ixx` and `.cppm` as C++ module interface units and compiles them accordingly. This only happens if a `ClCompile` tag is generated for them in the project file, but Premake doesn't emit this tag for these file extensions, meaning this default behavior is lost. This PR rectifies that.

Additionally, it's sometimes necessary to explicitly share modules between projects even though the exporting project is not a static library (for which this is enabled by default). So if you wanna use modules in another project from an executable project for example, you'd have to manually enable it by setting "All Modules are Public" to Yes in the VC++ directories settings in the project properties. I implemented a new Boolean project setting "allmodulespublic" to enable/disable this at the config level. A test is also implemented.

It only adds new behavior.

Don't think so.

### DIFF
--- a/modules/codelite/_preload.lua
+++ b/modules/codelite/_preload.lua
@@ -22,7 +22,7 @@
 
 		-- The capabilities of this action
 
-		valid_kinds     = { "ConsoleApp", "WindowedApp", "Makefile", "SharedLib", "StaticLib", "Utility" },
+		valid_kinds     = { "ConsoleApp", "WindowedApp", "Makefile", "None", "SharedLib", "StaticLib", "Utility" },
 		valid_languages = { "C", "C++" },
 		valid_tools     = {
 			cc = { "gcc", "clang", "msc" }
--- a/modules/codelite/codelite_project.lua
+++ b/modules/codelite/codelite_project.lua
@@ -86,6 +86,8 @@
 		ConsoleApp = "Console",
 		WindowedApp = "Console",
 		Makefile = "",
+		None = "",
+		Utility = "",
 		SharedLib = "Library",
 		StaticLib = "Library"
 	}
@@ -189,7 +191,7 @@
 	end
 
 	function m.compiler(cfg)
-		if configuration_iscustombuild(cfg) or configuration_isfilelist(cfg) then
+		if cfg.project.kind == p.NONE or configuration_iscustombuild(cfg) or configuration_isfilelist(cfg) then
 			_p(3, '<Compiler Required="no"/>')
 			return
 		end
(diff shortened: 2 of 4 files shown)


website/docs/allmodulespublic.md
```diff
@@ -0,0 +1,26 @@
...
+* `Off` - Not all C++ modules in the given project(s) will be public.
+
+## Applies To ###
+
+The `config` scope.
+
+### Availability ###
+
+Premake 5.0-beta2 or later.
+
```
Jarod42, line 20: And only for visual 2019 or superior

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1511,6 +1512,18 @@
 		end
 	end
 
+	function m.allModulesPublic(cfg)
+		if _ACTION >= "vs2019" then
+			if cfg.allmodulespublic ~= nil then
+				if cfg.allmodulespublic then
+					m.element("AllProjectBMIsArePublic", nil, "true")
+				else
+					m.element("AllProjectBMIsArePublic", nil, "false")
```
Jarod42, line 1521: Is there an equivalent flag for command line (for premake-core/src/tools/msc.lua)?

website/docs/allmodulespublic.md
```diff
@@ -0,0 +1,26 @@
...
+```lua
+allmodulespublic "value"
+```
+
+### Parameters ###
+
+`value` one of:
+* `On` - All C++ modules in the given project(s) will be public.
+* `Off` - Not all C++ modules in the given project(s) will be public.
```
Jarod42, line 11: How to make modules public without that?

hannes-harnisch: Using "Project properties > VC++ Directories > Public C++ module directories", which is certainly implementable in Premake. But that's more involved, and should probably be done in a separate PR.

modules/vstudio/vs2010_vcxproj.lua
hannes-harnisch: I don't think so.

website/docs/allmodulespublic.md
hannes-harnisch: Amended.

KyrietS (changes requested):
Please add `'allmodulespublic'` to the [website/sidebars.js](https://github.com/premake/premake-core/blob/master/website/sidebars.js)

KyrietS (approved):

starkos:
Ready to merge but could you please:

- make your changes on a branch, rather than on `master`
- [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) your changes against the latest Premake-core master branch, rather than merging master into your branch

hannes-harnisch:
@starkos I shall try.

hannes-harnisch:
Oops, looks like this was accidentally closed by reshuffling commits.

PR #1788, opened 2022-01-04
Add a new "ScanForModuleDependencies" flag.

LoneBoco:
**What does this PR do?**

This controls the Visual Studio "Scan Sources for Module Dependencies" option, which controls how it scans for C++20 modules.  This new flag sets the option in the generated project files.

This PR adds a new flag option that generates a new entry in the <ClCompile> group for each configuration.

Nope.

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -397,6 +397,7 @@
 			m.compileAsWinRT,
 			m.externalWarningLevel,
 			m.externalAngleBrackets,
+			m.scanSourceForModuleDependencies,
 		}
 
 		if cfg.kind == p.STATICLIB then
@@ -2868,6 +2869,15 @@
 				m.element("TreatAngleIncludeAsExternal", condition, "false")
 			elseif cfg.externalanglebrackets == p.ON then
 				m.element("TreatAngleIncludeAsExternal", condition, "true")
+			end
+		end
+	end
+
+
+	function m.scanSourceForModuleDependencies(cfg)
+		if _ACTION >= "vs2019" then
+			if cfg.flags.ScanForModuleDependencies then
+				m.element("ScanSourceForModuleDependencies", nil, "true")
 			end
 		end
 	end
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -2876,8 +2876,12 @@
 
 	function m.scanSourceForModuleDependencies(cfg)
 		if _ACTION >= "vs2019" then
-			if cfg.flags.ScanForModuleDependencies then
-				m.element("ScanSourceForModuleDependencies", nil, "true")
+			if cfg.scanformoduledependencies ~= nil then
+				if cfg.scanformoduledependencies then
+					m.element("ScanSourceForModuleDependencies", nil, "true")
+				else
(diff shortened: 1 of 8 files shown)


Jarod42:
Is there a equivalent flag for other IDEs which might use msvc as compiler?

LoneBoco:
> Is there a equivalent (command-line) flag for other IDEs which might use msvc as compiler?

No.  This isn't a command-line flag for the compiler.  This controls how the Visual Studio IDE scans for module dependencies and how it determines which files to compile.

This is related to using the `import` command.  By default, Visual Studio will only look at .ixx modules to build the dependency graph.  If it encounters something like `import <vector>;`, it will build that file as a header unit (.ifc), including any other headers it might `import`.

But say your main.cpp file had `import <iostream>;` and no .ixx files in your project tried to import the iostream header.  This would cause a build failure as iostream was never compiled into a header unit.

`Error C7612 could not find header unit for 'C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.30.30705\include\iostream'    MyProject    MyProject\src\main.cpp`

By enabling the `Scan Modules for Module Dependencies` flag, Visual Studio would have also scanned .cpp files and saw that we are trying to import iostream, so it would have built iostream as a header unit.  Then, when we compile, the build would be successful.

Other IDEs which use Visual Studio as a compiler would probably have their own way of determining which files to build.  They might scan .cpp files on their own, or maybe have a setting where you list additional files to compile.

The solutions given by Microsoft are:
1. Enable the `Scan Sources for Module Dependencies` option to scan every source file (increasing compile times).
2. Build a static shared library that imports header files, giving anything that links to it access to them.
3. Manually specifying that a .h file should be compiled as a header unit.  See PR #1570 for supporting CompileAs options.
4. Enable experimental STL modules and import them.

samsinsane (changes requested):
We don't add new flags, please create an API instead.

LoneBoco:
Okay.  It has been converted to an API call now.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2873,6 +2874,19 @@
 	end
 
 
+	function m.scanSourceForModuleDependencies(cfg)
+		if _ACTION >= "vs2019" then
+			if cfg.scanformoduledependencies ~= nil then
```
nickclark2016, line 2879: So if the value isn't specified in the premake project file, we just use the IDE default value?

website/docs/scanformoduledependencies.md
```diff
@@ -0,0 +1,26 @@
+Enables the `Scan Sources for Module Dependencies` option for Visual Studio projects.
+
+```lua
+scanformoduledependencies "value"
+```
+If no value is set for a configuration, Visual Studio will default the option to `No`.
```
nickclark2016, line 6: I don't think we need to specify this.  It should be noted that this is the default IDE value, not `No` based on my previous comment.

modules/vstudio/vs2010_vcxproj.lua
LoneBoco: Yes.  If the option isn't used then nothing gets written to the project file and the IDE's default is used.

website/docs/scanformoduledependencies.md
LoneBoco: I've removed the line.

nickclark2016 (approved):

PR #1790, opened 2022-01-04
Adding C++ module file extensions and implementing "allmodulespublic"

hannes-harnisch:
See #1773. That PR was closed accidentally due to reshuffling commits on request.

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -243,6 +243,7 @@
 				m.generateManifest,
 				m.extensionsToDeleteOnClean,
 				m.executablePath,
+				m.allModulesPublic,
 			}
 		end
 	end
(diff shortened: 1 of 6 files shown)


hannes-harnisch:
@starkos Is this okay?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1511,6 +1512,18 @@
 		end
 	end
 
+	function m.allModulesPublic(cfg)
+		if _ACTION >= "vs2019" then
+			if cfg.allmodulespublic ~= nil then
+				if cfg.allmodulespublic then
+					m.element("AllProjectBMIsArePublic", nil, "true")
```
samsinsane, line 1519: For future reference, you can use `iif(<cond>, "true", "false")` instead of duplicating the line.

samsinsane (approved):

modules/vstudio/vs2010_vcxproj.lua
hannes-harnisch: Which line?

nickclark2016: To follow up, that would take the place of the if-else block. It'd be:

```lua
m.element("AllProjectBMIsArePublic", nil, iif(cfg.allmodulespublic, "true", "false"))
```

PR #1842, opened 2022-03-24
Update shaderobjectfileoutput.md

TylerDahl:
**What does this PR do?**
Update the shaderobjectfileoutput documentation to replace the current auto-generated text.

Thanks for the contribution! Please provide a concise description of the problem this request solves.

Modifies the Premake documentation.

No.

No.

- [ CHECK] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [CHECK ] Add unit tests showing fix or feature works; all tests pass
- [CHECK ] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [ CHECK] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [CHECK ] Minimize the number of commits
- [ CHECK] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

### DIFF
--- a/website/docs/shaderobjectfileoutput.md
+++ b/website/docs/shaderobjectfileoutput.md
@@ -1,12 +1,12 @@
-shaderobjectfileoutput - This page was auto-generated. Feel free to help us improve the documentation by creating a pull request.
+Specifies the output object of compiled HLSL files.
 
 ```lua
-shaderobjectfileoutput (value)
+shaderobjectfileoutput ("path")
 ```
 
 ### Parameters ###
 
-`value` - needs documentation.
+`path` is the output path of HLSL files that have been compiled into Compiled Shader Objects.
 
 ## Applies To ###
 
@@ -18,7 +18,9 @@ Premake 5.0.0 alpha 14 or later.
 
 ### Examples ###
 
+This Visual Studio project will compile HLSL files to the shaders folder with a .cso extension.
+
 ```lua
-shaderobjectfileoutput (value)
+shaderobjectfileoutput "shaders/%%(Filename).cso"
 ```
 


website/docs/shaderobjectfileoutput.md
```diff
@@ -1,12 +1,12 @@
-shaderobjectfileoutput - This page was auto-generated. Feel free to help us improve the documentation by creating a pull request.
+Specifies the output object of compiled HLSL files.
 
 ```lua
-shaderobjectfileoutput (value)
```
nickclark2016, line 4: Let's keep it consistent with other docs. It should be `shaderobjectfileoutput ("path")`. See: https://premake.github.io/docs/objdir/

website/docs/shaderobjectfileoutput.md
```diff
@@ -18,7 +18,9 @@ Premake 5.0.0 alpha 14 or later.
 
 ### Examples ###
 
+This project will compile HLSL files to the shaders folder with a .cso extension.
+
 ```lua
-shaderobjectfileoutput (value)
+shaderobjectfileoutput "shaders/%%(Filename).cso"
```
nickclark2016, line 24: Nit: Choose a slightly simpler example that doesn't involve escaping.

nickclark2016 (changes requested):

website/docs/shaderobjectfileoutput.md
TylerDahl: For Visual Studio projects at least, this path will be dynamic based on the item being compiled.  Without the use of some form of macro all shader files would be compiled with the same file name.  I haven't used premake for other project types but I would assume this would also be the case for projects other than Visual Studio.  I've updated the example to make it clear that this would work specifically for a Visual Studio project which should be sufficient for most users?

KyrietS:
@TylerDahl will you resolve the issues pointed out by @nickclark2016?

TylerDahl:
@KyrietS I've addressed the first issue noted by @nickclark2016 however since DirectX can only be worked on a Windows machine and most users are likely using Visual Studio, I feel this example which took me some time to figure out (and I think would be very useful to many users) is the best example I can think of, as a static string will not do (the file output path needs to be dynamic).

KyrietS:
@nickclark2016 please approve if you agree 🙃

samsinsane (approved):
Going to merge this as-is, it's an improvement over what we had before, if @nickclark2016 has a specific idea around the example we can address that in a future PR.

PR #1850, opened 2022-04-03
! checking if each match found is either a file or folder appears to …

mihaisebea:
…be expensive on large projects (at least on windows)

! for a 5k files project GetAttributesW takes around 20% of the time

**What does this PR do?**

Thanks for the contribution! Please provide a concise description of the problem this request solves.

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -354,7 +354,7 @@
 --    A table containing the matched file or directory names.
 ---
 
-	function os.match(mask)
+	function os.match(mask, matchType)
 		mask = path.normalize(mask)
 		local starpos = mask:find("%*")
 		local before = path.getdirectory(starpos and mask:sub(1, starpos - 1) or mask)
@@ -370,14 +370,14 @@
 
 		if recurse then
 			local submask = mask:sub(1, starpos) .. mask:sub(starpos + 2)
-			results = os.match(submask)
+			results = os.match(submask, matchType)
 
 			local pattern = mask:sub(1, starpos)
 			local m = os.matchstart(pattern)
 			while os.matchnext(m) do
 				if not os.matchisfile(m) then
 					local matchpath = path.join(before, os.matchname(m), mask:sub(starpos))
-					results = table.join(results, os.match(matchpath))
+					results = table.join(results, os.match(matchpath, matchType))
 				end
 			end
 			os.matchdone(m)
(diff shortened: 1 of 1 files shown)


src/base/os.lua
```diff
@@ -354,7 +354,7 @@
 --    A table containing the matched file or directory names.
 ---
 
-	function os.match(mask)
+	function os.match(mask, matchType)
```
samsinsane, line 357: Can you please document the new parameter in the comment block above.

src/base/os.lua
```diff
@@ -387,10 +387,17 @@
 				while os.matchnext(m) do
 				if not (slashpos and os.matchisfile(m)) then
 					local matchpath = path.join(before, matchpath, os.matchname(m))
+					
 					if after then
-						results = table.join(results, os.match(path.join(matchpath, after)))
+						results = table.join(results, os.match(path.join(matchpath, after), matchType))
 					else
+						if matchType == "file" and os.matchisfile(m) then
 						table.insert(results, matchpath)
```
samsinsane, line 395: This should be indented.

samsinsane (changes requested):
LGTM.

samsinsane:
Noticed the tests fail on non-Windows platforms, it fails on `os.remove()` when attempting to remove a directory. `os.remove()` is supposed to fail, but it succeeds now. I can't really figure out what's different other than `os.matchisfile` will check the `dirent->d_type` if it has it before falling back to calling `os.isfile`.

mihaisebea:
> Noticed the tests fail on non-Windows platforms, it fails on `os.remove()` when attempting to remove a directory. `os.remove()` is supposed to fail, but it succeeds now. I can't really figure out what's different other than `os.matchisfile` will check the `dirent->d_type` if it has it before falling back to calling `os.isfile`.

I noticed and it has been puzzling me for a while :(
I need to get my hands on a mac or linux so this will take a bit.

src/base/os.lua
```diff
@@ -387,10 +391,17 @@
...
 					if after then
-						results = table.join(results, os.match(path.join(matchpath, after)))
+						results = table.join(results, os.match(path.join(matchpath, after), matchType))
 					else
-						table.insert(results, matchpath)
+						if matchType == "file" and os.matchisfile(m) then
+							table.insert(results, matchpath)
+						elseif matchType == "folder" and not os.matchisfile(m) then
+							table.insert(results, matchpath)
+						else -- keep previous behaviour
```
samsinsane, line 402: I just realised it's this here, it needs to check that `matchType == nil` or something, otherwise all folders when matching files are picked up, and all files when matching folders are picked up.

Jarod42: In addition the comment is not meaningful outside of the pull request.

src/base/os.lua
```diff
@@ -387,10 +391,17 @@
 				while os.matchnext(m) do
 				if not (slashpos and os.matchisfile(m)) then
 					local matchpath = path.join(before, matchpath, os.matchname(m))
+					
 					if after then
-						results = table.join(results, os.match(path.join(matchpath, after)))
+						results = table.join(results, os.match(path.join(matchpath, after), matchType))
 					else
-						table.insert(results, matchpath)
+						if matchType == "file" and os.matchisfile(m) then
```
Jarod42, line 398: typo? for below too
```suggestion
						if matchType == "file" and os.matchisfile(matchpath) then
```

Jarod42: We are in `if not (slashpos and os.matchisfile(m)) then` block...  so `os.matchisfile(m)` is always false...

samsinsane: That's not how `not (A and B)` works, either `A` or `B` can be false.

Jarod42: Oops, indeed, misread.

PR #1852, opened 2022-04-05
[*] gmake2: fix unescaped ' ' characters in target directory and target

ReeceSX:
Projects, target directories, and object directories didn't account for users including spaces into their respective values.  This commit should fix that by wrapping the values with the global premake.esc function as used by `_x`.

### DIFF
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -268,13 +268,13 @@
 
 
 	function gmake2.target(cfg, toolset)
-		p.outln('TARGETDIR = ' .. project.getrelative(cfg.project, cfg.buildtarget.directory))
-		p.outln('TARGET = $(TARGETDIR)/' .. cfg.buildtarget.name)
+		p.outln('TARGETDIR = ' .. p.esc(project.getrelative(cfg.project, cfg.buildtarget.directory)))
+		p.outln('TARGET = $(TARGETDIR)/' .. p.esc(cfg.buildtarget.name))
 	end
 
 
 	function gmake2.objdir(cfg, toolset)
-		p.outln('OBJDIR = ' .. project.getrelative(cfg.project, cfg.objdir))
+		p.outln('OBJDIR = ' .. p.esc(project.getrelative(cfg.project, cfg.objdir)))
 	end
 
 


nickclark2016:
Can we get a unit test that reflect this?

ReeceSX:
It'd require digging out the [boilerplate test suit](
https://github.com/premake/premake-core/blob/65deb619f8d5579487def157d7c369e5c6d18864/modules/self-test/test_helpers.lua), hacking it in [test_gmake2_makefile](https://github.com/premake/premake-core/blob/master/modules/gmake2/tests/test_gmake2_makefile.lua
) - a file that hasn't been touched since 2018, to validate that the widely used `p.esc` function can be used in this context. 

image

nickclark2016:
Why can we not make a new file for testing it, based off something like this: https://github.com/premake/premake-core/blob/master/modules/gmake2/tests/test_gmake2_ldflags.lua.  It'd be much easier to do than trying to put it into that monolithic test.

While yeah, we don't have a test for it right now, it'd be useful so we know if we break it in the future.  I also want to see testing for what the output would look like in cases where the path has no spaces, where the path has spaces, etc.

ReeceSX:
> Why can we not make a new file for testing it, based off something like this: https://github.com/premake/premake-core/blob/master/modules/gmake2/tests/test_gmake2_ldflags.lua. It'd be much easier to do than trying to put it into that monolithic test.

I have next to no experience with poking in premakes internal test suit; however it appears as though the [test_gmake2_makefile](https://github.com/premake/premake-core/blob/master/modules/gmake2/tests/test_gmake2_makefile.lua) file would cover `TARGET` and `TARGETDIR` part of the generator. The problem is, it depends on a core test util to generate the workspace and project. All the tests across premake assume a project name of `"MyProject"  (... index)`. It'd require modifying the [self-test](https://github.com/premake/premake-core/tree/65deb619f8d5579487def157d7c369e5c6d18864/modules/self-test) module - not just gmake2s test. Or perhaps you could get away with hacking the relevant parts out. Not sure how `getproject` in `prepare` works. I don't have enough experience to produce such a hacky test.

nickclark2016:
Again, that's why I recommended the `test_gmake2_ldflags` as a base instead.  While the change is minor, I'm not comfortable approving without a test, as this modified premake's outputs.

samsinsane:
@ReeceSX You don't need to use the helpers you can just call `workspace` and `project` yourself - they're just the normal API calls you make in a `premake5.lua` script. [`targetdir`](https://premake.github.io/docs/targetdir/), [`targetname`](https://premake.github.io/docs/targetname/) and [`objdir`](https://premake.github.io/docs/objdir) are the APIs you're looking for though as they can be used to control `TARGETDIR`, `TARGET` and `OBJDIR`, respectively.

nickclark2016:
Just wanted to poke on this to see if any progress was made.

ReeceSX:
> Just wanted to poke on this to see if any progress was made.

about as much progress made in the other issues where you expect new contributors to write tests in an alien framework without assistance. 

image

no, you're not getting a unit test for a 3 line change to verify your preface space with a backslash function works in a project with almost 4k commits and recent changes under gmake2/relevant subsytems with no tests to match. 

also you: https://github.com/premake/premake-core/commit/c36e48a10e6c2c69624ec51c44ab51167bf65fd3

PR #1867, opened 2022-04-28
Improve uuid generation on Linux

flakey5:
**What does this PR do?**

Makes use of Linux's uuid header to generate a uuid instead of reading from `/dev/urandom`.

No breaking changes or behavior differences should occur.

This does create the requirement of having the `libuuid`/`uuid-dev` package in order to build

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -248,7 +248,7 @@
 			defines     { "LUA_USE_MACOSX" }
 			links       { "CoreServices.framework", "Foundation.framework", "Security.framework", "readline" }
 			
-		filter "system:linux or macosx"
+		filter "system:linux"
 			links		{ "uuid" }
 
 		filter { "system:macosx", "action:gmake" }
--- a/premake5.lua
+++ b/premake5.lua
@@ -247,6 +247,9 @@
 		filter "system:macosx"
 			defines     { "LUA_USE_MACOSX" }
 			links       { "CoreServices.framework", "Foundation.framework", "Security.framework", "readline" }
+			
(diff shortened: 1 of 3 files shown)


premake5.lua
```diff
@@ -247,6 +247,9 @@
 		filter "system:macosx"
 			defines     { "LUA_USE_MACOSX" }
 			links       { "CoreServices.framework", "Foundation.framework", "Security.framework", "readline" }
+			
+		filter "system:linux or macosx"
```
samsinsane, line 251: Why `or macosx`?

samsinsane (approved):
I think this is fine, it looks like UUID has been available since June 2009, I imagine it's in all of the current LTS releases by now.

premake5.lua
flakey5: Sorry for the late reply. And that would be something I forgot to takeout while messing with it, will fix in a sec

flakey5: Should be fixed now

flakey5:
One more note I've just realized: the 6.x branch uses the [same generation code](https://github.com/premake/premake-core/blob/6.x/core/host/src/host/pmk_uuid.c#L43), so this could also be applied there. If y'all want I can do another pr implementing it. It would be basically the same code change and the only real difference I can see would be in the build files

nickclark2016 (approved):

nickclark2016:
Feel free to stick out a PR for v6.  Given UUID has been around since 2009, I think we're good. That was 13 years ago.

nickclark2016:
Merging.

PR #1899, opened 2022-06-25
Fixes incorrect C# File Mappings for VS

### LINKED ISSUE #1865: Error loading/configuring a C# .net project
Hi everyone! First of all thanks in advance for your help.
I'm trying to set up my C# .net project with premake5 and I'm having the next error:
**Current solution contains incorrect configurations mappings. It may causes projects to not work correctly. Open the Configuration Manager to fix them.**
image

This happens right after generating the project files when I open the Visual Studio solution. 
Which is quite weird is if I press the **"Open Configuration Manager"** button and close the pop up without changing anything. It works.
Apart from that, the bin and obj directories are not properly configurated. I got a bin/ and obj/ generated inside the project folder.

A C# .net project working fine. Also the bin and obj directories generated inside the project folder should be in the **targetdir** and **objdir** specificated locations

I've tried to write my own Directory.Build.props in order to patch some errors but it doesn't work as expected and as I said it's just a patch. I would like premake5 to do this.

You can reproduce this by replicating my little test project. Having installed NET core 6.0 and visual studio tools needed to create a net core project.
This is my directory hierarchy:

- LeafEditor (folder with C# code and project)
    - src (empty folder)
    - premake5.lua (I will call it Project premake)
- premake5.lua (I will call it Workspace premake)
- premake5.exe

The code of Workspace premake looks like this:
```
workspace "LeafEngine"
	startproject "LeafEditor"

	platforms { "x86", "x86_64" }
	flags { "MultiProcessorCompile" }
	
	outputdir = "%{cfg.buildcfg}-%{cfg.system}-%{cfg.platform}"

include "LeafEditor"
```
The code of Project premake looks like this:
```
project "LeafEditor"
	kind "WindowedApp"
	language "C#"
 	csversion "10"
	clr "Unsafe"

	targetdir   ("%{wks.location}/bin/" .. outputdir .. "/%{prj.name}")
 	objdir      ("%{wks.location}/bin-obj/" .. outputdir .. "/%{prj.name}")

	dotnetframework "net6.0"

	files
	{
		"src/**.cs",
		"src/**.xaml",
		"src/**.xaml.cs",
	}

	links
	{
		"Microsoft.CodeAnalysis.CSharp.NetAnalyzers",
		"Microsoft.CodeAnalysis.NetAnalyzers",
		"System.Text.Json.SourceGeneration",
		"Microsoft.NETCore.APP",
	}

filter "system:Windows"
 	defines "LF_WINDOWS"
	
 filter "system:Unix"
 	defines "LF_LINUX"

 filter "configurations:Debug"
 	defines "LF_DEBUG"
 	runtime "Debug"
 	symbols "on"
	 
 filter "configurations:Release"
 	defines "LF_RELEASE"
 	runtime "Release"
 	optimize "full"
```

That is all you need to reproduce the bug.

premake5 (Premake Build Script Generator) 5.0.0-beta1
I tried building for Visual Studio 2019 and 2022.

As I said the project builds once I hit the **"Open Configuration Manager"** but the platform for this build is **"Any CPU"** and I can't change that.

nickclark2016:
**What does this PR do?**

Closes #1865.  Without this, no platforms were ever emitted, which is invalid in VS2022.

Always emits a platform block. This is not tested in VS2019 or earlier.

N/A

### DIFF
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -81,6 +81,14 @@
 		_p(1,'<PropertyGroup>')
 		local cfg = project.getfirstconfig(prj)
 		p.callArray(dotnetbase.elements.projectProperties, cfg)
+		local platforms = prj.platforms
+		if platforms ~= nil and #platforms > 0 then
+			table.replace(platforms, 'Win32', 'x86')
+			table.replace(platforms, 'x86_64', 'x64')
+			_p(2, '<Platforms>%s</Platforms>', table.concat(platforms, ';'))
+		else
+			_p(2, '<Platforms>AnyCPU</Platforms>')
+		end
 		_p(1,'</PropertyGroup>')
 	end
 
--- a/modules/vstudio/tests/cs2005/projectsettings.lua
+++ b/modules/vstudio/tests/cs2005/projectsettings.lua
@@ -48,6 +48,7 @@
 		<AppDesignerFolder>Properties</AppDesignerFolder>
 		<RootNamespace>MyProject</RootNamespace>
 		<AssemblyName>MyProject</AssemblyName>
+		<Platforms>AnyCPU</Platforms>
 	</PropertyGroup>
 		]]
 	end
@@ -67,6 +68,7 @@
 		<AppDesignerFolder>Properties</AppDesignerFolder>
 		<RootNamespace>MyProject</RootNamespace>
 		<AssemblyName>MyProject</AssemblyName>
+		<Platforms>AnyCPU</Platforms>
 	</PropertyGroup>
 		]]
 	end
@@ -90,6 +92,7 @@
 		<TargetFrameworkProfile>
 		</TargetFrameworkProfile>
 		<FileAlignment>512</FileAlignment>
+		<Platforms>AnyCPU</Platforms>
 	</PropertyGroup>
 		]]
 	end
@@ -109,6 +112,7 @@
 		<AssemblyName>MyProject</AssemblyName>
 		<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
 		<FileAlignment>512</FileAlignment>
+		<Platforms>AnyCPU</Platforms>
 	</PropertyGroup>
 		]]
 	end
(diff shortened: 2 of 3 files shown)


modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -81,6 +81,14 @@
 		_p(1,'<PropertyGroup>')
 		local cfg = project.getfirstconfig(prj)
 		p.callArray(dotnetbase.elements.projectProperties, cfg)
+		local platforms = prj.platforms
+		if platforms ~= nil and #platforms > 0 then
+			table.replace(platforms, 'Win32', 'x86')
+			table.replace(platforms, 'x86_64', 'x64')
+			_p(2, '<Platforms>%s</Platforms>', table.concat(platforms, ';'))
```
Jarod42, line 88: I don't see tests to handle that case.

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -81,6 +81,14 @@
...
+		else
+			_p(2, '<Platforms>AnyCPU</Platforms>')
+		end
```
samsinsane, line 91: This should be in a function added to the `dotnetbase.elements.projectProperties` list. Additionally, perhaps the `AnyCPU` path should be limited to VS2022 and higher?

samsinsane (changes requested):

modules/vstudio/vs2005_dotnetbase.lua
nickclark2016: How is that going to interact with things like filters that work on configurations? I definitely agree we should move where it's handled, but we have to think about what else this would affect?  What I'm thinking is:
```lua
filter("configuration:Win32")
```
We would need to map that configuration to x86 somehow in the filter.

samsinsane: Perhaps this isn't the "right" place for this then? This looks to be "project level settings" but uses the first configuration, that's pretty weird. Seems like it should use the project, and then when something might differ on configuration, use the configurations correctly. For example, `dotnetbase.configurationCondition` seems like it would be outputting the wrong thing if you had debug and release output folders?

lolrobbe2: image
image

this is the premake5.lua for a school project, anyhow as you can see i just added the possible architectures for mac (arm64,x64) and the used the architecture prop and visual studio code and rider automatically pick up on this and select the correct architecture!

so maby it would be nice to be able to set a list of possible architectures? and then using filter for example you select the one you need for your platform.

nickclark2016:
Don't plan on working on this any time soon. Closing.

PR #1923, opened 2022-08-07
Add address sanitizer flags

### LINKED ISSUE #1595: Add sanitizers support
Currently you can't generate projects with any type of sanitizers enabled
I suggest, to add the ability to generate projects with sanitizers enabled.

  Sanitizers help with detecting a lot of bugs and issues, that occurs at runtime.

englercj:
**What does this PR do?**

Resolves #1595

1. Implement support for `fsanitize=address` in gcc, clang, msc, and Visual Studio backends.
2. Implement support for `fsanitize=fuzzer` in msc, Visual Studio, and clang backends.

Existing behavior should be unaffected.

There are a number of other `fsanitize` options that are supported by gcc/clang, but not by msc or visual studio. I did not add support for these, but can if it is desired.

    * **I did make a couple unnecessary format changes, but only in files I was already modifying and nothing major.**

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -195,6 +195,7 @@
 				m.clrSupport,
 				m.characterSet,
 				m.platformToolset,
+				m.sanitizers,
 				m.toolsVersion,
 				m.wholeProgramOptimization,
 				m.nmakeOutDirs,
@@ -2503,6 +2504,17 @@
 				end
 			else
 				m.element("PlatformToolset", nil, version)
+			end
+		end
+	end
+
+	function m.sanitizers(cfg)
+		if _ACTION >= "vs2022" then
+			if cfg.flags.AddressSanitizer then
+				m.element("EnableASAN", nil, "true")
+			end
+			if cfg.flags.Fuzzer then
+				m.element("EnableFuzzer", nil, "true")
 			end
 		end
 	end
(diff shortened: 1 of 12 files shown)


englercj:
Removed the `flags` and added a new `sanitize` API.

englercj:
Anything else need to be changed here?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2507,6 +2508,17 @@
 		end
 	end
 
+	function m.sanitizers(cfg)
+		if _ACTION >= "vs2022" and cfg.sanitize then
+			if table.contains(cfg.sanitize, "Address") then
```
nickclark2016, line 2513: Is ASAN 2022+ only as well? I thought it was VS 2019+

nickclark2016 (changes requested):
In general, I'd say it looks good.  As an aside, I see you added "address" sanitizer to GCC (and by proxy clang).  What about fuzzer on Clang? It should be supported as well.

modules/vstudio/vs2010_vcxproj.lua
englercj: Specifically VS2019 16.9, because it isn't supported in earlier VS versions and I have no way to check the specific version I rounded up to the next thing we can actually check.

englercj:
Requested changes have been made.

nickclark2016:
fuzzer = 2022
asan = 2019

nickclark2016 (approved):
LGTM. Leaving open for comments by @starkos or @samsinsane (or anyone else) for a few days, then I'll close. Thanks!

samsinsane (approved):

englercj:
Ah, sorry. They were originally flags which use CamelCase so I had followed that pattern. Forgot to change casing when I switched to a new API.

Issue #1346, opened 2019-10-10
Enable configuration reuse


starkos:
**What problem will this solve?**
Allow project authors to script reusable blocks of configuration, which can then be used for:

- Providing all necessary configuration to consumers of a library or project. Just linking a library is often not enough: preprocessor symbols, include directories, and additional linker dependencies may also be required. This feature would allow project authors to specify all of this configuration, including the `links()` call, in a block that can be reused by consumers.

- Specifying common configuration once, and reusing it across multiple projects and configurations. For instance, setting a standard for how a debug build should be configured, or the listing the settings necessary to use a third-party library.

We've also called this feature "usages", and have talked about them quite a lot (see *"What other alternatives…"* below).

**What might be a solution?**
After going through all of the discussions and my own notes over the years, the example below is how I think it should be implemented. Feel free to disagree with me in the comments; I'll update this write-up to summarize the discussion.

This is a contrived example to show the main points. Please use your imagination to scale it up to meaningful project sizes, with all the filters for platforms, etc. that you would normally need to script. All names are my proposals only and are open for discussion.

```lua
-- Projects are defined as before. Any configuration provided here is
-- "internal", and only used while building the project. To pull in
-- shared configuration, use the `uses` statement.
project "MyLibrary"
    kind "StaticLib"
    files { "**.h", "**.cpp" }
    uses { "MyLibrary:common" }

-- This shared block contains configuration that is used both to build
-- the library, and for consumers of the library. The name is arbitrary 
-- and assigned by the script author. Special characters are okay in 
-- block names.
block "MyLibrary:common"
    filter { "configurations:Debug" }
        defines { "MYLIBRARY_DEBUG" }
    filter { "configurations:Release" }
        defines { "MYLIBRARY_RELEASE" }

-- This shared block contains the configuration required only by 
-- consumers of the library, and not the library project itself.
-- It is okay to use the same name as the project.
block "MyLibrary"
    uses { "MyLibrary:common" }
    defines { "MYLIBRARY_API" }
    links { "MyLibrary", "Dependency1", "Dependency2" }

-- A project that wants to consume the library can pull in the shared
-- "public" configuration block.
project "MyExecutable"
    uses { "MyLibrary" }
```

Some usage points that might not be obvious, or would need to be considered during development:

- A block only gets "used" once. If it appears again in a `uses()` list it will be ignored. For example, if you try to "use" two libraries which share a common dependency.

[...]

PR #1927, opened 2022-08-12
6.x - Implementing project references

kellygravelyn:
**What does this PR do?**

Implements some of the baseline functionality for project references as discussed in https://github.com/premake/premake-core/discussions/1922.

Adds a new `setgroup` field type which is now used for `includeDirs` so that directories are sorted into `public` and `private` groups. Then a new `projectLinks` field (name totally placeholder for now) is used to propagate the `public` directories from projects to the projects referencing them.

This is still a draft. I'm opening it up to start getting some feedback on the overall approach before I go further. I think eventually I'd want to squash this into a few commits:

1. Just adding the `setgroup` field type.
2. Changing `includeDirs` to use `setgroup`.
3. Implementing `projectLinks` (or whatever we want to call it).

At that point I can open them as separate PRs or leave them together, whatever is easiest.

### DIFF
--- a/core/modules/field/src/kind_setgroup.lua
+++ b/core/modules/field/src/kind_setgroup.lua
@@ -57,8 +57,17 @@ local function receive(field, inner, currentValues, incomingValues, plain)
 	end
 
 	array.forEach(field.groups, function(k)
-		if incomingValues[k] then
-			incomingValues[k] = inner(field, nil, incomingValues[k], plain)
+		local value = incomingValues[k]
+		if value then
+			if type(value) == 'table' then
+				value = array.map(value, function(v)
+					return inner(field, nil, v, plain)
+				end)
+			else
+				value = inner(field, nil, value, plain)
+			end
+
+			incomingValues[k] = value
 		end
 	end)
 
--- a/core/modules/field/src/kind_setgroup.lua
+++ b/core/modules/field/src/kind_setgroup.lua
@@ -1,0 +1,85 @@
+---
+-- Stores items in a group of sets. The groups are defined when the field is
+-- registered. Also supports the idea of a defaultGroup where items are added
+-- if not explicitly put into a group.
+---
+
+local array = require('array')
+local set = require('set')
+local table = require('table')
+
+local Field = select(1, ...)
+
+
+local function default(field, inner)
+	local defaultValue = {}
+	array.forEach(field.groups, function(g)
+		defaultValue[g] = {}
+	end)
+	return defaultValue
+end
+
+
+local function match(field, inner, currentValues, pattern, plain)
+	error('todo')
+end
+
+
+local function merge(field, inner, currentValues, incomingValues, plain)
+	currentValues = currentValues or field:default()
+
+	if type(incomingValues) ~= 'table' then
+		incomingValues = { incomingValues }
+	end
+
+	array.forEach(field.groups, function(k)
+		if incomingValues[k] then
+			currentValues[k] = set.appendArrays(currentValues[k], incomingValues[k])
+		end
+	end)
+
+	if field.defaultGroup then
+		currentValues[field.defaultGroup] = set.appendArrays(currentValues[field.defaultGroup], incomingValues)
+	end
+
+	return currentValues
+end
+
+
+local function pattern(field, inner, pattern)
+	error('todo')
+end
+
+
+local function receive(field, inner, currentValues, incomingValues, plain)
+	if type(incomingValues) ~= 'table' then
+		incomingValues = { incomingValues }
+	end
+
+	array.forEach(field.groups, function(k)
+		if incomingValues[k] then
+			incomingValues[k] = inner(field, nil, incomingValues[k], plain)
+		end
+	end)
+
+	for i = 1, #incomingValues do
+		incomingValues[i] = inner(field, nil, incomingValues[i], plain)
+	end
+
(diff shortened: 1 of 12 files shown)


nickclark2016:
I should be able to take a look at this tomorrow. Can you provide a few example usages?

kellygravelyn:
There is a full test case for the VS exporter that shows the example usage: https://github.com/premake/premake-core/pull/1927/files#diff-fa88cf5607400dc83da753d5fe6142e4f3ff736f9bb23ff2442d08dd8bb86d4eR98-R139

nickclark2016:
That's awesome. I really like it. I'd love to get @starkos and @samsinsane to give their feedback. I'll try to do a code deep dive tomorrow.

core/modules/field/src/kind_setgroup.lua
```diff
@@ -0,0 +1,94 @@
+---
```
nickclark2016, line 1: Just to sanity check, this would be things like tasks in Makefiles or filters in VS, correct?

core/modules/field/tests/setgroup_field_tests.lua
```diff
@@ -0,0 +1,102 @@
+local Field = require('field')
```
nickclark2016, line 1: General comment on tests that's not directed at you, but we should probably go through the tests and comment them with their case, their preconditions, and their postconditions.

core/modules/helpers/helpers.lua
```diff
@@ -0,0 +1,52 @@
...
+
+function helpers.fetchAllIncludeDirs(configOrProject)
+	-- Determine if we're working on a project level or a configuration level.
+	-- The logic is pretty much the same, but for configurations we'll look up
+	-- the same named configuration on the linked project.
+	local isProject = configOrProject.project == nil
+
+	local includeDirs = set.join(configOrProject.includeDirs.public, configOrProject.includeDirs.private)
+
+	-- Use an array of projects to visit plus a set of projects we've visited
```
nickclark2016, line 21: Should we do some kind of prevalidation to enforce an acyclic graph?

core/modules/field/src/kind_setgroup.lua
kellygravelyn: set group isn't really constrained to any existing model. The idea of the field kind is to simply be a collection of sets that are named. So for `includeDirs` I defined them as `private` and `public`. The same set group field type could be used for `links` if you wanted a set of `static` and `shared` links. Anywhere there's a field where you want categories could be applicable here.

core/modules/helpers/helpers.lua
kellygravelyn: That'd probably be good if it's not already in there somewhere. I still think this is useful here for non-acyclic graphs in the case that you have something like A references B and C, and B also references C. Otherwise you'll visit C twice and duplicate include directories. I think the comment here could probably be improved to mention that since it's a good reason to have this set logic even if we guarantee acyclic graphs.

core/modules/field/src/kind_setgroup.lua
nickclark2016: Ahh, we may want to consider a different name then. `group` is already taken (or at least it was in v5 and I see no reason to change that API name in v6).

core/modules/helpers/helpers.lua
```diff
@@ -0,0 +1,52 @@
+local set = require('set')
```
nickclark2016, line 1: Not 100% sure that a new helpers file would be the correct place for this.  What about building this into the existing DOM?

nickclark2016:
Pretty happy with this as it stands. Feel free to remove the draft status whenever you're happy with it and I'll take another look at it.

core/modules/helpers/helpers.lua
kellygravelyn: I thought about putting it into the DOM but I wasn't sure how tightly we wanted to keep the DOM isolated from the various fields and stuff. Right now the dom files don't seem to operate on any of the fields so I didn't know if it'd be right to add this there. But I do agree it might be cleaner. Then you can just do `config:fetchAllIncludeDirs()` or `proj:fetchAllIncludeDirs()` without needing to use a new helpers module.

core/modules/field/src/kind_setgroup.lua
kellygravelyn: Yeah this one is `setgroup` but I can see the potential confusion. I'll see if there's a different name I can use.

core/modules/helpers/helpers.lua
nickclark2016: Very true. Especially once we move to C# and such, having includeDirs functionality in DOM may not make ass much sense.

nickclark2016:
Thinking more on this, I think this is a sizeable feature set for a PR and we may want to do "projectLinks" and "includeDirs" separately, just to keep the PRs smaller.  This sound reasonable?

kellygravelyn:
> Thinking more on this, I think this is a sizeable feature set for a PR and we may want to do "projectLinks" and "includeDirs" separately, just to keep the PRs smaller. This sound reasonable?

Yep I was at least going to squash down to fewer commits but I'll go ahead and do them as separate PRs when they're ready.

nickclark2016:
So I just had a chance to speak with @starkos and he pointed me to #1346.  I know it'd be a significant rework, but thoughts?  It's definitely another path we could take, which I believe he would rather we look at than the table based includeDir approach.

PR #1952, opened 2022-09-08
vstudio: add usestandardpreprocessor option

flakey5:
**What does this PR do?**

Adds the `usestandardpreprocessor` option for Visual Studio. Tells msbuild to use a standard-conforming preprocessor.

Doesn't really change much, just adds another config option.

Nope

### DIFF
--- a/modules/vstudio/tests/vc2019/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2019/test_compile_settings.lua
@@ -106,3 +106,27 @@
 	<ScanSourceForModuleDependencies>false</ScanSourceForModuleDependencies>
 		]]
 	end
+
+	function suite.UseStandardPreprocessorOn()
+		usestandardpreprocessor 'On'
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<Optimization>Disabled</Optimization>
+	<UseStandardPreprocessor>true</UseStandardPreprocessor>
+		]]
+	end
+
+	function suite.UseStandardPreprocessorOff()
+		usestandardpreprocessor 'Off'
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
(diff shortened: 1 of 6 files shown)


Jarod42:
Shouldn't src/tools/msc.lua be updated too.

flakey5:
Ah yeah, thanks for pointing that out. Added it to there as well.

modules/vstudio/tests/vc2019/test_compile_settings.lua
```diff
@@ -106,3 +106,27 @@
 	<ScanSourceForModuleDependencies>false</ScanSourceForModuleDependencies>
 		]]
 	end
+
+	function suite.UseStandardPreprocessorOn()
+		usestandardpreprocessor 'on'
```
nickclark2016, line 111: You have the API as "On" and "Off", then you're using it as "on" and "off" here. Is it case agnostic?  If so, can we add a test for each case (just as a sanity check).

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2955,6 +2956,16 @@
 		end
 	end
 
+	function m.useStandardPreprocessor(cfg)
+		if _ACTION >= "vs2019" and cfg.usestandardpreprocessor ~= nil then
+			if cfg.usestandardpreprocessor == 'On' then
+				m.element("UseStandardPreprocessor", nil, "true")
+			else
```
nickclark2016, line 2963: We force as "On" or "Off".  What happens if we get a value that isn't "On" or "Off"?

nickclark2016 (changes requested):

modules/vstudio/vs2010_vcxproj.lua
flakey5: It would just get defaulted to false, which is the default option on msvc anyways. I could add a check for it and error if that'd be better

flakey5:
Force pushed to squash commits

modules/vstudio/vs2010_vcxproj.lua
samsinsane: @nickclark2016 it shouldn't be possible to get a value that isn't `On` or `Off` due to the API having an `allowed` list with only those two values.

samsinsane (approved):

nickclark2016 (approved):

PR #1954, opened 2022-09-09
Add newoption.catagory to documentation, mark os.is as deprecated

### LINKED ISSUE #1859: List user defined options separated with premake system options when using "premake --help"
When I'm using "premake --help", the options I created in premake5.lua mixed up with the options provided by premake system.

**What might be a solution?**
List user defined options separated with premake system options when using "premake --help"

thomashope:
**What does this PR do?**

Closes #1859 

Adds a description of the `category` option when using `newoption`.

Also marks `os.is()` as deprecated which is mentioned in the same issue.

No. Should only be docs changes.

Nope.

### DIFF
--- a/website/docs/newoption.md
+++ b/website/docs/newoption.md
@@ -10,11 +10,12 @@ newoption { description }
 
 | Field       | Description                                                                        |
 |-------------|------------------------------------------------------------------------------------|
-| trigger     | What the user would type on the command line to select the option, e.g. "--name". |
+| trigger     | What the user would type on the command line to select the option, e.g. `--name`. |
 | description | A short description of the option, to be displayed in the help text. |
 | value       | Optional. If the option needs a value, provides a hint to the user what type of data is expected. |
 | allowed     | Optional. A list of key-value pairs listing the allowed values for the option. |
 | default     | Optional. Sets the default for this option if not specified on the commandline. |
+| category    | Optional. Places the option under a separate header when the user passes `--help` |
 
 
 ### Availability ###
@@ -32,6 +33,7 @@ newoption {
    value       = "API",
    description = "Choose a particular 3D API for rendering",
    default     = "opengl",
+   category    = "Build Options",
    allowed = {
       { "opengl",    "OpenGL" },
       { "direct3d",  "Direct3D (Windows only)" },
--- a/website/docs/os.is.md
+++ b/website/docs/os.is.md
@@ -1,3 +1,7 @@
+:::caution
+**This function has been deprecated.** Use [os.target()](os.target.md) or [os.host()](os.host.md) instead.
+:::
+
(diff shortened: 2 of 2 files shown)


website/docs/newoption.md
```diff
@@ -10,11 +10,12 @@ newoption { description }
 
 | Field       | Description                                                                        |
 |-------------|------------------------------------------------------------------------------------|
-| trigger     | What the user would type on the command line to select the option, e.g. "--name". |
+| trigger     | What the user would type on the command line to select the option, e.g. `--name`. |
 | description | A short description of the option, to be displayed in the help text. |
 | value       | Optional. If the option needs a value, provides a hint to the user what type of data is expected. |
 | allowed     | Optional. A list of key-value pairs listing the allowed values for the option. |
 | default     | Optional. Sets the default for this option if not specified on the commandline. |
+| category    | Optional. Places the option under a seperate header when the user passes `--help` |
```
Jarod42, line 18:
```suggestion
| category    | Optional. Places the option under a separate header when the user passes `--help` |
```

thomashope: Thanks! Fixed with a force push to minimise commits. Sorry if that means you miss out on credit...

nickclark2016 (approved):

nickclark2016:
Will leave this open for comment for a few days, then merge if no changes need to be made.

website/docs/os.is.md
```diff
@@ -4,6 +4,8 @@ Checks the current operating system identifier against a particular value.
 os.is("id")
 ```
 
+**This function has been deprecated.** Use [os.target()](os.target.md) or [os.host()](os.host.md) instead.
```
KyrietS, line 7: Maybe we could use fancy-dancy [Admonitions](https://premake.github.io/docs/Style-Guide/#admonitions) for the first time on our website? 😀

It would be:
```
:::caution
**This function has been deprecated.** Use [os.target()](os.target.md) or [os.host()](os.host.md) instead.
:::
```

I've tested it on my local demo branch for `configuration`: https://kyriets.github.io/premake-core/docs/configuration/

What do you think?

thomashope: Ah yes, I was actually looking for something like this.

Happy to make this change and also make a separate PR that uses them for all deprecation notices?

thomashope: On second thoughts, maybe it's clearer to leave this PR as is and make a new one that does the change everywhere

nickclark2016: I think we use it here now, in case no one gets around to the follow up PR to do it everywhere immediately.

thomashope:
Bump, any other change requests?

KyrietS:
Thanks for your contribution 🤗

PR #1975, opened 2022-10-04
Strip linking decorators for sibling projects

### LINKED ISSUE #1685: Links with `:static` suffix do not generate correct target dependencies
Given this sample project where `links` is used with `:static` suffix to enable putting the static library in static section of link libraries (`-Wl,-Bstatic`):
```lua
workspace "MyWorkspace"
    configurations { "Debug" }
    toolset "gcc"

    project "MyLib"
        kind "StaticLib"
        files { "lib.cpp" }

    project "MyApp"
        kind "ConsoleApp"
        libdirs { "bin/Debug" }
        links { "MyLib:static" }
        files { "app.cpp" }
```
When premake is called to generate gmake Makefiles, and make is invoked to build MyApp:
```
$ premake5 gmake
$ make MyApp
```

Then it fails to build with the compiler not finding the MyLib:
```
==== Building MyApp (debug) ====
Creating obj/Debug/MyApp
app.cpp
Creating bin/Debug
Linking MyApp
/usr/bin/ld: cannot find -lMyLib
collect2: error: ld returned 1 exit status
make[1]: *** [MyApp.make:54: bin/Debug/MyApp] Error 1
make: *** [Makefile:31: MyApp] Error 2
```

This is because MyApp in generated Makefiles does not include a dependency on MyLib target. Therefore, MyLib is never built and the linking phase of MyApp fails.

Generated Makefiles should contain the correct dependency of MyApp on MyLib and `make MyApp` should build first MyLib, then MyApp successfully.

The only workaround is to manually build MyLib first, then MyApp.

Run the example given above.

premake5 (Premake Build Script Generator) 5.0.0-alpha16

When not using the `:static` suffix, the dependencies are correct and the build works as expected:

However, that is not a valid workaround for the project I am working on. In this particular case, putting the library in `-Wl,-Bstatic` section is quite important for building the project correctly.

samsinsane:
- Added unit tests for linking decorators

**What does this PR do?**

Fixes #1685 by ignoring the linking decorators for sibling projects.

Specifying a linking decorator on a sibling project will result in it treating it as a sibling project not a system library.

N/A

### DIFF
--- a/tests/config/test_links.lua
+++ b/tests/config/test_links.lua
@@ -222,3 +222,45 @@
 		local r = prepare("all", "fullpath")
 		test.isequal({ "bin/Debug/MyProject2.lib" }, r)
 	end
+
+--
+-- Linking decorators need to be passed through
+--
+
+	function suite.canLink_StaticDecoratorSystemLib()
+		links { "SystemLibrary:static" }
+		local r = prepare("all", "fullpath")
+		test.isequal({ "SystemLibrary:static" }, r)
+	end
+
+	function suite.canLink_SharedDecoratorSystemLib()
+		links { "SystemLibrary:shared" }
+		local r = prepare("all", "fullpath")
+		test.isequal({ "SystemLibrary:shared" }, r)
+	end
+
+--
+-- Linking decorators need to be stripped for sibling projects
+--
+
+	function suite.canLink_StaticDecoratorSiblingLib()
+		links { "SiblingLibrary:static" }
+
+		project "SiblingLibrary"
+		kind "StaticLib"
+		language "C++"
+
+		local r = prepare("all", "fullpath")
+		test.isequal({ "bin/Debug/SiblingLibrary.lib" }, r)
+	end
+
+	function suite.canLink_SharedDecoratorSiblingLib()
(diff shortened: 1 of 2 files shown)


tests/config/test_links.lua
```diff
@@ -222,3 +222,45 @@
...
+
+	function suite.canLink_StaticDecoratorSiblingLib()
+		links { "SiblingLibrary:static" }
+
+		project "SiblingLibrary"
+		kind "StaticLib"
+		language "C++"
+
+		local r = prepare("all", "fullpath")
+		test.isequal({ "bin/Debug/SiblingLibrary.lib" }, r)
```
nickclark2016, line 254: ".lib" feels hard coded to Windows file extension conventions.

nickclark2016, line 254: ".lib" feels hard coded to Windows file extension conventions.

samsinsane: Yeah, the whole file uses Windows as the target, I decided to keep with that since the logic applies to all platforms not just Linux.

nickclark2016: Sounds good.

nickclark2016 (approved):

PR #1997, opened 2022-11-11
Add validation for toolset.

Jarod42:
**What does this PR do?**

Add validation for toolset.
So `valid_tools` is now useful.

Add a message for invalid configuration for toolset

No.

### DIFF
--- a/src/base/action.lua
+++ b/src/base/action.lua
@@ -257,6 +257,45 @@
 		return false
 	end
 
+---
+-- Determines if an action supports a particular toolset.
+--
+-- @param language
+--    The language that toolset belongs.
+-- @param toolset
+--    The toolset to check.
+-- @returns
+--    True if the toolset is supported, false otherwise.
+---
+	function action.supportsToolset(language, toolset)
+		if not language or not toolset then
+			return true
+		end
+		local self = action.current()
+		if not self then
+			return false
+		end
+		local language_keys_map = {
+			["C"] = "cc",
+			["C++"] = "cc",
+			["C#"] = "dotnet",
+			["D"] = "dc",
+		}
+		local language_key = language_keys_map[language]
+		if not language_key then
+			p.warn("Unknown mapping for language %s", language)
+			return true
+		end
+		if not self.valid_tools then
+			return true
+		end
+		local valid_tools = self.valid_tools[language_key]
+		if not valid_tools then
+			return true
+		end
+
+		return table.contains(valid_tools, toolset)
+	end
 
 --
 -- Determines if an action supports a particular configuration.
--- a/src/base/validation.lua
+++ b/src/base/validation.lua
@@ -56,6 +56,7 @@
 		return {
 			m.projectHasLanguage,
 			m.actionSupportsLanguage,
+			m.actionSupportsToolset,
 			m.actionSupportsKind,
 			m.projectRulesExist,
 			m.projectValuesInScope,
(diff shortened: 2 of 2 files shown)


src/base/validation.lua
```diff
@@ -174,6 +175,13 @@
 	end
 
 
+	function m.actionSupportsToolset(prj)
+		if not p.action.supportsToolset(prj.language, prj.toolset) then
+			p.warn("Unsupported toolset '%s' used for language '%s' for project '%s'", prj.toolset, prj.language, prj.name)
```
Jarod42, line 180: Most other validations use `warn` and not `error`...
So I stay coherent with above checks. (but I think all should be `error`).

src/base/action.lua
```diff
@@ -257,6 +257,47 @@
...
+-- @param language
+--    The language that toolset belongs.
+-- @param toolset
+--    The toolset to check.
+-- @returns
+--    True if the toolset is supported, false otherwise.
+---
+
+	function action.supportsToolset(language, toolset)
+		local function buildLanguageKeysMap()
```
Jarod42, line 272: Not sure how it should be done with extern registered languages. Apply to `valid_tools` entry or registered action anyway...

src/base/action.lua
```diff
@@ -257,6 +257,47 @@
...
+			local language_keys_map = {C = "cc", D = "dc" }
+			language_keys_map["C++"] = "cc"
+			language_keys_map["C#"] = "dotnet"
```
samsinsane, line 275: I'd like to see this be consistent:
```lua
local language_keys_map = {
  ["C"] = "cc",
  ["C++"] = "cc",
  ["C#"] = "dotnet",
  ["D"] = "dc",
}
```

src/base/action.lua
```diff
@@ -257,6 +257,47 @@
...
+			return language_keys_map
+		end
+		if not language or not toolset then
+			return true
+		end
```
samsinsane, line 280: Can you comment the logic here? If neither is specified, why does it return true?

Can you do the same for the other `if not <thing> then` blocks that return true?

src/base/action.lua
```diff
@@ -257,6 +257,47 @@
...
+		local self = action.current()
+		if not self then
+			return false
+		end
+		local language_keys_map = buildLanguageKeysMap()
+		local language_key = language_keys_map[language]
+		if not language_key then
+			print("unknown mapping for language %s", language)
```
samsinsane, line 288: Was the choice of `print` based on another similar function, or should this be `p.warn`?

samsinsane (approved):
Overall, this looks good. I'm a bit concerned that it might be emitting the warning on valid uses since we probably haven't been updating the `valid_tools` along the way, but we can just make sure they're all fixed up before the next release.

src/base/action.lua
Jarod42: As above function in fact. we pass `prj.language`, `prj.toolset` which might be not provided, so `nil`.
So handle that case.
For unknown/unsupported languages (rust, F#), it is just to be future proof and avoid warning.

src/base/action.lua
Jarod42: Changed to `p.warn`

src/base/action.lua
Jarod42: Indeed, changed.

Jarod42:
Tests actually fail as master fails.

PR #2011, opened 2022-12-13
Add support for unity builds

Sharlock93:
**What does this PR do?**

Adds Support for Unity builds in visual studio 2019+. Adds one new flag "enableUnityBuild" with options On and Off.

Adds Unity Build/Jumbo Build support.

Unity builds really speed up compile time, being able to set them per config and project is a nice feature and something I personally needed.

Honestly I did this in a hurry and was a neat feature that I wanted myself, if there is anything in particular I should be adding then feedback is welcome.


Jarod42:
Documentation and unit test are missing.

Wonder if it should be integrated to https://github.com/dcourtois/premake-compilationunit which does [Single compilation unit](https://en.wikipedia.org/wiki/Single_Compilation_Unit) support.

src/_premake_init.lua
```diff
@@ -1512,6 +1512,17 @@
 		tokens = true,
 	}
 
+	api.register {
+		name = "enableUnityBuild",
+		scope = { "config", "rule" },
```
Jarod42, line 1517: `"rule"` seems suspicious...

src/_premake_init.lua
```diff
@@ -1512,6 +1512,17 @@
 		tokens = true,
 	}
 
+	api.register {
```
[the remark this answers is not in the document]
Sharlock93: I honestly didn't see that. Thanks for pointing it out.

src/_premake_init.lua
Sharlock93: Initially when I added this, "config" didn't work, so I added "rule" and that seemed to work. didn't find enough documentation around what it really means but also didn't look that hard.

Sharlock93:
> Documentation and unit test are missing.
> 
> Wonder if it should be integrated to https://github.com/dcourtois/premake-compilationunit (listed in https://premake.github.io/community/modules/#tool-modules) which does [Single compilation unit](https://en.wikipedia.org/wiki/Single_Compilation_Unit) support.

The difference here being that Visual Studio itself does that kind of "Unity/Single Compilation Unit" automatically. For example if you enable it in a project it will generate a file called "unity_<random_suffix>.cpp" that basically does what that module is doing.

I haven't added more flags (honestly because I didn't need them myself) but a more complete set of flags exist for Unity/Jumbo builds that specify how many files should be packed up in one Unity file etc.

image

That module is more useful in a context where native support for Unity builds doesn't exist.

Jarod42:
> > Documentation and unit test are missing.
> 
> If you could point me to where I should be adding those it would be great.

I would say somewhere in premake-core\modules\vstudio\tests\vc2010
 
> > Wonder if it should be integrated to https://github.com/dcourtois/premake-compilationunit (listed in https://premake.github.io/community/modules/#tool-modules) which does [Single compilation unit](https://en.wikipedia.org/wiki/Single_Compilation_Unit) support.
> 
> The difference here being that Visual Studio itself does that kind of "Unity/Single Compilation Unit" automatically. For example if you enable it in a project it will generate a file called "unity_<random_suffix>.cpp" that basically does what that module is doing.
> 
> That module is more useful in a context where native support for Unity builds doesn't exist.

I meant: use special flag for msvc which can do it natively, but keep the module as is for other actions.

Sharlock93:
> I meant: use special flag for msvc which can do it natively, but keep the module as is for other actions.

I mean you could do that but this feature really should be part of visual studio as this is a native feature on visual studio itself so in essence premake is missing it.

nickclark2016:
Can you add the documentation so the CI can pass that step?

website/docs/unitybuilds.md
```diff
@@ -0,0 +1,29 @@
+Enables Unity Builds in Visual Studio, also known as Jumbo Builds
+
+```lua
+enableUnityBuilds "value"
+```
+
+If no toolset is specified for a configuration, the system or IDE default will be used.
```
Jarod42, line 7: "toolset"? Seems to be a copy paste...

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -2550,6 +2551,12 @@
 		end
 	end
 
+	function m.enableJumboBuild(cfg)
+		if _ACTION > "vs2017" and cfg.enableUnityBuild then
```
Jarod42, line 2555:
```suggestion
		if _ACTION >`= "vs2019" and cfg.enableUnityBuild then
```
To be coherent with surrounding conditions.

website/docs/unitybuilds.md
Sharlock93: yeah 100%, I wanted to keep the same structure and easiest was to copy paste something, been so busy lately that this ended up being a rushed job.

modules/vstudio/vs2010_vcxproj.lua
Sharlock93: Ah yes my bad, unity builds exist on 2017, updated the docs.

Jarod42: So 
```suggestion
		if _ACTION >= "vs2017" and cfg.enableUnityBuild then
```

nickclark2016:
Missing one more step in the documentation (and maybe this is something we should better document in the contributing doc), but you need to add it to the `sidebar.js` since it's a new API.

Sharlock93:
> Missing one more step in the documentation (and maybe this is something we should better document in the contributing doc), but you need to add it to the `sidebar.js` since it's a new API.

yeah sorry about that, been super busy with work and it could  be that its written somewhere but I didn't see it.

website/docs/Project-API.md
```diff
@@ -72,6 +72,7 @@
 | [editandcontinue](editandcontinue.md)                     |  |
 | [editorintegration](editorintegration.md)                 | Enable or disable IDE integration |
 | [enablewarnings](enablewarnings.md)                       |  |
+| [enableunitybuilds](unitybuilds.md)                       |  |
```
KyrietS, line 75: Our current convention in docs is that file names are the same as the function that they describe. You have added the `enableUnityBuilds` API, so the file should be named `enableUnityBuilds.md`. Lua is case-sensitive, so the above line is also misleading 😉 Please change it to:

```markdown
| [enableUnityBuilds](enableUnityBuilds.md)                       |  |
```

KyrietS (changes requested):

website/docs/Project-API.md
Sharlock93: I fixed up all the naming to match what is already there, I hope this is more consistent now.

website/sidebars.js
```diff
@@ -275,6 +275,7 @@ module.exports = {
 						'toolsversion',
 						'transition',
 						'undefines',
+                        'unitybuilds',
```
KyrietS, line 278: `enableunitybuild`

KyrietS (changes requested):
Docs

website/sidebars.js
Sharlock93: one of these days...

samsinsane (approved):

KyrietS (approved):

sparkskapil:
When I try to use this option I get this error.
```attempt to call a nil value (global 'enableunitybuild')```

Am I missing something here?

nickclark2016:
What version of premake are you using?

sparkskapil:
> What version of premake are you using?

Premake 5.0.0-beta2

nickclark2016:
Try one of the CI builds. They should have it.

PR #2023, opened 2023-01-17
Fix support of openmp for visual studio with clang toolset.

Jarod42:
**What does this PR do?**

Fix support of openmp for visual studio with clang toolset.

only visual generator are impacted

Also tested with [my testing repo](https://github.com/Jarod42/premake-sample-projects)
especially action https://github.com/Jarod42/premake-sample-projects/actions/runs/3941604370

### DIFF
--- a/modules/vstudio/vs2010_vcxproj.lua
+++ b/modules/vstudio/vs2010_vcxproj.lua
@@ -1645,6 +1645,10 @@
 		elseif _ACTION >= "vs2019" and cfg.toolset and cfg.toolset == "clang" then
 			local value = iif(cfg.unsignedchar, "On", "Off")
 			table.insert(opts, p.tools.msc.shared.unsignedchar[value])
+			-- <OpenMPSupport>true</OpenMPSupport> is unfortunately ignored with clang toolset
+			if cfg.openmp == "On" then
+				table.insert(opts, 1, '/openmp')
+			end
 		end
 
 		if #opts > 0 then
--- a/modules/vstudio/tests/vc2019/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2019/test_compile_settings.lua
@@ -79,6 +79,20 @@
 		]]
 	end
 
+	function suite.openmpOnWithClang()
+		toolset "clang"
+		openmp "On"
+		prepare()
+		test.capture [[
+<ClCompile>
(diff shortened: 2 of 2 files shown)


modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1644,6 +1644,10 @@
 		elseif _ACTION >= "vs2019" and cfg.toolset and cfg.toolset == "clang" then
 			local value = iif(cfg.unsignedchar, "On", "Off")
 			table.insert(opts, p.tools.msc.shared.unsignedchar[value])
+			-- <OpenMPSupport>true</OpenMPSupport> is unfortunately ignored with clang toolset
+			if cfg.openmp == "On" then
+				table.insert(opts, 1, '-openmp')
+			end
```
samsinsane, line 1650: Could this use the same as the above two lines and use the MSC toolset value? It looks like Clang is fine with `/J`, so it might be fine with `/openmp` and `/openmp-`? It looks like the "real" Clang flag is `-fopenmp` based on [this](https://github.com/premake/premake-core/blob/f4e88bbd925433181f52ec31a2a96c95ddfef238/src/tools/gcc.lua#L70-L72).

Thoughts?

samsinsane (changes requested):

modules/vstudio/vs2010_vcxproj.lua
Jarod42: Changed to `'/openmp'` which works too. (`-fopenmp` doesn't work IIRC, msvc translates option to clang).

samsinsane: Sorry, what I meant is could this be written like the above lines which don't duplicate the flag but instead get it out of MSC? So, instead of:
```lua
table.insert(opts, 1, '/openmp')
```
it would be:
```lua
table.insert(opts, 1, p.tools.msc.shared.openmp[cfg.openmp])
```

I'm just not sure if it would accept `/openmp-` (from `openmp "Off"`) as a valid flag?

Jarod42:
Tests actually fail as master fails.

modules/vstudio/vs2010_vcxproj.lua
samsinsane: @Jarod42 to clarify, that's not me requesting this change but asking if you think it's worth it?

samsinsane (approved):

PR #2028, opened 2023-01-30
Add support for idirafter flag in GCC/Clang

### LINKED ISSUE #2021: Add support for vendored includes on linux.
There is a problem with some (most) linux distro-provided mozjs-*/js*package (or whatever spidermonkey (any version) is packaged as on that system), in that it contains a copy of icu headers that don't match the version of icu installed on the system. This mismatch causes a linking issue in [0 A.D.'s Empire Ascendant](play0ad.com) build process, preventing us to use system provided libraries.

**What might be a solution?**
[This revision ](https://code.wildfiregames.com/D3127#change-rhfEiPNVQqRG )contains a possible solution, by telling gcc (and clang, as it follows enough of the same code path within premake) to look through the mozjs-* includes directory after the system standard directories. This means that gcc/clang will look for icu in the system standard location (and find it) before searching the mozjs-* location.

Keeping a bundled copy of the library forever.

nickclark2016:
**What does this PR do?**

Adds support for the `-idirafter` flag with GCC and Clang in the gmake2 and codelite exporters.

No changes to the existing Premake behavior.  Introduces `includedirsafter` API.  For GCC and Clang utilizing the gmake2 or Codelite exporters, the directories specified by `includedirsafter` will utilize `-idirafter` flag for includes.  For XCode using GCC or Clang, directories defined by `includedirsafter` will be at the end of the `SYSTEM_HEADER_SEARCH_PATHS` block.  For Visual Studio using MSVC or Clang, directories defined by `includedirsafter` will at the end of the `ExternalIncludePath` element if in VS2019 or later or in the `IncludePath` element if 2017 or before. Similarly, if the MSC toolset is used, `/external:I` will be utilized for `includedirsafter` when using `msc-142` or newer, or `/I` for `msc-141` or earlier.

Closes #2021


src/tools/gcc.lua
```diff
@@ -318,6 +336,11 @@
 			table.insert(result, '-isystem ' .. p.quoted(dir))
 		end
 
+		for _, dir in ipairs(includedirsafter or {}) do
```
Jarod42, line 339:
```suggestion
		for _, dir in ipairs(includedirsafter or cfg.includedirsafter or {}) do
```
"Future" proof for generators not in core (might also apply to previous parameters too).

nickclark2016: I tend to disagree there. That'll change the behavior of those exporters, potentially in unexpected ways. I see this as a responsibility of the caller, not this function.

Jarod42: `includedirsafter` would be ignored by unmodified generators.
With my proposal, there is (a big) chance to have expected behavior.
Else it would be broken too (in another way) and requires changes anyway (pass `cfg.includedirsafter` for your version, pass `{}` for mine, and handle `includedirsafter` another way.

codelite/gmake2 might be unmodified ;-)
xcode/visual handle it differently anyway

gmake support seems missing BTW.

nickclark2016: `includedirsafter` would be ignored by unmodified generators in this implementation, since arguments are optional.  Correct, XCode and Visual Studio handle it differently, as they don't support functionality like `idirafter` provides. GMake support is intentionally missing, as it's deprecated.

If we do change the behavior of how this works, I think it'd be better to have this as a separate PR, as it's out of scope of the current functionality.  As written, the new behavior matches the expected behavior from previous `includedirs` and friends.

nickclark2016: On top of this, if any 3rd party module passing `nil` to the function with your mechanism to attempt "skip" a parameter (rather than use `{}`) would break, as it would now check `cfg.####` before just putting in `{}`.  Given we're to beta, I'm a bit hesitant to change existing behavior.

Jarod42: As usual, we disagree ;-) on expected behavior.
But ok.
(I will have to create 3 PRs to add support to the external generators ^_^)

Shouldn't doc specify that it is unsupported by gmake?
(even if deprecated, some features only works with gmake and not gmake2 (D, premake-qt, ...)
(fixing it seems simpler btw).

nickclark2016: I think the plan is to completely drop gmake support in 5.0 full release.  We can update the documentation to explicitly call out lack of gmake support, though.

Jarod42: As long as gmake is here, either:
- Update it (add `cfg.includedirsafter` to `toolset.getincludedirs` in gmake)
- Specify in includedirsafter.md than gmake is unsupported.

First solution seems simpler/better to me.

tests/tools/test_msc.lua
```diff
@@ -308,6 +308,23 @@
 		test.contains("/external:I/usr/local/include", msc.getincludedirs(cfg, cfg.includedirs, cfg.externalincludedirs))
 	end
 
+--
+-- Check handling includedirsafter.
+--
+
+function suite.cflags_onIncludeDirsAfter()
+	includedirsafter { "/usr/local/include" }
```
Jarod42, line 316: Inhabituel path for msc ^_^
(but ok).

website/docs/includedirsafter.md
```diff
@@ -0,0 +1,39 @@
...
+does not support include directory ordering, these directories are added to the external include directory path.
+
+```lua
+includedirsafter { "paths" }
+```
+
+### Parameters ###
+
+`paths` specifies a list of include file search directories. Paths should be specified relative to the currently running script file.  Search order is evaluated from
+right to left.
```
Jarod42, line 11: According to https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html
it is left-to-right (for gcc at least).

website/docs/includedirsafter.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Applies To ###
+
+Project configurations.  GCC and Clang are the only toolsets supporting the ordering functionality in the gmake, gmake2 and Codelite exporters.  All exporters and toolsets
```
Jarod42, line 15: Shouldn't "remarks" go in Availability?

website/sidebars.js
```diff
@@ -235,6 +236,7 @@ module.exports = {
 						'rules',
 						'runpathdirs',
 						'runtime',
+						'sanitize',
```
Jarod42, line 239: This fix is unrelated to this PR ;-)

website/docs/includedirsafter.md
nickclark2016: Good catch. I'm pretty sure it was in my head correct, I just wrote it down wrong.

tests/tools/test_msc.lua
nickclark2016: Not gonna lie, just stole these from above it in the file

website/sidebars.js
nickclark2016: Yeah, I just had it locally and didn't catch it.  Want me to back this out and put in a separate PR?

Jarod42: As you prefer.

but if you keep it, you have to uncheck:
 ;-)

nickclark2016: Unchecked

Jarod42:
For info, UT added in my [repo](https://github.com/Jarod42/premake-sample-projects/tree/includedirsafter_test) (using this branch), and as expected, core-generators pass, external module generators fail (see [premake-sample-projects/actions includedirsafter_test](https://github.com/Jarod42/premake-sample-projects/actions))

samsinsane (approved):

Jarod42:
@samsinsane : UTs are broken with that merge (due to previous merges).

samsinsane:
@Jarod42 thanks for pointing that out, I'll get that fixed up. Sorry about that!

PR #2038, opened 2023-03-03
wrap OBJDIR creation in prevent gmake warning

### LINKED ISSUE #726: Make warnings if objdir and targetdir are the same

thomashope:
## Wrap OBJDIR creation in conditional to prevent defining two rules with the same name

**What does this PR do?**

Make doesn't want you to create two rules with the same name and will emit a warning if you do (unless they are double colon rules which have some caveats?).

If you set the `targetdir` and `objdir` to the same location in Premake two rules are produced with the same name, as described in #726, causing make to complain.

Wraps the `OBJDIR` rule in a conditional so it is not run if `OBJDIR` and `TARGETDIR` are the same.

closes #726

I'm not an expert on makefiles but in my research I saw two possible solutions to this issue.

1. Change the rules to [double colon rules](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html) which did work but possibly has some caveats.

2. Wrap a conditional around one of the rules. Apparently make has the concept of *immediate* and *deferred* contexts which I looked into a bit but seems to not be an issue for this case. I was reading about the concepts [here](https://www.gnu.org/software/make/manual/html_node/Reading-Makefiles.html#index-ifeq_002c-expansion).

I selected option 2 because as far as I could tell I didn't have any relevant caveats and seemed the least surprising for someone reading the makefile / Lua code.

Tests were not included because the warning was only visible when you run the makefile and I wasn't sure how / if it was worthing testing that.

### DIFF
--- a/modules/gmake2/gmake2.lua
+++ b/modules/gmake2/gmake2.lua
@@ -129,10 +129,12 @@
 		_p('endif')
 	end
 
-	function gmake2.mkdirRules(dirname)
+	function gmake2.mkdirRules(dirname, before, after)
+		if before then _p(before) end
 		_p('%s:', dirname)
 		_p('\t@echo Creating %s', dirname)
 		gmake2.mkdir(dirname)
+		if after then _p(after) end
 		_p('')
 	end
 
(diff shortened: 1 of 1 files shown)


nickclark2016:
I believe Visual Studio may actually have an issue with the intermediate and target directories being the same. We may want to do this at a core level, rather than a per-exporter level.

thomashope:
Hmm I thought I'd had a VS project setup in the past that put them in the same place and it didn't mind, although I didn't test when investigating this issue.

Just tried on Xcode and it seemed happy to use the same dir for both

nickclark2016:
It's very possible I'm thinking of another quirk with intermediate and target paths.

Jarod42:
(deprecated) gmake can have similar fix.

PR #2045, opened 2023-03-22
Update checkaction.md

Stolas:
Quick and Lazy documentation patch

### DIFF
--- a/website/docs/checkaction.md
+++ b/website/docs/checkaction.md
@@ -1,4 +1,5 @@
-checkaction - This page was auto-generated. Feel free to help us improve the documentation by creating a pull request.
+checkaction - What should happend when an assertion fails (D Lang)
+
 
 ```lua
 checkaction (value)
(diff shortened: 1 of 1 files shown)


website/docs/checkaction.md
```diff
@@ -1,4 +1,5 @@
-checkaction - This page was auto-generated. Feel free to help us improve the documentation by creating a pull request.
+checkaction - What should happend when an assertion fails (D Lang)
```
KyrietS, line 1: - [ ] The `checkaction -` is redundant. We use filename as a title then you can start right away with a description.
- [ ] It's important to be more explicit. Instead of `(D Lang)` add: _This option applies only to **D Programming Language**_

website/docs/checkaction.md
```diff
@@ -9,10 +10,12 @@ checkaction (value)
...
-* `D`: needs documentation
-* `C`: needs documentation
-* `Halt`: needs documentation
-* `Context`: needs documentation
+* `D`: call D assert on failure
+* `C`: call C assert on failure
+* `Halt`: cause program halt on failure
+* `Context`: call D assert with the error context on failure
+
+See https://dlang.org/library/dmd/globals/checkaction.html
```
KyrietS, line 18: Use proper Markdown signature for hyperlinks: `[text](url)`

website/docs/checkaction.md
```diff
@@ -9,10 +10,12 @@ checkaction (value)
...
 
 ## Applies To ###
```
KyrietS, line 21: You need to align the example, because it is not valid now. Use: `checkaction "Halt"`

KyrietS (changes requested):
Thank you for taking your time! Please use `git commit --amend` when resolving my comments

PR #2047, opened 2023-03-30
Fix for premake-qt

### LINKED ISSUE #1916: gmake2 action looses custom build rules added programmatically.
I'm "maintaining" a small addon used to add support for Qt (https://github.com/dcourtois/premake-qt/)
In this addon, I'm overriding a few Premake functions to scan project files, add generated ones, and setup their build commands. It works fine for most actions, but the custom build commands are completely ignored by the gmake2 one.

I've narrowed down the problem to this line: https://github.com/premake/premake-core/blob/ea2971dd50d048582d646d12d9e50253f6f6731c/modules/gmake2/gmake2_cpp.lua#L178

This line "loses" the `buildOutputs` in the case of build commands added by my addon: if I print the number of `buildOutputs` of `filecfg`, it shows 1, but on `shadowContext`, it's 0.
Thing is if I add a custom rule in my project (using the usual `filter` and `buildcommands`, etc.) then `shadowContext` still retain the `buildOutputs`.

I'm rusty in Lua and Premake and my Decoda stopped working for some reason, so I'm having a bit of trouble to understand if the issue is a bug in gmake2 or in my addon, and I'm not sure how to fix this.

For reference, here is how my addon adds custom build rules: https://github.com/dcourtois/premake-qt/blob/a79003c5ae2589a1d6ea84838d92ca30116ad4b2/qt.lua#L470
And the generated files are added directly to the list of files in my override of `premake.oven.bakeFiles`: https://github.com/dcourtois/premake-qt/blob/a79003c5ae2589a1d6ea84838d92ca30116ad4b2/qt.lua#L337

Any help, pointer, etc. appreciated!

Jarod42:
**What does this PR do?**

closes #1916
Fix for premake-qt for gmake2
- `shadowContext` is wrong (case typo of `PathVars` BTW).
- Remove duplicated generated `node` entries.

gmake2 generator fixed.
modify `table.indexof` to allow to have predicate too.

Tested with https://github.com/Jarod42/premake-sample-projects/actions/runs/4563673420

### DIFF
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -163,31 +163,19 @@
 
 		-- process custom build commands.
 		if fileconfig.hasCustomBuildRule(filecfg) then
-			local env = table.shallowcopy(filecfg.environ)
-			env.PathVars = {
-				["file.basename"]     = { absolute = false, token = node.basename },
-				["file.abspath"]      = { absolute = true,  token = node.abspath },
-				["file.relpath"]      = { absolute = false, token = node.relpath },
-				["file.name"]         = { absolute = false, token = node.name },
-				["file.objname"]      = { absolute = false, token = node.objname },
-				["file.path"]         = { absolute = true,  token = node.path },
-				["file.directory"]    = { absolute = true,  token = path.getdirectory(node.abspath) },
-				["file.reldirectory"] = { absolute = false, token = path.getdirectory(node.relpath) },
-			}
-
-			local shadowContext = p.context.extent(filecfg, env)
-
-			local buildoutputs = p.project.getrelative(cfg.project, shadowContext.buildoutputs)
+			local buildoutputs = p.project.getrelative(cfg.project, filecfg.buildoutputs)
 			if buildoutputs and #buildoutputs > 0 then
 				local file = {
 					buildoutputs  = buildoutputs,
 					source        = node.relpath,
-					buildmessage  = shadowContext.buildmessage,
-					buildcommands = shadowContext.buildcommands,
-					buildinputs   = p.project.getrelative(cfg.project, shadowContext.buildinputs)
+					buildmessage  = filecfg.buildmessage,
+					buildcommands = filecfg.buildcommands,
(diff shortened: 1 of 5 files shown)


modules/gmake2/gmake2_cpp.lua
```diff
@@ -163,31 +163,19 @@
...
-				["file.abspath"]      = { absolute = true,  token = node.abspath },
-				["file.relpath"]      = { absolute = false, token = node.relpath },
-				["file.name"]         = { absolute = false, token = node.name },
-				["file.objname"]      = { absolute = false, token = node.objname },
-				["file.path"]         = { absolute = true,  token = node.path },
-				["file.directory"]    = { absolute = true,  token = path.getdirectory(node.abspath) },
-				["file.reldirectory"] = { absolute = false, token = path.getdirectory(node.relpath) },
-			}
-
-			local shadowContext = p.context.extent(filecfg, env)
```
samsinsane, line 178: It's probably better to fix this than to remove it. `%{file.directory}` and `%{file.abspath}` should give an absolute path, but won't without the shadow context changes above - the `env.PathVars` changing to `env.pathVars` will fix that.

samsinsane (changes requested):
I'm not sure this PR meets the requirements of `Focus on a single fix or feature; remove any unrelated formatting or code changes` - how does removing the shadow context fix premake-qt? What was broken in premake-qt?

modules/gmake2/gmake2_cpp.lua
Jarod42: Not sure what `p.context.extent` does, but I go from `#filecfg.buildOutputs == 1` to `#shadowContext.buildOutputs == 0`.

Other generators don't have those... (and indeed absolute path is not respected :/ )

Jarod42:
https://github.com/dcourtois/premake-qt/pull/27 should fix the issue.

Need to replace 
```
fcfg.buildcommands	= { command }
```
by
```
premake.configset.store(fcfg._cfgset, premake.field.get("buildcommands"), { command })
```

Not sure if there is a proper way (especially to not use *private* `_cfgset`).

PR #2053, opened 2023-04-07
Added "see also" for `files` for common use case.

WraithGlade:
**What does this PR do?**

I was initially confused about whether `files` included its files in the search path or not (I assumed yes at first), but then found that you have to use `includedirs` to do that.

So, I added a comment here for that, so people will figure this out more easily.

It doesn't. This is just a documentation suggestion.

No.

- [ Y] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [ N/A] Add unit tests showing fix or feature works; all tests pass
- [ N/A] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [ N/A] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [ Y] Minimize the number of commits
- [ Y] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

### DIFF
--- a/website/docs/files.md
+++ b/website/docs/files.md
@@ -50,3 +50,4 @@ filter "system:MacOSX"
 * [Adding Source Files](Adding-Source-Files.md)
 * [Removing Values](Removing-Values.md)
 * [vpaths](vpaths.md)
+* [includedirs](https://premake.github.io/docs/includedirs/) (Whereas `files` is used to determine which files are included in a generated project's view of what files are associated with that project, `includedirs` is used to determine *where to automatically search for source files* when including/importing them. Being aware of this distinction is important if you want to use a non-standard project folder structure.) 
--- a/website/docs/conformancemode.md
+++ b/website/docs/conformancemode.md
@@ -24,6 +24,7 @@ Premake 5.0.0 beta 1 or later.
 conformancemode(true)   -- better for cross-platform projects
 conformancemode(false)  -- has access to more platform-specific features
 
--- Parentheses are required in this case because Lua only supports omitting them if the parameter is a single string or table.
+-- Parentheses are required in this case because Lua only supports 
+-- omitting them if the parameter is a single string or table.
 ```
 
--- a/website/docs/conformancemode.md
+++ b/website/docs/conformancemode.md
@@ -1,12 +1,14 @@
-conformancemode - This page was auto-generated. Feel free to help us improve the documentation by creating a pull request.
+This function sets a boolean flag that controls whether or not the generated projects should adhere ("conform") to a *stricter* interpretation of the language standard or not. 
 
 ```lua
-conformancemode (value)
+conformancemode(is_strict)
 ```
 
 ### Parameters ###
 
-`value` - needs documentation.
+If `is_strict` is true (and applicable to the generated project's capabilities), the generated project will use stricter language standard rules (e.g. C and C++ ISO standards) when building the project. This will increases the chances of the build system and compiler reporting problems with the code you write and build that may not be as portable between compilers or target systems. However, it may also decrease how many platform-specific build system and compiler features are available to you.
+
+On the other hand, if `is_strict` is false then the opposite tradeoff will apply: more features/permissiveness, but less cross-platform compatibility.
 
 ## Applies To ###
 
@@ -19,6 +21,9 @@ Premake 5.0.0 beta 1 or later.
 ### Examples ###
 
 ```lua
-conformancemode (value)
+conformancemode(true)   -- better for cross-platform projects
+conformancemode(false)  -- has access to more platform-specific features
+
+-- Parentheses are required in this case because Lua only supports omitting them if the parameter is a single string or table.
 ```
 


website/docs/files.md
```diff
@@ -50,3 +50,4 @@ filter "system:MacOSX"
 * [Adding Source Files](Adding-Source-Files.md)
 * [Removing Values](Removing-Values.md)
 * [vpaths](vpaths.md)
+* [includedirs](https://premake.github.io/docs/includedirs/) (Whereas `files` is used to determine which files are included in a generated project's view of what files are associated with that project, `includedirs` is used to determine *where to automatically search for source files* when including/importing them. Being aware of this distinction is important if you want to use a non-standard project folder structure.)
```
KyrietS, line 53: > `files` is used to determine which files are included in a generated project's view of what files are associated with that project

I don't understand this sentence.  Generally, `files` is used to determine which c/cpp files **will be compiled and then linked** into a binary. From a compilation point of view, it has nothing to do with "project's view" or header files. I know that you can use this option to "add" header files to see them in Visual Studio project's view, but please remember that Premake is not only for generating Visual Studio projects.

In CMake you have this distinction as well:

```cmake
add_executable(my_binary
    src/file1.cpp
    src/file2.cpp
)

target_include_directories(my_binary
    include/
    vendor/include/
)
```

If you look at the command that is executed during the compilation of your project (msvc, gcc, clang) you will see that _source files_ and _include directories_ are separate parameters passed to the compiler.

I would even say that "_being aware of this distinction_" is not only important but also crucial to survive in C/C++ world 😄

KyrietS (changes requested):

WraithGlade:
Alright, thanks for the info and response.

I only just started using Premake, so the only thing I had as a basis for
the `conformancemode` was the fact that I used grep on Windows to find the
only occurrence of "conformancemode" that I could find in the whole source
file folder and it said that it was a "boolean". I couldn't even find
wherever the substantive implementation of any of `conformancemode`
actually is/was. Perhaps this is a case of abstraction making that opaque.

In any case, I might not be using Premake for a while after all, given that
the widespread lack of basic feature documentation is quite disconcerting
for a project that has been around for so long.

Lua is certainly a better language than CMake though, and I plan on looking
into the Premake project again sometime in the future.

I've only used premake a little bit yesterday and a little bit today.

*I'm not in a position to make the changes that you want myself since I
don't really know what you preferences are and have little familiarity with
Premake.*


*Feel free to go ahead and edit the commit or whatever else you feel like
doing. I don't mind at all.*

KyrietS:
Actually, my fondness for Premake was born out of a hatred of the CMake language and its syntax 😄

> I couldn't even find wherever the substantive implementation of any of `conformancemode` actually is/was. Perhaps this is a case of abstraction making that opaque.

Every parameter in Premake should have unit tests. If you used grep on Premake repo you would find it. They describe what parameters would be set on *.vcxproj level. Here is an example:

```lua
--
-- If ConformanceMode flag is set, add <ConformanceMode> element (supported from VS2017)
--

	function suite.onConformanceModeOn_VS2017()
		conformancemode "On"
		p.action.set("vs2017")
		prepare()
		test.capture [[
<ClCompile>
	<WarningLevel>Level3</WarningLevel>
	<ConformanceMode>true</ConformanceMode>
		]]
	end
```

I will close this PR now but thanks for getting involved 🙃

WraithGlade:
Thanks you for your efforts and time.

When I used grep on the repo I searched the `src` folder of the Premake
source code, assuming there would very likely be implementation details for
it in there besides just the small spec item I found.

Searching through the tests folder too would have been a smarter move
admittedly, but I just randomly didn't think of that at the time.

I don't appreciate the snark via the " 🙃 " etc though. I took time and
effort to help you to document something that is a basic feature and never
should have been left undocumented. Approximate documentation is still much
better than none. I have little/no Premake experience, but it's like they
say "if you want something done, you (often) have to do it yourself". Why
is a beginner needing to document such a fundamental feature *themselves*?
There is no productive reason to be sarcastic/rude. I had just decided to
*probably* not use Premake for the next project within an hour of getting
your requested for changes back. I'm not in a position to know your
preferences well and don't want to have to dance around them when I have no
familiarity with the project. This entire scenario wouldn't have happened
if all the basics of your system that a newbie would be likely to read had
at least a small amount of documentation already. That's not unreasonable
to say the least.

Personally I love writing documentation and it always seems strange to me
how many programmers avoid writing it.

Anyway though, I do hope Premake unseats CMake and also improves its
documentation to levels that would make people feel far more likely to be
comfortable adopting it.

Have a great day and a great weekend! The spring weather sure is nice so
I'm looking forward to it personally! 🏞😎

KyrietS:
> I don't appreciate the snark via the " 🙃 " etc though.

Sorry, I didn't think you can interpret it that way. I use it a lot 😅

It's true that Ptemake has a lot of documentation debt. At the time, before I got involved into the project, many many parameters were added without proper documentation. I wouldn't allow it now but it happened and we have what we have.

Please note that _conformancemode_ is not a basic feature. It is a single flag that you can switch on or off. It will only generate `<ConformanceMode>true/false</ConformanceMode>` in your *.vcxproj file. In CMake they don't have it. You just have to pass the raw flag that will be used by the compiler:

```cmake
add_compile_options(/permissive)
```

You can do it in Premake too:

```lua
buildoptions { "/permissive" }
```

Many functions in Premake are there just for convenience but you don't have to use them to achieve the result that you want.

Premake is really nice once you get to know it better. I share your hopes for improved Premake documentation 👍

PR #2057, opened 2023-04-12
Fix building on OpenBSD

nekobbbbbbit:
**What does this PR do?**
Fixes building on OpenBSD

It fixes building on OpenBSD

Curl piece works

- [x] Fix building on OpenBSD

I will write a port on OpenBSD this weekend for premake5. maybe

Also worth noting:

```
os_match.c(obj/Release/Premake5/os_match.o:(os_matchstart)): warning: strcpy() is almost always misused, please use strlcpy()
function function> os_uuid.c(obj/Release/Premake5/os_uuid.o:(os_uuid)): warning: sprintf() is often misused, please use snprintf()
function function function function> path_getabsolute.c(obj/Release/Premake5/path_getabsolute.o:(do_getabsolute)): warning: strcat() is almost always misused, please use strlcat()
function function function function function function> lmathlib.c(lmathlib.o:(math_random) in archive bin/Release/liblua-lib.a): warning: random() may return deterministic values, is that what you want?
```

You can peek the first 3 cases for yourself if that matters. PS: Should use internal Curl in anyway?

### DIFF
--- a/contrib/curl/lib/strerror.c
+++ b/contrib/curl/lib/strerror.c
@@ -187,8 +187,10 @@ curl_easy_strerror(CURLcode error)
   case CURLE_TELNET_OPTION_SYNTAX :
     return "Malformed telnet option";
 
+#if LIBCURL_VERSION_NUM < 0x073e00
   case CURLE_PEER_FAILED_VERIFICATION:
     return "SSL peer certificate or SSH remote key was not OK";
+#endif
 
   case CURLE_GOT_NOTHING:
     return "Server returned nothing (no headers, no data)";
@@ -214,9 +216,11 @@ curl_easy_strerror(CURLcode error)
   case CURLE_SSL_CIPHER:
     return "Couldn't use specified SSL cipher";
 
+#if LIBCURL_VERSION_NUM >= 0x073e00
   case CURLE_SSL_CACERT:
     return "Peer certificate cannot be authenticated with given CA "
       "certificates";
+#endif
 
   case CURLE_SSL_CACERT_BADFILE:
     return "Problem with the SSL CA cert (path? access rights?)";


samsinsane (changes requested):
This change seems wrong. What issue are you trying to resolve?

nekobbbbbbit:
> This change seems wrong. What issue are you trying to resolve?

```
# gmake -f Bootstrap.mak bsd
telnet.c
   ...snip!
Linking zip-lib
tftp.c
timeval.c
transfer.c
../../contrib/curl/lib/strerror.c:217:8: error: duplicate case value 'CURLE_PEER_FAILED_VERIFICATION'
  case CURLE_SSL_CACERT:
       ^
/usr/local/include/curl/curl.h:659:26: note: expanded from macro 'CURLE_SSL_CACERT'
#define CURLE_SSL_CACERT CURLE_PEER_FAILED_VERIFICATION
                         ^
../../contrib/curl/lib/strerror.c:190:8: note: previous case defined here
  case CURLE_PEER_FAILED_VERIFICATION:
       ^
1 error generated.
url.c
vauth.c
version.c
vtls.c
warnless.c
gmake[2]: *** [curl-lib.make:655: obj/Release/curl-lib/strerror.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [Makefile:86: curl-lib] Error 2
gmake[1]: Leaving directory '/home/nekobit/junk/premake-core/build/bootstrap'
gmake: *** [Bootstrap.mak:122: bsd] Error 2
```

i wouldn't be surprised if this is supposed to be a warning that GCC or whatever on other systems permits. OpenBSD's default makeopts causes an error.

The fix is presumably correct, considering if you look at the definitions at curl.h:659 and see how things happened... Actually, the fix is overkill, but I did this and decided it was still correct so I left it.

(shouldn't hard-rely on contrib anyway)

nekobbbbbbit:
> [Update curl to 7.53.1](https://github.com/premake/premake-core/commit/af7187407a5f1c1ed93ea492b262d96b5f8781a2)

Even worse, actually. This is a security issue too. Libcurl has had up to 77 security vulnerabilities since that release...

samsinsane:
Looks like a configuration issue that results in the system libcurl headers being used instead of the ones in contrib.

nekobbbbbbit:
So what is the *correct* way to fix this issue in your opinion/regarding this projects goals? There's multiple paths. The best way is to slowly deprecate this "contrib" path for things like libcurl and all that, or find a way to automate the downloading process for libcurl. Everything seems unusual here, specifically the outdated libcurl copy.

What is the intention of it?

samsinsane:
> So what is the correct way to fix this issue in your opinion/regarding this projects goals?

To me, the correct fix would be to use the contrib headers. I'm not really sure how to fix that though, but maybe you have some ideas as an OpenBSD user?

> What is the intention of it?

I assume you mean the contrib folder as a whole not just curl, I'm not sure what the initial intention was. However, having worked on Premake for awhile, the benefit is that everything "just works" on all platforms - obviously excluding configuration issues like you're experiencing. Beyond your preferred compiler and make, there's no additional dependencies required to build Premake, you just `<make> -f Bootstrap.mak <os>` and you have a build of Premake.

nekobbbbbbit:
> To me, the correct fix would be to use the contrib headers.

What about outdated libraries for example? In this case, libcurl. Why can't we just extract a tarball at the start instead of bundling and maintaining our own libcurl library

samsinsane:
> What about outdated libraries for example? In this case, libcurl.

Use the contrib headers - we build the 7.53.1 source, it should use the 7.53.1 headers.

> Why can't we just extract a tarball at the start instead of bundling and maintaining our own libcurl library

Thanks for the suggestion.

nekobbbbbbit:
> Use the contrib headers - we build the 7.53.1 source, it should use the 7.53.1 headers.

This specific contrib header is outdated though, should we bump it?

Interestingly, I asked for bagder's (libcurl dev) input on this, he obviously agreed the old version was bad, but defaulting to premake's libcurl over systems libcurl is also bad. He also mentioned that some distributions and OS's ship modified versions of libcurl. 

I think the best solution would be to make the contrib headers act as a fallback if the system headers cannot be found.

samsinsane:
@nekobbbbbbit I don't want to be rude here, but I've told you twice now what my expectation for the correct solution is. I'm only one of a few people that make decisions regarding Premake, I'm not going to make a drastic decision around the contrib dependencies because of your opinion.

Whether we upgrade libcurl, or download the source from another location, the problem on OpenBSD is still the same. It is using the wrong files. That problem doesn't go away just because we use a newer set of files, it just probably won't cause compilation problems - for now.

nekobbbbbbit:
> However, having worked on Premake for awhile, the benefit is that everything "just works" on all platforms

And yet, it doesn't work on a _default_ configuration of OpenBSD. I couldn't even tell you how things are going on NetBSD or other situations. Premake should be able to work with different configurations.

> it just probably won't cause compilation problems - for now.

I understand what's wrong here, but that isn't my point for now. The fix I submitted was a quick hack.

> because of your opinion.

You are now bundling an outdated libcurl. This has nothing to do with opinion. Clearly nobody seems motivated to want to bump the outdated libcurl and organize the directories. It is assumed to be maintained, and not 6 years out of date.

Go back to how I mentioned that many distributions have their own libcurl patches built in. Does this help portability to you? I understand that the intention is to make it "just work," but it may hurt portability for the nicher configurations.

Maybe this benefits Windows users and macOS users, sure. But nearly all Unix-like operating systems have their own ways of handling libraries already, and it's supposed to assume the distributor knows what they are doing and the packager should offer ways for distributors to work with that. If not, patches are required, but some patches are nice when sent upstream, so everyone is benefited. More importantly, a build system made for... well, portable building... should offer those ways. Providing ways for distributors to easily use their own libcurl libraries _by default, when detected (!!)_ would not only eliminate this issue from occuring, but would free worries of outdated libraries and allow for easier distribution.

Please remember I am offering to put premake5 in the OpenBSD ports tree, since I use it for my projects. I am trying to help you guys here so I don't have to write a bunch of patches. I am willing to write another patch and close this PR which can ease that transition. I just need an OK from you or someone so I can ensure I'm not wasting my time.

nekobbbbbbit:
> Can you figure out that problem first?

That's because Premake only generates gmake files. It's not a problem really, it's working as intended. I just run `gmake` to build premake. Are you asking if you want a BSD-make and/or POSIX-make generator?

samsinsane:
Your compiler ignores flags because it was invoked from gmake?

nekobbbbbbit:
> Your compiler ignores flags because it was invoked from gmake?

Elaborate?

samsinsane:
> Elaborate?

I was replying to an earlier edit of your comment - you really shouldn't edit your comments so many times.

Since it seems like this was an issue on your end, and this patch isn't going to be merged. I'm going to close this PR. You can open an issue if you want to discuss the dependencies situation further.

nekobbbbbbit:
Sure, apologies for the edits.

PR #2060, opened 2023-04-13
Add unistd.h to libzip to prevent build failure with Xcode 14.3

shinra-electric:
**What does this PR do?**

Without this PR, Premake will fail to build using Xcode 14.3. This PR adds `unistd.h` to three files in libzip which will allow the build to succeed. 

Allows the build to succeed. 

**Are there any breaking changes? Will any existing behavior change?**
Not that I'm aware of.

From my understanding, `unistd.h` is for posix-compliant OSes, so ideally I should ifdef these changes so they are not added to non-posix OSes. 

`zip_close.c` has the check `#ifdef HAVE_UNISTD_H` which I thought would be appropriate, but building still fails when using this. 

If anyone can suggest a better define, let me know and I'll add it.

### DIFF
--- a/contrib/libzip/mkstemp.c
+++ b/contrib/libzip/mkstemp.c
@@ -43,6 +43,7 @@
 #endif
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #ifndef O_BINARY
 #define O_BINARY 0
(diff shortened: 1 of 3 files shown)


Jarod42:
Isn't contrib 3rd party? so existing repo? shouldn't we upgrade to more up-to-date and propose fix to there repo?

shinra-electric:
Well yes, but it is part of this repo. That usually means that there are changes from upstream. 

If we are just using stock libzip shouldn't we have it as a submodule or something?

shinra-electric:
Closing, as PR #2117 does the same thing and has already been merged.

Issue #350, opened 2015-12-01
Support TCC


purger:
I'd like to request the support for tinyc (tcc) as it's gaining a lot of use lately for code generation.

tvandijck:
I think we need a little more information on this, a few links etc would help. Doing a cursory search on google for tincy yields pictures of indian women, and a link to the urban dictionary for tincy...

I quote:
"tincy
noun - Often correlated to an has obsession with hijinx; known to serve animals of short stature; 
adj. - resembling a behavior of craziness and picking up royalty's poop
"

samsinsane:
Do you mean [tinyc](http://bellard.org/tcc/)? It appears to be thoroughly dead, no updates in over 2 years and the main developer has officially dropped support for it. Is there an active fork or something?

purger:
@samsinsane Yeah, sorry, I managed to misspell it :) The active fork is at http://repo.or.cz/w/tinycc.git

purger:
I would like to add that in CMake/gmake etc, I can simply set CC=tcc and everything works. Is there anything similar I can do in premake5?

TurkeyMan:
I actually did this last night... just didn't push cus I didn't test it properly ;)

purger:
@TurkeyMan That was fast. Push and I will test :)

TurkeyMan:
@purger https://github.com/TurkeyMan/premake-tcc

It makes makefiles that look okay... haven't tested it much more than that.
This is very niche; you're probably best to tweak it till it works right for you and submit PR's for fixes.
Don't consider this officially supported ;)

purger:
Doesn't seem to work on Win

TurkeyMan:
Never tried it on windows. What's the problem? Pathing?
tools/tcc.lua should be pretty self evident, you can tweak it til it works
and submit a PR if you like?

PR #2068, opened 2023-04-25
Add TCC support

swagtoy:
**What does this PR do?**

Add's TCC support

Add's TCC support

Needs review and tests (?)

bumps #350

### DIFF
--- a/packages/debian/premake.1
+++ b/packages/debian/premake.1
@@ -21,6 +21,7 @@ Remove all binaries and build scripts.
 Choose a C/C++ compiler, if supported by target; one of:
       gcc       GNU gcc compiler
       dmc       Digital Mars C/C+ compiler (experimental)
+      tcc       Tiny C Compiler
 .TP 8
 .B "\-\-dotnet name"
 Choose a .NET compiler set, if supported by target; one of:
--- a/src/tools/tcc.lua
+++ b/src/tools/tcc.lua
@@ -1,0 +1,498 @@
+---
+-- tcc.lua
+-- Provides Tiny C Compiler-specific configuration strings.
+-- Copyright (c) 2002-2023  Perkins and the Premake project
+---
+
+	local p = premake
+
+	p.tools.tcc = {}
+	local tcc = p.tools.tcc
+
+	local project = p.project
+	local config = p.config
+
+
+--
+-- Returns list of C preprocessor flags for a configuration.
+--
+
+	tcc.cppflags = {
+		system = {
+			haiku = "-MMD",
+			wii = { "-MMD", "-I$(LIBOGC_INC)", "$(MACHDEP)" },
+			_ = { "-MD" }
+		}
+	}
+
+	function tcc.getcppflags(cfg)
+		local flags = config.mapFlags(cfg, tcc.cppflags)
+		return flags
+	end
+
+
+--
+-- Returns string to be appended to -g
+--
+	function tcc.getdebugformat(cfg)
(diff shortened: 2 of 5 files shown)


nekobbbbbbit:
I like TCC for testing programs due to it's incredible speed, so that is why I made this PR

src/tools/tcc.lua
```diff
@@ -0,0 +1,500 @@
...
+
+	local project = p.project
+	local config = p.config
+
+
+--
+-- Returns list of C preprocessor flags for a configuration.
+--
+
+	tcc.cppflags = {
```
Jarod42, line 20: Most stuff seems to be a copy of (old) gcc functions/tables.
if true, maybe do as clang and just do `tcc.XXX = gcc.XXX`
I said *old* as some stuff seems missing (standard "C17", openmp, ...)

packages/debian/premake.1
```diff
@@ -21,6 +21,7 @@ Remove all binaries and build scripts.
 Choose a C/C++ compiler, if supported by target; one of:
```
Jarod42, line 21: > **if supported by target**

`$(action).valid_tools` si checked now, So it should be added there properly.

src/tools/tcc.lua
nekobbbbbbit: It is indeed, and while I agree that basing off of GCC could work due to similarities, it would strip out a lot more so i figured it was better to maintain a separate file. Is this okay?

Jarod42: If it is different, then separate fields are ok.

nickclark2016:
The issue I'd have with adding in TCC to core is that the core team doesn't use TCC thus doesn't have the expertise to maintain it (one of the issues with the D stuff right now).  Is it possible to make this a plugin/external script at least initially?

Jarod42:
D module seems written in modular way though.
You might use D as model to write premake-tcc as external module.

nekobbbbbbit:
> The issue I'd have with adding in TCC to core is that the core team doesn't use TCC thus doesn't have the expertise to maintain it (one of the issues with the D stuff right now). Is it possible to make this a plugin/external script at least initially?

I'm willing to maintain it myself, I use premake for many of my projects. If I have to stand up to maintain something it shouldn't be difficult

nickclark2016:
> > The issue I'd have with adding in TCC to core is that the core team doesn't use TCC thus doesn't have the expertise to maintain it (one of the issues with the D stuff right now). Is it possible to make this a plugin/external script at least initially?
> 
> I'm willing to maintain it myself, I use premake for many of my projects. If I have to stand up to maintain something it shouldn't be difficult

Agreed that it shouldn't be difficult. It's why external scripts are intentionally as pain-free as possible. We've recently had a similar discussion on a PR of NVCC into the core repository.  It's a hard balance to strike with adding new exporters and toolsets, as there's only half a dozen people actively working on the repo, usually no more than 1 or 2 at a time (this doesn't count those who send in the one off PRs, for which we are extremely appreciative of).

nekobbbbbbit:
Hi, regardless of conflicts I would like to bump this. Any thoughts?

nickclark2016:
I personally hold the same stance, but commentary from other maintainers would be appreciated.

samsinsane:
Sorry for not replying sooner, I somehow missed this along the way. I agree with @nickclark2016 on this, an external module is the way to go.

swagtoy:
:(

PR #2076, opened 2023-05-08
Allow to select specific version of msc in command line.

Jarod42:
**What does this PR do?**

Allow to select specific version of msc in command line

No changes, except option "cc"

No

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1790,7 +1790,15 @@
 			{ "clang", "Clang (clang)" },
 			{ "gcc", "GNU GCC (gcc/g++)" },
 			{ "mingw", "MinGW GCC (gcc/g++)" },
-			{ "msc", "Microsoft compiler" }
+			{ "msc-v80", "Microsoft compiler (Visual Studio 2005)" },
+			{ "msc-v90", "Microsoft compiler (Visual Studio 2008)" },
+			{ "msc-v100", "Microsoft compiler (Visual Studio 2010)" },
+			{ "msc-v110", "Microsoft compiler (Visual Studio 2012)" },
+			{ "msc-v120", "Microsoft compiler (Visual Studio 2013)" },
+			{ "msc-v140", "Microsoft compiler (Visual Studio 2015)" },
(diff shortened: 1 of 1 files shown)


redorav:
I have a question, how is this different to choosing the compiler via the toolset keyword?

Jarod42:
@redorav:
It is from the **command line**,
else I might add extra option in premake5.lua to mimic existing `--cc=..`
(I use same premake5.lua script but with different generators/toolsets).

redorav:
Ah I see, I haven't configured projects much from the command line. How would you choose the clang or gcc version in a similar fashion? Or would that be unrelated?

Jarod42:
Never select gcc/clang version from premake (pretty sure it is not possible from command line).
except for supported standard, I don't know if there are flags different depending on version for gcc/clang.
There is for msvc.

redorav:
I see, looks good to me then. How does the process normally work, write the review? I'd like to be a bit more active with Premake and get to know the normal workflows you guys are using

Jarod42:
Simple contributor here :-)
So creating PRs, answer questions and do some reviews.
but there are owner/dev of premake which do the final approve (if any).

nickclark2016:
I'm currently abroad on vacation.  Looks like I have a few reviews and issues to triage when I get back if someone else hasn't. Should be able to look around the 15th or so.

nickclark2016:
Quick question: What was the previous behavior of just putting "msc" as an argument?  Is there behavior there that we should preserve rather than remove?

Also, are there existing unit tests for CLI arguments? Or is this an area of test deficiency?

Jarod42:
> What was the previous behavior of just putting "msc" as an argument? 

Putting simply msc, we don't have version so all version checks assume the oldest. so problematic for some flags as `externalincludedirs`.

> Is there behavior there that we should preserve rather than remove?

msc was added recently by me.
I don't thing keeping simple `msc` is useful.

> Also, are there existing unit tests for CLI arguments? Or is this an area of test deficiency?

I don't think there is UTs for CLI arguments except maybe to check that `--options=value` feeds `_OPTION` table.

PR #2081, opened 2023-05-13
Add some missing flags for msc toolset

Jarod42:
**What does this PR do?**

For msc toolset:
- Add some missing flags (compileas/c(pp)dialect).
- Fix option dependency (D9007: '/external:I' requires '/external:W').
- Fix toolset version comparison (`"msc-v142" < "msc-v80"`!) (so vs2005/vs2008) (for external warnings).

Just change msc toolset,
so indirectly affects Codelite (and extern modules such as Code::Blocks, ninja) when using that toolset

One of the fix needed to make ninja works with msvc in [my testing repository](https://github.com/Jarod42/premake-sample-projects)

### DIFF
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -13,7 +13,20 @@
 	local msc = p.tools.msc
 	local project = p.project
 	local config = p.config
+	local string = require("string")
 
+	-- string comparison `toolset >= "msc-v142"` won't work with "msc-v80"
+	local function isVersionGreaterOrEqualTo(lhs, rhs)
+		if lhs == nil or rhs == nil then
+			return false
+		end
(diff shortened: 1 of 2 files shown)


src/tools/msc.lua
```diff
@@ -158,6 +192,12 @@
 --
 
 	msc.cxxflags = {
+		cppdialect = {
+			["C++14"] = "/std:c++14",
```
redorav, line 196: How does the cppdialect keyword work currently in VS? I use it in my projects and it seems to work properly. Some other codepath?

Jarod42: vs20XX generator handle flags internally.
msc.lua is not used from vs20XX generators.

msc.lua is used for Codelite/Code::Blocks/Ninja (gmakeX doesn't support it) generator with msc toolset.

nickclark2016 (approved):

PR #2110, opened 2023-07-16
Add vslinux module

redorav:
**What does this PR do?**

I've added a new module that handles Linux on Visual Studio, it's able to compile and run Linux remotely on another machine through Visual Studio

It will add support for using system("linux") in VS actions and generate the proper project configuration

This module refers to the native support Visual Studio now has for these kinds of projects. I've tested it in VS2022, VS2019 in theory supports it but I've only been able to compile with it, not run it. I've tried to be as thorough as possible with some of the most common settings, but there are a few others that may be useful that are only available to Linux projects. I'd have to create new keywords for that.

I have a question regarding the extensibility of modules. In the vcxproj generator I've had to add many table.remove(elements to be able to remove a lot of properties that VS is adding to this kind of project that it doesn't actually support. Having them there doesn't actually do anything, it just clutters the project file and can be confusing to read. It's a bit annoying because adding settings to one type of project means it can affect the module.

I've added a few simple unit tests following the Android example but I think they are not enough. I can improve this once these simple ones are passing the CI.


modules/vslinux/vslinux_sln2005.lua
```diff
@@ -0,0 +1,10 @@
+--
+-- vslinux/vslinux_sln2005.lua
+-- vslinux integration for vstudio.
+-- Copyright (c) 2012-2015 Manu Evans and the Premake project
```
Jarod42, line 4: Copy pasted, wrong years...

modules/vslinux/vslinux_sln2005.lua
```diff
@@ -0,0 +1,10 @@
...
+--
+
+	local p = premake
+
+	local vslinux = p.modules.vslinux
+	local sln2005 = p.vstudio.sln2005
```
Jarod42, line 10: This file seems useless, no code...

modules/vslinux/vslinux_vstudio.lua
```diff
@@ -0,0 +1,10 @@
+--
+-- vslinux/vslinux_vstudio.lua
+-- vslinux integration for vstudio.
+-- Copyright (c) 2012-2015 Manu Evans and the Premake project
+--
+
+	local p = premake
+
+	local vslinux = p.modules.vslinux
+	local vstudio = p.vstudio
```
Jarod42, line 10: Same here...

Jarod42:
Some UTs fail:
> [  FAILED  ] oven_filtering.onFilterLinuxIsPosix
> .../premake-core/premake-core/tests/oven/test_filtering.lua:131: expected POSIX but was NOTPOSIX

modules/vslinux/vslinux_sln2005.lua
redorav: Fixed

modules/vslinux/vslinux_vstudio.lua
redorav: Fixed

modules/vslinux/vslinux_sln2005.lua
redorav: Deleted file so fixed

redorav:
Looks like somehow Linux means POSIX in some implicit way, I'll dig in see if I can figure out how it happens

redorav:
Right seems like I figured that out, now I need to document these new fields properly. Any pointers would be appreciated

Jarod42:
Out of curiosity, do you have a sample github action for your module, so I can add it to my [testing repo](https://github.com/Jarod42/premake-sample-projects).

redorav:
By example action, do you mean a script? Something like the following works for me

```lua
filter { 'platforms:'..PlatformLinux64 }
    system('linux')
    architecture('x64')
    toolset('gcc')
			
    -- custom
    remoterootdir("~/projects/crstl")
    remoteprojectrelativedir("")
    toolchainversion('remote')
			
    -- Builtin
    rtti('off') -- Works with no overrides
    exceptionhandling ('on') -- Works with overrides
    cppdialect('C++14') -- Works with overrides
    cdialect("C89") -- Works with overrides
    pic('on') -- Works with overrides
    strictaliasing ('off') -- Works with overrides
    -- warnings work
    -- warnings as errors work
    -- symbols work with overrides
    -- optimize works
    floatingpoint('fast')
			
    flags { 'fatalcompilewarnings' }
    staticruntime('off')
```

It's just an example, many of the options I was turning on was just to see how the vcxproj reacted

redorav:
The WSL toolchains need a couple more options that I'll need to add as well. There is a deploy directory and a project directory as well, need to properly figure out what each option does

Jarod42:
I meant CI/workflow script, not premake ones ;-)
or instructions to test locally. I'm not familiar with WSL...

I wonder if "regular" premake scripts might be used as `premake5 vs2022 --os=linux --cc=gcc`, and build it (and run it).

redorav:
> I meant CI/workflow script, not premake ones ;-)

Ah I see, I don't have a script or anything like that, I have a project that does the building through AppVeyor but since this isn't merged yet it's not being tested.

> or instructions to test locally. I'm not familiar with WSL...

There's two ways testing this. Either install WSL and compile with the gcc compiler + wsl toolchain, or do the compilation remotely via a Linux machine or a VM with Linux installed, via gcc + remote toolchain. It's not an instant setup process but you can have it ready in an hour or two.

E.g.

- Install WSL: https://learn.microsoft.com/en-us/windows/wsl/install
- Install GCC and G++ on WSL: https://stackoverflow.com/questions/50729550/installing-g-on-windows-subsystem-for-linux
- Run the premake script with the requirements
- Build 'locally' through WSL

> I wonder if "regular" premake scripts might be used as premake5 vs2022 --os=linux --cc=gcc, and build it (and run it).

Premake already knows about the linux keyword and compilers, so running that would probably default to the GCC Remote toolchain in VS as that's what happens when you don't do anything. The problem with that is now you have to set up some VM or something like that

nickclark2016:
Really late response, but any reason that this can't be part of the `vstudio` module instead of its own?

redorav:
@nickclark2016 I was just following how the android module works. If you'd rather I make it part of the main module I can change it. Is that more desirable? Would it make sense to merge the Android module into the vstudio as well?

nickclark2016:
> @nickclark2016 I was just following how the android module works. If you'd rather I make it part of the main module I can change it. Is that more desirable?
Can it be cleanly integrated in the existing module. If so, then I think it makes sense to integrate it.

> Would it make sense to merge the Android module into the vstudio as well?
That is a question for @samsinsane. I don't have much interaction with the android side of things.

redorav:
@nickclark2016 I've updated the PR with the initial port of the functionality. Before I go ahead and fix/add tests, is this kind of what you were expecting?

nickclark2016:
I think this is closer to what I was picturing. Though a nitpick from me, would it be possible to have separate functions for the linux and windows targets, then just have a couple of system checks, rather than having system checks littered throughout various functions. May be nice to separate out the code like that for maintaining it.

redorav:
Could you specify which functions you were thinking of having separate or what kind of separation you were thinking? For example is it that you would have m.warningLevel as that and m.linuxWarningLevel? At what level of the code would you do these checks?

nickclark2016:
This is me spitballing, but I'd have windows function and a linux function. Then, have a common function that delegates to the windows or linux variant, depending on which is active in the config.  That way, we don't have deeply nested if checks for the systems, version, etc, in the same function.  Feel free to push back if you have other thoughts, as I'd rather get this right the first time.

redorav:
Ok, I think what I can do is keep the Linux check all the way up in the config/system functions, as it's where we have a lot of shared functions I wouldn't want to replicate. Let me prototype this. It also means that I need to fix up the unit tests in many places, as the order in which things appear will no longer be the same. As long as we aren't checking lots of properties in some sorted fashion maybe it's not a lot of work

redorav:
Sorry I made a mess of the repository and I changed too many things to fix the unit tests properly. I'll try again in a clean branch and redo the PR

PR #2025, opened 2023-01-22
Fix missing targets file issue in some C++ nuget packages

hanagasira:
**What does this PR do?**

Projects using the C++ nuget package import the targets file `packages\\{package_id}.{version}\\build\\native\\{package_id}.targets` is imported,
but there are some packages that have `.targets` files in other locations or need to import `.props` files other than `.targets`.

Example.

The targets file for the [WinPixEventRuntime](https://www.nuget.org/packages/WinPixEventRuntime) package targets file exists in
`packages\WinPixEventRuntime.1.0.220810001\build\WinPixEventRuntime.targets`. not
`packages\WinPixEventRuntime.1.0.220810001\build\native\WinPixEventRuntime.targets`.

[Microsoft.Direct3D.D3D12](https://www.nuget.org/packages/Microsoft.Direct3D.D3D12) package needs to import
`packages\Microsoft.Direct3D.D3D12.1.608.2\build\native\Microsoft.Direct3D.D3D12.props` file as well as
`packages\Microsoft.Direct3D.D3D12.1.608.2\build\native\Microsoft.Direct3D.D3D12.targets` file.

**How does this PR change Premake's behavior?**

* Create `packageAPIInfo.packageEntries` variable for package nuget even when it is not a C# project
* In a project using the nuget package, look for `.props` and `.targets` files in the `packageAPIInfo.packageEntries` variable and import them.

**Anything else we should know?**

May slow down project generation to search for targets and props files.

Closes #1497

**Did you check all the boxes?**

- [x] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [x] Add unit tests showing fix or feature works; all tests pass
- [x] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] Minimize the number of commits
- [ ] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

*You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to requests like these!*

PR #2112, opened 2023-07-27
Allow adding any property to a VisualStudio project

hanagasira:
**What does this PR do?**

At #2025 I was able to use the nuget package from Microsoft.Direct3D.D3D12.

But to parameterize this package, we need to set properties such as [Microsoft_Direct3D_D3D12_D3D12SDKPath and Microsoft_Direct3D_D3D12_SkipLibraryCopy](https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/#2-set-agility-sdk-parameters) etc.

Also in charp, we need to set properties for [nullable](https://learn.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types#create-the-application-and-enable-nullable-reference-types) and [other settings](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#publish-related-properties).

Instead of creating dedicated functions for each of these settings, I would like to create a single function that can set any property.

Add a `props` command to allow setting arbitrary properties

no

### DIFF
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -8,6 +8,7 @@
 	p.vstudio.dotnetbase = {}
 
 	local vstudio = p.vstudio
+	local vs2005 = p.vstudio.vs2005
 	local dotnetbase  = p.vstudio.dotnetbase
 	local project = p.project
 	local config = p.config
@@ -241,6 +242,17 @@
 		end
 	end
 
+--
+-- Write out the additional props.
+--
+
+	function dotnetbase.additionalProps(cfg)
+		for i = 1, #cfg.vsprops do
+			for key, value in spairs(cfg.vsprops[i]) do
+				_p(2, '<%s>%s</%s>', key, vs2005.esc(value), key)
+			end
+		end
+	end
 
 --
 -- Write the compiler flags for a particular configuration.
@@ -739,6 +751,7 @@
 			_p(2,'<LangVersion>%s</LangVersion>', cfg.csversion)
 		end
 	end
+
 
 	function dotnetbase.targetFrameworkProfile(cfg)
 		if _ACTION == "vs2010" then
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -216,6 +216,13 @@
 		}
 	}
 
+	p.api.register {
+		name = "vsprops",
+		scope = "config",
+		kind = "list:table",
+		tokens = true,
+	}
+
 --
 -- Decide when the full module should be loaded.
 --
--- a/modules/vstudio/tests/_tests.lua
+++ b/modules/vstudio/tests/_tests.lua
@@ -6,6 +6,7 @@ return {
 	"dotnet2005/test_nuget_framework_folders.lua",
 
 	-- Visual Studio 2005+ C# projects
+	"cs2005/test_additional_props.lua",
 	"cs2005/test_assembly_refs.lua",
 	"cs2005/test_build_events.lua",
 	"cs2005/test_common_props.lua",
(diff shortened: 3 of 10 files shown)


modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -241,6 +241,17 @@
...
 	end
 
+--
+-- Write out the additional props.
+--
+
+	function dotnetbase.additionalProps(cfg)
+		for i = 1, #cfg.props do
+			for key, value in spairs(cfg.props[i]) do
+				_p(2, '<%s>%s</%s>', key, value, key)
```
Jarod42, line 251: Wonder if we have a xml escaper for `value`...
It seems not according to above function (which is more likely to have `<` or `>`.)

src/_premake_init.lua
```diff
@@ -1062,6 +1062,13 @@
 		kind = "list:table",
 	}
 
+	api.register {
+		name = "props",
+		scope = "config",
+		kind = "list:table",
+		tokens = true,
+	}
```
Jarod42, line 1070: vstudio/_premake_init.lua seems a more appropriate place.

website/docs/props.md
```diff
@@ -0,0 +1,47 @@
...
+```lua
+	props {
+		Name1 = "value1",
+		Name2 = "value2",
+	}
+```
+
+### Parameters ###
+
+Name and value is string
```
Jarod42, line 15:
```suggestion
Name and value are strings
```

website/docs/props.md
```diff
@@ -0,0 +1,47 @@
+Add any property to your visual studio project
+This allows you to set properties that premake does not support without extending it
+
+Values set at one time are sorted alphabetically
```
Jarod42, line 4: if order is important for msvc, adding in the doc the way to specify order (several calls) should be added IMO.
else that sentence can be removed.

src/_premake_init.lua
```diff
@@ -1062,6 +1062,13 @@
 		kind = "list:table",
 	}
 
+	api.register {
+		name = "props",
```
Jarod42, line 1066: I wonder if the name should not reflect that it is visual specific (`vsprops`)?

src/_premake_init.lua
hanagasira: Moved function definitions to `modules\vstudio\_preload.lua`

website/docs/props.md
hanagasira: I added it to the document because it is sometimes necessary to specify the order when reusing properties defined before.
By the way, the reason for sorting by name is that the order of the lua table is not always the same.

modules/vstudio/vs2005_dotnetbase.lua
hanagasira: There are multiple xml escaping functions, which one should I use?
For now, I use `vs2005.esc` for csharp, `vs2010.esc` for c++

src/_premake_init.lua
hanagasira: Renamed to vsprops

nickclark2016: Late to the party, but can this life in `modules/vstudio/_preload.lua` instead, since it's specific to VS?

hanagasira: Thanks for pointing this out. I have moved the definition to `modules/vstudio/_preload.lua`.

nickclark2016:
I think this needs to be added in the sidebar still. Other than that, I think I'm happy with the code (though a merge conflict needs to be resolved, if you could do that, it would be very appreciated).

hanagasira:
@nickclark2016 
I rebased and resolved the conflict.

KyrietS:
Please add `vsprops` to the sidebar: https://github.com/premake/premake-core/blob/master/website/sidebars.js#L293
Some day when we upgrade Docusaurus to the latest version we might have auto-generated sidebar, but for now we need to maintain it manually 🙃

hanagasira:
@KyrietS @Jarod42 
Thanks for the explanation. I have recorrected it.

PR #2124, opened 2023-08-11
Integrate Linux support for Visual Studio as part of the vstudio module

redorav:
**What does this PR do?**

I've added a new module that handles Linux on Visual Studio, it's able to compile and run Linux remotely on another machine through Visual Studio

It will add support for using system("linux") in VS actions and generate the proper project configuration

This module refers to the native support Visual Studio now has for these kinds of projects. I've tested it in VS2022, VS2019 in theory supports it but I've only been able to compile with it, not run it. I've tried to be as thorough as possible with some of the most common settings, but there are a few others that may be useful that are only available to Linux projects. I'd have to create new keywords for that.


redorav:
@nickclark2016 Ok I'm back with this PR. I think what I've done is more in line with what you were looking for. I've split the following functions in two:

globals
globalsCondition
configurationProperties
outputProperties
clCompile
link
fileConfigFunction

I then checked which of the original properties are relevant to the Linux projects, added them to a common section, then added Linux-specific functions a appropriate

Jarod42:
CI fails for missing doc:
```
Missing documentation file for: 	remoterootdir
Missing documentation file for: 	remoteprojectdir
Missing documentation file for: 	remoteprojectrelativedir
toolchainversion missing value `remote`
toolchainversion missing value `wsl`
toolchainversion missing value `wsl2`
Missing documentation file for: 	remotedeploydir
```

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -139,34 +139,66 @@
...
 	function m.globals(prj)
 
 		-- Write out the project-level globals
 		m.propertyGroup(nil, "Globals")
-		p.callArray(m.elements.globals, prj)
+		if prj.system == p.LINUX then
+			p.callArray(m.elements.linuxGlobals, prj)
+		else
+			p.callArray(m.elements.globals, prj)
+		end
```
Jarod42, line 178: Possibly
```suggestion
		p.callArray(iif(prj.system == p.LINUX, m.elements.linuxGlobals, m.elements.globals), prj)
```
Apply to others `if prj.system == p.LINUX then`.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -412,9 +508,48 @@
 		return calls
 	end
 
+	m.elements.linuxClCompile = function(cfg)
+		local calls = {
+			-- Common - TODO REVIEW LIST
```
Jarod42, line 513: TODO?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -799,6 +953,57 @@
...
+
+		return {
+			-- Common
+			m.excludedFromBuild,
+			m.objectFileName,
+			m.clCompilePreprocessorDefinitions,
+			m.clCompileUndefinePreprocessorDefinitions,
+			m.optimization,
+			m.forceIncludes,
+			--m.enableEnhancedInstructionSet,
```
Jarod42, line 994: Why commented?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3214,3 +3402,189 @@
...
+			else
+				m.element("DebugInformationFormat", nil, "Minimal")
+			end
+
+		end
+	end
+
+	function m.linuxExceptionHandling(cfg, condition)
+		if cfg.exceptionhandling then
+			m.element("ExceptionHandling", condition,  iif(cfg.exceptionhandling == p.OFF, "Disabled", "Enabled"))
```
Jarod42, line 3435:
```suggestion
			m.element("ExceptionHandling", condition, iif(cfg.exceptionhandling == p.OFF, "Disabled", "Enabled"))
```

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3214,3 +3402,189 @@
...
+			["C++17"]   = "c++17",
+			["C++2a"]   = "c++2a",
+			["C++20"]   = "c++20",
+			["C++latest"] = "c++2a",
+			["gnu++98"] = "gnu++98",
+			["gnu++03"] = "gnu++03",
+			["gnu++11"] = "gnu++11",
+			["gnu++14"] = "gnu++14",
+			["gnu++17"] = "gnu++17",
+			["gnu++17"] = "gnu++20",
```
Jarod42, line 3462:
```suggestion
			["gnu++20"] = "gnu++20",
```

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3214,3 +3402,189 @@
...
+	function m.linuxLanguageStandardCpp(cfg)
+		local cpp_langmap = {
+			["C++98"]   = "c++98",
+			["C++03"]   = "c++98",
+			["C++11"]   = "c++11",
+			["C++14"]   = "c++14",
+			["C++17"]   = "c++17",
+			["C++2a"]   = "c++2a",
+			["C++20"]   = "c++20",
+			["C++latest"] = "c++2a",
```
Jarod42, line 3456: Should it be c++20?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3214,3 +3402,189 @@
...
+	function m.linuxFloatingPointModel(cfg)
+
+		if cfg.floatingpoint then
+			m.element("RelaxIEEE", nil, iif(cfg.floatingpoint == p.OFF, "false", "true"))
+		end
+
+	end
+
+	function m.linuxLanguageStandardCpp(cfg)
+		local cpp_langmap = {
```
Jarod42, line 3448: Wonder if we can/should use `gcc.cxxflags.cppdialect` (and remove prefix `--std=`).

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3214,3 +3402,189 @@
...
+
+			local map = iif(cfg.toolset == "gcc", gcc_map, clang_map)
+			version  = map[cfg.toolchainversion]
+
+		end
+
+		if version then
+			m.element("PlatformToolset", nil, version)
+		else
+			p.error('Invalid toolchainversion for the selected toolset (%s).', cfg.toolset)
```
Jarod42, line 3544:
```suggestion
			p.error('Invalid toolchainversion (%s) for the selected toolset (%s).', cfg.toolchainversion, cfg.toolset)
```
?

modules/vstudio/vs2010_vcxproj.lua
redorav: If you don't mind I think the `if then` makes it clearer, if I end up giving the Android module integration a go I'd have to do it anyway

modules/vstudio/vs2010_vcxproj.lua
redorav: I need to review this one as well. I'm not sure how these should be handled, I'll review it too

modules/vstudio/vs2010_vcxproj.lua
redorav: Where can I find this? I took this code essentially from the Android module

modules/vstudio/vs2010_vcxproj.lua
redorav: Indeed, fixed

modules/vstudio/vs2010_vcxproj.lua
redorav: Whoops, fixed

modules/vstudio/vs2010_vcxproj.lua
redorav: Fixed

redorav:
> CI fails for missing doc:
> 
> ```
> Missing documentation file for: 	remoterootdir
> Missing documentation file for: 	remoteprojectdir
> Missing documentation file for: 	remoteprojectrelativedir
> toolchainversion missing value `remote`
> toolchainversion missing value `wsl`
> toolchainversion missing value `wsl2`
> Missing documentation file for: 	remotedeploydir
> ```

Fixed the documentation as well

modules/vstudio/vs2010_vcxproj.lua
Jarod42: Was just a proposal ;-) ok with `if then`.

modules/vstudio/vs2010_vcxproj.lua
Jarod42: I meant that mapping is similar to the one from gcc. Wonder how we can factorize that.

Probably outside of the scope of that PR though.

redorav: I see, fair enough

modules/vstudio/vs2010_vcxproj.lua
nickclark2016: Apologies, I've been out for the last 2 weeks. LGTM. Thoughts on a slight change to have 3 tables: 1 for common, one for linux, one for windows, then just join the function tables rather than duplicating?

redorav: @nickclark2016 good to have your approval, I'll take that bit of feedback onboard and I still need to finish up a couple other things that were reported by @Jarod42, I'll try to find some good time this week

redorav: @nickclark2016 The problem with having a common table is that the ordering changes and fixing the tests is quite a pain. I think I'll keep it as it is for the moment until I nail down what the common parts should be, then we can do that in another PR?

nickclark2016: Works for me. Appreciate the leg work on investigating that.

modules/vstudio/vs2010_vcxproj.lua
redorav: I have left this one out for now, as I'm not sure how to best tackle it. The builtin xml tags that the MSVC compiler uses don't work, but are also very poor in terms of expressing multiple versions of SSE. If we wanted this to be correct we'd have to have a function that properly passes in -msse4.1 and all the rest. I'll leave it for a future PR

modules/vstudio/vs2010_vcxproj.lua
redorav: @nickclark2016 This I am still a bit unsure of. Do you have any particular opinion on it? What is the difference between toolsversion and toolchainversion?

redorav:
Ok I think this is ready for review now. I've fixed up the functions and checked that the project supports a common subset of them with the main MSVC projects. The only thing that's missing is the vectorextensions which needs a bit more work as the normal tags that MSVC projects use don't work here. We'd have to have a function that translates those into clang/gcc options like the gmake module does, probably, with some VS changes

modules/vstudio/_preload.lua
```diff
@@ -195,6 +195,48 @@
...
+-- Register Linux properties
+--
+
+	p.api.addAllowed("toolchainversion", { "remote", "wsl", "wsl2" })
+
+	-- Directory in the remote machine where our files will be copied before compilation
+	p.api.register {
+		name = "remoterootdir",
+		scope = "config",
+		kind = "string",
```
Jarod42, line 215:
```suggestion
		kind = "path",
```
?
Apply to other added api.

redorav: Aha, I wasn't aware we had thing kind of thing, will do and test. Is it because path is able to resolve tokens?

redorav: Hmm unfortunately it tries to form a path on the local machine, so for example something like `~/projects/crstl` turns into `C:/Users/redorav/Desktop/crstl/~/projects/crstl` and that's not what we want

Jarod42: ok

nickclark2016:
Is this ready for me to look at again?

redorav:
Yes, other than this question I think it's good to go

> @nickclark2016 This I am still a bit unsure of. Do you have any particular opinion on it? What is the difference between toolsversion and toolchainversion?

Jarod42:
You should also add the new entries in premake-core/website/sidebars.js

redorav:
> You should also add the new entries in premake-core/website/sidebars.js

Ah I wasn't aware of it, I'll do that this evening, thanks

modules/vstudio/_preload.lua
```diff
@@ -197,7 +197,7 @@
 	}
 
 	p.api.register {
-		name = "enablemodules",
+	name = "enablemodules",
```
Jarod42, line 200: This style change seems wrong.

redorav: Let me check this again, it is completely unintentional, I merged in the web interface and it introduced differences that I tried to fix locally after, won't be a minute

redorav: Fixed now, thanks

modules/android/_preload.lua
```diff
@@ -19,6 +19,15 @@
 	api.addAllowed("flags", { "Thumb" })
 	api.addAllowed("kind", p.PACKAGING)
 
+	api.addAllowed("toolchainversion", {
```
nickclark2016, line 22: Just wondering why we moved the order of this?

website/docs/remotedeploydir.md
```diff
@@ -0,0 +1,23 @@
...
+
+`path` specifies the directory on the remote machine where the project is deployed
+
+### Applies To ###
+
+The `config` scope.
+
+### Availability ###
+
+Premake 5.0.0 alpha 14 or later, only applies to Visual Studio Linux projects.
```
nickclark2016, line 17: This availability is wrong. It would be the next beta version or later.

website/docs/remoteprojectdir.md
```diff
@@ -0,0 +1,23 @@
...
+
+`path` specifies the directory on the remote machine that WSL sees the project in
+
+### Applies To ###
+
+The `config` scope.
+
+### Availability ###
+
+Premake 5.0.0 alpha 14 or later, only applies to Visual Studio Linux projects.
```
nickclark2016, line 17: Same as above.

website/docs/remoteprojectrelativedir.md
```diff
@@ -0,0 +1,23 @@
...
+
+`path` specifies the directory on the remote machine where the source files of a single project will be copied to before compiling, relative to the root path
+
+### Applies To ###
+
+The `config` scope.
+
+### Availability ###
+
+Premake 5.0.0 alpha 14 or later, only applies to Visual Studio Linux projects.
```
nickclark2016, line 17: Same as above.

website/docs/remoterootdir.md
```diff
@@ -0,0 +1,23 @@
...
+
+`path` specifies the directory on the remote machine where the source files will be copied to before compiling
+
+### Applies To ###
+
+The `config` scope.
+
+### Availability ###
+
+Premake 5.0.0 alpha 14 or later, only applies to Visual Studio Linux projects.
```
nickclark2016, line 17: Same as above.

website/docs/toolchainversion.md
```diff
@@ -17,13 +19,19 @@ toolchainversion ("value")
...
+* `wsl2`: Windows Subsystem for Linux 2
+
 ## Applies To ###
 
 The `config` scope.
 
 ### Availability ###
 
-Premake 5.0.0 alpha 14 or later, only applies to Android projects.
+Premake 5.0.0 alpha 14 or later, only applies to Android or Linux projects.
```
nickclark2016, line 34: Split this into two statements. One for Android, one for Linux projects (describing correct versions for each)

nickclark2016 (changes requested):
Just those doc changes. Code LGTM.

modules/android/_preload.lua
redorav: I wanted to put them next to the other addAllowed functions, since it has changed from using the register function. Now the main preload registers the api, and both linux and android add their allowed versions

website/docs/remotedeploydir.md
redorav: Fixed via b16716229a1312b82434add4b53dfe6e66fc88ae

website/docs/remoteprojectdir.md
redorav: b16716229a1312b82434add4b53dfe6e66fc88ae

redorav (commented):
I didn't realize I had to leave a global comment for the proposed changes. All feedback addressed

PR #2127, opened 2023-08-27
Fixes for using debugger under Linux

vadz:
This PR fixes using `--debugger` option under Linux. Currently it can't be used at all for the reasons explained in the commit messages.

It's a bit surprising that it was so broken, but without these fixes following [the instructions for debugging](https://premake.github.io/docs/Debugging-Scripts/) simply doesn't work currently.

### DIFF
--- a/binmodules/luasocket/premake5.lua
+++ b/binmodules/luasocket/premake5.lua
@@ -44,6 +44,9 @@ project "luasocket"
 		
 		defines { "LUASOCKET_API=__attribute__((visibility(\"default\")))" }
 		
+		-- Do not prepend with "lib" prefix.
+		targetprefix ""
+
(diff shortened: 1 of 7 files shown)


redorav:
It's not that surprising, there used to be many instructions for debugging and they all kind of fell into a state where it was no longer possible to debug. I replaced it with this new approach but I only ever used it on Windows. Thanks for taking the time to update it

nickclark2016 (approved):
LGTM, but I'm not able to test Linux debugging (and to test to make sure it doesn't break Windows debugging) at the moment.  Will leave open for a few days to allow people to review and comment.

vadz:
FWIW I don't think it affects Windows debugging in any way, but I didn't test it neither.

PR #2128, opened 2023-08-28
Slightly improve MSVS filters generation

vadz:
**What does this PR do?**

This PR brings the generated `.vcxproj.filters` files closer to the files created by MSVS itself. This is considered to be a good thing because it facilitates comparing Premake-generated files with the existing ones.

There are no breaking changes, but the contents of the generated projects does change slightly. It shouldn't result in any problems.

**Documentation**

I'm not sure if it's appropriate to mention the special treatment of these filters in `vpaths` documentation. Currently documentation seems to be mostly module-agnostic, so I didn't change it, but I could add an MSVS-specific note, of course, if it's considered to be useful.

### DIFF
--- a/modules/vstudio/vs2010_vcxproj_filters.lua
+++ b/modules/vstudio/vs2010_vcxproj_filters.lua
@@ -52,7 +52,32 @@
 --
 
 	function m.uniqueIdentifiers(prj)
-		local tr = project.getsourcetree(prj)
+		-- This map contains the sort key for the known filters.
+		local knownFilters = {
+			['Source Files'] = 1,
+			['Header Files'] = 2,
+			['Resource Files'] = 3,
+		}
+
+		local sortInFilterOrder = function (a,b)
+			if #a.children > 0 or #b.children > 0 then
+				local orderA = knownFilters[a.name] or 999
+				local orderB = knownFilters[b.name] or 999
+				if orderA < orderB then
+					return true
+				end
+				if orderA > orderB then
+					return false
+				end
+				-- This can only happen if both filters are
+				-- unknown, fall back on default comparison.
+			end
+
+			-- Use default order
+			return a.name < b.name
+		end
+
+		local tr = project.getsourcetree(prj, sortInFilterOrder)
 		local contents = p.capture(function()
 			p.push()
 			tree.traverse(tr, {
--- a/src/base/project.lua
+++ b/src/base/project.lua
@@ -270,9 +270,16 @@
 --
 
 	function project.getsourcetree(prj, sorter)
-
+		-- Reuse the previously generated tree if we have it.
 		if prj._.sourcetree then
-			return prj._.sourcetree
+			tr = prj._.sourcetree
+
+			-- But sort it if necessary, as it is always unsorted.
+			if sorter then
+				tree.sort(tr, sorter)
+			end
+
+			return tr
 		end
 
 		local tr = tree.new(prj.name)
@@ -324,9 +331,11 @@
 		end)
 
 		tree.trimroot(tr)
-		tree.sort(tr, sorter)
-
 		prj._.sourcetree = tr
+
+		if sorter then
+			tree.sort(tr, sorter)
+		end
 		return tr
 	end
 
--- a/modules/vstudio/tests/vc2010/test_filter_ids.lua
+++ b/modules/vstudio/tests/vc2010/test_filter_ids.lua
@@ -97,11 +97,11 @@
 		prepare()
 		test.capture [[
 <ItemGroup>
-	<Filter Include="Header Files">
-		<UniqueIdentifier>{21EB8090-0D4E-1035-B6D3-48EBA215DCB7}</UniqueIdentifier>
-	</Filter>
 	<Filter Include="Source Files">
 		<UniqueIdentifier>{E9C7FDCE-D52A-8D73-7EB0-C5296AF258F6}</UniqueIdentifier>
+	</Filter>
+	<Filter Include="Header Files">
+		<UniqueIdentifier>{21EB8090-0D4E-1035-B6D3-48EBA215DCB7}</UniqueIdentifier>
 	</Filter>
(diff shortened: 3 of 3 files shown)


modules/vstudio/vs2010_vcxproj_filters.lua
```diff
@@ -47,18 +47,55 @@
...
--- map vpaths like "**.h" to an <Extensions>h</Extensions> element.
+-- directory or virtual group.
 --
 
 	function m.uniqueIdentifiers(prj)
-		local tr = project.getsourcetree(prj)
+		local knownFilters = {
+			['Source Files'] = {
+				order = 1,
+				extensions = 'cpp;c;cc;cxx;def;odl;idl;hpj;bat;asm;asmx',
```
nickclark2016, line 57: My initial thought is that these extensions should match what we have in the generator for determining what's a source file for the vcxproj file. Where did these lists come from?

modules/vstudio/vs2010_vcxproj_filters.lua
```diff
@@ -47,18 +47,55 @@
...
+				local filterA = knownFilters[a.name]
+				local filterB = knownFilters[b.name]
+				if filterA ~= nil and filterB ~= nil then
+					return filterA.order < filterB.order
+				end
+				if filterA ~= nil and filterB == nil then
+					return true
+				end
+				if filterA == nil and filterB ~= nil then
+					return false
```
Jarod42, line 80: Possibly:
```suggestion
				local unknown_filter = { order = 4 }
				local filterA = knownFilters[a.name] or unknown_filter 
				local filterB = knownFilters[b.name] or unknown_filter 

				return filterA.order < filterB.order
```

modules/vstudio/vs2010_vcxproj_filters.lua
vadz: These lists correspond to the contents of the generated files if you do "New > Project" in the IDE itself and so are, IMO, the "canonical" extensions for the corresponding kinds of files. So for me using the actual extensions used in vpaths would result in gratuitous (because they don't really affect anything) differences between Premake-generated and "native" project files and I'd like to avoid this because I want to review the diffs when replacing the existing build system with Premake.

AFAIK these extensions are mostly cosmetic anyhow, i.e. they affect the list of files shown when adding files to this group in the IDE, but this is never going to happen anyhow for the generated projects.

modules/vstudio/vs2010_vcxproj_filters.lua
vadz: Nice, thanks! Sorry for my "baby Lua", unfortunately Lua is not a language I'm really familiar/comfortable with.

If this PR is accepted, I can apply this suggestion (perhaps with `order = 999` instead of `4`?) or you could apply it when merging, of course.

Jarod42: Not equivalent though, when both are not found, I didn't use name comparison :/

modules/vstudio/vs2010_vcxproj_filters.lua
nickclark2016: From the perspective of Premake, I think it's more correct to make the filters match what's in the `ClCompile`, `ClInclude`, etc. lists in the `vcxproj` generator.  We'll end up with filter mismatches with the file extensions in the project file, which isn't something we want.

modules/vstudio/vs2010_vcxproj_filters.lua
Jarod42:
```suggestion
			local filterA = knownFilters[a.name] or { order = 999 + iif(a.name < b.name, 0, 1) } 
			local filterB = knownFilters[b.name] or { order = 999 + iif(a.name < b.name, 1, 0) }

			return filterA.order < filterB.order
```

Jarod42: > or you could apply it when merging, of course.

Simple contributor here, So cannot apply any changes ;-)

modules/vstudio/vs2010_vcxproj_filters.lua
vadz: OK, what about applying this PR without the last commit then? This would just change the order of the filters, without changing anything else.

nickclark2016: I'm fine with leaving out the last commit.

vadz: Should I force push to this PR to remove the last commit or will you just merge `FETCH_HEAD~`?

AFAICS no new tests are needed as the existing tests cover this already (and had to be updated).

vadz: Force pushed now.

nickclark2016 (approved):
Leaving open for a few days for the chance for others to review and comment.

src/base/project.lua
```diff
@@ -326,7 +326,9 @@
 		tree.trimroot(tr)
 		tree.sort(tr, sorter)
 
-		prj._.sourcetree = tr
+		if sorter then
```
Jarod42, line 329: Should it be `not sorter` instead? (for cache consistency)
Or even better move it before the `table.sort`.
So `prj._.sourcetree` still have unsorted tree, and we sort the result afterward and `local tr = tree.new(prj.name)` creation should be done only when `prj._.sourcetree` is not `Nil`

vadz: Oops, you're absolutely right, this should be `not sorter`.

But I also agree that your suggestion to cache unsorted tree is even better, done (and force pushed) now.

Thanks!

nickclark2016: Is this resolved?

vadz: Yes, I think the latest version is optimal: it caches the unsorted tree and sorts it if necessary before returning it.

Jarod42 (approved):

PR #2133, opened 2023-09-01
[Draft] [VS] Add UAC configuration

### LINKED ISSUE #2132: [VS] Add UAC configuration

In Windows, you can set the required privilege level for an application by including an [Application Manifest](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#trustinfo).

The field *requestedExecutionLevel* under *trustInfo* controls this, and has three levels: *asInvoker*, *highestAvailable*, *requireAdministrator*.

At its' simplest, such a manifest file can look like this:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
    <security>
      <requestedPrivileges>
        <requestedExecutionLevel
          level="requireAdministrator"
          uiAccess="false"/>
      </requestedPrivileges>
    </security>
  </trustInfo>
</assembly>
```

Currently there is no option to set this with premake.

You can include a manifest simply by adding it to the *files* option, as such:
```lua
files { "myapp.exe.manifest" }
```
And it gets correctly added as *Additional Manifest Files* entry:

:+1:
image

The problem however is, that it's added as *additional manifest*, while the actual configuration property in the .vcxproj file isn't set accordingly.

:-1:
image

Besides the confusion, the real problem is that the build will fail:
```
1>manifest authoring error c1010001: Values of attribute "level" not equal in different manifest snippets.
1>LINK : fatal error LNK1327: failure during running mt.exe
```
This is because how manifests work in modern VS, which is that the additional manifests will be merged with whatever options you have set in the configuration, unlike earlier when manifests weren't controlled by configuration options and were embedded directly as raw resources instead. So now VS will have the *requestedExecutionLevel* field set twice, as both *asInvoker* and *requireAdministrator* and gets angry :angry: 

**What might be a solution?**

Add configuration options *uacexecutionlevel* and *uacuiaccess* for premake and the corresponding configuration properties in the .vcxproj file under &lt;Project&gt; → &lt;ItemDefinitionGroup&gt; → &lt;Link&gt;
```xml
<Project ...>
  <ItemDefinitionGroup ...>
    <Link>
      <UACExecutionLevel>AsInvoker</UACExecutionLevel>
      <UACUIAccess>false</UACUIAccess>
    </Link>
  </ItemDefinitionGroup>
</Project>
```
UACExecutionLevel = `AsInvoker` (default) / `HighestAvailable` / `RequireAdministrator`
UACUIAccess = `false` (default) / `true`

I have seen other projects use *linkoptions* as a workaround, as any command line options take precedence over whatever is configured in the project. This will make the project build, however it overrides the configuration property which will still show the wrong value (and makes changing it have no effect).
```lua
linkoptions { "/MANIFESTUAC:\"level='requireAdministrator' uiAccess='false'\"" }
```

The correct workaround to this is to use a [call array override](https://premake.github.io/docs/Overrides-and-Call-Arrays/) in premake like this.
```lua
if _ACTION and _ACTION >= "vs2010" then
  require "vstudio"
  premake.override(premake.vstudio.vc2010.elements, "link", function(base, prj)
    local calls = base(prj)
    table.insert(calls, function(prj) premake.vstudio.vc2010.element("UACExecutionLevel", nil, "RequireAdministrator") end)
    return calls
  end)
end
```
This however isn't very user-friendly nor super obvious especially if you are not familiar with Lua. ( If you are not, go learn it now, lua is amazing :smile: )

So I think the best option is to add this as a proper feature so no workarounds are needed.

anzz1:
**What does this PR do?**

Adds the UAC (User Account Control) configuration property for VS targeting Windows applications.
Refer to issue below for in-depth explanation.

- Closes #2132 

This PR is purely additional and will not change existing behaviour nor introduce breaking changes.

**TODO**

- [x] VS2010 functionality implemented
- [ ] VS200x functionality implemented

This is a draft PR where just the base functionality for VS2010-format projects is implemented and tested to work under VS2019.

The [patch branch](https://github.com/anzz1/premake-core/tree/patch-vs-add-uac-props) is open for PR's and any help with adding documentation, unit tests, VS200x functionality, testing other versions than VS2019, or whatever else is appreciated.

### DIFF
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -152,23 +152,6 @@
 		},
 	}
 
-	p.api.register {   -- DEPRECATED 2019-10-21
-		name = "debuggerflavor",
-		scope = "config",
-		kind = "string",
-		allowed = {
-			"Local",
-			"Remote",
-			"WebBrowser",
-			"WebService"
-		}
-	}
-
-	p.api.deprecateField("debuggerflavor", 'Use `debugger` instead.',
-	function(value)
-		debugger('VisualStudio' .. value)
-	end)
-
 	p.api.register {
 		name = "scanformoduledependencies",
 		scope = "config",
@@ -194,6 +177,60 @@
 			"Off"
 		}
 	}
+
+	api.register {
+		name = "uacexecutionlevel",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"AsInvoker",
+			"HighestAvailable",
+			"RequireAdministrator"
+		},
+		aliases = {
+			Invoker = 'AsInvoker',
+			Highest = 'HighestAvailable',
+			Admin = 'RequireAdministrator'
+		}
+	}
+
+	api.register {
+		name = "uacuiaccess",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Default",
+			"On",
+			"Off"
+		},
+		aliases = {
+			Bypass = 'On',
+			True = 'On',
+			False = 'Off'
+		}
+	}
+
+--
+-- Deprecated fields
+--
+
+	p.api.register {   -- DEPRECATED 2019-10-21
+		name = "debuggerflavor",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Local",
+			"Remote",
+			"WebBrowser",
+			"WebService"
+		}
+	}
+
+	p.api.deprecateField("debuggerflavor", 'Use `debugger` instead.',
+	function(value)
+		debugger('VisualStudio' .. value)
+	end)
 
 --
 -- Decide when the full module should be loaded.
--- a/modules/vstudio/_preload.lua
+++ b/modules/vstudio/_preload.lua
@@ -186,12 +186,14 @@
 			"Default",
 			"AsInvoker",
 			"HighestAvailable",
-			"RequireAdministrator"
+			"RequireAdministrator",
+			"Disabled"
 		},
 		aliases = {
 			Invoker = 'AsInvoker',
 			Highest = 'HighestAvailable',
-			Admin = 'RequireAdministrator'
+			Admin = 'RequireAdministrator',
+			None = 'Disabled'
 		}
 	}
 
--- a/modules/vstudio/vs200x_vcproj.lua
+++ b/modules/vstudio/vs200x_vcproj.lua
@@ -533,6 +533,7 @@
 				m.largeAddressAware,
 				m.optimizeReferences,
 				m.enableCOMDATFolding,
+				m.userAccountControl,
 				m.entryPointSymbol,
 				m.importLibrary,
 				m.targetMachine,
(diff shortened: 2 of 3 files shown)


src/_premake_init.lua
```diff
@@ -1512,6 +1512,39 @@
 		tokens = true
 	}
 
+	api.register {
```
nickclark2016, line 1515: Given these are VS only APIs, they should be registered by the VS module, rather than the global API registry.

src/_premake_init.lua
```diff
@@ -1540,6 +1573,7 @@
 	api.alias("fileextension", "fileExtension")
 	api.alias("propertydefinition", "propertyDefinition")
 	api.alias("removefiles", "excludes")
+	api.alias("uacexecutionlevel", "uaclevel")
```
nickclark2016, line 1576: I don't see the need for this alias. Let's give it a good name the first time, rather than adding an alias for a name.

nickclark2016 (changes requested):
Initial thoughts while you're working on this.

src/_premake_init.lua
anzz1: Good call, I definitely didn't put the feature in the global table without even noticing that a module-specific one exists :wink:

src/_premake_init.lua
anzz1: I was on the fence about this too, since the property is called "uacexecutionlevel" which is quite a mouthful for which "uaclevel" I thought would be a good shorthand, but then again calling the same thing multiple names is bad practice. Shouldn't have included this in the first place. :+1:

modules/vstudio/_preload.lua
```diff
@@ -152,23 +152,6 @@
 		},
 	}
 
-	p.api.register {   -- DEPRECATED 2019-10-21
```
nickclark2016, line 155: Curious why Git thinks these lines were deleted, rather than just inserting new lines above it. Can you check to make sure your line endings are LF (Unix) on the commit?

anzz1: The line endings are correct, it's because I did some spring cleaning and moved the deprecated fields downwards below the non-deprecated ones to be in line how the fields are arranged elsewhere in the project.

nickclark2016:
Just bumping this draft

modules/vstudio/_preload.lua
```diff
@@ -195,6 +178,62 @@
...
+			Highest = 'HighestAvailable',
+			Admin = 'RequireAdministrator',
+			None = 'Disabled'
+		}
+	}
+
+	api.register {
+		name = "uacuiaccess",
+		scope = "config",
+		kind = "string",
```
Jarod42, line 203:
```suggestion
		kind = "boolean",
```
It might simplify `aliases`

anzz1: did you try this? iirc there was a reason I did it this way as aliases didn't work with 'boolean' or there was some other problem.

Jarod42: Didn't try it, was just a (bad) suggestion.

nickclark2016:
Given the amount of time since this was last modified or replied to, I will be closing this draft PR. Feel free to reopen at a later date if changes are made.

PR #2135, opened 2023-09-07
Fix libzip missing a library

KanuX-14:
Fix premake compiling from undefined functions in libzip dependency.

Tested on `amd64`:
- Arch Linux
- Kernel: 6.4.12-arch1-1
- Clang: 16.0.6
- GNU

Tested on `armv7l`:
- postmarketOS
- Kernel: 6.3.0-postmarketos-qcom-msm8974-g65fa893de431
- Clang: 16.0.6
- Musl

### DIFF
--- a/contrib/libzip/premake5.lua
+++ b/contrib/libzip/premake5.lua
@@ -13,6 +13,7 @@ project "zip-lib"
 
 	filter "system:linux or bsd or solaris or haiku"
 		defines { "HAVE_SSIZE_T_LIBZIP", "HAVE_CONFIG_H" }
+		forceincludes { "unistd.h" }
 
 	filter "system:windows"
 		defines { "_WINDOWS" }


nickclark2016:
Instead of modifying source, can we use a forceinclude in the premake file?

KanuX-14:
> can we use a forceinclude in the premake file

Sorry, I don't know how to include it in a premake file.

If you know how to, feel free to modify so I can test it.

nickclark2016:
This is something that's already in the premake file for the lib. You'd just need to update the filter to include linux instead of only MacOS.

KanuX-14:
> in the premake file for the lib

Now I understand. Thank you.

KanuX-14:
Confirmed, compiled on both machines.
Thanks for the tip!

nickclark2016:
No problem. I have an issue open to go in and update the library so we can get rid of the hack completely, just need to find time to do it.

nickclark2016 (approved):

PR #2141, opened 2023-09-18
Enable msbuild batching for custom rules

cathei:
**What does this PR do?**

This PR lets users to write custom rules compatible with [msbuild batching](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-batching?view=vs-2022). By providing `@(Rule)` as input and `@(Rule->'%(Output)')` as output, msbuild can recognize this task can be batched and built incrementally.

Existing rules should be unaffected because the behavior should be same when using tokens like `${file.relpath}` since properties like `%(Identity)` or `%(FullPath)` breaks batching.

To enable batching, user should write command with newly added tokens `${rule.inputs}` or `${rule.outputs}` which mapped to string list. For non-msbuild environments like `gmake2` or `codelite`, these tokens will be substituted as single input and output as there're no batching.

Let me know if introducing new tokens is undesirable.

### DIFF
--- a/modules/gmake2/_preload.lua
+++ b/modules/gmake2/_preload.lua
@@ -62,7 +62,12 @@
 
 		onCleanProject = function(prj)
 			p.clean.file(prj, p.modules.gmake2.getmakefilename(prj, true))
-		end
+		end,
+
+		pathVars = {
+			["rule.inputs"]	 = { absolute = false, token = "$<" },
+			["rule.outputs"] = { absolute = false, token = "$@" }
+		}
 	}
 
 --
--- a/website/docs/Custom-Rules.md
+++ b/website/docs/Custom-Rules.md
@@ -72,3 +72,14 @@ project "MyProject"
       StripDebugInfo = true
     }
 ```
+
+## Rule Batching
+
+With msbuild, custom rules can be batched if same properties are used. To enable this, use `rule.inputs` or `rule.outputs` tokens in `buildcommands`. If corresponding output is up-to-date to the input, then the input will be omitted.
+
+```lua
+rule "BatchRule"
+  fileextension ".xyz"
+  buildcommands 'MyProcessor.exe %{rule.inputs}'
+```
+
--- a/modules/codelite/_preload.lua
+++ b/modules/codelite/_preload.lua
@@ -46,6 +46,11 @@
 		onCleanTarget = function(prj)
 			p.modules.codelite.cleanTarget(prj)
 		end,
+
+		pathVars = {
+			["rule.inputs"]	 = { absolute = false, token = "$&lt;" },
+			["rule.outputs"] = { absolute = false, token = "$@" }
+		}
 	}
 
 
--- a/modules/vstudio/tests/vc2010/test_rule_props.lua
+++ b/modules/vstudio/tests/vc2010/test_rule_props.lua
@@ -38,6 +38,10 @@
 				'%{output_path}%{file.basename}.example.cc',
 				'%{output_path}%{file.basename}.example.h'
 			}
+			builddependencies {
+				'dependency_1.lib',
+				'dependency_2.lib',
+			}
 	end
 
 
@@ -68,3 +72,16 @@
 		]]
 	end
 
+
+--
+-- additionalDependencies
+--
+
+	function suite.additionalDependencies()
+		local r = test.getRule("example")
+		m.additionalDependencies(r)
+
+		test.capture [[
+<AdditionalDependencies>dependency_1.lib;dependency_2.lib</AdditionalDependencies>
+		]]
+	end
--- a/modules/vstudio/tests/vc2010/test_rule_targets.lua
+++ b/modules/vstudio/tests/vc2010/test_rule_targets.lua
@@ -95,3 +95,56 @@
 </UsingTask>
 		]]
 	end
+
+
+
+--
+-- ruleTask
+--
+
+	function suite.ruleTask()
+		local r = test.getRule("example")
+		m.rule(r)
+
+		test.capture [[
+<example
+	Condition="'@(example)' != '' and '%(example.ExcludedFromBuild)' != 'true'"
+	CommandLineTemplate="%(example.CommandLineTemplate)"
+	output_path="%(example.output_path)"
+	AdditionalOptions="%(example.AdditionalOptions)"
+	Inputs="@(example)"
+	Outputs="@(example->'%(Outputs)')"
+	StandardOutputImportance="High"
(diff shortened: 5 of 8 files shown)


website/docs/Custom-Rules.md
```diff
@@ -72,3 +72,14 @@ project "MyProject"
       StripDebugInfo = true
     }
 ```
+
+## Rule Batching
+
+With msbuild, custom rules can be batched if same properties are used. To enable this, use `rule.inputs` or `rule.outputs` tokens in `buildcommands`. If corresponding output is up-to-date to the input, then the input will be omitted.
```
nickclark2016, line 78: Nit: Rename `msbuild` to MSBuild

modules/gmake2/_preload.lua
```diff
@@ -62,7 +62,12 @@
 
 		onCleanProject = function(prj)
 			p.clean.file(prj, p.modules.gmake2.getmakefilename(prj, true))
-		end
+		end,
+
+		pathVars = {
+			["rule.inputs"]	 = { absolute = false, token = "$<" },
+			["rule.outputs"] = { absolute = false, token = "$@" }
```
Jarod42, line 69: > For non-msbuild environments like gmake2 or codelite, these tokens will be substituted as single input and output as there're no batching.

What would be the value without `pathVars` ? (i.e for gmake/xcode/external modules)
Does they need update?

website/docs/Custom-Rules.md
```diff
@@ -72,3 +72,14 @@ project "MyProject"
       StripDebugInfo = true
     }
 ```
+
+## Rule Batching
```
Jarod42, line 76: Those tokens should also go in Tokens.md IMO.

modules/gmake2/_preload.lua
cathei: Those tokens are only valid in `buildcommands` in custom rule context.
For built-in modules it seems like gmake2, codelite and vs are only modules support custom rules.

However I do think it's better to have fallback for any modules. I was thinking of using `fcfg_mt` for fallback and name tokens something like `file.ruleinputs`/`file.ruleoutputs`. Does that make more sense?

Jarod42: Custom rules are indeed not supported by gmake and xcode.
Some external module such as codeblocks/ninja does support it.

How is `rule.outputs` "concatenated"? with space? (and `$@` `$<` may probably be removed).
`rule.inputs` is a little more complex, Does it include `%{file.relpath}`?

cathei: Yes, in VS modules,
`rule.inputs` is space-concatenated `file.relpath`s for the batch. (Same as item included in `.vcxproject`)
`rule.outputs` is space-concatenated `path.normalize(buildoutputs)`s for the batch. (Same as outputs defined in `.targets`)

cathei:
So I went with `fcfg_mt.ruleinputs` as fallback, which just returns `fcfg.relpath`.
On supported platform (MSBuild only for now) this should expand to proper list of inputs (depends on build system).
As a result there's no modification needed for other existing modules.

I excluded output for now, as it's bit complex to get value and users probably can deduce it from inputs.
Can be improved in future.

Also docs are updated as reviewed.

Jarod42:
> This PR lets users to write custom rules compatible with [msbuild batching](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-batching?view=vs-2022). By providing @(Rule) as input and @(Rule->'%(Output)') as output, msbuild can recognize this task can be batched and built incrementally.

Unclear for me what it means...
From the link I quickly read, it seems more a code factorization. and as you write the rule in Premake, it is not really a concern.
(not even sure the factorization takes place with generated code).

What do you mean by *"can be batched and built incrementally"*?
as I understand inputs/ouputs are the important part for individual build, not the `buildcommands`...

cathei:
> What do you mean by "can be batched and built incrementally"?

I'll give brief example. When you have custom rule,
```lua
rule 'MyCustomRule'
  fileextensions { '.txt' }
  buildcommands { 'echo %(file.ruleinputs)'  } -- using new ruleinputs token
```

Without batching msbuild will invoke each command per file separately.
```
echo a.txt
echo b.txt
echo c.txt
```

When batching is possible, msbuild can invoke command as a whole. (Incremental just means that if only a.txt changed only `echo a.txt` will be executed unless it's clean build.)
```
echo a.txt b.txt c.txt
```
This can boost performance for the commands that have long startup time or use shared resorces.

Jarod42:
OK, so your code is a shorthand for:

```lua
rule "myrule"
	--location(LocationDir)
	--display "My custom rule"
	fileextension ".txt"

	-- propertydefinition { .. }
	-- buildinputs { "%{file.relpath}.in" } -- Possible extra inputs
	-- buildoutputs { path.join(LocationDir, "%{file.basename}.out") } -- All outputs
if "vs2000" <= _ACTION and _ACTION <= "vs3000" then -- Batch only handled by vs*
	buildmessage 'custom rule: [Inputs]'
	buildcommands { "MyCommand [Inputs]" }
else
	buildmessage 'custom rule: %{file.relpath}'
	buildcommands { "MyCommand %{file.relpath}" }
end
-- if block replaced by
--	buildmessage 'custom rule: %{file.inputs}'
--	buildcommands { "MyCommand %{file.inputs}" }
```

`%{file.ruleinputs}` seems a wrong naming: it is the name of several `%{file}`. `rule.inputs` is better IMO (but might be confused with `buildinputs` though (as I did several post above)).

Wonder if a flag `batchbuild "on"` would be better in that case.

cathei:
> OK, so your code is a shorthand for:

Yes pretty much.

> %{file.ruleinputs} seems a wrong naming: it is the name of several %{file}. rule.inputs is better IMO (but might be confused with buildinputs though (as I did several post above)).

I agree that `rule.inputs` is better in terms of naming, but I can't find any good customization point like `fcfg_mt` without making seemingly big change that could require external modules to update.

> Wonder if a flag batchbuild "on" would be better in that case.

I think it is better to be separate token, since even if there is flag, many tokens can break batch if MSBuild can't detect common pattern.

nickclark2016:
Do you mind pushing an empty commit just to get the actions to trigger? I think I'm good to merge it, assuming all tests pass.

nickclark2016:
Just wanted to poke this one more time to see if we can get commit to trigger the workflow so I can go ahead and merge this @cathei

cathei:
@nickclark2016 Sorry for the delay. Also pushed this one / tested locally

Jarod42 (approved):

PR #2142, opened 2023-09-18
Remove unused configuration from gmake2

cathei:
**What does this PR do?**

Remove `env.PathVars` (notice the capitalization) that had no effect.

Shouldn't change any behavior as it has wrong name and not referenced.

N/A


nickclark2016:
Any reason to remove this, rather than rename is properly to `pathVars`?

cathei:
> Any reason to remove this, rather than rename is properly to `pathVars`?

This was my thought process:
* If the values are identical to tokens from fileconfig, then there would be no point of re-defining here.
* If the values are not identical, then it might break existing scripts that use tokens, due to behavior change.

Jarod42:
As I remember, it might be useful only for cases with `absolute = true` (as other values are identical as without pathvars)...

But I would say our handling of absolute is strange anyway (we have `%{!path}` to turn/keep absolute path, `%{absolute_path_token}` might turn into relative path (inside `%[path]`)).

Created [project-tokens](https://github.com/Jarod42/premake-sample-projects/tree/project-tokens/projects/project-tokens) to test tokens.

nickclark2016 (approved):

nickclark2016:
@cathei do you mind pushing up an empty commit to retrigger actions? Once they're triggered and pass I should be able to merge this.

nickclark2016:
Poking on this to see if we can get a commit to retrigger the workflow @cathei

cathei:
@nickclark2016 Sorry for the delay, done now :)

Jarod42 (approved):

Issue #1390, opened 2020-01-17
set precompiled header output file


xzores:
Can we have an option for setting precompiled header output file. 

visual studio equivalent:
![image](https://user-images.githubusercontent.com/17770917/72648340-18ea7280-397b-11ea-95f2-226e8b86a2b5.png)

starkos:
See [Precompiled Headers](https://github.com/premake/premake-core/wiki/Precompiled-Headers). Give a shout if you have any questions.

starkos:
Apologies, I misread. You're correct, I don't believe there is a way to do that currently. That should be a pretty easy one to add though, if you feel like having a go. You'd need to define a new API for it in `_premake_init.lua`, and then use it to emit the appropriate elements. You should be able to copy from the PCH related functions [here](https://github.com/premake/premake-core/blob/master/modules/vstudio/vs2010_vcxproj.lua#L2310). Would be happy to approve a pull request for such a feature!

xzores:
Ok, I might give it a go ;)

PR #2152, opened 2023-10-21
Setting precompiled header output file.

Sharlock93:
**What does this PR do?**

Implements `pchoutputfile`, closes : https://github.com/premake/premake-core/issues/1390

Not that I'm aware of.

This is currently only implemented for Visual Studio.


website/docs/pchoutputfile.md
```diff
@@ -0,0 +1,28 @@
...
+
+On MSVC this is the `/FP` compiler flag.
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0 and up.
```
Jarod42, line 21: Not sure it is the correct version.

Jarod42 (commented):
There is also tools msvc.lua to update to have `/FP` flags

website/docs/pchoutputfile.md
Sharlock93: I assumed 5.0 is still in beta and wasn't "released yet" but once its released it would have this feature, I'm honestly not sure either.

Jarod42:
For other tools, modifying `function gcc.getpch(cfg)` should be enough IMO.

website/docs/pchoutputfile.md
Sharlock93: I hope its the correct way to change msc.lua

Sharlock93:
> There is also tools msvc.lua to update to have `/FP` flags
I didn't know this file existed, I noticed that Yu/Yc flags aren't in it


> For other tools, modifying `function gcc.getpch(cfg)` should be enough IMO.

I unfortunately do not have these tools like gcc or clang setup on my current machine so I won't be able to test them

Jarod42:
For testing, I had setup some testing projects as [project-03-pch](https://github.com/Jarod42/premake-sample-projects/tree/master/projects/project-03-pch) with actions for all built-in generators.

BTW, you broke UTs of premake.

website/docs/pchoutputfile.md
nickclark2016: It would be 5.0.0 beta 3 or later

nickclark2016:
Please fix the UTs. This cannot be merged until they are passing.

Sharlock93:
> Please fix the UTs. This cannot be merged until they are passing.

Yeah sorry about that, I have been exceptionally busy the last few weeks, and I took a quick look to see why the UTs broke besides the api registration, so I would have to set aside some time for it to really give it a go. Might have some time this weekend.

nickclark2016:
> > Please fix the UTs. This cannot be merged until they are passing.
> 
> Yeah sorry about that, I have been exceptionally busy the last few weeks, and I took a quick look to see why the UTs broke besides the api registration, so I would have to set aside some time for it to really give it a go. Might have some time this weekend.

Sounds good.  I thoroughly understand life gets busy, been in the same boat lately.  If there's any questions you have, just toss them on a comment here and I'll try to get you an answer quickly.

Sharlock93:
Hi ya, unfortunately I have to close the PR because I don't have the time these days and due to a bug
in the Xcode project generation I had to use other build systems.

PR #2154, opened 2023-10-28
Added NativeWithManagedCore as an allowed value for debuggertype

peter1745:
**What does this PR do?**
Added `NativeWithManagedCore` as an allowed value to `debuggertype`, this value is the .NET Core equivalent of the `Mixed` value, allowing debugging of both native and .NET Core code

This changes the vs2010 generator, specifically the vcxproj.user generator

I'm unsure when support for the `NativeWithManagedCore` value was added to Visual Studio, it seems to be there in vs2019, but I was unable to find exactly when it was added.

### DIFF
--- a/website/docs/debuggertype.md
+++ b/website/docs/debuggertype.md
@@ -10,6 +10,7 @@ debuggertype "value"
 * `Mixed` - needs documentation.
 * `NativeOnly` - needs documentation.
 * `ManagedOnly` - needs documentation.
+* `NativeWithManagedCore` - needs documentation.
 
 ### Applies To ###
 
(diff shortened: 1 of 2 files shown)


website/docs/debuggertype.md
```diff
@@ -10,6 +10,7 @@ debuggertype "value"
 * `Mixed` - needs documentation.
 * `NativeOnly` - needs documentation.
 * `ManagedOnly` - needs documentation.
+* `NativeWithManagedCore` - needs documentation.
```
Jarod42, line 13: Would be fine to fix docu...

peter1745: Documentation has been updated for all of the supported values

nickclark2016 (approved):

PR #2188, opened 2024-02-19
Handle `buildaction "Copy"` for gmake2.

Jarod42:
**What does this PR do?**

Handle `buildaction "Copy"` for gmake2.

Handle `buildaction "Copy"` for gmake2.

Tested with https://github.com/Jarod42/premake-sample-projects/tree/master/projects/project-buildaction_copy
https://github.com/Jarod42/premake-sample-projects/actions/runs/7961205332/job/21731877841

### DIFF
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -192,6 +192,17 @@
 					cpp.addGeneratedFile(cfg, node, output)
 				end
 			end
+		elseif filecfg.buildaction == "Copy" then
+			local output = '$(TARGETDIR)/' .. node.name
+			local file = {
+				buildoutputs  = { output },
+				source        = node.relpath,
+				buildmessage  = '$(notdir $<)',
+				verbatimbuildcommands = gmake2.copyfile_cmds('"$<"', '"$@"'),
+				buildinputs = {'$(TARGETDIR)'}
+			}
+			table.insert(cfg._gmake.fileRules, file)
+			cpp.addGeneratedFile(cfg, node, output)
 		else
 			cpp.addRuleFile(cfg, node)
 		end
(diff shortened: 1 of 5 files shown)


modules/gmake2/gmake2_cpp.lua
```diff
@@ -162,7 +162,19 @@
 		end
 
 		-- process custom build commands.
-		if fileconfig.hasCustomBuildRule(filecfg) then
+		if filecfg.buildaction == "Copy" then
```
nickclark2016, line 165: Would it be desirable to support having both?

nickclark2016: Yeah, what I'm asking is would it be desirable to be an if instead of else if so a script could provide both a build action and a custom build rule and both are invoked?

Jarod42: Unsure what other generators do in that case.

nickclark2016: Yeah, just wondering if this may break existing code that uses "Copy" (but does nothing if gmake2 is used) and has a custom build rule. If there are scripts like that, "Copy" will be used and the custom build rule will be dropped.

nickclark2016: I'd expect that to be the case, as well. Is there any downside to allowing both custom rules and copy build actions? If not, I think I'd prefer both are there/copy doesn't "overwrite" the custom build rule.

Jarod42: Ok, I will setup a project which does both, and check current behavior of other generators.
and report back.

Jarod42: I have checked:
- vs*/ninja only does the custom build part.
- codelite does only the copy part (add recently by #2171)
- Others doesn't support `buildaction "Copy"`.

`buildaction` comes from vs, so I propose to mimic vs behavior (also for codelite, so change order of the if/elseif).
Note that directly from vs IDE, we cannot have both at the same time.

Jarod42: updated to have same behavior than vs.
test project
https://github.com/Jarod42/premake-sample-projects/tree/customcommand-with-buildaction-copy/projects/project-customcommand-with-buildaction-copy added.

nickclark2016 (approved):
LGTM. Will leave it open for a few days in case others want to comment, then I'll merge (please feel free to ping me if I forget)

PR #2187, opened 2024-02-17
Enable code analysis via clang-tidy in Visual Studio

theComputeKid:
### What does this PR do?

This PR enables the selection of clang-tidy as a code analyser when the code analyser action of Visual Studio is run in the IDE. Further details can be found in the [official documentation](https://learn.microsoft.com/en-us/cpp/code-quality/clang-tidy?view=msvc-170).

### How does this PR change Premake's behavior?

Add configuration-scoped option `clangtidy` for Visual Studio 2019 and above.

I have tried to follow pre-existing conventions:
- The option name was a lowercase boolean based on `compilebuildoutputs` as an example.
- The test_output_props tests for vs2019 were based on vs2022 test_output_props.
- A test was also added to ensure pre-vs2019 configs were not affected.
- Documentation was added based on the `buildstlmodules` example, which was a similar boolean.

### Did you check all the boxes?

- [x]  Focus on a single fix or feature; remove any unrelated formatting or code changes

- [x]  Add unit tests showing fix or feature works; all tests pass
- [ ]  Mention any [related issues](https://github.com/premake/premake-core/issues) (put closes #XXXX in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] Minimize the number of commits
- [x] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

PR #2190, opened 2024-02-28
Add option to enforce build-time code analysis for VS projects

theComputeKid:
**What does this PR do?**

This is a follow up to a previous PR: #2187 

After enabling clang-tidy as an option in the VS IDE (#2187), this patch allows us to enforce it at build time. This can significantly increase build time, as you would expect from any static analysis tool. When building VS projects using this option, along with #2187, the log will show:

```
     2>ClangTidy:
         C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\Llvm\x64\bin\clang-tidy.exe -p="..\myProj.ClangTidy"
         Suppressed 3 warnings (3 in non-user code).
         Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

A few people have asked for compile_command.json generation. On Linux/Macos, tools exist to generate this (e.g. `bear` etc). This is tricky on windows. As an extra side-benefit of this change, as seen in the log above, a compile_commands.json file is generated in the location specified with the `-p` flag. However, while useful, this is a side-effect of this change and not its primary purpose.

Add configuration-scoped option `runcodeanalysis` for Visual Studio 2019 and above.


Jarod42:
> A few people have asked for compile_command.json generation. 

There is 2 (external) premake modules which should do that:
- https://github.com/tarruda/premake-export-compile-commands
- https://github.com/MattBystrin/premake-ecc

Jarod42 (approved):

nickclark2016 (changes requested):
This is missing the API entry in sidebars.js

theComputeKid:
@nickclark2016 : If you don't mind, could you also review the addition of `clangtidy` to the sidebar, as it is related to this change. I was unaware of the need to add this and I missed it in #2187. In any case, I will make the addition you requested. Thank you.

nickclark2016 (approved):
LGTM

PR #2192, opened 2024-03-05
Adjust Tokens.md to correspond to codebase

UAVXP:
Tokens page shows that {MKDIR} only calls mkdir Windows command, so it would be confusing to people who write "if not exist {MKDIR} blabla" when they happen to see an error while compiling

**What does this PR do?**

It fixes a documentation error

It doesn't

This misbehavior happened to me after I wrote something like "if not exist path {MKDIR} path",
 so it translated into "if not exist path IF NOT EXIST path (mkdir path)" and got an error.


KyrietS (approved):
You're right. The docs should explicitly say what command is going to be executed 👍

nickclark2016 (approved):

PR #2205, opened 2024-04-28
Fix wrong system include when targeting Emscripten.

tritao:
Fixes an extra system include when targetting Emscripten with the Emscripten Premake extension module.

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -505,14 +505,16 @@
 		architecture = {
 			x86 = function (cfg)
 				local r = {}
-				if not table.contains(os.getSystemTags(cfg.system), "darwin") then
+				if not (table.contains(os.getSystemTags(cfg.system), "darwin")
+					or table.contains(os.getSystemTags(cfg.system), "emscripten")) then
 					table.insert (r, "-L/usr/lib32")
 				end
 				return r
 			end,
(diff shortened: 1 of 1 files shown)


src/tools/gcc.lua
```diff
@@ -505,14 +505,16 @@
 		architecture = {
 			x86 = function (cfg)
 				local r = {}
-				if not table.contains(os.getSystemTags(cfg.system), "darwin") then
+				if not (table.contains(os.getSystemTags(cfg.system), "darwin")
+					or table.contains(os.getSystemTags(cfg.system), "emscripten")) then
```
samsinsane, line 509: Core doesn't support Emscripten as a target, and shouldn't contain fixes for it. There are two things the Emscripten module should fix:
1. It should add appropriate architectures (`asmjs` and/or `wasm32`, maybe `wasm64` too) instead of hijacking existing architectures. This is likely a lot of work with a lot of testing required, so doing the following would suffice.
2. It should override the outputs from the `clang` toolset, the architecture library directories override should be done [here](https://github.com/TurkeyMan/premake-emscripten/blob/ffcd11ef70c19bc4b2f7a65a8d7b39d25adc683e/emscripten_emcc.lua#L208-L214).

samsinsane (changes requested):

PR #2206, opened 2024-04-29
Update Emscripten extension fork in modules.md

tritao:
This just updates the URL to my fork, which is an up to date and mantained version.


KyrietS:
Could we ask the original author (@TurkeyMan) to mention your fork as an actively maintained successor in his README.md, if that is his wish? I think that would be more fair.

Also, I could use some help in verifying the legitimacy of your fork (which is 2 days old).

BTW you seem to be very fond of forking projects. You have over 100 of them...

KyrietS:
Sorry if I sounded harsh. I've never used emscripten before, so even though I can see the code, I just have to trust that it will work. I just want to be careful, hence my skepticism 😅

Let's give @TurkeyMan a week to respond and if there is no response, we will just merge your PR.

Anyway, thanks for sharing your patched premake-emscripten extension 🤗

KyrietS (approved):

Issue #2207, opened 2024-04-29
Premake system toolset selection issue


tritao:
# Problem

Lets take a simple of a Premake script with a library and `toolset "gcc"` and generate with `gmake2` action.

The generated Makefile will generate with the default `$(CC)` and `$(CXX)` Make implicit variables.

From the [Make documentation](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html):

**CC** Program for compiling C programs; default `cc`.
**CXX** Program for compiling C++ programs; default `g++`.

Notice is uses `g++` instead of what I would expect `c++``.

On my system `cc` and `c++` are mapped to Clang, so I end up with a weird build mixing both Clang for C and GCC for C++.

This also ends up being a problem with the `premake-export-compile-commands` extension.
https://github.com/tritao/premake-export-compile-commands/commit/495800463587cc8fe817aefcae2d41a10c1e8b89

Even though it uses `toolset.gettoolname`, it ends up generating the JSON file as: `command": "cxx -MD -MP`

Which is wrong, since it will look for system includes with the `cxx` command which is Clang, while the build uses the implicit `g++`.

# Proposed Solution

If I choose `toolset "gcc"`, then `gmake` action should generate the `Makefile` with `C = gcc` and `CXX = g++`.

For the system compiler, a new `toolset "system"`, can be added, default to C = cc` and `CXX = c++`.

Basically the gmake action should always generate explicit `C` and `CXX` rules or we may end up mixing toolsets.

@samsinsane Any thoughts on this?

samsinsane:
> If I choose `toolset "gcc"`, then `gmake` action should generate the `Makefile` with `C = gcc` and `CXX = g++`.

I agree, but even if you don't explicitly pick GCC, if Premake uses GCC it should output that stuff. Otherwise problems like this will persist.

> For the system compiler, a new `toolset "system"`, can be added, default to C = cc` and `CXX = c++`.

I like the idea, but I'm not sure I agree with the usage of `cc` and `c++`. Clang and GCC don't support the same set of flags (there is considerable overlap, but still different) and using `cc` and `c++` would prevent users from sharing their Makefiles (which Premake does for the source releases), so we would need to know what they point to and emitting the actual toolset names would allow users to continue sharing their Makefiles.

Jarod42:
My UT https://github.com/Jarod42/premake-sample-projects/tree/master/projects/project-toolset doesn't catch it directly :-/
[...]

PR #2240, opened 2024-09-01
Fix `gmake`/`gmake2` action mixing up compilers

tritao:
Due to Make behavior, specifically its default values to the CC and CXX implicit rules, we could end up in a scenario where the gcc and clang compilers would be used in the same project.

Be more explicit and do not rely on using the system default compilers, and use the `gcc` and `g++` compilers explicitly.

Fixes https://github.com/premake/premake-core/issues/2207.

### DIFF
--- a/modules/gmake2/tests/test_gmake2_tools.lua
+++ b/modules/gmake2/tests/test_gmake2_tools.lua
@@ -28,9 +28,34 @@
 -- Make sure that the correct tools are used.
 --
 
-	function suite.usesCorrectTools()
+	function suite.usesCorrectTools_gcc()
 		gmake2.cpp.tools(cfg, p.tools.gcc)
 		test.capture [[
+ifeq ($(origin CC), default)
+  CC = gcc
+endif
+ifeq ($(origin CXX), default)
+  CXX = g++
+endif
+ifeq ($(origin AR), default)
+  AR = ar
+endif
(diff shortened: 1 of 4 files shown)


src/tools/gcc.lua
```diff
@@ -684,5 +684,5 @@
 		if ((cfg.gccprefix  or version ~= "") and gcc.tools[tool]) or tool == "rc" then
 			return (cfg.gccprefix or "") .. gcc.tools[tool] .. version
 		end
-		return nil
+		return gcc.tools[tool]
```
Jarod42, line 687: Previous condition might be removed completely and just return
`return (cfg.gccprefix or "") .. gcc.tools[tool] .. version`

Jarod42 (approved):

modules/gmake2/tests/test_gmake2_tools.lua
```diff
@@ -31,6 +31,15 @@
 	function suite.usesCorrectTools()
 		gmake2.cpp.tools(cfg, p.tools.gcc)
 		test.capture [[
+ifeq ($(origin CC), default)
```
nickclark2016, line 34: Would you mind also adding test cases for clang and no tool specified?

nickclark2016 (approved):

PR #2242, opened 2024-09-02
Add git integration.

### LINKED ISSUE #221: Premake interaction with source control
We need to think of strategies to make premake work nice with various source control systems.
The classic problem with premake is that the user does a pull and forgets to re-run premake, resulting in a build with a stale project. Happens way too often to ignore! (in the order of a couple of times a day at my work)

With Git, it's possible to run premake with even hooks, but it's pretty awkward and requires configuration by each user manually. There's no way to commit that to the repo.
Another option I've seen is to re-run premake as a pre-build step, but that just sucks; it's a slow step before every build.

What other strategies have people tried?

I'm wondering if we may be able to improve the situation in premake? Perhaps we could have premake detect that there's no work to do and skip the rebuild if nothing is dirty (making it more viable to run as a pre-build step), but this is hard to detect, and it's still pretty lame.

Jarod42:
**What does this PR do?**

Add `gitintegration` api to have post-checkout hooks

That allows to run premake automatically when switching branch with git.
2 mains usages:
- regenerate a version.h from `git describe` to match current branch
- regenerate solution when files has been added/removed (or premake script changed) between branches

No changes of old behavior.

closes #221

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -1,7 +1,7 @@
 --
 -- _premake_main.lua
 -- Script-side entry point for the main program logic.
--- Copyright (c) 2002-2015 Jason Perkins and the Premake project
+-- Copyright (c) 2002-2024 Jason Perkins and the Premake project
 --
 
 	local shorthelp     = "Type 'premake5 --help' for help"
@@ -333,6 +333,13 @@
 		end
 	end
 
+---
+-- Run git integration part.
+---
+
+	function m.gitHookInstallation()
+		p.git_integration.gitHookInstallation()
+	end
 
 ---
 -- Override point, for logic that should run after validation and
@@ -387,6 +394,7 @@
 		m.bake,
 		m.postBake,
 		m.validate,
+		m.gitHookInstallation,
 		m.preAction,
 		m.callAction,
 		m.postAction,
--- a/website/docs/gitintegration.md
+++ b/website/docs/gitintegration.md
@@ -1,0 +1,66 @@
+Enable git integration to run premake on checkout.
+
+```lua
+gitintegration ("value")
+```
+
+### Parameters ###
+
+| Action          | Description                                                                      |
+|-----------------|----------------------------------------------------------------------------------|
+| Off             | Disable git integration.                                                         |
+| Always          | Run premake on checkout.                                                         |
+| OnNewFiles      | Run premake only when files are added/removed or if premake script has changed.  |
+
+### Applies To ###
+
+Global scope.
+
+### Availability ###
+
+Premake 5.0.0 beta 3 or later.
+
+### Examples ###
+
+Regenerate autoversion.h with git tag when checkout to another branch.
+
+```lua
+gitintegration "Always"
+
+local locationDir = _OPTIONS["to"]
+
+local function autoversion_h()
+	local git_tag, errorCode = os.outputof("git describe --tag --always")
+	if errorCode == 0 then
+		print("git description: ", git_tag)
+		local content = io.readfile("src/autoversion.h.in")
+		content = content:gsub("${GIT_DESC}", git_tag)
+
+		os.mkdir(locationDir)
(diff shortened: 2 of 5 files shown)


website/docs/gitintegration.md
```diff
@@ -0,0 +1,66 @@
...
+gitintegration ("value")
+```
+
+### Parameters ###
+
+| Action          | Description                                                                      |
+|-----------------|----------------------------------------------------------------------------------|
+| Off             | Disable git integration.                                                         |
+| Always          | Run premake on checkout.                                                         |
+| OnNewFiles      | Run premake only when files are added/removed or if premake script has changed.  |
```
Jarod42, line 13: Don't hesitate if you have better naming for value and/or for api

src/_premake_main.lua
```diff
@@ -387,6 +443,7 @@
 		m.bake,
 		m.postBake,
 		m.validate,
+		m.nonActionGeneration,
```
Jarod42, line 446: Don't hesitate if you have better naming.

src/_premake_main.lua
```diff
@@ -333,6 +333,62 @@
...
+		while not os.isdir(path.join(current_path, ".git")) do
+			current_path = path.getdirectory(current_path)
+			if current_path == "" then
+				error("No git root path")
+			end
+		end
+		return current_path
+	end
+
+	local function print_git_post_checkout_hooks(root_path, mode)
```
Jarod42, line 351: What would be a better place for those functions?

Jarod42 (commented):
Wonder if we need other API to avoid to overwrite `.git/hook/post-checkout` if user already has one. How to have both user's scripts and premake's scripts.

As a side note, I tested it on Windows with TortoiseGit, so testing it with other configuration would be fine.

The one who add/Remove files would have to run premake manually though, as it is not a checkout, but a commit. (and hook for pre-commit seems a little late for regeneration).
It seems less problematic than switching branch though, as it is known that files configuration has changed.

src/_premake_main.lua
nickclark2016: Maybe a new file dedicated to git integration?

src/_premake_main.lua
nickclark2016: Maybe git hook installation?

website/docs/gitintegration.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Applies To ###
+
+Global scope.
+
+### Availability ###
+
+Premake 5.0.0 beta 3 or later for Visual Studio 2019 and later.
```
nickclark2016, line 21: Why is this logic specific to VS2019 and later?

src/_premake_main.lua
```diff
@@ -333,6 +333,62 @@
...
+		local args = {}
+		for _, arg in ipairs(_ARGV) do
+			if not (arg:startswith("--file") or arg:startswith("/file")) then
+				table.insert(args, arg);
+			end
+		end
+
+		local indent = ''
+		p.outln('#!bin/sh')
```
nickclark2016, line 360: Should this be `#!/bin/sh`?

nickclark2016 (changes requested):

website/docs/gitintegration.md
Jarod42: Bad copy paste to get *"Premake 5.0.0 beta 3 or later"*.

src/_premake_main.lua
Jarod42: Fixed

website/docs/gitintegration.md
Jarod42: Fixed.

src/_premake_main.lua
Jarod42: New file created

src/_premake_main.lua
Jarod42: Changed

nickclark2016 (approved):

PR #2244, opened 2024-09-05
Introduce a tools `getrelative` function to allow relative path customization.

tritao:
This just centralizes all `project.getrelative` in the GCC toolset, allowing for customization by external code, by overriding this function.

This is needed for instance in Ninja generators, when the location of a project is not the workspace location, but the code of the Ninja file itself should still refer to workspace-relative paths.

### DIFF
--- a/src/tools/dotnet.lua
+++ b/src/tools/dotnet.lua
@@ -286,7 +286,7 @@
 		table.insert(flags, '/noconfig')
 
 		if cfg.project.icon then
-			local fn = project.getrelative(cfg.project, cfg.project.icon)
+			local fn = p.tools.getrelative(cfg.project, cfg.project.icon)
 			table.insert(flags, string.format('/win32icon:"%s"', fn))
 		end
 
--- a/src/base/tools.lua
+++ b/src/base/tools.lua
@@ -59,3 +59,14 @@
 		local parts = identifier:explode("-", true, 1)
 		return p.tools[parts[1]], parts[2]
 	end
+
+
+--
+-- Returns the relative path to passed value for the given project.
+-- This can be overriden by actions to allow for more customized relative
+-- path behaviors.
+--
+
+	function p.tools.getrelative(prj, value)
+		return p.project.getrelative(prj, value)
+	end
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -289,7 +289,7 @@
 		local result = {}
 
 		table.foreachi(cfg.forceincludes, function(value)
-			local fn = project.getrelative(cfg.project, value)
+			local fn = p.tools.getrelative(cfg.project, value)
 			table.insert(result, string.format('-include %s', p.quoted(fn)))
 		end)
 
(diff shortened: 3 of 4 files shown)


Jarod42:
Should also apply to other toolsets (so currently dotnet and msc, as others use `toolset.gcc`)
Whereas premake-ninja doesn't support C# yet, it supports msc.

nickclark2016 (changes requested):
This should also be applied to other toolsets (clang, msc)

src/tools/dotnet.lua
```diff
@@ -286,7 +297,7 @@
 		table.insert(flags, '/noconfig')
 
 		if cfg.project.icon then
-			local fn = project.getrelative(cfg.project, cfg.project.icon)
+			local fn = dotnet.getrelative(cfg.project, cfg.project.icon)
```
Jarod42, line 300: Results in `project.getrelative(cfg.project.project, cfg.project.icon)`, so it should probably be
```suggestion
			local fn = dotnet.getrelative(cfg, cfg.project.icon)
```
Apply to all other places too.

if not, then 

```
function dotnet.getrelative(cfg, value)
```
should probably be

```
function dotnet.getrelative(project, value)
```

Jarod42: > Thanks for the review, surprised the test suite did not catch anything. 😱

Possibly that `cfg.project.project` is also `cfg.project`; Testsuite uses indirectly `gcc.getrelative`, so it behave correctly.
BTW, `gcc.getrelative` has same "issue".

> Also this would have probably have easily been caught with your test suite.

Both can catch this.

> Has there been any discussion in getting that integrated recently?

Premake's team doesn't want integration tests, as they think it is too much work and maintenance (I disagree with them).

nickclark2016 (approved):

nickclark2016:
Will leave open for comment for a few days. Barring any change requests, will merge. Thank you!

tests/tools/test_clang.lua
```diff
@@ -122,3 +122,16 @@ function suite.cxxflags_onSanitizeFuzzer()
...
+--
+-- Check the handling of `getrelative` hook.
+--
+
+function suite.onGetRelative()
+	toolset "clang"
+	prepare()
+	local toolset = p.config.toolset(cfg)
+	print(toolset)
+	test.isnotnil(toolset.getrelative)
```
Jarod42, line 135: Your test is really unit,
and I think it doesn't test what you want to check, i.e `clang.some_function_which_uses_getrelative()` calls actually `clang.getrelative` (especially with gcc redirection)
And I think that test would currently fail :-(

Jarod42: I meant, if user (or premake.ninja) does `clang.getrelative = function (cfg, path) return p.workspace.getrelative(cfg.workspace, value) end`
then `clang.getincludedirs(cfg.includedirs, ..)` would call  `gcc.getincludedirs(cfg.includedirs, ..)` which will call
`gcc.getrelative` and not `clang.getrelative`.

As bug happens with premake.ninja, can you test your changes with problematic case (project location != workspace location) with clang toolset? (https://github.com/Jarod42/premake-sample-projects/tree/location/projects/project-location is one such project (not the main branch))

Jarod42: A priori, I don't see issue with `p.tools.getrelative`

Jarod42 (approved):

PR #2247, opened 2024-09-11
Add clangtidy for gmake2

### LINKED ISSUE #2245: Enable clang-tidy checks for clang gmake2
Some time ago, I added build-time code checking for MSVC (#2190) and static analysis via clang-tidy (#2187). I also wanted to add this for the gmake2 generator, specifically for the clang toolset. 

However, having read through the code, there are multiple different ways to achieve this, and I am not sure what to do. I am happy to add this feature myself if I can get some guidance from people familiar with the codebase:

1. How can I limit enabling clang-tidy to just clang?
2. What changes should be made in the gmake2 generator to make this happen?

To mimic cmake, should I add a second build command here, protected by a conditional:

```
	function cpp.initialize()
		rule 'cpp'
			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
			buildmessage  '$(notdir $<)'
			buildcommands {'$(CXX) %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
```

The second build line will contain the clang-tidy invocation:
```
	function cpp.initialize()
		rule 'cpp'
			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
			buildmessage  '$(notdir $<)'
			buildcommands {
                           '$(CXX) %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"',
                           'clang-tidy -- all necessary clang-tidy arguments along with compiler flags.'}
```
What would a conditional look like? Or, do I add a prebuild command instead? 

Nice to haves: 

1. We would also need to inform clang-tidy where to read the .clang-tidy file from as it may not necessarily be in a standard location or inside the output project folder.
2. Specify location of clang-tidy. Or perhaps this can be taken from llvmdir? But that is only supported for msvc.

theComputeKid:
ClangTidy code checks have been added to the Visual Studio generator in #2187. A request for the equivalent in gmake2 has been made in #2245 as well as in [this](https://github.com/premake/premake-core/pull/2187#discussion_r1494649407) comment. Static analysis is a necessary tool for software developers.

This PR enables clang tidy for gmake2 at the configuration level. It adds a build command to the cpp rule enforcing the checks.

When clangtidy is Off, the build command enforcing clang tidy expands to a no-op, and it is skipped. When clangtidy is On, the command first runs clang tidy on the file, and then compiles it if the checks are successful. This method of working allows finishing the build early if clang tidy fails. This is also how the Visual Studio generator works, as well as how CMake implements clang tidy.

Closes #2245 

I may need more help with this PR. Due to the invasive nature of the changes (a new build command is added to the makefile, which may understandably make the code reviewers uncomfortable), the following alternatives were considered:

1. Add a pre/post build command performing clang tidy checks on all files. This is not possible as it requires a compile_commands.json file, and would probably run every time the build is run. Not a good idea.
2. Conditionally add the build rule only if `clangtidy` is On. I don't know how to do this, because `cfg.clangtidy` does not seem to be available during `cpp.initialize`.

Because of these reasons, I instead decided to go with the implementation in this PR.

Future work is possible to make clang tidy in premake more powerful:

1. Enable per-file clang tidy.
2. Set clang tidy location.
3. Set clang tidy additional args, including config file location.
4. etc etc.

### DIFF
--- a/modules/gmake2/gmake2_cpp.lua
+++ b/modules/gmake2/gmake2_cpp.lua
@@ -60,7 +60,13 @@
 			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
 			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
 			buildmessage  '$(notdir $<)'
-			buildcommands {'$(CXX) %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
+			buildcommands {
+				-- code checking must be before the build, so that if the code checking 
+				-- fails, the build is not completed, and hence, a re-run will trigger 
+				-- the code check again.
+				'$(CLANG_TIDY) "$<" -- %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"',
+				'$(CXX) %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'
+			}
 
 		rule 'cc'
 			fileExtension {".c", ".s", ".m"}
@@ -326,6 +332,7 @@
 			cpp.cppFlags,
 			cpp.cFlags,
 			cpp.cxxFlags,
+			cpp.clangtidy,
 			cpp.resFlags,
 			cpp.libs,
 			cpp.ldDeps,
@@ -425,6 +432,13 @@
 		p.outln('ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS)' .. flags)
 	end
 
+	function cpp.clangtidy(cfg, toolset)
+		if cfg.clangtidy ~= nil then
+			p.outln('CLANG_TIDY = clang-tidy')
+		else
+			p.outln('CLANG_TIDY = \\#')
+		end
+	end
 
 	function cpp.resFlags(cfg, toolset)
 		local resflags = table.join(toolset.getdefines(cfg.resdefines), toolset.getincludedirs(cfg, cfg.resincludedirs), cfg.resoptions)
--- a/modules/gmake2/tests/test_gmake2_clangtidy.lua
+++ b/modules/gmake2/tests/test_gmake2_clangtidy.lua
@@ -1,0 +1,45 @@
+--
+-- test_gmake2_clangtidy.lua
+-- Test ClangTidy support in Makefiles.
+--
+
+local suite = test.declare("gmake2_clangtidy")
+local gmake2 = premake.modules.gmake2
+
+local wks, prj, cfg
+
+function suite.setup()
+	wks = workspace("MyWorkspace")
+	configurations { "Debug", "Release" }
+  targetname "blink"
+	kind "StaticLib"
+	language "C++"
+	prj = test.createProject(wks)
+end
+
+local function prepare()
+	wks = test.getWorkspace(wks)
+	prj = test.getproject(wks, 1)
+	cfg = test.getconfig(prj, "Debug")
+  gmake2.cpp.clangtidy(cfg, toolset)
+  files { "src/hello.cpp", "src/hello2.c" }
+
+end
+
+function suite.clangtidyOn()
+  clangtidy "On"
+
+	prepare()
+
+  test.capture [[
+CLANG_TIDY = clang-tidy
+  ]]
+end
+
+function suite.clangtidyOff()
+	prepare()
+
+  test.capture [[
+CLANG_TIDY = \#
+  ]]
+end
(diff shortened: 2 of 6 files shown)


modules/gmake2/gmake2_cpp.lua
```diff
@@ -60,7 +60,13 @@
 			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
 			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
 			buildmessage  '$(notdir $<)'
-			buildcommands {'$(CXX) %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
+			buildcommands {
+				-- code checking must be before the build, so that if the code checking 
+				-- fails, the build is not completed, and hence, a re-run will trigger 
+				-- the code check again.
+				'$(CLANG_TIDY) "$<" -- %{premake.modules.gmake2.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"',
```
nickclark2016, line 67: How difficult would it be to make it so this isn't emitted when clang-tidy isn't requested, rather than always emitting and just "ignoring" it sometimes?

theComputeKid: @nickclark2016 : I knew someone would ask this question, so I already pre-empted the answer in my original PR description: Unfortunately, at the point where the rules are decided, we do not have access to the value of the clangtidy property. You can see it here in this function, there is no information to play with. And then we have other issues as well like making these rules configuration specific.

Then again, I am an outsider to the premake team, you probably know more about the code and can suggest better.

Jarod42, line 67: `-o "$@" -MF "$(@:%.o=%.d)" -c "$<"` seems superfluous/wrong.

Jarod42: From "`rule`" point of view, we would need
- 2 rules: `rule 'cpp'` and `rule 'cpp_with_clang_tidy'`
- remove the `fileExtensions` and do the dispatch "manually".

modules/gmake2/gmake2_cpp.lua
```diff
@@ -425,6 +432,13 @@
 		p.outln('ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS)' .. flags)
 	end
 
+	function cpp.clangtidy(cfg, toolset)
+		if cfg.clangtidy ~= nil then
```
Jarod42, line 436: "Off" (or false if there is some translation phases for that parameter) would wrongly enter in that case.

modules/gmake2/tests/test_gmake2_clangtidy.lua
```diff
@@ -0,0 +1,45 @@
...
+  clangtidy "On"
+
+	prepare()
+
+  test.capture [[
+CLANG_TIDY = clang-tidy
+  ]]
+end
+
+function suite.clangtidyOff()
```
Jarod42, line 39:
```suggestion
function suite.clangtidyDefaulted()
```
currently, and you might add a dedicated test for `"Off"`, especially when code suggest that it is wrongly handled ;-)

modules/gmake2/gmake2_cpp.lua
theComputeKid: Could you please give some psuedo code to explain what you are saying? I'm not sure I understand.

Jarod42: I mix with customcommand as fileextension cannot be removed...
Idea is to customize rule to handle clangtidy:
if you read https://premake.github.io/docs/Custom-Rules/
we might add `propertydefinition` to handle clangtidy, then we have to activate that property for each `filecfg` with clangtidy.
Not sure at which point it is doable without hack.
I will try to add clang-tidy support to premake-ninja, but it doesn't use (premake) `rule` internally but has "intrinsic" rule.
Above (adapted) method should do the trick for me.

theComputeKid: I see. This sounds very messy. Not sure if there is an acceptable solution.

theComputeKid: Sounds like there is no elegant fix for adding clang tidy for gmake2. It was fun trying to get it to work but no worries, feel free to close this PR.

Jarod42: I succeeded with ninja, and failed with gmake currently (to have a correct per-file per-config)
Some features are already partially working (per-file not taken into account for some generator).
Might be better to have partial working solution than no solution at all...
BTW, I will see if I can improve your PR with my suggestion.

theComputeKid: Awesome, thanks!

PR #2252, opened 2024-09-19
Document `os.rename` and `os.getenv` APIs

tritao:
**What does this PR do?**

Documents the `os.rename` and `os.getenv` APIs


Jarod42 (approved):

website/docs/os.getenv.md
samsinsane, line 1: This function comes from the Lua runtime, I don't think we normally document those functions unless we modify them. On one hand, I can see the value in including functions like `os.getenv`, but on the other, I don't know where we draw the line on documenting Lua functions.

website/docs/os.rename.md
```diff
@@ -0,0 +1,35 @@
+Rename file system files or directories.
+
+```lua
+os.rename("path", "newpath")
+```
+
+### Parameters ###
+
+`path` is the file system path to be renamed.
```
samsinsane, line 9: Missing `newpath`.

website/docs/os.rename.md
samsinsane, line 1: This is similar to the above, though we have extended `os.rename` to support paths with wide characters on Windows. Is that enough to warrant documenting the API?

website/sidebars.js
```diff
@@ -399,6 +400,7 @@ module.exports = {
 						'os.outputof',
 						'os.pathsearch',
 						'os.realpath',
+						'os.rename',
 						'os.remove',
```
samsinsane, line 404: These are around the wrong way.

Jarod42 (approved):

samsinsane (approved):
I'm going to approve and merge this, though I'm not sure I agree with documenting Lua APIs like this, but I do agree we should have _something_ for them.

PR #2259, opened 2024-09-25
Use `stdbool.h` instead of defining custom bool types.

tritao:
**What does this PR do?**

Just modernizes the codebase a little bit as I don't think Premake needs to define its own bools anymore?

### DIFF
--- a/src/host/os_compile.c
+++ b/src/host/os_compile.c
@@ -46,7 +46,7 @@ int os_compile(lua_State* L)
 			return 2;
 		}
 
-		lua_dump(P, writer, outputFile, FALSE);
+		lua_dump(P, writer, outputFile, false);
 		fclose(outputFile);
 
 		lua_close(P);
--- a/src/host/os_copyfile.c
+++ b/src/host/os_copyfile.c
@@ -29,7 +29,7 @@ int os_copyfile(lua_State* L)
 		return lua_error(L);
 	}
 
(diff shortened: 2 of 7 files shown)


Jarod42 (approved):

samsinsane (changes requested):
This is a tricky PR. I agree with the idea of using `bool`, `true` and `false` where applicable. However, I'm not sure it's applicable to any instances in this PR.

Firstly, the Lua functions return an `int` not a `bool`, so we should continue to use an `int` value - from C23 onwards, `true` and `false` will be of type `bool` not `int` and also won't be defined in the `stdbool.h` header. Similarly, the Lua functions that are used, such as `lua_dump`, take an `int` not a `bool`, same logic.

The key issue here is getting type warnings when C23 or later becomes the default. Also, there's more than 6 C files in `src/host`, so we're obviously not returning `TRUE` and `FALSE` on all of the Lua functions, consider just changing them to `1` and `0`. If they're internal functions then change the return type to `bool` - `getKernelVersion` and `getversion` in `os_getversion.c` are internal functions that could return `bool` instead.

Secondly, Windows API functions use `BOOL` and we should use `TRUE` and `FALSE` as is intended as [some functions](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmessage#return-value) can have a third value. The same logic should apply to the Objective-C `Boolean` type used in `os_getversion.c`.

PR #2260, opened 2024-09-25
Adds a new toolset `driver` flag to signal driver compatibility.

tritao:
This allows external consumers of those toolsets to know what kind of driver the tool uses.

This can be important in some cases, like in Ninja, which provides some dependency handling features that need to change depending if the compiler is GCC or MSVC-like.

The problem is that due to lack of this information, the Ninja generator ends up having hardcoded logic for this depending on the toolset, https://github.com/jimon/premake-ninja/blob/master/ninja.lua#L401.

Which this is not extensible to other modules that define their own toolset (like the Emscripten module).

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -44,6 +44,7 @@
 
 	clang.shared = {
 		architecture = gcc.shared.architecture,
+		driver = gcc.shared.driver,
 		flags = gcc.shared.flags,
 		floatingpoint = {
 			Fast = "-ffast-math",
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -51,6 +51,7 @@
 			x86 = "-m32",
 			x86_64 = "-m64",
 		},
+		driver = "gcc",
 		flags = {
 			FatalCompileWarnings = "-Werror",
 			LinkTimeOptimization = "-flto",
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -61,6 +61,7 @@
 			["C"] = "/TC",
 			["C++"] = "/TP",
 		},
+		driver = "msvc",
 		flags = {
 			FatalCompileWarnings = "/WX",
 			LinkTimeOptimization = "/GL",
--- a/tests/tools/test_clang.lua
+++ b/tests/tools/test_clang.lua
@@ -33,6 +33,7 @@
 
 	function suite.tools_onDefaults()
 		prepare()
+		test.isequal("gcc", clang.shared.driver)
 		test.isequal("clang", clang.gettoolname(cfg, "cc"))
 		test.isequal("clang++", clang.gettoolname(cfg, "cxx"))
 		test.isequal("ar", clang.gettoolname(cfg, "ar"))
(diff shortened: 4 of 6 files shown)


samsinsane:
Can you provide an example of how this is intended to be used? Based on what you have linked, it seems like the conditions would still exist but the Clang/GCC condition would reduce to a single check. That's not really a good enough improvement to me to justify this change.

samsinsane:
> It really has nothing to do with cleaning up code but just allowing proper extensibility.

You could achieve pretty much the same result by changing Ninja from `if/elseif` to `if/else` and all GCC-compatible toolsets would work without any modifications. Lots of things seem to be like GCC, but nothing else seems to be like MSVC. However, I think a better solution would be to actually allow for extensibility by allowing a toolset that isn't compatible with either to work without necessarily modifying/extending Ninja.

Having a really quick glance at the Ninja code, it seems like there's two differences between the MSVC and GCC versions. The `command` and `deps`, seems like `deps` is just `msvc` or `gcc` these seem like they're Ninja-specific(?) and should live in the Ninja module regardless. Which leaves `command`, both versions largely just boil down to wanting to specify the input file, the output file, the intermediate file, and the compile flag - these flags are duplicated in the gmake and gmake2 modules.

Perhaps starting with sharing the `-c -o <output> -MF <intermediate> <input>` pattern would give other toolsets a better path towards Ninja support. It would also allow make/ninja competitors to not have to duplicate everything again.

samsinsane:
> but that is a different problem and seems unrelated to the issue at hand here.

Then you might need to explain what the issue is because I'm not getting it from what you've said and linked.

To me, it seems like the issue is that the branch conditions are too strict, you want to loosen them and I'm saying just get rid of them. How is getting rid of them not the solution?

samsinsane:
Okay, it looks like we're misunderstanding each other a bit here. The main solution I've been talking about is getting rid of both the `if` and the `elseif`, I don't think they're actually needed but I have also never used Ninja, so I don't know what I'm looking at.

Here's what I'm seeing:
```lua
if toolset == p.tools.msc then
  p.outln("CFLAGS=" .. all_cflags)
  p.outln("rule cc")
  p.outln("  command = " .. cc .. " $CFLAGS" .. " /nologo /showIncludes -c /Tc$in /Fo$out")
  p.outln("  description = cc $out")
  p.outln("  deps = msvc")
  p.outln("")
  -- More rules
elseif toolset == p.tools.clang or toolset == p.tools.gcc then
  -- pch stuff
  p.outln("CFLAGS=" .. all_cflags)
  p.outln("rule cc")
  p.outln("  command = " .. cc .. " $CFLAGS" .. force_include_pch .. " -x c -MF $out.d -c -o $out $in")
  p.outln("  description = cc $out")
  p.outln("  depfile = $out.d")
  p.outln("  deps = gcc")
  p.outln("")
  -- More rules
end
```

Why can't these be merged into a single path? Something like this?
```lua
-- pch stuff
p.outln("CFLAGS=" .. all_cflags)
p.outln("rule cc")
p.outln("  command = " .. cc .. " $CFLAGS" .. force_include_pch .. toolset.getccinoutflags(cfg, "$in", "$out.d", "$out"))
p.outln("  description = cc $out")
p.outln("  depfile = $out.d")
p.outln("  deps = " .. ninja.getdepsname(cfg, toolset)) -- This seems like it requires knowledge of Ninja? So, it shouldn't live in the toolset.
p.outln("")
-- More rules
```

PR #2263, opened 2024-09-26
Adds a new `os.targetarch()` function.

tritao:
**What does this PR do?**

This PR introduces the concept of a target architecture `_TARGET_ARCH`, analogous to the existing `_TARGET_OS`.

So in all my previous Premake projects over the years, I have ended up hacking something comparable to this concept of a target architecture, but it really should be a core Premake concept IMHO.

There is another change that I think should also be done, to default to the target architecture by default. At the moment Premake will default to 32-bits architecture, at least on Windows for example. This behaviour should be updated to target the architecture of the running system by default.

I left that line commented in the PR (in `oven.lua`), would like to understand what's the community's opinion on that before making further changes as I'll need to update some tests.

I'll also add some new tests specifically for the new APIs if people are ok with this change.


samsinsane:
So, I can understand how this would work for a workspace that doesn't specify any `platforms` or `architecture`, but how does this work for workspaces that do? Additionally, how does this work for workspaces that specify multiple `platforms` or `architecture`s?

> There is another change that I think should also be done, to default to the target architecture by default. At the moment Premake will default to 32-bits architecture, at least on Windows for example. This behaviour should be updated to target the architecture of the running system by default.

How do you expect this to work with the workspaces that specify their `platforms` or `architecture`(s)?

samsinsane:
> In this case I think it should work like it currently does with `system` and `os.target()`, in that you have a default value that is assumed, but you can override it with `system`. In current Premake, you can also set `--os` but pass an explicit value to `system` in your configs. The same would be true with `--arch` and `architecture`.

This sounds good, but doesn't quite answer my question because I don't know if `--os` has priority over `system` or not. 😆 Do you happen to know?

> seems like being as most explicit about it would be best, but since it's a pre-existing behaviour, I think it's reasonable to keep it as-is.

I agree with being explicit, but there is also a requirement for files generated by Premake to be shared by default. The only limit on this currently is the target OS, you can't go from Windows to Linux unless you specify `--os=linux`. I remember a few occasions that @starkos mentioned sharing the files was a requirement, but I think most of these were related to people trying to introduce changes like having `links` search for the libraries and error if they're not found. Obviously, that would prevent you from being able to generate a Linux Makefile from Windows since it wouldn't have access to the system libraries by default.

I think your approach to maintaining the existing behaviour is a good call for now.

premake5.lua
```diff
@@ -112,10 +112,9 @@
 			--
 			{ "Win32", "Same as x86" },
 			{ "x64", "Same as x86_64" },
-			--
-			{ "default", "Generates default platforms for targets, x86 and x86_64 projects for Windows." }
 		},
-		default = "default",
+		-- "Generates default platforms for targets, x86 and x86_64 projects for Windows." }
```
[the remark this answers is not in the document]
samsinsane: Line 170 uses the default architecture, I think to preserve that behaviour you need to do something like `options:not arch` instead? We should probably also reconsider the other filters that are duplicating the work you've done in this PR.

samsinsane (approved):

website/docs/premake_OS.md
```diff
@@ -2,6 +2,10 @@
 id: _OS
 ---
 
+:::caution_
+**This has been deprecated in Premake 5.0 alpha 12.** Use the new [_TARGET_OS](premake_TARGET_OS.md) instead.
```
nickclark2016, line 6: Is this really deprecated that far back, or is it deprecated at the current version?

website/docs/premake_TARGET_OS.md
```diff
@@ -0,0 +1,20 @@
...
+
+The current OS may be overridden on the command line with the `--os` option.
+
+```
+$ premake5 --os=linux gmake
+```
+
+### Availability ###
+
+Premake 5.0 alpha 12 or later.
```
nickclark2016, line 15: Is this the correct version?

nickclark2016 (commented):
Please hold on merging until the questions are addressed

website/docs/premake_OS.md
samsinsane: This commit introduces the change, and was first included in alpha 12. 5f589ad5

premake5.lua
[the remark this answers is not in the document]
samsinsane: Oops, I should have had a better look at the filters before commenting. A future PR for the macOS ones makes perfect sense.

PR #2268, opened 2024-09-28
Re-structure common docs files into sub-folders.

tritao:
As title says, just some re-organization to group common files into sub-folders.


website/sidebars.js
```diff
@@ -321,34 +320,36 @@ module.exports = {
...
+						'globals/_TARGET_ARCH',
+						'globals/_TARGET_OS',
+						'globals/_WORKING_DIR',
+						'globals/dofileopt',
+						'globals/iif',
+						'globals/include',
+						'globals/includeexternal',
+						'globals/printf',
+						'globals/require',
+						'globals/verbosef',
```
samsinsane, line 341: I believe you can replace the list of items with a single item:
```js
{
  type: 'autogenerated',
  dirName: 'globals',
},
```
And it should expand to all of the files in the folder. See [here](https://docusaurus.io/docs/sidebar/autogenerated) under the "A real-world example" collapsed section.

@KyrietS I recall when you first introduced this doc system that we had discussed autogenerating the sidebar, was there a reason we didn't do this before? I vaguely recall that it might not have been available at the time, is that right?

samsinsane (approved):

samsinsane:
@tritao I'd like to give @KyrietS some time to provide feedback before I merge this.

PR #2270, opened 2024-09-28
Default to compiling `.m` and `.mm` files as `Objective-C/C++`.

tritao:
As the title says, defaults to compiling `.m` and `.mm` files as `Objective-C/C++`.

Another setting I've often had to set in my projects, and AFAICT there should be no problem setting this as the default.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -2013,6 +2013,12 @@
 	filter { "system:darwin" }
 		toolset "clang"
 
+	filter "files:**.m"
+		compileas(p.OBJECTIVEC)
+
+	filter "files:**.mm"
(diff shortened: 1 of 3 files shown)


src/_premake_init.lua
```diff
@@ -2013,6 +2013,12 @@
 	filter { "system:darwin" }
 		toolset "clang"
 
+	filter "files:**.m"
+		compileas(p.OBJECTIVEC)
+
+	filter "files:**.mm"
+		compileas(p.OBJECTIVECPP)
```
samsinsane, line 2020: Is it possible to have these work the same as the other files instead of using `compileas`?

PR #2271, opened 2024-09-28
Add `linker` flag and `LLD` support.

tritao:
Adds a `linker` flag and implements `LLD` support for `gcc` and `clang` toolset.

### DIFF
--- a/website/docs/linker.md
+++ b/website/docs/linker.md
@@ -1,0 +1,29 @@
+Specifies the linker.
+
+```lua
+linker("value")
+```
+
+### Parameters ###
+
+`value` string, one of:
+
+* `Default` - uses the toolset platform default linker.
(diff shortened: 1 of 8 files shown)


website/docs/linker.md
```diff
@@ -0,0 +1,28 @@
...
+
+```lua
+linker("value")
+```
+
+### Parameters ###
+
+`value` is a string value.
+
+By default, `Default` and `LLD` are allowed, but toolsets or modules canprovide support for other linkers.
```
samsinsane, line 11: This isn't consistent with other documented APIs, which just have a list of the values available in core.

samsinsane (changes requested):

samsinsane (approved):

PR #2274, opened 2024-09-29
Port Premake to Cosmopolitan Libc

tritao:
**What does this PR do?**

Decided to have a little bit of fun tonight and try to port Premake to https://justine.lol/cosmopolitan/index.html

It outputs a single binary that runs natively on Linux + Mac + Windows + FreeBSD + OpenBSD + NetBSD + BIOS on AMD64 and ARM64. Premake at BIOS time anyone? :laughing: 

It was relatively painless and all tests are passing besides the binary module test.

It might be possible to support these with `cosmos_dlopen`, but doesn't seem worth it to spend time on it.

I think this would be most valuable for those who want to bundle a Premake binary with their code as a single binary is needed on Git.

Seems like it would be worth to support this as the changes are pretty minor and maintenance going forward should be minimal, but let me know if you have other thoughts.

If it seems like it would be worth, I can probably add it to the CI build.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -109,6 +109,29 @@ jobs:
       with:
         name: premake-${{ matrix.msystem }}-${{ matrix.platform }}
         path: bin\${{ matrix.config }}\
+  cosmopolitan:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        config: [debug, release]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - uses: tritao/setup-cosmopolitan@v1.3
+      with:
+        version: '3.9.2'
+    - name: Build
+      run: make -f Bootstrap.mak cosmo CONFIG=${{ matrix.config }}
+    - name: Test
+      run: bin/${{ matrix.config }}/premake5 test --test-all
+    - name: Docs check
+      run: bin/${{ matrix.config }}/premake5 docs-check
+    - name: Upload Artifacts
+      if: matrix.config == 'release'
+      uses: actions/upload-artifact@v4
+      with:
+        name: premake-cosmopolitan-universal
+        path: bin/${{ matrix.config }}/premake5
 
   # This job will be required for PRs to be merged.
   # This should depend on (via needs) all jobs that need to be successful for the PR to be merged.
--- a/modules/gmake/_preload.lua
+++ b/modules/gmake/_preload.lua
@@ -20,7 +20,7 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Utility", "Makefile", "None" },
 		valid_languages = { "C", "C++", "C#" },
 		valid_tools     = {
-			cc     = { "clang", "gcc" },
+			cc     = { "clang", "gcc", "cosmocc" },
 			dotnet = { "mono", "msnet", "pnet" }
 		},
 
--- a/premake5.lua
+++ b/premake5.lua
@@ -256,7 +256,7 @@
 			defines     { "LUA_USE_MACOSX" }
 			links       { "CoreServices.framework", "Foundation.framework", "Security.framework", "readline" }
 
-		filter "system:linux"
+		filter { "system:linux", "toolset:not cosmocc" }
 			links		{ "uuid" }
 
 		filter { "system:macosx", "action:gmake" }
@@ -289,13 +289,14 @@
 			include "contrib/curl"
 		end
 
-	group "Binary Modules"
-		include "binmodules/example"
+	if _OPTIONS["cc"] ~= "cosmocc" then
+		group "Binary Modules"
+			include "binmodules/example"
 
-		if not _OPTIONS["no-luasocket"] then
-			include "binmodules/luasocket"
-		end
-
+			if not _OPTIONS["no-luasocket"] then
+				include "binmodules/luasocket"
+			end
+	end
 --
 -- A more thorough cleanup.
 --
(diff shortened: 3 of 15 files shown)


premake5.lua
```diff
@@ -289,13 +289,16 @@
 			include "contrib/curl"
 		end
 
-	group "Binary Modules"
-		include "binmodules/example"
+	filter { "toolset:not cosmocc" }
```
samsinsane, line 292: Did we change `filter` to work on groups and projects? I didn't think that had happened?

src/host/premake.h
```diff
@@ -38,12 +38,15 @@
 #elif defined (__GNU__)
 #define PLATFORM_HURD  (1)
 #define PLATFORM_OS  "hurd"
+#elif defined (__COSMOPOLITAN__)
+#define PLATFORM_COSMO  (1)
+#define PLATFORM_OS  "cosmopolitan"
```
samsinsane, line 43: Won't this mean the default target OS will be this instead of the users host OS?

samsinsane:
I'm not necessarily opposed to this support, but in context of this project moving forward, this may not be viable for Premake to use. The lack of support for binary modules is concerning.

There has been a number of users who have been wanting Premake to stop using the contrib folder as-is. Some just want them to be the latest version available at the time of building, while others want them to be the system library. I think some also want them to be shared libraries so they can update/patch them.

Adding it the CI is fine, but I would be hesitant to make it one of the builds we ship in releases.

samsinsane:
@tritao I should have been clearer in my points, sorry about that. My concerns are from the perspective of this being the only binary we offer, or even one of the binaries we offer. If this is something users have to go out of their way to build, then the extent of my concern is "I'd like to have feature parity, if it's an option".

I'm happy to see this merged in once the CI stuff is sorted.

src/tools/cosmocc.lua
```diff
@@ -0,0 +1,27 @@
...
+cosmocc.ldflags.kind.SharedLib = cosmocc.getsharedlibarg
+
+cosmocc.tools = {
+    cc = "cosmocc",
+    cxx = "cosmoc++",
+    ar = "cosmoar",
+}
+
+function cosmocc.gettoolname(cfg, tool)
+    return cosmocc.tools[tool]
```
Jarod42, line 26: Not sure if it is relevant, but does there is version as cosmoscc-42? (gcc handles it there)

modules/gmake/_preload.lua
```diff
@@ -20,7 +20,7 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Utility", "Makefile", "None" },
 		valid_languages = { "C", "C++", "C#" },
 		valid_tools     = {
-			cc     = { "clang", "gcc" },
+			cc     = { "clang", "gcc", "cosmocc" },
```
Jarod42, line 23: gmake is the *"deprecated"* generator, in favor to gmake2

.github/workflows/ci-workflow.yml
```diff
@@ -109,3 +109,26 @@ jobs:
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - uses: tritao/setup-cosmopolitan@v1.0
+      with:
+        version: '3.9.2'
+    - name: Build
+      run: make -f Bootstrap.mak cosmo CONFIG=${{ matrix.config }}
+    - name: Test
+      run: bin/${{ matrix.config }}/premake5.com test --test-all
```
Jarod42, line 126: `.com` seems to be a typo

Jarod42:
[CI fails](https://github.com/premake/premake-core/actions/runs/11144058827/job/30970477554?pr=2274):
> cosmocc -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_USE_POSIX -DLUA_USE_DLOPEN -I"contrib/lua/src" -I"contrib/luashim" src/host/*.c contrib/lua/src/lapi.c contrib/lua/src/lbaselib.c contrib/lua/src/lbitlib.c contrib/lua/src/lcode.c contrib/lua/src/lcorolib.c contrib/lua/src/lctype.c contrib/lua/src/ldblib.c contrib/lua/src/ldebug.c contrib/lua/src/ldo.c contrib/lua/src/ldump.c contrib/lua/src/lfunc.c contrib/lua/src/lgc.c contrib/lua/src/linit.c contrib/lua/src/liolib.c contrib/lua/src/llex.c contrib/lua/src/lmathlib.c contrib/lua/src/lmem.c contrib/lua/src/loadlib.c contrib/lua/src/lobject.c contrib/lua/src/lopcodes.c contrib/lua/src/loslib.c contrib/lua/src/lparser.c contrib/lua/src/lstate.c contrib/lua/src/lstring.c contrib/lua/src/lstrlib.c contrib/lua/src/ltable.c contrib/lua/src/ltablib.c contrib/lua/src/ltm.c contrib/lua/src/lundump.c contrib/lua/src/lutf8lib.c contrib/lua/src/lvm.c contrib/lua/src/lzio.c  -lm -ldl -lrt
/bin/sh: 1: cosmocc: not found
make: *** [Bootstrap.mak:159: cosmo] Error 127

samsinsane:
> I have put a bit more work into this and everything is "ready", except I am getting a CI-specific http error which I cannot repro locally 😭

Are you testing with Ubuntu 22.04 as well? If so, then maybe just add into the error message in `http_get.c:45` the URL, might be easiest to get it back from the curl object - should be `curl_easy_getinfo` with `CURLINFO_EFFECTIVE_URL`.

samsinsane:
Oh, the HTTP tests require you to explicitly test everything. `premake5 test --test-all`

samsinsane (approved):

PR #2277, opened 2024-09-30
Add C++23 cppdialect

### LINKED ISSUE #2273: C++23 support
`cppdialect` stops at c++20. Is there something blocking adding c++23?

Are you interested in getting a pull request for that?

jlaumon:
**What does this PR do?**

It adds C++23, C++2b, gnu++23 and gnu++2b to `cppdialect`.

It adds the new values for cppdialect and changes the behavior of C++Latest to C++23 on several platforms. I used 4a627cc7642fb379790cc2ac2de9e51d8044d3e8 (which added C++20) as a model for what to do.

Some notes though:
- I did not change DMD and LDC because I'm not sure they support C++23 at all yet.
- I did not change the behavior for vs2015. It has a separate code path from vs2017+ and I don't know if it supports C++23. My thinking was that it's old enough that C++23 support is probably not a concern.
- I did not add the Draft versions (C++2b/gnu++2b) for Visual Studio because the other Drafts weren't in the list. But maybe they should be? (And remap to the regular C++20 instead of being ignored.)

Closes #2273


nickclark2016:
DMD/LDC can be ignored. They are D toolchains, not C++.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1746,6 +1746,8 @@
 				m.element("LanguageStandard", condition, 'stdcpp17')
 			elseif (cfg.cppdialect == "C++20") then
 				m.element("LanguageStandard", condition, iif(_ACTION == "vs2017", 'stdcpplatest', 'stdcpp20'))
+			elseif (cfg.cppdialect == "C++23") then
+				m.element("LanguageStandard", condition, 'stdcpp23')
```
nickclark2016, line 1750: We should probably figure out a better, long term way of handling this, but for now, please change `'stdcpp23'` here to the `iif` statement above. I believe these statements should likely be `<= 'vs2017'` for C++20 and `<= 'vs2019'` for C++23.

In addition, this will require an additional test to be written in the vstudio module.

nickclark2016 (changes requested):

modules/vstudio/vs2010_vcxproj.lua
jlaumon: I'll make the change but I'm curious what is the point of it? For example you might request C++23 on vs2017 and because you get c++latest instead, your code does not compile perhaps in a non-obvious way?
Wouldn't it be better to explicitly use C++23 and let vs2017 give an error if it does not support it? (Or have premake spit an error, but it's harder to know exactly what vs supports).

samsinsane: > I'm curious what is the point of it?

Consistency and scope. As @nickclark2016 said, we need to figure out a better, long term way of handling this. Your argument also applies to C++20 and vs2017, so it goes beyond the scope of this PR. Getting C++23 support in is a good first step, and we can just log an issue to deal with this remapping stuff.

nickclark2016 (approved):

tests/tools/test_gcc.lua
```diff
@@ -991,10 +991,24 @@ end
 		test.contains({ }, gcc.getcflags(cfg))
 	end
 
+	function suite.cxxflags_onCpp2b()
+		cppdialect "C++2b"
+		prepare()
+		test.contains({ "-std=c++2b" }, gcc.getcxxflags(cfg))
+		test.contains({ }, gcc.getcflags(cfg))
```
Jarod42, line 998: I know it is the same as others, but `test.contains({}, any_string)` would always be true, and can be removed.
`test.excludes({ "" -std_c++2b}, gcc.getcflags(cfg))` might make more sense though

Jarod42 (approved):

tests/tools/test_gcc.lua
samsinsane: Nice catch @Jarod42, `test.contains({}, ...)` should trigger an error since it does nothing. I'll submit a PR that introduces an error for doing that, _and_ fix up any existing instances of that. I think we can merge this as-is though.

modules/vstudio/vs2010_vcxproj.lua
jlaumon: I just noticed that vs2022 doesn't actually support stdcpp23 yet, so we should probably change it to use stdcpplatest (ie. replace the <= 2019 by < 2022).

If that sounds OK, I can a new PR tonight.

samsinsane: You could probably just make it emit `stdcpplatest`, either works. Does MSC support C++23, or will that also need to be changed?

nickclark2016: Oh that's a good catch. It does not.

jlaumon: PR #2281 changes both vs2022 and MSC

PR #2280, opened 2024-10-02
Upgrade `libcurl` to latest.

tritao:
**What does this PR do?**

Upgrades libcurl to https://github.com/curl/curl/commit/0acf0bcedad9208703cdbe91b06cacb1065a6e53.

Main motivation is to get the Cosmopolitan build to work, since old libcurl is not working with it.

~Freshly CMake-generated `config-linux.h` with some manual changes for Premake-defined CA paths and mbedTLS defines.~

Using the same configs as previously, just some minor adjustments for some necessary new defines.

Hopefully no behaviour changes.

There were some minor changes around progress function signatures, now it uses `CURLOPT_XFERINFOFUNCTION`, while keeping backwards compatibility with older curl versions.


src/host/curl_utils.h
```diff
@@ -1,7 +1,7 @@
 /**
 * \file   curl_utils.h
 * \brief  curl utilities for the http library.
-* \author Copyright (c) 2017 Tom van Dijck, Jo�o Matos and the Premake project
+* \author Copyright (c) 2017 Tom van Dijck, Jo�o Matos and the Premake project
```
samsinsane, line 4: Your name got a bit broken here.

samsinsane (approved):

PR #2283, opened 2024-10-03
Miscelanneous cleanups

tritao:
**What does this PR do?**

Just a couple of minor cleanups and improvements I've accumulated while working on the codebase.

None intended.

### DIFF
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -1143,25 +1143,6 @@
 
 
 ---
--- Start a new block of configuration settings, using the old, "open"
--- style of matching without field prefixes.
----
-
-	function configuration(terms)
-		-- Sep 16 2021
-		premake.warnOnce("configuration", "`configuration` has been deprecated; use `filter` instead (https://premake.github.io/docs/Filters/)")
-		if terms then
-			if (type(terms) == "table" and #terms == 1 and terms[1] == "*") or (terms == "*") then
-				terms = nil
-			end
-			configset.addblock(api.scope.current, {terms}, os.getcwd())
-		end
-		return api.scope.current
-	end
-
-
-
----
 -- Start a new block of configuration settings, using the new prefixed
 -- style of pattern matching.
 ---
--- a/src/host/path_join.c
+++ b/src/host/path_join.c
@@ -5,6 +5,7 @@
 */
 
 #include "premake.h"
+#include <assert.h>
 #include <string.h>
 #include "path_isabsolute.h"
 
@@ -150,6 +151,17 @@ int path_has_deferred_join(lua_State* L)
 	return 1;
 }
 
+// Copy string "in" with at most "insz" chars to buffer "out", which
+// is "outsz" bytes long. The output is always 0-terminated. Unlike
+// strncpy(), strncpy_t() does not zero fill remaining space in the
+// output buffer:
+// Credit: https://stackoverflow.com/a/58237928
+static char* strncpy_t(char* out, size_t outsz, const char* in, size_t insz){
+    assert(outsz > 0);
+    while(--outsz > 0 && insz > 0 && *in) { *out++ = *in++; insz--; }
+    *out = 0;
+    return out;
+}
 
 int path_resolve_deferred_join(lua_State* L)
 {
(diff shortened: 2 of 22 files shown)


src/host/path_join.c
```diff
@@ -150,6 +150,11 @@ int path_has_deferred_join(lua_State* L)
 	return 1;
 }
 
+#ifdef __GNUC__
+/* avoid GCC warning below: 'strncpy' output truncated before terminating nul copying... */
```
samsinsane, line 154: I think this warning might be highlighting a problem with our usage of `strncpy`. `strncpy(inBuffer, path, len)` is 100% wrong since `len` is based on the length of `path`, which makes it identical to `strcpy(inBuffer, path)`. What we want is to limit the copy by the length of the buffer, `strcpy(inBuffer, path, sizeof(inBuffer))`. If you change it to that, does it still emit the warning? Might need a `-1` on the size, but I don't think so.

src/base/api.lua
```diff
@@ -1148,8 +1148,12 @@
 ---
 
 	function configuration(terms)
+		local caller =  filelineinfo(2)
 		-- Sep 16 2021
-		premake.warnOnce("configuration", "`configuration` has been deprecated; use `filter` instead (https://premake.github.io/docs/Filters/)")
+		premake.warnOnce("configuration", "`configuration` has been deprecated; use `filter` instead (https://premake.github.io/docs/Filters/) @%s\n", caller)
+		if api._deprecations == "error" then
+			error("deprecation errors enabled", 3)
```
samsinsane, line 1155: I think 5 years might be long enough for the deprecation, you can probably just remove the function now...

src/_premake_init.lua
```diff
@@ -274,7 +274,7 @@
 		kind = "string",
 		allowed = {
 			"Default",
-			"c7",
+			p.C7,
```
samsinsane, line 277: Is there a reason to have `p.C7` over treating it like most other API values? Since nothing else seems to work like that, I assume it's a relic of premake4 or even earlier and can probably just be removed instead (the variable `p.C7`, not the value `"c7"`).

samsinsane (changes requested):
As annoying as it might be to submit these individually, I think that might have been the better option. Happy to keep working within the one PR though, but up to you.

src/host/premake.c
```diff
@@ -287,12 +287,14 @@ static void setErrorColor(lua_State* L)
...
-void printLastError(lua_State* L)
+void premake_handle_lua_error(lua_State* L)
 {
 	const char* message = lua_tostring(L, -1);
 	int oldColor = term_doGetTextColor();
 	setErrorColor(L);
-	printf(ERROR_MESSAGE, message);
+	/* avoid printing a double Error: prefix for premake.error() messages */
+	int has_error_prefix = strncmp(message, "** Error:", 9) == 0;
+	printf(has_error_prefix ? "%s\n" : ERROR_MESSAGE, message);
```
Jarod42, line 297: Possibly
```suggestion
	int has_error_prefix = strncmp(message, "** Error:", 9) == 0;
	printf(ERROR_MESSAGE, has_error_prefix  ? message + 9 : message);
```

Jarod42 (approved):

src/host/path_join.c
samsinsane: `strncpy` should copy the null terminator, so the only time that `inBuffer[x] = '\0'` is needed is when `path` is larger than `inBuffer`, so `inBuffer[sizeof(inBuffer) - 1] = '\0'` should always be correct.

So, I'm pretty sure the final solution would have been:
```c
strncpy(inBuffer, path, sizeof(inBuffer));
inBuffer[sizeof(inBuffer) - 1] = '\0';
```

I am concerned with potential bugs in the `strncpy_t` you've got, maybe not in the current version but if someone comes along in the future and changes it at all. However, I'm happy to merge it as-is since there is a comment there with a link to the original version of the function which should be a hint to not modify the function.

> Also C11 has `strcpy_s` which is also interesting, but doesn't provide enough guarantees as the new proposed fix.

My understanding of the `_s` functions is that they're optional and they're mostly not implemented. In addition to that, they're different to the Microsoft ones which have existed longer, and obviously, won't be replaced.

samsinsane:
> But I'll keep it in mind for the next ones and figure out a better workflow, I think using the GH tool it's possible to open PRs directly from the CLI, which should ease the pain a bit.

I've been using TortoiseGit for about a decade now, I'd recommend checking it out if you haven't already. It integrates with the Windows shell, but I generally just operate out of the log window when doing git operations. After pushing it provides a link to create a PR but this should be pretty standard now since I'm pretty sure it's part of the server response message now.

samsinsane (approved):

PR #2284, opened 2024-10-03
Fix `os.host` for Cosmopolitan build

tritao:
**What does this PR do?**

Fixes `os.host` for Cosmopolitan build

Also a minor cleanup where `_TARGET_OS` is moved to Lua which would have prevented this bug in the first place.

Should we still leave an assignement to `_TARGET_ARCH = nil` just for documentation sake that it exists?

### DIFF
--- a/src/_premake_main.lua
+++ b/src/_premake_main.lua
@@ -8,8 +8,11 @@
 	local versionhelp   = "premake5 (Premake Build Script Generator) %s"
 	local startTime     = os.clock()
 
--- set a global.
+-- set main globals.
 	_PREMAKE_STARTTIME = startTime
+
+	-- default the target OS to the host OS
+	_TARGET_OS = os.host()
 
 -- Load the collection of core scripts, required for everything else to work
 
--- a/src/host/os_host.c
+++ b/src/host/os_host.c
@@ -6,8 +6,30 @@
 
 #include "premake.h"
 
+#if PLATFORM_COSMO
+#include <cosmo.h>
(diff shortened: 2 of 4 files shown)


src/_premake_main.lua
```diff
@@ -179,7 +182,14 @@
...
 	function m.prepareAction()
-		p.action.set(_ACTION)
+		if _ACTION then
+			p.action.set(_ACTION)
+		else
+			local action = p.action.default()
+			if action then
+				p.action.set(action.trigger)
+			end
+		end
```
samsinsane, line 192: This seems unrelated to the rest of the PR?

src/_premake_main.lua
```diff
@@ -245,7 +255,7 @@
 
 			local action = p.action.current()
 			if not action then
-				print("Error: no such action '" .. _ACTION .. "'")
+ 				print("Error: no such action '" .. _ACTION .. "'")
```
samsinsane, line 258: Accidental space at the start of the line?

src/_premake_main.lua
```diff
@@ -300,6 +310,10 @@
 ---
 
 	function m.postBake()
+		if not p.action.isConfigurable() then
+			return
+		end
+
```
samsinsane, line 316: Seems unrelated to the rest of the PR?

samsinsane (changes requested):

Jarod42 (approved):

samsinsane (approved):

PR #2289, opened 2024-10-07
.NET C# no way to enable different .NET sdk's on .netcore projects!

### LINKED ISSUE #2288: .NET C# no way  to enable different .NET sdk's on .netcore projects!
Add support for setting .NET sdk using netsdk option. (only on new format projects otherwise just use the defautl)
 - Web
 - Razor
 - Worker
 - Blazor (BlazorWebAssembly)
 - WindowsDesktop
 - default (Microsoft.NET.Sdk)

i don't see an alternative at the moment to my knowledge! 

this change would remove wpf from flags and move it to netsdk option (could leave it in for backwards compatibility)

*(You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to requests like these!)*

lolrobbe2:
**What does this PR do?**

Add support for setting .NET sdk using dotnetsdk option. (only on new format projects otherwise just use the defautl)

Web
Razor
Worker
Blazor (BlazorWebAssembly)
WindowsDesktop
default (Microsoft.NET.Sdk)

adds a new option called dotnetsdk

**for some reason it took the commits from my previous pr (add C# documentation file to) i didn't change these files tho!**

`closes #2288`


modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -820,6 +820,30 @@
...
+		elseif cfg.dotnetsdk == "worker" then
+			return "Microsoft.NET.Sdk.Worker"
+		elseif cfg.dotnetsdk == "blazor" then
+			return "Microsoft.NET.Sdk.BlazorWebAssembly"
+		elseif cfg.dotnetsdk ==  "windowsdesktop" then
+			return "Microsoft.NET.Sdk.WindowsDesktop"
+		elseif cfg.dotnetsdk == "mstest" then
+			return "MSTest.Sdk"
+		elseif not cfg.dotnetsdk then
+			return "Microsoft.NET.Sdk"
```
Jarod42, line 837:
```suggestion
        if not cfg.dotnetsdk then
			return "Microsoft.NET.Sdk"
        end
        local map = {
            ["web"] = "Microsoft.NET.Sdk.Web",
            ["razor"] = "Microsoft.NET.Sdk.Razor",
            ["worker"] = "Microsoft.NET.Sdk.Worker",
            ["blazor"] = "Microsoft.NET.Sdk.BlazorWebAssembly",
            ["windowsdesktop"] = "Microsoft.NET.Sdk.WindowsDesktop",
            ["mstest"] = "MSTest.Sdk",
        }
        return map[cfg.dotnetsdk]
```

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -820,6 +820,30 @@
...
+		end
+	end
+
+	function dotnetbase.netcore.dotnetsdk(cfg)
+		if cfg.dotnetsdk == "mstest" then
+			_p(2,'')
```
Jarod42, line 843: Is it intended to print nothing?

src/_premake_init.lua
```diff
@@ -772,6 +772,20 @@
 		kind = "string",
 	}
 
+	api.register {
```
Jarod42, line 775: Better to place it in the vs file which  calls the `api.register` as msc specific

website/docs/dotnetsdk.md
```diff
@@ -0,0 +1,47 @@
...
+ * [web](https://learn.microsoft.com/en-us/aspnet/core/razor-pages/web-sdk?toc=%2Fdotnet%2Fnavigate%2Ftools-diagnostics%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json)
+ * [razor](https://learn.microsoft.com/en-us/aspnet/core/razor-pages/sdk?toc=%2Fdotnet%2Fnavigate%2Ftools-diagnostics%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json)
+ * [worker](https://learn.microsoft.com/en-us/dotnet/core/extensions/workers)
+ * [blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/)
+ * [windowsdesktop](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop?view=aspnetcore-8.0)
+ * [mstest](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk)
+
+
+## mstest ##
+to make the MSTest sdk work you need to add the version to `global.json`:
```
Jarod42, line 20: Would it make sense to generate that file too?

modules/vstudio/vs2005_dotnetbase.lua
lolrobbe2: i forgot to remove this! this was a remnent of trying to get the version fot mstest sdk specified it does not work when it is not specified! but that issue got fixed by using global.json

website/docs/dotnetsdk.md
lolrobbe2: yes i was thinking about that but how would i make it so you could specify the version? or do i just have a default version and tell people where they can change it! and code for writing the file as i am not familliar with the file api's of premake

src/_premake_init.lua
lolrobbe2: i put it next to the dotnetframework register one

samsinsane (changes requested):

website/docs/dotnetsdk.md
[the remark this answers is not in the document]
lolrobbe2: yes that is correct but specifiying it becomes a little hard so i decided on the global.json route

src/_premake_init.lua
```diff
@@ -618,6 +618,21 @@
...
+			"Default",
+			"Web",
+			"Razor",
+			"Worker",
+			"Blazor",
+			"WindowsDesktop",
+			"MSTest"
+		}
+	}
+
```
Jarod42, line 635: modules/vstudio/_preload.lua or src/tools/dotnet.lua  seems a better place

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -820,6 +816,31 @@
...
+		if cfg.flags.WPF then
+			return map["WindowsDesktop"]
+        end
+
+        return map[cfg.dotnetsdk or "Default"]
+	end
+
+	function dotnetbase.netcore.dotnetsdk(cfg)
+		local globalpath = string.format("%s/global.json", cfg.workspace.location)
+		if cfg.dotnetsdk == "mstest" and not os.isfile(globalpath) then
```
Jarod42, line 839:
```suggestion
		if cfg.dotnetsdk == "MSTest" and not os.isfile(globalpath) then
```

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
```diff
@@ -0,0 +1,121 @@
...
+	function suite.testWindowsDesktop()
+		prepare()
+		dotnetsdk "WindowsDesktop"
+		setConfig()
+
+		test.capture [[
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+		]]
+	end
+	function suite.testMSTest()
```
Jarod42, line 102: global.json part is not tested

lolrobbe2: how would i go about that?

Jarod42: Issue is probably that global.json doesn't follow other file generation and so `dotnetbase.netcore.dotnetsdk` would be misplaced...

lolrobbe2: would you happen to know where i can find another example of this file generating?

lolrobbe2:
Could it be that tests are blocked from creating files? wich would explain why the file is not found!

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
Jarod42: I though about `premake.generate`, something like:
```lua
local content = p.capture(function() generate_global_json_content(prj) end)
if content ~= nil and #content > 0 then
	p.generate(prj, "global.json", function() p.outln(content) end)
end
```
That way, `generate_global_json_content` can be tested.
User is notified of the creation of the file.

Jarod42:
> Could it be that tests are blocked from creating files? what would explain why the file is not found!

There is indeed test_runner.lua which installs custom hooks to avoid file creation.
Test should not use file, but check what they output to them.

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
lolrobbe2: i got premake.generate to work except the test is not working probably somthing stupid that i am missing!

lolrobbe2: image
i got this far but it seems the content is not getting captured properly

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -838,7 +838,10 @@
 		-- do not remove the file generation for global.json otherwise MSTest will not work!!!
 		local globalpath = path.join(cfg.workspace.location, "global.json")
 		if cfg.dotnetsdk == "MSTest" and not os.isfile(globalpath) then
-			io.writefile(globalpath, '{"msbuild-sdks": {"MSTest.Sdk": "3.6.1"}}')
+			local content = p.capture(function() generate_global_json_content(prj) end)
+			if content ~= nil and #content > 0 then
+				p.generate(cfg.workspace, path.join(cfg.workspace.location,"global.json"), function() p.outln(content) end)
+			end
 		end
 	end
```
Jarod42, line 846: I meant:

```suggestion
function dotnetbase.output_global_json(prj) -- unsure how you handle project with different config
		if cfg.dotnetsdk == "MSTest" then
			_p('{"msbuild-sdks": {"MSTest.Sdk": "3.6.1"}}')
    end
end

function dotnetbase.generate_global_json(prj)
			local content = p.capture(function() dotnetbase.output_global_json(prj) end)
			if content ~= nil and #content > 0 then
				p.generate(prj, path.join(prj.workspace.location,"global.json"), function() p.outln(content) end)
			end
end
```
In  vs2005.generateProject(prj)
Add `dotnetbase.generate_global_json(prj)` in `iscsharp` and in `isfsharp` branch.

In test, just check output of `dotnetbase.output_global_json`

lolrobbe2: ill give it try!

lolrobbe2: getting somewhere but now other test is broken

lolrobbe2: fixed it it is now working

lolrobbe2:
the error in mingw is because it cannot connect to nuget services!

lolrobbe2:
question: when is beta 3 planned to be released because i would like to have a look a the <platforms> bug and see if i can fix it potentially?

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -820,8 +816,49 @@
...
+		dotnetbase.generate_global_json(cfg)
+	end
+
 	function dotnetbase.allowUnsafeBlocks(cfg)
 		if cfg.clr == "Unsafe" then
 			_p(2,'<AllowUnsafeBlocks>true</AllowUnsafeBlocks>')
 		end
 	end
+
+	function dotnetbase.output_global_json(prj) -- unsure how you handle project with different config
```
Jarod42, line 847: The comment was for you ;-)
Does it make sense that user have something like

```
filter "configuration:Release"
    dotnetsdk "Web"
filter "configuration:Test"
    dotnetsdk "MSTest"
```
How do you handle it?
As it seems global.json would be ignored with other `dotnetsdk`, checking that at least one cfg has this flag might be the solution

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -820,8 +816,49 @@
...
+			"MSTest.Sdk": "3.6.1"
+		}
+}
+		]])
+		end
+	end
+	function dotnetbase.generate_global_json(prj)
+				local content = p.capture(function() dotnetbase.output_global_json(prj) end)
+				if content ~= nil and #content > 0 then
+					p.generate(prj, path.join(prj.workspace.location,"global.json"), function() p.outln(content) end)
```
Jarod42, line 862:
```suggestion
					p.generate(prj, path.join(prj.workspace.location, "global.json"), function() p.outln(content) end)
```
BTW, I removed your check to not overwrite existing file for simplicity in precedent suggestion.
Unsure of your expected behavior with existing file.
If you keep existing code, doc should be updated as existing file would be overwritten.

modules/vstudio/vs2005_dotnetbase.lua
lolrobbe2: no not really as they are seperate frameworks i see absolutly no reason ass it imports different namespaces etc and would break code!

modules/vstudio/vs2005_dotnetbase.lua
lolrobbe2: preferebly you would just append it to the global.json file or update it but have no idea on how to acomplish that with existing json api's

lolrobbe2: gonna have a look at json.decode and encode to make it so it just updates the version or generates it when missing!

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -845,20 +845,24 @@
...
+			if globaljson == nil then
+				globaljson = {}
+				globaljson["msbuild-sdks"] = {}
+				globaljson["msbuild-sdks"]["MSTest.Sdk"] = "3.6.1"
+			elseif globaljson["msbuild-sdks"] ~= nil then
+				globaljson["msbuild-sdks"]["MSTest.Sdk"] = "3.6.1"
+			else
+				globaljson["msbuild-sdks"] = {}
+				globaljson["msbuild-sdks"]["MSTest.Sdk"] = "3.6.1"
+			end
```
Jarod42, line 859:
```suggestion
			globaljson = json.decode(io.readfile(path.join(prj.workspace.location, "global.json")))
			globaljson = globaljson or {}
			globaljson["msbuild-sdks"] = globaljson["msbuild-sdks"] or {}
			globaljson["msbuild-sdks"]["MSTest.Sdk"] = "3.6.1"
```

Jarod42 (approved):

samsinsane (changes requested):
Remove everything related to `global.json` and just specify the version number in the csproj. There's no point in generating the separate file if you're just going to stamp over the top of it every time Premake is executed. Furthermore, if there's concern about specifying the SDK version then provide a mechanism for doing so (either a new API, or make it part of the SDK API), don't offload it to modifying generated files after Premake has run.

lolrobbe2:
fair enough but how would i go about that? being able to specify the version but only allow it to be specified with the sdks? but there is also customs sdks. wich would then not work and it would be prefereble that custom sdks work to.

samsinsane:
> fair enough but how would i go about that? being able to specify the version but only allow it to be specified with the sdks? but there is also customs sdks. wich would then not work and it would be prefereble that custom sdks work to.

I might not be understanding what you're ask, but there's a few ways to handle the versioning, each has it's own drawbacks.

Firstly, you could just drop the entire `allowed` part of the API and allow users to specify anything. You could also define a format of something like `<sdk>/<version>` which would allow you to still perform the lookup using `<sdk>` to save users having to type out `Microsoft.NET.` for all of the APIs. This approach would give the most flexibility, but it also means any spelling mistakes from a user would go unnoticed by Premake and they would be relying on VS to provide a meaningful error.

Secondly, you could create a second API to handle `dotnetsdkversion` and just ignore it if a user doesn't specify `dotnetsdk`. We ignore `cppdialect` when a user generates a C# project, so there's no real difference. This approach probably works the best, but it does mean the two values are specified separately and that can sometimes cause problems. This isn't an issue for `language "C++"` and `cppdialect "C++14"`, so shouldn't be an issue here either.

Thirdly, you could do something similar to `toolset` which has a restricted set of values but also allows for `msc-<version>` where `<version>` isn't restricted. This one is probably the trickier one to implement while still allowing users to add new values that support versions. This can be achieved with something like this:
```lua
allowed = {
  "Default",
  "Web",
  "Razor",
  "Worker",
  "Blazor",
  "WindowsDesktop",
  "MSTest",
  function (value, kind)
    -- value is expected to be in the format <sdk>/<version>
    local parts = value:explode("/", true, 1)

    -- Only perform the check when a version is provided to avoid infinite recursing.
    if parts[2] == nil then
      return nil
    end

    -- Check that the specified sdk is in the allowed list
    if api.checkValue(p.field.get("dotnetsdk", parts[1], kind) then
      return value
    end

    -- The specified sdk isn't valid
    return nil
  end,
},
```

Let me know if that doesn't answer your questions, or if any of it didn't make sense.

lolrobbe2:
i went with the toolset way of doing things, there is also support for providing custom sdks and the previous way of just specifying the sdk still works!
thanks for the help and advice @Jarod42 & @samsinsane!!!

lolrobbe2:
now that dotnedSdk can specify WPF projects the WPF flag can be deprecated if i am not mistake just don't know how to go about that!

Jarod42 (approved):

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -826,36 +826,35 @@
...
 
-		if cfg.flags.WPF then
-			return map["WindowsDesktop"]
-        end
+		local sdk = (parts and #parts > 0 and parts[1]) or cfg.dotnetsdk
+		if not parts or #parts < 2 then
+			-- prioritize WPF flag for compatibility
+			if cfg.flags.WPF then
+				return map["WindowsDesktop"]
+			end
```
samsinsane: This probably won't work as expected - it doesn't trigger if a user doesn't specify a `dotnetsdk` value, which is the current behaviour.

Deprecating the flag would eliminate this need. In this block:
https://github.com/premake/premake-core/blob/6f0cb4a7ea2aa5d7d1b6024c7d15b82f039efb76/src/_premake_init.lua#L1112-L1129

You'll want to add something like:
```lua
api.deprecateValue("flags", "WPF", 'Use `dotnetsdk "WindowsDesktop"` instead.',
function(value)
  dotnetsdk "WindowsDesktop"
end,
function(value)
  dotnetsdk "Default"
end)
```

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -826,36 +826,35 @@
...
+				p.error("MSTest requires a version to be specified!")
+				return nil
+			end
+
+			return map[sdk or "Default"]
 
-	function dotnetbase.netcore.dotnetsdk(cfg)
-		dotnetbase.generate_global_json(cfg)
+		elseif parts and #parts == 2 then
+			return string.format("%s/%s",map[parts[1]] or parts[1],parts[2])
```
samsinsane:
```suggestion
			return string.format("%s/%s", map[parts[1]] or parts[1], parts[2])
```

samsinsane (changes requested):

lolrobbe2:
the provide snippet wont work because of how it is setup currently ill try and change it so it works

```lua
	function api.addAllowed(fieldName, value)
		local field = p.field.get(fieldName)
		if not field then
			error("No such field: " .. fieldName, 2)
		end

		if type(value) == "table" then
			for i, item in ipairs(value) do
				api.addAllowed(fieldName, item)
			end
		else
			field.allowed = field.allowed or {}

			-- If we are trying to add where the current value is a function,
			-- put the function in a table
			if type(field.allowed) == "function" then
				field.allowed = { field.allowed }
			end

			if field.allowed[value:lower()] == nil then
				table.insert(field.allowed, value)
				field.allowed[value:lower()] = value
			end
		end
	end
```

```
	function api.register(field)
		-- verify the name
		local name = field.name
		if not name then
			error("missing name", 2)
		end

		if rawget(_G, name) then
			error("name '" .. name .. "' in use", 2)
		end

		-- add this new field to my master list
		field, err = p.field.new(field)
		if not field then
			error(err)
		end


		-- Flag fields which contain filesystem paths. The context object will
		-- use this information when expanding tokens, to ensure that the paths
		-- are still well-formed after replacements.

		field.paths = p.field.property(field, "paths")

		-- Add preprocessed, lowercase keys to the allowed and aliased value
		-- lists to speed up value checking later on.

		if type(field.allowed) == "table" then
			for i, item in ipairs(field.allowed) do
				field.allowed[item:lower()] = item
			end
		end

		if type(field.aliases) == "table" then
			local keys = table.keys(field.aliases)
			for i, key in ipairs(keys) do
				field.aliases[key:lower()] = field.aliases[key]
			end
		end

		-- create a setter function for it
		_G[name] = function(value)
			return api.storeField(field, value)
		end

		if p.field.removes(field) then
			_G["remove" .. name] = function(value)
				return api.remove(field, value)
			end
		end

		return field
	end
```

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
```diff
@@ -0,0 +1,188 @@
...
+--
+
+	local wks, prj, cwd
+
+--
+-- Setup and teardown
+--
+	function suite.setup()
+		cwd = os.getcwd()
+		os.chdir(_TESTS_DIR)
```
Jarod42, line 20: @tritao: Is it still needed?
Apply also to
```
function suite.teardown()
		os.chdir(cwd)
	end
```

lolrobbe2:
that is the issue for some reason the value is nill

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
lolrobbe2: removed it!

modules/vstudio/tests/cs2005/test_dotnetsdk.lua
```diff
@@ -0,0 +1,182 @@
...
+-- Copyright (c) 2012-2024 Jason Perkins and the Premake project
+--
+	local p = premake
+	local suite = test.declare("vstudio_cs2005_dotnetsdk")
+	local dn2005 = p.vstudio.dotnetbase
+--
+-- Setup
+--
+
+	local wks, prj, cwd
```
Jarod42, line 13:
```suggestion
	local wks, prj
```

Jarod42 (approved):

Jarod42 (approved):

modules/vstudio/_preload.lua
```diff
@@ -608,6 +608,35 @@
...
+			"WindowsDesktop",
+			"MSTest",
+			function (value)
+				-- value is expected to be in the format <sdk>/<version>
+				local parts = value:explode("/", true, 1)
+				local sdk = parts[1] or value
+
+
+				if parts and #parts == 2 then
+					if p.api.checkValue(p.field.get("dotnetsdk"), parts[1],"string") then
```
samsinsane, line 630:
```suggestion
					if p.api.checkValue(p.field.get("dotnetsdk"), parts[1], "string") then
```

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -57,11 +57,7 @@
 
 	function dotnetbase.projectElement(prj)
 		if dotnetbase.isNewFormatProject(prj) then
-			if prj.flags.WPF then
-				_p('<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">')
-			else
-				_p('<Project Sdk="Microsoft.NET.Sdk">')
-			end
+			_p('<Project Sdk="%s">', dotnetbase.netcore.getsdk(prj))
```
samsinsane, line 60: You need to change all instances of `flags.WPF` to check `dotnetsdk == "WindowsDesktop"`, otherwise when specifying `dotnetsdk "WindowsDesktop"` instead of `flags { "WPF" }` users won't get the expected output.

src/_premake_init.lua
```diff
@@ -355,12 +355,20 @@
...
 		},
 	}
+	--25 November 2024
+	api.deprecateValue("flags", "WPF", 'Use `dotnetsdk "WindowsDesktop"` instead.',
+	function(value)
+		dotnetsdk "WindowsDesktop"
+	end,
+	function(value)
+		dotnetsdk "Default"
+	end)
```
samsinsane, line 371: Deprecations should live further down where I linked to in my previous comment.

src/base/api.lua
```diff
@@ -265,7 +265,11 @@
 
 		if type(field.allowed) == "table" then
 			for i, item in ipairs(field.allowed) do
-				field.allowed[item:lower()] = item
+				if(type(item) == "function") then
+					field.allowed["dynamicvalidator"] = item
+				else
```
samsinsane, line 270: I don't think this is required at all, it also doesn't support multiple functions in the allowed field. Please just change it to this as this block is just about adding lowercase variants to the list.
```suggestion
				if type(item) == "string" then
```

website/docs/dotnetsdk.md
```diff
@@ -0,0 +1,52 @@
+Selects a .NET SDK
+
+```lua
+dotnetsdk "SDK"
+```
+
+For more information see the MSDN documentation [here](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview)
+
+## parameters ##
```
samsinsane, line 9:
```suggestion
## Parameters ##
```

website/docs/dotnetsdk.md
```diff
@@ -0,0 +1,52 @@
...
+ * [Web](https://learn.microsoft.com/en-us/aspnet/core/razor-pages/web-sdk?toc=%2Fdotnet%2Fnavigate%2Ftools-diagnostics%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json)
+ * [Razor](https://learn.microsoft.com/en-us/aspnet/core/razor-pages/sdk?toc=%2Fdotnet%2Fnavigate%2Ftools-diagnostics%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json)
+ * [Worker](https://learn.microsoft.com/en-us/dotnet/core/extensions/workers)
+ * [Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/)
+ * [WindowsDesktop](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop?view=aspnetcore-8.0)
+ * [MSTest](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk)
+
+
+## mstest ##
+when using MSTest sdk a version has to be specified!
```
samsinsane, line 22:
```suggestion
 * [MSTest](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk):  Requires a version be specified.
```

website/docs/dotnetsdk.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Examples ###
+```lua
+dotnetsdk "Web"
+```
+
+```lua
+dotnetsdk "Web/3.4.0"
+
+```
```
samsinsane, line 41: These two lines should be swapped around - the empty line should be outside the snippet.

website/docs/dotnetsdk.md
```diff
@@ -0,0 +1,52 @@
...
+
+```lua
+premake.api.addAllowed("dotnetsdk", "CustomSDK") -- add the customSDK to allowed values for dotnetsdk
+dotnetsdk "CustomSDK/3.4.0"
+```
+
+```lua
+premake.api.addAllowed("dotnetsdk", "CustomSDK") -- add the customSDK to allowed values for dotnetsdk
+dotnetsdk "CustomSDK"
+```
```
samsinsane, line 52: The extra title is a bit weird and would be better served as just a comment like this:
````suggestion
A custom SDK can be specified using the following:

```lua
premake.api.addAllowed("dotnetsdk", "CustomSDK") -- add the custom SDK to allowed values for dotnetsdk
dotnetsdk "CustomSDK"

-- Specifying a version with a custom SDK is also supported
dotnetsdk "CustomSDK/3.4.0"
```
````

**Review by samsinsane — CHANGES_REQUESTED** [review]

**Comment by nickclark2016** [comment]
Can you resolve the merge conflict?

File: src/base/api.lua (hunk 19 above)
Comment (samsinsane) [reply src/base/api.lua:270]: I think there might be a bit of a misunderstanding here. The block here is just adding lowercase versions of items so that you can have "Item" and "item" but when you put "item" you get "Item". There's no lowercase version of a function, so doing anything for a function is wrong. Furthermore, I'm pretty sure if you called `dotnetsdk "dynamicvalidator"` it will accept it and probably error when calling the `explode` function on `dotnetsdk` in `vs2005_dotnetbase.lua` because `explode` probably expects a string.

So to reiterate my comment above, this lowercase thing should only happen when `type(item) == "string"` and nothing should happen for any other type.

**Comment by lolrobbe2** [comment]
update: i think the issue is with the fact that it does not do dictionary values:
@samsinsane 

so the solution would probaly be to have a portion where it handles the dictionary values because if i looked it up right ipairs only returns the indexed pairs
````lua
for i, item in ipairs(field.allowed) do
    if type(item) == "string" then
         field.allowed[item:lower()] = item
    end	
end
``

**Review by Jarod42 — APPROVED** [review]

File: modules/vstudio/_preload.lua [hunk 24]
```diff
@@ -608,6 +608,35 @@
...
+			"Web",
+			"Razor",
+			"Worker",
+			"Blazor",
+			"WindowsDesktop",
+			"MSTest",
+			function (value)
+				-- value is expected to be in the format <sdk>/<version>
+				local parts = value:explode("/", true, 1)
+				local sdk = parts[1] or value
```
samsinsane, line 626: This is unused.

src/base/api.lua
```diff
@@ -265,7 +265,11 @@
 
 		if type(field.allowed) == "table" then
 			for i, item in ipairs(field.allowed) do
-				field.allowed[item:lower()] = item
+				if type(item) == "string" then
+            		field.allowed[item:lower()] = item
+				elseif type(item) == "function" then
+					field.allowed["_function_" .. i .."_"] = item
```
samsinsane, line 271: This still serves no purpose. Just remove it.

samsinsane (changes requested):
There are still outstanding changes requested from previous reviews.

src/base/api.lua
lolrobbe2: well where would the change have to be made so it can pick up the function? because there is support for it in the checkvalue function.
so i am at a loss.

because i don't see anywhere dictionary values are processed.

lolrobbe2:
```lua
	function api.addAllowed(fieldName, value)
		local field = p.field.get(fieldName)
		if not field then
			error("No such field: " .. fieldName, 2)
		end

		if type(value) == "table" then
			for i, item in ipairs(value) do
				api.addAllowed(fieldName, item)
			end
		else
			field.allowed = field.allowed or {}

			-- If we are trying to add where the current value is a function,
			-- put the function in a table
			if type(field.allowed) == "function" then
				field.allowed = { field.allowed }
			end

			if field.allowed[value:lower()] == nil then
				table.insert(field.allowed, value)
				field.allowed[value:lower()] = value
			end
		end
	end
	```
	
	this either allows a single function or  a table of values but not a combination

modules/vstudio/_preload.lua
lolrobbe2: it is used for adding versions for the sdk

src/base/api.lua
lolrobbe2: for some reason it started working?
 I am confused, how?

samsinsane: > apperently it is working and i don't know why

There's a lot of things to explain here. Firstly, the block we're talking about is adding lowercase variants of each item - in addition to the original items, not instead of. This means that `{ "Item" }` will be stored as `{ "Item", ["item"] = "Item" }`. So why do we do this? So that we can do [this](https://github.com/premake/premake-core/blob/1420eae5e6326c69ece4678699873bc95fac21c7/src/base/api.lua#L646-L695).

We do everything using the lowercase variant. The first thing we do is resolve any aliases. The next thing we do depends on what `allowed` is, if it's _just_ a function then we call that, otherwise it's a table and we perform a lookup against the lowercase variant. If either results in a value, we have confirmed that the requested value is valid. Otherwise, we move on to iterating the `allowed` table, if it is one, and execute any functions it contains within `ipairs`. `ipairs` ignores dictionary entries, so the items stored in the snippet above are ignored.

The `addAllowed` function that you posted above allows functions and tables, it just explicitly handles `allowed` being only a function and converts it to be in a table so the new value can be added. However, `addAllowed` does not currently allow another function to be added, but that's unrelated to this PR.

lolrobbe2: @samsinsane thanks for the explanation. it think i understand now why it does work!

lolrobbe2: gonna do the last touchups after the design patterns exam

lolrobbe2:
all outstanding changes should be fixed.

lolrobbe2:
> There are still outstanding changes requested from previous reviews.

@samsinsane all outstanding changes should have been take care of.
Can you pls look over everthing pls?

Issue #2129, opened 2023-08-28
Allow specifying root directory for the project sources


vadz:
For various reasons, I can't have Premake files in the root directory of the project, but have to put it into a subdirectory, let's call it `build/premake`. This means that even my simplest project definitions look like this:

```lua
project "foo"
  kind "StaticLib"
  symbols "On"

  local srcdir = "../../foo/"

  vpaths {
     ["Source Files"] = srcdir .. "**.cpp",
     ["Header Files"] = srcdir .. "**.h",
  }

  includedirs { srcdir .. "include" }

  files {
    srcdir .. "include/foo/bar.h",
    -- a dozen more of them...
    srcdir .. "src/foo/foo.cpp",
    -- and a dozen more of those...
  }
```

Having to write `srcdir ..` everywhere is aggravating.

My existing build system allows to do this instead
```
srcdir ../../foo; // This line makes all the paths relative to this directory.

library foo {
  includedirs = include;

  headers {
    include/foo/foo.h
    ...
  }

  sources {
    src/foo/foo.cpp
    ...
  }

}
```

and I find this much more convenient.

So I wonder if a similar option could be added to Premake? I'd be willing to implement it, but I'd appreciate any hints about how to do it because I am not sure how to get an option set at the current project level in `translate()` function which is where, I think, prepending of `srcdir` should be implemented.

Jarod42:
I wonder if [`os.chdir`](https://premake.github.io/docs/os.chdir) would do the job.
Else I would look at `premake.field.kind("path",`.

vadz:
I thought I could use `basedir()` for this somehow, but I didn't find a way to make it work.

Jarod42:
[`basedir`](https://premake.github.io/docs/basedir/) seems indeed a good track.

PR #2303, opened 2024-10-16
Allow using basedir as a root base directory for files and paths.

tritao:
**What does this PR do?**

This adds some improvements to `basedir` so it can be used to specify a base directory for `file`, `path` and `directory` fields.

Previously it only could be used for tokens, but now it can also affects paths, so depending on the order it's set, it might provide different results.

Another option would be to go with a new `srcdir`, that also seems like a reasonable approach to take.

So feel free to provide some feedback if you have preferences how this should work.

Closes https://github.com/premake/premake-core/issues/2129.

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -51,7 +51,8 @@
 	api.register {
 		name = "basedir",
 		scope = "project",
-		kind = "path"
+		kind = "path",
+		cwdAsBase = true
 	}
 
 	api.register {
--- a/src/base/api.lua
+++ b/src/base/api.lua
@@ -148,7 +148,8 @@
 		-- if I have an existing instance, create a new configuration
 		-- block for it so I don't pick up an old filter
 		if instance then
-			configset.addFilter(instance, {}, os.getcwd())
+			local basedir = p.api.scope.current.basedir or os.getcwd()
+			configset.addFilter(instance, {}, basedir)
 		end
 
 		-- otherwise, a new instance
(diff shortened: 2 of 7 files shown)


nickclark2016:
This reads as an extension of existing behavior, rather than breaking the existing behavior.  Would this be a correct interpretation of your goals?

nickclark2016:
> This reads as an extension of existing behavior, rather than breaking the existing behavior. Would this be a correct interpretation of your goals?

@tritao Just pinging for response here

nickclark2016:
We are still in API break territory since we haven't hit 5.0. It could be worth some investigation into projects using Premake to see if they're using the undocumented behavior (hopefully no Hyrum's Law yet).

nickclark2016:
The more I think about this, I think the better it is to get the @premake/full-team thoughts, since it would be a potentially breaking change for some.

samsinsane:
I don't think I'm really following what `basedir` currently does from what has been said, and the docs suggest it should do what the current PR implements - so I'm not sure what `basedir` is for, but if the functionality is different then maybe we should have a new API to avoid the inevitable complaints.

I haven't tested anything, but does `files { "**.cpp" }` give different results after each call to `basedir`? For example:
```lua
basedir "a"
files { "**.cpp" }
basedir "b"
files { "**.cpp" }
```

Is this the equivalent of `files { "a/**.cpp", "b/**.cpp" }`? If not, should it be? I didn't see a test for this, so should probably include that regardless of which way it works.

What do the changes in `oven.lua` do? I might be wrong, but I don't think it's required for the `api.lua` changes?

nickclark2016:
Personally, I'm fine with breaking the API here. At a minimum, it's a pretty minor break and we can put it in the release notes. Can you provide a minimum example of what would break?

PR #2323, opened 2024-11-11
Post Release 5.0.0-beta3 Cleanup

nickclark2016:
Reverts the version back to 5.0.0-dev in the build, updates copyright year in `src/host/premake.h`

### DIFF
--- a/src/host/premake.h
+++ b/src/host/premake.h
@@ -1,7 +1,7 @@
 /**
  * \file   premake.h
  * \brief  Program-wide constants and definitions.
- * \author Copyright (c) 2002-2021 Jess Perkins and the Premake project
+ * \author Copyright (c) 2002-2024 Jess Perkins and the Premake project
  */
 
 #define lua_c
@@ -12,8 +12,8 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-#define PREMAKE_VERSION        "5.0.0-beta3"
-#define PREMAKE_COPYRIGHT      "Copyright (C) 2002-2021 Jess Perkins and the Premake Project"
+#define PREMAKE_VERSION        "5.0.0-dev"
+#define PREMAKE_COPYRIGHT      "Copyright (C) 2002-2024 Jess Perkins and the Premake Project"
 #define PREMAKE_PROJECT_URL    "https://github.com/premake/premake-core/wiki"
 
 #if defined(__linux__)


Jarod42 (approved):

src/host/premake.h
```diff
@@ -12,8 +12,8 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-#define PREMAKE_VERSION        "5.0.0-beta3"
-#define PREMAKE_COPYRIGHT      "Copyright (C) 2002-2021 Jess Perkins and the Premake Project"
+#define PREMAKE_VERSION        "5.0.0-dev"
```
Jarod42, line 15: How about having version.h.in using `git describe`?

Something like:
```
local function template(file_in, dict, file_out)
	local content = io.readfile(file_in)
	for k,v in pairs(dict) do
		content = content:gsub(k, v)
	end
	os.mkdir(path.getdirectory(file_out))
	local f, err = os.writefile_ifnotequal(content, file_out)

	if (f == 0) then
		-- file not modified
	elseif (f < 0) then
		error(err, 0)
		return false
	elseif (f > 0) then
		print("Generated " .. file_out)
	end
	return true
end

local function autoversion_h()
	local git_tag, errorCode = os.outputof("git describe --tag --always")
	if errorCode ~= 0 then
        error("`git describe --tag` failed with error code", errorCode, git_tag)
    end
    print("git description: ", git_tag)
	return template(
		"src/autoversion.h.in",
		{["@GIT_DESC@"] = git_tag},
		path.join("solution", _ACTION, "autoversion.h")
end

gitintegration "Always" -- to add git hook

if _ACTION ~= nil then -- Do nothing for premake --version :-)
    autoversion_h()
end
```
So no longer need to modify the code, it just use git tag.

nickclark2016: Let's table this for a follow up, just so we can complete the current release cycle under the existing release guide. Then we can discuss updating this for the future in maybe a GH discussion?

Jarod42: No pb to postpone the discussion (I approved BTW).

nickclark2016: Unfortunately, you're not a reviewer with write access so I have to wait for another review :(

starkos (approved):

nickclark2016:
Thanks @starkos :)

PR #2347, opened 2024-11-28
Deprecate FatalWarnings flags in favor of fatalwarnings API

nickclark2016:
**What does this PR do?**

Moves the existing `FatalWarnings`, `FatalCompileWarnings`, and `FatalLinkWarnings` flags to the existing `fatalwarnings` API.

Deprecates the `FatalWarnings` (and friends) flags. Translates the calls to:

```lua
flags { "FatalWarnings" }
-- becomes
fatalwarnings { "Compile", "Link" }

flags { "FatalCompileWarnings" }
-- becomes
fatalwarnings { "Compile" }

flags { "FatalLinkWarnings" }
-- becomes
fatalwarnings { "Link" }
```

Deprecation of flags ahead of 5.0 release.

### DIFF
--- a/modules/vstudio/tests/vc2010/test_compile_settings.lua
+++ b/modules/vstudio/tests/vc2010/test_compile_settings.lua
@@ -158,6 +158,20 @@
 <ClCompile>
 	<PrecompiledHeader>NotUsing</PrecompiledHeader>
 	<WarningLevel>TurnOffAllWarnings</WarningLevel>
+	<Optimization>Disabled</Optimization>
+		]]
+	end
+
+
+	function suite.warningLevel_onNoWarningsOverOtherWarningsAPI()
+		fatalwarnings { "Compile" }
+		warnings "Off"
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>TurnOffAllWarnings</WarningLevel>
+	<Optimization>Disabled</Optimization>
 		]]
 	end
 
@@ -544,8 +558,20 @@
 -- Add <TreatWarningAsError> if FatalWarnings flag is set.
 --
 
-	function suite.treatWarningsAsError_onFatalWarnings()
+	function suite.treatWarningsAsError_onFatalWarningsViaFlag()
 		flags { "FatalCompileWarnings" }
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>Level3</WarningLevel>
+	<TreatWarningAsError>true</TreatWarningAsError>
+		]]
+	end
+
+
+	function suite.treatWarningsAsError_onFatalWarningsViaAPI()
+		fatalwarnings { "Compile" }
 		prepare()
 		test.capture [[
 <ClCompile>
--- a/modules/vstudio/vs2005_dotnetbase.lua
+++ b/modules/vstudio/vs2005_dotnetbase.lua
@@ -286,7 +286,7 @@
 			dotnetbase.allowUnsafeBlocks(cfg)
 		end
 
-		if cfg.flags.FatalCompileWarnings then
+		if p.hasFatalCompileWarnings(cfg.fatalwarnings) then
 			_p(2,'<TreatWarningsAsErrors>true</TreatWarningsAsErrors>')
 		end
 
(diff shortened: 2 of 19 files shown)


Jarod42 (approved):
Testing `filterFatalWarnings` might make sense, i.e:
```
fatalwarnings { "Link", "4996" }
```

src/_premake_init.lua
```diff
@@ -1113,6 +1111,40 @@
 		linktimeoptimization("Default")
 	end)
 
+	api.deprecateValue("flags", "FatalWarnings", "Use `fatalwarnings` instead.",
```
samsinsane, line 1114: Should these perhaps say explicitly what settings to use? ``"Use `fatalwarnings { "Compile", "Link" }` instead."``

src/_premake_init.lua
```diff
@@ -1113,6 +1111,40 @@
...
+
+	premake.filterFatalWarnings = function(tbl)
+		if type(tbl) == "table" then
+			return table.filter(tbl, function(warning)
+				return not (warning == "Compile" or warning == "Link")
+			end)
+		else
+			return tbl
+		end
+	end
```
samsinsane, line 1146: Is this the best location for this?

Similarly, I had a thought that we could also do something during the bake stage to remove and convert compile and link to something like `cfg._fatalcompilewarnings = true` and `cfg._fatallinkwarnings = true`. It would remove any mistakes around using `fatalwarnings` directly (especially in external modules), but it is probably a bit more complicated to make it work like this.

samsinsane (changes requested):
I can see that `src/tools/clang.lua` hasn't been modified, I assume it needs to be updated?

modules/vstudio/vs2005_dotnetbase.lua
```diff
@@ -286,7 +286,7 @@
 			dotnetbase.allowUnsafeBlocks(cfg)
 		end
 
-		if cfg.flags.FatalCompileWarnings then
+		if table.contains(cfg.fatalwarnings, "Compile") then
```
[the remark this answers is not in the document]
nickclark2016: Works for me

website/docs/fatalwarnings.md
```diff
@@ -10,6 +10,13 @@ fatalwarnings { "warnings" }
 
 For Visual Studio, the MSC warning number should be used to specify the warning. On other compilers, the warning should be identified by name.
 
+In addition, Premake provides two special values to turn on all compiler and linker warnings.
+
+| Value   | Description                   |
+-------------------------------------------
+| Compile | Turn on all compiler warnings |
+| Link    | Turn on all linker warnings   |
+
```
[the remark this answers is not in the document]
nickclark2016: Yes, this is much better.

modules/vstudio/tests/vc2010/test_compile_settings.lua
```diff
@@ -161,6 +161,18 @@
...
+
+	function suite.warningLevel_onNoWarningsOverOtherWarningsAPI()
+		fatalwarnings { "Compile" }
+		warnings "Off"
+		prepare()
+		test.capture [[
+<ClCompile>
+	<PrecompiledHeader>NotUsing</PrecompiledHeader>
+	<WarningLevel>TurnOffAllWarnings</WarningLevel>
+		]]
```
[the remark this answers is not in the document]
nickclark2016: Sounds good. I'll fix the other tests. I just copy-pasted the other test using the flags here.

src/_premake_init.lua
nickclark2016: Let's circle back on figuring out a robust way of doing something like this in the bake process. I like the idea, though.

src/_premake_init.lua
```diff
@@ -1113,6 +1111,56 @@
 		linktimeoptimization("Default")
 	end)
 
+	api.deprecateValue("flags", "FatalWarnings", "Use `fatalwarnings { \"Compile\", \"Link\" }` instead.",
```
Jarod42, line 1114: Possibly, to avoid escaping with `\`
```suggestion
	api.deprecateValue('flags', 'FatalWarnings', 'Use `fatalwarnings { "Compile", "Link" }` instead.',
```
(apply to below code too)

nickclark2016: The only reason I'm against that is because we pretty much use quotes everywhere here and that breaks the convention. I could just use them around the warning message string, if we feel that strongly about escaping them

samsinsane (approved):
For a future PR, we should consider an `All` value for those that used `FatalWarnings` before.

PR #2357, opened 2024-12-04
Added API for Action Deprecation

nickclark2016:
**What does this PR do?**

This adds an API for action deprecation. This is in preparation for the official deprecation of gmake.

No breaking changes with this PR. This will be as a stepping stone to introduce breaking changes to actions later.

N/A

- [ ] ~~Add unit tests showing fix or feature works; all tests pass~~ No existing action tests exist

### DIFF
--- a/src/base/action.lua
+++ b/src/base/action.lua
@@ -38,6 +38,8 @@
 --
 
 	action._list = {}
+	action._aliases = {}
+	action._deprecatedaliases = {}
 
 
 ---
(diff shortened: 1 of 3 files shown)


src/base/action.lua
```diff
@@ -68,6 +70,20 @@
...
+			table.foreachi(act.aliases, function(alias)
+				action._aliases[alias] = act.trigger
+			end)
+		end
+
+		-- Add deprecated aliases table
+		if act.deprecatedaliases then
+			for key, value in pairs(act.deprecatedaliases) do
+				action._deprecatedaliases[key] = value
+			end
```
samsinsane, line 85: I think this needs to do one of two things; either automatically register deprecated aliases so they can be specified without the `aliases` table, or, emit a warning/error when the deprecated alias isn't in the aliases table.

My vote is for the auto registration because it's easier to work with.

samsinsane (changes requested):

src/base/action.lua
nickclark2016: Sounds good to me

nickclark2016: I think I'm going to go with both a notify and patching it into the table.

samsinsane (approved):

PR #2366, opened 2024-12-08
Integrated Android module into vstudio module

redorav:
**What does this PR do?**

Integrates VS Android more tightly into the vstudio module, along the lines of the Linux integration. It moves all the functionality of the existing module into vstudio-related files. It also fixes some issues with the previous integration such as tags and options that don't make sense for the Android project type, as well as fixing others (e.g. Android debug PDBs)

It's meant to have no breaking changes. I have tested in my own project that makes use of it and I can build, deploy and debug on VS2019 with an Ant-based project (Ant-based projects were deprecated for VS022 so that would be further work to integrate Gradle projects if I can get them to work). All unit tests are passing locally too.

The idea is to integrate core functionality as tightly as possible with Premake. I personally think the module approach is quite complicated to follow and edit, especially if they are in core already. My hope is that we can bring in other extensions into premake core


website/docs/flags.md
```diff
@@ -36,7 +36,8 @@ flags { "flag_list" }
 | WPF                   | Mark the project as using Windows Presentation Framework, rather than WinForms. |
 | DebugEnvsDontMerge    | Needs documentation                                                        |
 | DebugEnvsInherit      | Needs documentation                                                        |
-| Thumb                 | Needs documentation                                                        |
+| Thumb                 | Android Only. Generate code that executes for thumb microarchitecture. This is applicable for arm architecture only. |
+| DataLevelLinking      | Android Only. Enables linker optimizations to remove unused data by emitting each data item in a separate section. |
```
Jarod42, line 40: Seems it is just docu update,
but notice that flags are deprecating for dedicated api.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +4016,466 @@
...
+	function m.androidClCompilePreprocessorDefinitions(cfg, condition)
+		m.preprocessorDefinitions(cfg, cfg.defines, false, condition)
+	end
+
+	function m.androidLanguageStandardCpp(cfg)
+		local cpp_langmap = {
+			["C++98"]   = "c++98",
+			["C++11"]   = "c++11",
+			["C++14"]   = "c++1y",
+			["C++17"]   = "c++1z",
```
Jarod42, line 4125: Isn't real/recent standard not supported (i.e c++17, c++20, ...)?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +4016,466 @@
...
+
+			if not cfg.architecture or string.startswith(cfg.architecture, "arm") then
+				-- we might want to define the arch to generate better code
+--				if not alreadyHas(cfg.buildoptions, "-march=") then
+--					if cfg.architecture == "armv6" then
+--						table.insert(cfg.buildoptions, "-march=armv6")
+--					elseif cfg.architecture == "armv7" then
+--						table.insert(cfg.buildoptions, "-march=armv7")
+--					end
+--				end
```
Jarod42, line 4172: Comment code to remove?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +4016,466 @@
...
+		if c_langmap[cfg.cdialect] ~= nil then
+			m.element("CLanguageStandard", nil, c_langmap[cfg.cdialect])
+		end
+	end
+
+	function m.androidAdditionalCompileOptions(cfg)
+
+		if _ACTION >= "vs2015" then
+
+		else
```
Jarod42, line 4154: Missing code?
else `if _ACTION < "vs2015" then`

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +4016,466 @@
...
+					mips = "mipsel-linux-android-",
+					mips64 = "mips64el-linux-android-",
+					x86 = "x86-",
+					x86_64 = "x86_64-",
+				}
+				local toolset = defaultToolsetMap[arch]
+
+				if cfg.toolset == "clang" then
+					error("The clang toolset is not yet supported by vs-android", 2)
+					toolset = toolset .. "clang"
```
Jarod42, line 4323: unreachable code.

Jarod42 (approved):

modules/vstudio/vs2010_vcxproj.lua
redorav: Bear in mind a lot of this is just copy pasting the code from the original module, unaltered. There is a possibility to clean it up but it's kind of out of scope of this I think

modules/vstudio/vs2010_vcxproj.lua
redorav: Part of the original code

modules/vstudio/vs2010_vcxproj.lua
redorav: Part of the original code

modules/vstudio/vs2010_vcxproj.lua
redorav: Part of the original code

website/docs/flags.md
redorav: Yeah I get that, there was a stray option in the original vsandroid module and I decided to bring it over. I get that it needs a specific thing but that's probably for another PR

modules/vstudio/vs2010_vcxproj.lua
samsinsane: These aren't the values that the compiler will receive, these are the values that the UI uses. The support also goes back to VS2015, so we can't just change these values as it will break those older project files.

modules/vstudio/vs2010_vcxproj.lua
samsinsane: I'm in favour of just dropping the pre-VS2015 support, we can save that for a future PR.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +3970,451 @@
...
+	-- Shared project generation functions
+	--
+
+	function m.androidLinuxIncludePath(cfg)
+		-- Disable usage of ExternalIncludePath, Android project format ignores it
+		local dirs = vstudio.path(cfg, cfg.externalincludedirs)
+		if #dirs > 0 then
+			m.element("IncludePath", nil, "%s;$(IncludePath)", table.concat(dirs, ";"))
+		end
+	end
```
samsinsane, line 4416: Not sure I agree with this function. The way it works isn't wrong, but extending the original function is probably better. Then just add a function like:
```lua
function m.hasExternalIncludePath(cfg)
  if ... then -- The inverse of the current check
    return true
  end

  return false
end
```
This function allows the extensibility without requiring the entire function contents to be duplicated.

modules/vstudio/vstudio.lua
```diff
@@ -279,8 +282,12 @@
 			arch = iif(isnative, "x86", "Any CPU")
 		end
 
-		if win32 and isnative and arch == "x86" then
-			arch = "Win32"
+		if cfg.system == p.WINDOWS or cfg.system == p.UWP then
```
samsinsane, line 285: I don't think this really helps fix the problem with the `win32` override. I'd rather see this condition used instead of the `true` that is currently used when calling `archFromConfig` or whichever function has the `true` value. Or, just remove the `win32` argument altogether and maybe replace it with a function that can be overridden as required but the initial impl is just `return cfg.system == p.WINDOWS or cfg.system == p.UWP`.

Ultimately, we still want this module to be compatible with being overridden even if Android and Linux support are brought into it instead of being separate modules.

src/base/os.lua
```diff
@@ -836,6 +836,7 @@
 		["solaris"]  = { "solaris", "posix", "desktop" },
 		["uwp"]      = { "uwp", "windows", "desktop" },
 		["windows"]  = { "windows", "win32", "desktop" },
+		[premake.ANDROID]  = { "android", "mobile" },
```
samsinsane, line 839: Alphabetical order, also none of the others use the premake variables, so this probably shouldn't either for consistency. I think this was overlooked for "posix", perhaps it should have it too.

website/docs/flags.md
samsinsane: FWIW, I think `Thumb` already has a dedicated API but I'm not sure if the flag is deprecated.

samsinsane (changes requested):

website/docs/flags.md
redorav: If thumb has an api already I can try to hook it up too

modules/vstudio/tests/cs2005/test_build_events.lua
```diff
@@ -19,6 +19,8 @@
 		p.action.set("vs2005")
 		p.escaper(p.vstudio.vs2010.esc)
 		wks = test.createWorkspace()
+		language "C#"
+		architecture("x86")
```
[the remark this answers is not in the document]
redorav: Not really, the problem is that the unit test was wrong as it was trying to test C# but doesn't specify the C# platform, so was defaulting to Windows, which by default outputs Win32. No architecture in C# defaults to AnyCPU, which is an alternative fix to this unit test. Let me know if you prefer that instead

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -434,6 +518,15 @@
 				m.ruleVars,
 				m.buildLog,
 			}
+
+			if cfg.system == p.ANDROID and _ACTION < "vs2015" then
+				elements = table.join(elements, {
+					m.androidAntBuildPreVS2015,
+				})
+			end
```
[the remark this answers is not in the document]
redorav: I'm all for removing the pre-VS2015 support, it did seem kind of custom to me but I didn't want to go too far. It can be tackled after

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1305,6 +1484,104 @@
...
+	function m.androidManifestSubType(cfg, file)
+		m.element("SubType", nil, "Designer")
+	end
+
+	m.categories.AndroidManifest = {
+		name = "AndroidManifest",
+		priority = 99,
+
+		emitFiles = function(prj, group)
+			m.emitFiles(prj, group, "AndroidManifest", {m.generatedFile, m.androidCategoryLink, m.androidManifestSubType})
```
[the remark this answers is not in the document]
redorav: What do you mean by "only supposed to be done with one value in the table."

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3501,14 +3805,14 @@
 	function m.linuxApplicationType(cfg)
 		m.element("Keyword", nil, "Linux")
 		m.element("RootNamespace", nil, "%s", cfg.project.name)
-		m.element("MinimumVisualStudioVersion", nil, "17.0")
+		m.element("MinimumVisualStudioVersion", nil, "15.0")
```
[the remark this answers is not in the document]
redorav: I thought the MinimumVisualStudioVersion was there to prevent opening a project in earlier versions. The reason I changed it is that I was testing Linux on VS2019 and it wasn't letting me open, so I went to VS2017 to see if it let me create the project, which it did. I can do a mapping from action to number no problem. Something like

vs2017 -> 15
vs2019 -> 16
vs2022 -> 17

Not sure what you mean by "so setting this below the current action version is probably wrong". If I generate a vs2022 Linux project and the minimum is 15, that means it all works fine. If I open it in VS2015 instead, the project won't load. That means you'd have to have a check in premake.lua to exclude it from VS2015

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +3970,451 @@
 		end
 	end
 
+	function m.linuxMultiProcNumber(cfg)
+		-- Linux equivalent of 'MultiProcessorCompilation'
+		-- Default to 8 parallel jobs
+		if cfg.flags.MultiProcessorCompile then
+			m.element("MultiProcNumber", nil, "8")
+		end
+	end
```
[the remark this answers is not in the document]
redorav: I'll remove it and add to another PR

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +3970,451 @@
...
+		else
+			m.element("ApplicationTypeRevision", nil, "1.0")
+		end
+	end
+
+	function m.androidProjectVersion(cfg)
+		_p(2, "<RootNamespace>%s</RootNamespace>", cfg.project.name)
+		_p(2, "<MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>")
+		_p(2, "<ProjectVersion>1.0</ProjectVersion>")
+	end
```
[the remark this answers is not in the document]
redorav: Done

modules/vstudio/vstudio.lua
redorav: The reason I haven't been able to remove the win32 boolean is that some of the unit tests still need fixing and it's non-trivial for me and out of the scope of this I feel to address it. So I fix the most egregious part (it breaks the other platforms) and I'll create an issue to fix it properly

src/base/os.lua
redorav: Sorted

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1000,14 +1147,44 @@
...
-			m.additionalCompileOptions,
 			m.compileAs,
 			m.runtimeTypeInfo,
 
 			-- Linux
 			m.linuxWarningLevel,
 			m.linuxExceptionHandling,
-			m.linuxPIC
+			m.linuxPIC,
+			m.gccClangAdditionalCompileOptions,
```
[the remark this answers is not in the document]
redorav: Removed and made a note for another PR

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -1000,14 +1147,44 @@
 			m.clCompileUndefinePreprocessorDefinitions,
 			m.optimization,
 			m.forceIncludes,
-			m.additionalCompileOptions,
```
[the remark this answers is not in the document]
redorav: Done

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -347,6 +400,34 @@
...
+		else
+			return {
+				m.intDir,
+				m.targetName,
+				m.targetExt,
+				m.includePath,
+				m.libraryPath,
+				m.extensionsToDeleteOnClean,
+				m.executablePath,
+
```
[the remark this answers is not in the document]
redorav: Just to separate Android-specific functionality, I've added a comment above like I do in the other Android and Linux functions

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -347,6 +400,34 @@
 				m.libraryPath,
 				m.extensionsToDeleteOnClean,
 				m.executablePath,
+
+				m.androidLinuxIncludePath,
+				m.linuxMultiProcNumber,
```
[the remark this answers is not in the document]
redorav: Resolved, the extra line is there to clarify it's project-specific as opposed to common functions but I've deleted it for another PR so all good

modules/vstudio/vs2010.lua
```diff
@@ -80,6 +80,19 @@
 						p.touch(prj, ".vcxitems")
 					end
 				end
+
+			elseif prj.kind == p.PACKAGING then
+
+				if project.iscpp(prj) then
```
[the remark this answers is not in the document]
redorav: In my project I've not had to set the system for the packaging project. If I add this check now it breaks the project and is unable to load it. It's not a big issue but also not trivial to diagnose if one forgets. Perhaps I should add the check and an error message? How would I go about doing that?

modules/vstudio/vs2005_solution.lua
```diff
@@ -254,7 +254,12 @@
 
 
 	function sln2005.deploy0(cfg, context)
-		if context.prjCfg.system == p.UWP and not context.excluded and (context.prjCfg.kind == p.WINDOWEDAPP or context.prjCfg.kind == p.CONSOLEAPP) then
+
+		local allowDeployment =
+			(context.prjCfg.system == p.UWP and not context.excluded and (context.prjCfg.kind == p.WINDOWEDAPP or context.prjCfg.kind == p.CONSOLEAPP))
+		or	(cfg.system == p.ANDROID and context.prjCfg.kind == p.PACKAGING and _ACTION >= "vs2015")
```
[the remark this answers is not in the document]
redorav: Can you review the new version of this and let me know if it's what you wanted?

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3666,6 +3970,451 @@
...
+		_p(2, "<RootNamespace>%s</RootNamespace>", cfg.project.name)
+		_p(2, "<MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>")
+		_p(2, "<ProjectVersion>1.0</ProjectVersion>")
+	end
+
+	function m.androidDebugInformation(cfg)
+		if cfg.flags.Symbols then
+			m.element("GenerateDebugInformation", "true")
+		end
+	end
```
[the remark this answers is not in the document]
redorav: You're right, not sure how that slipped by, it's also taken care of by the function below androidDebugInformationFormat, so I'll just remove it

modules/vstudio/vs2010_vcxproj.lua
redorav: Hmm that won't have the same behavior. This is an interesting thing though, it seems as though in e.g. VS2019 these tags are both allowed, I can have IncludePath and ExternalIncludePath, but the existing common code won't allow it. Do you know the rationale for this?

```lua
if #dirs > 0 then
	if _ACTION < "vs2019" then
		m.element("IncludePath", nil, "%s;$(IncludePath)", table.concat(dirs, ";"))
	else
		m.element("ExternalIncludePath", nil, "%s;$(ExternalIncludePath)", table.concat(dirs, ";"))
	end
end
```

Could you suggest how to merge both of these functions into something that makes sense to you?

redorav: On closer study I think I just need to delete the androidLinuxIncludePath and just add the regular one to the Android project so it supports the externalincludedirs api, I'll do that right away

website/docs/flags.md
samsinsane: `thumbmode`, uncertain why it doesn't deprecate the flag though both come from the Android module.

modules/vstudio/tests/cs2005/test_build_events.lua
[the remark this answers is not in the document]
samsinsane: That makes sense, I'll mark this as resolved.

modules/vstudio/vs2005_solution.lua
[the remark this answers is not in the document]
samsinsane: That is exactly what I was thinking. Marking as resolved.

modules/vstudio/vs2010.lua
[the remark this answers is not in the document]
samsinsane: Hmm, this is now outside the scope of this PR, but if you try to use `kind "Packaging"` on Windows, it just generates a `.vcxproj`. I assume with your change it might generate a `.androidproj` instead? That's not necessarily worse, but we should probably emit an error (in a future PR) - that can be done using `p.error`.

modules/vstudio/vs2010_vcxproj.lua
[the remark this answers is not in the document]
samsinsane: Like this: https://github.com/premake/premake-core/blob/8f174c409ab88b3c744441fb7e15085e9c014678/modules/vstudio/vs2010_vcxproj.lua#L1131-L1147

The tables are moved to local variables, which allows the function to remain "clean" and we can easily read what functions are going to be used.

modules/vstudio/vs2010_vcxproj.lua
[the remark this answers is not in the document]
samsinsane: The mapping already exists in each VS `newaction` call: https://github.com/premake/premake-core/blob/8f174c409ab88b3c744441fb7e15085e9c014678/modules/vstudio/vs2015.lua#L65

Which we use like this: https://github.com/premake/premake-core/blob/8f174c409ab88b3c744441fb7e15085e9c014678/modules/vstudio/vs2010_vcxproj.lua#L66-L74

> Not sure what you mean by "so setting this below the current action version is probably wrong".

I don't have a specific example to provide, but this kind of thing does exist in the normal VS path. When VS introduces new settings, old versions end up ignoring it. With the VS Linux implementation in particular, they have made improvements since their vs2015 extension, including a number of new settings which old versions won't understand. So, while an older VS _could_ open it, it may not work due to those missing settings - or may work in a broken or unexpected way.

modules/vstudio/vs2010_vcxproj.lua
samsinsane: The issue here is that Android projects don't support the `ExternalIncludePath` element, so projects with `includedirs` or `externalincludedirs` both fail to build. VS2022 Android projects might support it now, but I'm fairly certain they didn't before - I submitted the override to make Android projects keep using `IncludePath`.

I don't know why for VS2019 and above we only use `ExternalIncludePath` instead of splitting the two lists across the two elements.

> Hmm that won't have the same behavior. 

It should, but I may not have explained my suggestion well enough. Changing the `if _ACTION < "vs2019" then` to `if m.hasExternalIncludePath(cfg) then`, you can make that function return `_ACTION < "vs2019" or cfg.system == p.LINUX or cfg.system == p.ANDROID`. This helper function allows users and modules to easily override the behaviour when/if required.

modules/vstudio/vstudio.lua
samsinsane: That's fair enough. Happy for this conversation to be resolved when you're ready.

website/docs/flags.md
redorav: I addressed this already, it was already using the correct api

modules/vstudio/vs2010.lua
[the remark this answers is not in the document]
redorav: That's interesting, the code I have for creating a packaging project is

```lua
project (AndroidProject)
	removeplatforms("*")
	platforms { PlatformAndroidARM, PlatformAndroidARM64 }
	kind("Packaging") -- This type of project builds the apk
	links (UnitTestProject) -- Android needs to link to the main project which was built as a dynamic library
	androidapplibname(UnitTestProject)
	files
	{
		unitTestDir.."/android/AndroidManifest.xml",
		unitTestDir.."/android/build.xml",
		unitTestDir.."/android/project.properties",
		unitTestDir.."/android/res/values/strings.xml",
	}
```

and it creates the file correctly. I can add the system and the error when I get a minute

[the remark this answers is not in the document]
samsinsane: > I can add the system and the error when I get a minute

I'm happy for this to be done in a follow up PR if you'd prefer.

modules/vstudio/vs2010_vcxproj.lua
redorav: I see what you mean and it could be useful. However in this instance, in the interest of simplicity, I have confirmed that Android VS2019 and 2022 both support the external directory, and VS2017 does not. Therefore the default behavior of the VS module is correct and doesn't need the patched version

modules/vstudio/vs2010.lua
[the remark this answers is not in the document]
redorav: I have done that now, it was easy enough. It's easy enough to update my script and makes sense

nickclark2016 (approved):

website/docs/datalevellinking.md
```diff
@@ -0,0 +1,20 @@
+Enables linker optimizations to remove unused data by emitting each data item in a separate section.
```
Jarod42, line 1:
```suggestion
Emit each data item in a separate section. This help linker optimizations to remove unused data.
```
It look like `-ffunction-sections`/`-fdata-sections` gcc flags.

redorav: Yes, you are correct that these map to -ffunction-sections/-fdata-sections. I have been looking to add the flags in the way you describe, and realized the default is to turn these on. The only reason I was adding this is that in the old module there was a remnant of this option and sounded useful, but given it's on by default, does it still make sense to have it? Is there any reason to turn it off?

Jarod42: https://stackoverflow.com/questions/4274804/query-on-ffunction-section-fdata-sections-options-of-gcc
There are some possible drawbacks

redorav: I've gone ahead and added that last one, let me know

Jarod42: Last one seems better, but that might probably go in another PR.

samsinsane (approved):

p0358:
@redorav Regression:
```diff
	vstudio.vs2010_architectures =
	{
		win32   = "x86",
	}

+	if _ACTION < "vs2015" then
+		vstudio.vs2010_architectures.android = "Android"
+	end

	local function architecture(system, arch)
		local result
		if _ACTION >= "vs2010" then
			result = vstudio.vs2010_architectures[arch] or vstudio.vs2010_architectures[system]
		end
		return result or vstudio.vs200x_architectures[arch] or vstudio.vs200x_architectures[system]
	end
```
this causes:
```
Error: [string "vstudio/vstudio.lua"]:20: attempt to compare nil with string
```

Util scripts can `require("vstudio")` at any point, in order to extend the built-in functions such as `premake.override(premake.vstudio.sln2005, "projects", function(base, wks) ...`, and `_ACTION` is apparently not something defined at this stage. But putting such a check in global seems invalid regardless, there would be circumstances otherwise where it'd be `nil` silently. And wouldn't vstudio.lua be re-used multiple times as a common/base script for multiple vs actions?

redorav:
@p0358 What kind of script causes your issue? I'm surprised _ACTION is not defined at the point where vstudio is loaded, I thought it was always set to something when you run Premake. This was ported from the Android module so perhaps it needs a different place

redorav:
It's probably not a problem to remove it, VS2015 doesn't support Android projects properly anyway

redorav:
I've tried removing the line and one of the unit tests is failing now. I'm trying to figure out what is broken, the test or the line. It's kind of weird because we are setting the action as vs2015 in this line p.action.set("vs2015")

Yet somehow the unit test changes, which makes me think that wasn't really working correctly

p0358:
I kinda figured out what was happening I think!

* `premake5.exe --version` => action = nil
* `premake5.exe version` => action = "version"

It's set in advance indeed, it's just that there isn't always action when a premake command is ran!

As such this is the only workaround that actually works:
```lua
if _ACTION == nil then _ACTION = "" end -- workaround Premake bug, TODO: remove
require("vstudio")
if _ACTION == "" then _ACTION = nil end -- workaround Premake bug, TODO: remove
```

redorav:
I don't know how to add a unit test that checks for that as our unit tests generally check for output in a correctly generated file, I'll ask around. It's good to know you got to the bottom of it for now at least.

p0358:
> I'm trying to figure out what is broken, the test or the line. It's kind of weird because we are setting the action as vs2015 in this line p.action.set("vs2015")

```lua
local p = premake
local suite = test.declare("test_android_files")
local vc2010 = p.vstudio.vc2010


--
-- Setup
--

	local wks, prj

	function suite.setup()
		p.action.set("vs2015")
		wks = test.createWorkspace()
	end
```

If premake.vstudio is already defined, then you're changing the action after the global check for _ACTION from vstudio has already ran it'd seem. So you can see how this preemptive check is problematic and surprising xd

redorav:
Yes that makes sense, it's obvious in hindsight but I'm still learning about premake. We don't care about the workaround for pre-vs2015 so I've fixed the unit tests and hopefully it can get approved soon

p0358:
> I don't know how to add a unit test that checks for that as our unit tests generally check for output in a correctly generated file, I'll ask around. It's good to know you got to the bottom of it for now at least.

If they run multiple tests within a single process where they change _ACTION multiple times (and it seems so), then nothing in globals can depend on a value of _ACTION even moreso, only stuff in functions.

But in such case indeed I also have no idea how a case like this could be covered there without spawning a new process or using some kind of separate/virtual/fake global scope... It's also non-trivial because something like this could potentially happen in any file. I think they'd just have to kinda dry-load every file to a clean premake scope and ensure it doesn't error out, including with both _ACTION being nil and non-nill, and then also set up some kind of getter proxies to raise an error if anything tries to *access* any global that might potentially be changing later, at the moment of loading

redorav:
It's probably the case that existing code is already correct, I'll just have to pay more attention in the future. Thanks for reporting and sorry for the breakage!

PR #2378, opened 2024-12-12
Fixed configuration properties on VS Linux

redorav:
**What does this PR do?**

Fixes configuration properties on VS Linux

No breaking changes

### DIFF
--- a/modules/vstudio/tests/linux/test_linux_files.lua
+++ b/modules/vstudio/tests/linux/test_linux_files.lua
@@ -22,10 +22,10 @@ local vc2010 = p.vstudio.vc2010
 
 
 --
--- Test filtering of source files into the correct categories.
+-- Test link time optimization.
 --
 
-	function suite.none_onLinkTimeOptimization()
+	function suite.linkTimeOptimization_On()
 		linktimeoptimization('on')
 		prepare()
 		test.capture [[
--- a/modules/vstudio/tests/linux/test_linux_files.lua
+++ b/modules/vstudio/tests/linux/test_linux_files.lua
@@ -14,12 +14,17 @@ local vc2010 = p.vstudio.vc2010
 		wks, prj = test.createWorkspace()
 	end
 
-	local function prepare()
+	local function prepareOutputProperties()
+		system "linux"
+		local cfg = test.getconfig(prj, "Debug")
+		vc2010.outputProperties(cfg)
+	end
+
+	local function prepareConfigProperties()
 		system "linux"
 		local cfg = test.getconfig(prj, "Debug", platform)
 		vc2010.configurationProperties(cfg)
 	end
-
 
 --
 -- Test link time optimization.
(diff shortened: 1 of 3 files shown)


modules/vstudio/tests/linux/test_linux_files.lua
```diff
@@ -0,0 +1,38 @@
...
+
+	local function prepare()
+		system "linux"
+		local cfg = test.getconfig(prj, "Debug", platform)
+		vc2010.configurationProperties(cfg)
+	end
+
+
+--
+-- Test filtering of source files into the correct categories.
```
samsinsane, line 25: This comment wasn't updated.

modules/vstudio/tests/linux/test_linux_files.lua
```diff
@@ -0,0 +1,38 @@
...
+--
+
+	function suite.none_onLinkTimeOptimization()
```
samsinsane, line 28: This function would probably be better named as `suite.linkTimeOptimization_On`.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -400,6 +400,9 @@
 				m.libraryPath,
 				m.extensionsToDeleteOnClean,
 				m.executablePath,
+
+				-- Linux
+				m.linuxMultiProcNumber
```
samsinsane, line 405: This needs tests.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3967,6 +3972,20 @@
 		end
 	end
 
+	function m.linuxMultiProcNumber(cfg)
+		-- Linux equivalent of 'MultiProcessorCompilation'
+		-- Default to 8 parallel jobs
+		if cfg.flags.MultiProcessorCompile then
+			m.element("MultiProcNumber", nil, "8")
+		end
```
samsinsane, line 3980: @nickclark2016 worth keeping this in mind when doing the conversion to API, might be nice to be able to specify a number for the systems that allow it.

modules/vstudio/vs2010_vcxproj.lua
```diff
@@ -3967,6 +3972,20 @@
...
+	end
+
+	function m.linuxLinkTimeCodeGeneration(cfg)
```
samsinsane, line 3983: I have a few questions:
1. Was there any reason to not extend the `linkTimeCodeGeneration` function instead?
2. Is this Linux only, or does Android support it too?
3. The API and element are both "link time optimization", why opt for "link time code generation"?

samsinsane (changes requested):

modules/vstudio/tests/linux/test_linux_files.lua
redorav: Makes sense

modules/vstudio/tests/linux/test_linux_files.lua
redorav: Sorry about that, I'll be more careful. Done

modules/vstudio/vs2010_vcxproj.lua
redorav: I was a bit in doubt as I've chosen the number 8 for it which is a bit arbitrary, it's better than not testing anything I suppose

modules/vstudio/vs2010_vcxproj.lua
redorav: Yes agreed that's a nice extension to it

modules/vstudio/vs2010_vcxproj.lua
redorav: 1. This is a good question. On Linux, for whatever reason, the LTO flag is applied to the compiler settings and not to the linker settings, and therefore I need to apply them in a different function. I'm assuming those somehow get passed later to the linker
2. I took a look at Android (on VS2019 as 2022 support for Ant projects is deprecated) and this flag isn't available that I could see
3. I opted for this name because the original function we use is called like that, linkTimeCodeGeneration. And that one is called like that because the vcxproj XML tag is called like that. I can change them all for consistency with the public premake api if we'd rather do that.

samsinsane: I'm happy with the answers you've given, going to resolve the conversation.

samsinsane (approved):

PR #2396, opened 2024-12-26
Remove mingw toolset

Jarod42:
**What does this PR do?**

It removes toolset mingw
which is mostly an alias of "gcc on windows" (for filtering).

`filter { "toolset:mingw" }` is broken, and alternative should be used, as `filter { "system:windows", "toolset:not msc" }`

- Detecting correct mingw-w64/llvm-mingw/msys2 is currently not done anyway
- Some mingw uses clang and not gcc

So some works should be done if we want to handle mingw correctly.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -253,8 +253,8 @@
 			flags       { "NoIncrementalLink" }
 
 		-- MinGW AR does not handle LTO out of the box and need a plugin to be setup
-		filter { "system:windows", "configurations:Release", "toolset:not mingw" }
-			flags		{ "LinkTimeOptimization" }
+		filter { "system:windows", "configurations:Release", "toolset:msc" }
+			flags { "LinkTimeOptimization" }
 
 		filter { "system:uwp" }
 			systemversion "latest:latest"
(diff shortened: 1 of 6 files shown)


premake5.lua
```diff
@@ -312,8 +312,8 @@
 			links       { "ole32", "ws2_32", "advapi32", "version" }
 			files { "src/**.rc" }
 
-		filter "toolset:mingw"
-			links		{ "crypt32", "bcrypt" }
+		filter { "system:windows", "toolset:not msc" }
```
nickclark2016, line 315: What about LLVM with Visual Studio? How should this be handled?

Jarod42: So big build matrix!

I honestly don't know
- if clang under visual should have the link
- if clang under mingw should have the link.
It seems to be `usage` from curl.

I can use `filter { "system:windows", "toolset:gcc" }` to not change those configs

redorav: This feels like a bigger change like we discussed, does this solve an immediate problem?

Jarod42: Discussion was adding extra API which would be a big change IMO.
Here I (propose to) remove the strange toolset mingw (which is no even web documented)

Jarod42: Added https://github.com/premake/premake-core/pull/2576 for LLVM with visual.
Once merged, I would rebase this PR which should answer the original question.

src/_premake_init.lua
```diff
@@ -1280,7 +1280,7 @@
 		allowed = {
 			{ "clang", "Clang (clang)" },
 			{ "gcc", "GNU GCC (gcc/g++)" },
-			{ "mingw", "MinGW GCC (gcc/g++)" },
+			{ "mingw", "MinGW GCC (gcc/g++)" }, -- deprecated
```
nickclark2016, line 1283: Requesting a follow-up PR to allow for deprecation of "allowed" values in "newoption" (via an API)

PR #2417, opened 2025-01-12
fix warnings function for the xcode4 generator

artamonovoleg:
**What does this PR do?**

fixes -Weverything being overriden by xcode defaults

Might break the build for those who used `warnings('Everything')` together with `fatalwarnings`

Xcode by default disables a lot of useful warnings by passing `-Wno-<warning>`


modules/xcode/xcode_common.lua
```diff
@@ -1610,6 +1610,62 @@
 			settings['WARNING_CFLAGS'] = '-Wall -Wextra'
 		elseif cfg.warnings == "Everything" then
 			settings['WARNING_CFLAGS'] = '-Weverything'
+			settings["CLANG_WARN_ASSIGN_ENUM"] = "YES"
```
nickclark2016, line 1613: This feels a bit brittle. Is there a way to disable XCode from disabling warnings, rather than explicitly turning a massive list on and hoping Apple doesn't decide to change the list in a some future release?

Jarod42 (approved):

modules/xcode/xcode_common.lua
samsinsane: @nickclark2016 doesn't look like there's anything. The XCode docs doesn't mention anything about it and for `WARNING_CFLAGS` it just says:
> Space-separated list of additional warning flags to pass to the compiler. Use this setting if Xcode does not already provide UI for a particular compiler warning flag.

However, I'm also not particularly sold on the list in this PR. Specifically, these:
- `GCC_NO_COMMON_BLOCKS` because it is not exactly a warning:
  > In C, allocate even uninitialized global variables in the data section of the object file, rather than generating them as common blocks. This has the effect that if the same variable is declared (without `extern`) in two different compilations, you will get an error when you link them.

- `GCC_WARN_PEDANTIC` because it's more than warnings:
  > Issue all the warnings demanded by strict ISO C and ISO C++; reject all programs that use forbidden extensions, and some other programs that do not follow ISO C and ISO C++. For ISO C, follows the version of the ISO C standard specified by any `-std` option used.

It is also weird that `GCC_WARN_ABOUT_RETURN_TYPE` and `GCC_WARN_UNUSED_VARIABLE` are set further up this function without any context as to why they're always set. I assume that they're the default values of a new project, but at least one of the settings added in this PR would likely also be defaulted to `On` in a new project too. Perhaps we should consider updating that list as it should reduce the list specified for the "everything" block.

artamonovoleg: `GCC_NO_COMMON_BLOCKS` mistakenly copy pasted from xcodeproj, will remove it

`GCC_WARN_PEDANTIC` - from clang page https://clang.llvm.org/docs/UsersManual.html#diagnostics-enable-everything

> In addition to the traditional -W flags, one can enable all diagnostics by passing [-Weverything](https://clang.llvm.org/docs/UsersManual.html#cmdoption-Weverything). This works as expected with [-Werror](https://clang.llvm.org/docs/UsersManual.html#cmdoption-Werror), and also includes the warnings from [-pedantic](https://clang.llvm.org/docs/UsersManual.html#cmdoption-pedantic). Some diagnostics contradict each other, therefore, users of [-Weverything](https://clang.llvm.org/docs/UsersManual.html#cmdoption-Weverything) often disable many diagnostics such as -Wno-c++98-compat and -Wno-c++-compat because they contradict recent C++ standards.

I didn't check other warnings that might match the defaults, my intention was to enable every warning that exists in Xcode as a separate setting.

nickclark2016 (approved):

PR #2418, opened 2025-01-14
Fix regression trying to use _ACTION in the global scope

redorav:
**What does this PR do?**

Fixes a reported issue when calling require(vstudio) at the top of a file. Mentioned here https://github.com/premake/premake-core/pull/2366#issuecomment-2590777784

This is simpler than trying to add the proper fix as versions < 2015 don't support Android anyway

Not really, vs2015 doesn't support Android natively

### DIFF
--- a/modules/vstudio/vstudio.lua
+++ b/modules/vstudio/vstudio.lua
@@ -36,10 +36,6 @@
 		win32   = "x86",
 	}
 
-	if _ACTION < "vs2015" then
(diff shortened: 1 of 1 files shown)


p0358:
Seems it's failing some tests modified in the original PR now. I would suggest using the opportunity to also extend the unit tests by covering the override scenario from docs to avoid something similar in the future: https://premake.github.io/docs/Overrides-and-Call-Arrays/#introducing-overrides

PR #2432, opened 2025-02-19
Usages and Uses API

nickclark2016:
# Problem Solved

Currently, there is no standardized way to handle exposing properties to dependent projects in Premake. Users have been required to build their own mechanisms for sharing properties or rely on third party modules, like [premake-usage](https://github.com/moomalade/premake-usage).  This PR attempts to remedy this by providing a standardized solution to project dependency management in Premake ahead of the 5.0 stable release.

# Proposed API

This PR uses the API implemented by [premake-usage](https://github.com/moomalade/premake-usage) and extends it with CMake-like words of power.  At a high level, a `usage` container defines a set of properties to be consumed at a later point, and a `uses` field consumes a set of `usage` blocks.

In the current implementation, a `usage` may only be defined from within the scope of a project when no filters are active. Similarly, a `uses` block may only be defined from within the scope of a project, but it may be applied while filters are active.

The following is an example of two projects with the proposed API:

```lua
project "A"
    defines { "A_PRIVATE_MACRO" }
    usage "MyUsage"
        defines { "A_MACRO_IN_A_USAGE" }

project "B"
    uses { "MyUsage" }
```

In the above example, `A` will only receive the `A_PRIVATE_MACRO` define, as `A` does not consume the `MyUsage` usage. Project `B` will only receive `A_MACRO_IN_A_USAGE`, as it only consumes the usage `MyUsage`. Properties defined at a project level are explicitly not inherited by the `usage`  containing them.

CMake provides some extremely useful functionality in the form of their `PRIVATE`, `PUBLIC`, and `INTERFACE` keywords. This proposal and PR leverages these words of power in ways that match their meanings in CMake to make implemented Premake scripts based off of CMake easier.

* `PUBLIC` - A usage block with this name will be applied to both the project defining the `usage` and any project consuming it via `uses`.
* `PRIVATE` - A usage block with this name will only be applied to the project defining the `usage` and is inaccessible to other projects.
* `INTERFACE` - A usage block with this name will only be applied to the project consuming the `usage`.

This slightly changes how we think of `uses`, as we could have many projects defining these blocks within the same Premake build. To access these blocks, scripts should specify the project name in the `uses` list.

```lua
project "A"
    usage "PUBLIC"
        defines { "A_PUBLIC" }
    usage "INTERFACE"
        defines { "A_INTERFACE" }
    usage "PRIVATE"
        defines { "A_PRIVATE" }

project "B"
    uses { "A" }
```
In this case, `A` has the `A_PUBLIC` and `A_PRIVATE` defines applied to it, while `B` has the `A_PUBLIC` and `A_INTERFACE` defines applied to it.

# Transitive Dependencies

By default, dependencies consumed via `uses` are not applied transitively. However, this is quite trivial to enforce via `PUBLIC` usages. By specifying a `uses` field in the `PUBLIC` usage block, the usage can be applied to dependency chains indefinitely.

# Other Notes

The `uses` field will not automatically apply `links` and `dependson` relationships between projects. In order to enforce this, authors of projects can utilize `INTERFACE` blocks to add these relationships like so:

```lua
project "A"
    usage "INTERFACE"
        depdendson { "A" }
        links { "A" }
```

For non-magic names, the behavior of `uses` is not well specified when there are multiple of the same name. There is further discussion in the Further Development section.

# Known Limitations and Further Development

Token expansion is an area of concern. In the current iteration of this functionality, avoiding path tokens in custom build commands within usage blocks is recommended until that behavior can be agreed upon and refined.

The primary target of further development would be to allow "usage" blocks at any scope (global, workspace, group). 

There is also room for improvements with `usage` resolution in `uses`. Currently, the behavior of `uses` is not well defined if there are multiple `usage` blocks with the same name (excluding the names of power). To resolve this, one of the following options may be viable (these are not mutually exclusive of each other and not an exhaustive list of options):
* Warn or error out when multiple usage blocks with the same name are detected
* Warn or error out when an ambiguous `uses` is applied
* Provide a scoping operator, similar to C++'s `operator::`, to select a specific `usage` when there is ambiguity. (Example: `:workspace:project:usage`


src/base/oven.lua
```diff
@@ -554,10 +649,12 @@
 			wks = wks,
 			sln = wks,
 			prj = prj,
+      usage = usage,
```
redorav, line 652: tabs v spaces? github is showing these weirdly

website/docs/usage.md
```diff
@@ -0,0 +1,42 @@
...
+```lua
+usage 'MyUsage'
+    defines { 'MY_DEFINE' }
+```
+
+The `usage` API is used to define configuration to be consumed by the `uses` API.  Usages must have unique names, except for magic usage block names (as described below).
+
+### Magic Usage Blocks ###
+
+1. `PRIVATE` - Private specifies a configuration to be automatically applied to the project defining the `usage` block. This block is inaccessible to be consumed by `uses`.
```
redorav, line 12: Would it make sense to say it's equivalent to adding it outside of the usage block?

website/docs/usage.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Example ###
+
+```lua
+project 'A'
+    usage 'PRIVATE'
+        defines { 'A_PRIVATE' }
+    usage 'PUBLIC'
+        defines { 'A_PUBLIC' }
+    usage 'NotMagic'
```
Jarod42, line 26: To enforce correct usage with unique name, using some sort of namespace might be used.
```suggestion
    usage 'A:NotMagic'
```

src/base/oven.lua
```diff
@@ -782,11 +879,188 @@
...
+								if type(absPaths) == 'table' then
+									for i, v in ipairs(absPaths) do
+										relPaths[i] = path.getrelative(prj.location, v)
+									end
+								else
+									relPaths = path.getrelative(prj.location, absPaths)
+								end
+
+								if type(relPaths) == 'table' then
+									verbosef('Translated paths - %s = %s', key, table.concat(relPaths, ","))
```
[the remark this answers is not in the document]
nickclark2016: Just autocomplete. Will normalize.

src/base/usage.lua
```diff
@@ -0,0 +1,95 @@
...
+-- Checks if a usage name is a magic name (PUBLIC, PRIVATE, INTERFACE)
+--
+-- @param name
+--   The name of the usage.
+-- @return
+--   True if the usage is a special usage.
+---
+
+	function usage.isSpecialName(name)
+		return name == usage.PUBLIC or name == usage.PRIVATE or name == usage.INTERFACE
```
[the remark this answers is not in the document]
nickclark2016: Changing to be case sensitive. Will update documentation to note case sensitivity.

tests/oven/test_usages.lua
```diff
@@ -0,0 +1,148 @@
...
+            
+            uses {  "MyProject1" }
+
+        p.oven.bake()
+
+        local cfg = test.getconfig(prj2, "Debug", "x86")
+
+        -- Order is important here, as the test treats different order as a failure
+        -- However, the order of the defines is not important in the actual implementation
+        test.isequal({ "PROJECT_2_PUBLIC", "PROJECT_1_PUBLIC", "PROJECT_1_INTERFACE", "PROJECT_2_PRIVATE" }, cfg.defines) -- Interface defines should not be included
```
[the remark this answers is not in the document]
nickclark2016: Yeah, autocorrect. Fixing docs.

tests/oven/test_usages.lua
```diff
@@ -0,0 +1,148 @@
...
+                defines { "PROJECT_2_PRIVATE" }
+            
+            uses {  "MyProject1" }
+
+        p.oven.bake()
+
+        local cfg = test.getconfig(prj2, "Debug", "x86")
+
+        -- Order is important here, as the test treats different order as a failure
+        -- However, the order of the defines is not important in the actual implementation
```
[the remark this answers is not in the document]
nickclark2016: This is a tricky problem. For now, I will add documentation advising to turn on `linkgroups`.

website/docs/usage.md
nickclark2016: This is the future work proposed in the PR body itself. I'm trying to get us a minimum usable product in this PR.

Jarod42 (approved):

website/docs/uses.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+The `uses` API is used to consume `usage` blocks from within a project. The `usage` blocks are case sensitive.
+
+### Usage resolve priority ###
+
+1. `PUBLIC` and `INTERFACE` usage scopes within a project of the corresponding name.
+2. `usage` blocks with the corresponding name in any scopes.
+
+Note: If there are duplicate usage blocks with the same resolved name, the selected usage block is unspecified. `usage` blocks should have unique names if they are not specified as `PUBLIC`, `PRIVATE`, or `INTERFACE`. If `links` or `linkoptions` are defined within a usage block, it is recommended that `linkgroups` is also turned on.
```
samsinsane, line 14: Duplicate usage blocks: This should error in Premake, or "reopen" the usage block.

`linkgroups`: I don't think I really agree with re-documenting this. Will every API that ends up having edge cases be re-documented here? `uses` doesn't create any issues with `links`, if anything, it resolves them by having projects state their dependencies which can alleviate the ordering issues.

samsinsane (changes requested):
Might also be worth having one of those generic docs, similar to [Linking](https://premake.github.io/docs/Linking) or [Filters](https://premake.github.io/docs/Filters). This would give the opportunity to provide better examples of the intended usage without bloating the individual API docs with long examples.

website/docs/uses.md
nickclark2016: Duplicate usage blocks (custom names) will be addressed in a follow on PR.

samsinsane (approved):

PR #2446, opened 2025-03-13
Added Instrumented Profiling Support

nickclark2016:
**What does this PR do?**

Adds an API for enabling instrumented profiling in gcc, clang, cosmocc, and msc, as well as the Visual Studio exporter.

No breaking changes. Only a functionality add.

N/A

### DIFF
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -386,6 +386,12 @@
 			table.insert(flags, '/NODEFAULTLIB:' .. ignore)
 		end
 
+		if cfg.kind == "ConsoleApp" or cfg.kind == "WindowedApp" or cfg.kind == "SharedLib" then
+			if cfg.profile then
+				table.insert(flags, "/PROFILE")
+			end
+		end
+
 		return flags
 	end
 
--- a/website/docs/profile.md
+++ b/website/docs/profile.md
@@ -1,0 +1,27 @@
+Enable or disable instrumented performance profiling support for binaries.
+
+```lua
+profile "value"
+```
+
+### Parameters ###
+| Value   | Description                                                             |
+-------------------------------------------------------------------------------------
+| On      | Turn on instrumented performance profiling.                             |
+| Off     | Turn off instrumented performance profiling.                            |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0-beta6 or later.
+
+### Examples ###
+
+```lua
+project "MyProject"
+    kind "ConsoleApp"
+    profile "On"
+```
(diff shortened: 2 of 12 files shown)


redorav (commented):
So it's both a property in the property sheet and a manually set flag? Sounds a bit of an odd setup but I guess that's the way it works, looks good to me

nickclark2016:
> So it's both a property in the property sheet and a manually set flag? Sounds a bit of an odd setup but I guess that's the way it works, looks good to me

Can you elaborate? This is something that we can set via the MSBuild API and if we are using msc from a non Visual Studio exporter.

nickclark2016:
The "msc.lua" is completely separate from the VS exporter. They both pipe data to the linker.

redorav:
Ah I see, I misunderstand many parts of Premake still

src/tools/msc.lua
```diff
@@ -386,6 +386,13 @@
 			table.insert(flags, '/NODEFAULTLIB:' .. ignore)
 		end
 
+		-- If the type is ConsoleApp, WindowedApp, or SharedLib and PROFILE is set, add the appropriate flag.
```
Jarod42, line 389: I find that kind of comment useless, it just repeats the below code (which is clear).

Jarod42 (approved):
Would it be `-pg` flags for gcc/clang to be usable with gprof?

src/tools/msc.lua
nickclark2016: Yeah, I wrote the comment for myself before writing all the code. Will remove.

nickclark2016:
> Would it be `-pg` flags for gcc/clang to be usable with gprof?

I completely forgot about this flag. I'll need to investigate where all this would need to go.

website/docs/profile.md
```diff
@@ -0,0 +1,28 @@
...
+### Parameters ###
+| Value   | Description                                                             |
+-------------------------------------------------------------------------------------
+| Default | Use the toolset's default instrumentated performance profiling setting. |
+| On      | Turn on instrumented performance profiling.                             |
+| Off     | Turn off instrumented performance profiling.                            |
+
+### Applies To ###
+
+Project configurations where the `kind` is an executable or a shared library.
```
Jarod42, line 16: What happens to static libraries (used by those "executables")?
I assume that profiling instrumentation should be added (at least for gcc/clang), no?

Jarod42 (approved):

website/docs/profile.md
```diff
@@ -0,0 +1,28 @@
+Enable or disable instrumented performance profiling support for binaries.
+
+```lua
+profile "value"
+```
+
+### Parameters ###
+| Value   | Description                                                             |
+-------------------------------------------------------------------------------------
+| Default | Use the toolset's default instrumentated performance profiling setting. |
```
samsinsane, line 10: Does the boolean kind support default? I thought it was a hack we achieved via the string kind.

samsinsane (approved):

samsinsane (approved):

PR #2458, opened 2025-04-09
fix optimize('Debug') for xcode project generator

artamonovoleg:
**What does this PR do?**

This pr fixes debug optimization level for xcode generator

`-O1` changed to `-O0` when `Debug` is used as the parameter for `optimize`

### DIFF
--- a/modules/xcode/xcode_common.lua
+++ b/modules/xcode/xcode_common.lua
@@ -1490,7 +1490,7 @@
 			settings['GCC_ENABLE_OBJC_EXCEPTIONS'] = 'NO'
 		end
 
-		local optimizeMap = { On = 3, Size = 's', Speed = 3, Full = 'fast', Debug = 1 }
+		local optimizeMap = { On = 3, Size = 's', Speed = 3, Full = 'fast', Debug = 'g' }
 		settings['GCC_OPTIMIZATION_LEVEL'] = optimizeMap[cfg.optimize] or 0
 
 		if cfg.pchheader and not cfg.flags.NoPCH then
--- a/modules/xcode/tests/test_xcode_project.lua
+++ b/modules/xcode/tests/test_xcode_project.lua
@@ -2266,6 +2266,29 @@
 		]]
 	end
 
+	function suite.XCBuildConfigurationProject_OnOptimizeDebug()
+		optimize "Debug"
+		prepare()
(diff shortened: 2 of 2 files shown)


modules/xcode/xcode_common.lua
```diff
@@ -1490,7 +1490,7 @@
 			settings['GCC_ENABLE_OBJC_EXCEPTIONS'] = 'NO'
 		end
 
-		local optimizeMap = { On = 3, Size = 's', Speed = 3, Full = 'fast', Debug = 1 }
```
nickclark2016, line 1493: Should this be `g` instead?

artamonovoleg: minimum optimization via `GCC_OPTIMIZATION_LEVEL` is `-O0`. Probably because dSYMs are standard, and embedding is something unusual in Xcode

nickclark2016: Sure, but why 1 instead of g? `-Og` is typically considered debug. Does Xcode do something different by default for debug?

artamonovoleg: Sorry, I'm not quite following you. I replaced -O1 with -O0, which is the lowest available option for the Xcode parameter. I can pass -Og, but not for this parameter, and I'm not sure if it's the right thing to do since Xcode will pass its own setting anyway

nickclark2016: Sorry, mistyped 1 for 0 here. Yeah, "Off" or "O0" typically denotes no optimizations, whereas "Og" is optimizations compatible with debugging. I would check if that works first, rather than just turning optimizations off.

artamonovoleg: Checked, it's possible to use -g, and it will work as intended. Not sure why it's not listed in the documentation. Though, it feels like something forced—it's not presented in the UI dropdown by default, and it's not listed in the docs. PR updated

Jarod42 (approved):

nickclark2016 (changes requested):
I missed this on my first review. Code itself looks good, but please add a unit test. Here is a good candidate for where to put it: https://github.com/premake/premake-core/blob/master/modules/xcode/tests/test_xcode_project.lua

artamonovoleg:
@nickclark2016  done

nickclark2016 (approved):

PR #2460, opened 2025-04-13
Add basic support for tvOS

sbeca:
**What does this PR do?**

This PR adds basic support for targeting tvOS.

For reference, when you're writing code that targets both iOS and tvOS, there are a bunch of small differences you have to worry about, but from an Xcode configuration point of view, there are only a few differences between iOS and tvOS:
1. Need to target the `appletvos` SDK instead of the `iphoneos` SDK.
2. Need to use `TVOS_DEPLOYMENT_TARGET` to set the min OS version supported, instead of `IPHONEOS_DEPLOYMENT_TARGET`.
3. There is no tvOS equivalent to `iosfamily`, so that can just be ignored.

So the amount of changes required for Premake to support tvOS is relatively small, with a large percent of the additions in this PR being duplications of iOS unit tests with minor changes.

A new target OS has been added.

To show how similar iOS and tvOS are from a builds perspective, I was literally using this small snippet of bash code to smash the small amount of changes needed from an iOS build, and that got things building successfully for tvOS:
```bash

function smash_tvos_changes {
  # HACK: Premake doesn't currently support tvOS as a separate target OS, so we passed iOS to Premake earlier in this script, and now we're going to smash
  # a small amount of changes needed to change the Xcode project from targeting iOS to targeting tvOS, that can't be overridden with Premake config

  # Need to convert lines like:
  #   "CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "Apple Development";
  #   SDKROOT = iphoneos;
  # to:
  #   "CODE_SIGN_IDENTITY[sdk=appletvos*]" = "Apple Development";
  #   SDKROOT = appletvos;
  sed -i '' 's/iphoneos/appletvos/' "$1"

  # Need to convert lines like:
  #   IPHONEOS_DEPLOYMENT_TARGET = 12.0;
  # to:
  #   TVOS_DEPLOYMENT_TARGET = 12.0;
  sed -i '' 's/IPHONEOS_DEPLOYMENT_TARGET/TVOS_DEPLOYMENT_TARGET/' "$1"
}

if [ "${TARGET_OS}" == "tvos" ]; then
  smash_tvos_changes "${SOURCE_DIR}/Generated/${PROJECT}.xcodeproj/project.pbxproj"
fi
```

I would love to get tvOS support upstreamed, so I don't need to do ugly hacks like this anymore haha

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -834,6 +834,7 @@
 		["linux"]      = { "linux",   "posix", "desktop" },
 		["macosx"]     = { "macosx",  "darwin", "posix", "desktop" },
 		["solaris"]    = { "solaris", "posix", "desktop" },
+		["tvos"]       = { "tvos",    "darwin", "posix", "mobile" },
 		["uwp"]        = { "uwp", "windows", "desktop" },
 		["windows"]    = { "windows", "win32", "desktop" },
 	}
--- a/website/docs/systemversion.md
+++ b/website/docs/systemversion.md
@@ -10,7 +10,7 @@ systemversion ("value")
 
 Ranges are currently only supported by the Windows targets with the Visual Studio actions.
 
-Otherwise, only a minimum version can be set for macOS/iOS targets with `xcode` and `gmake`-based actions.
+Otherwise, only a minimum version can be set for macOS/iOS/tvOS targets with `xcode` and `gmake`-based actions.
 
 ### Applies To ###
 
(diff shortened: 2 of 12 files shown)


src/base/os.lua
```diff
@@ -834,6 +834,7 @@
 		["linux"]      = { "linux",   "posix", "desktop" },
 		["macosx"]     = { "macosx",  "darwin", "posix", "desktop" },
 		["solaris"]    = { "solaris", "posix", "desktop" },
+		["tvos"]       = { "tvos",    "darwin", "posix", "mobile" },
```
sbeca, line 837: This is something I'm not quite sure on, what tags should tvOS have?

An Apple TV isn't a desktop device and it isn't a mobile device, but an Apple TV's internals are very similar to an iPhone, and iOS and tvOS share a lot of common history and functionality. So because of the familiarity to iOS I tagged it as `mobile` but I'm open to other suggestions

website/docs/systemversion.md
```diff
@@ -44,15 +44,21 @@ filter "system:macosx"
 
 ### Apple Targets ###
 
-Under macOS this sets the minimum version of the operating system required for the app to run and  is equivalent to setting the `-macosx-version-min` (or newer `-macos-version-min`) compiler flag.
+Under macOS this sets the minimum version of the operating system required for the app to run and is equivalent to setting the `-mmacosx-version-min` (or newer `-mmacos-version-min`) compiler flag.
 
-The same is true for iOS, iPadOS, tvOS, and watchOS system targets except it is equivalent to setting the `-miphoneos-version-min` (or newer `ios-version-min`) compiler flag.
+The same is true for iOS, iPadOS, and watchOS system targets except it is equivalent to setting the `-miphoneos-version-min` (or newer `-mios-version-min`) compiler flag.
```
sbeca, line 49: This is a pre-existing issue, and not directly relevant to this PR so I haven't touched it, but I want to say that mentioning `watchOS` in this sentence is not super correct in my opinion.

First, Premake doesn't directly support watchOS, so mentioning it at all without also saying Premake doesn't support it seems wrong.

Second, watchOS doesn't even use this compiler flag as Clang has a separate `-mwatchos-version-min` compiler flag just for watchOS, unless Clang has some weird internal logic where it falls back to using the iOS flag when the watchOS flag isn't specified

nickclark2016 (approved):

PR #2462, opened 2025-04-19
Add toolsetpath API for specifying tool executable paths

tritao:
Introduces a new API function `toolsetpath(toolsetName, toolName, toolPath)` that allows users to specify custom paths for tool executables within their Premake scripts.

This provides a mechanism to override the default tool lookup behavior for specific toolsets and tools, offering greater flexibility in build configurations.

Changes include:
- Registering the `toolsetpaths` field in `_premake_init.lua` to store the custom paths.
- Implementing the `toolsetpath` function in `base/api.lua` to process and store the provided paths.
- Modifying the `gettoolname` functions in the GCC, Clang, MSVC, SNC, .NET, and Emscripten toolset modules to check for and use paths defined via `toolsetpaths` before falling back to default lookup logic.
- Adding tests in the corresponding toolset test files (`tests/tools/`) to verify that the `toolsetpath` setting correctly overrides the default behavior.
- Creating a new documentation page (`website/docs/toolsetpath.md`) detailing the API function, its parameters, and usage.
- Adding an entry for the new documentation page in the website sidebar (`website/sidebars.js`).

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -350,6 +350,11 @@
 	}
 
 	function clang.gettoolname(cfg, tool)
+		-- Check toolsetpaths first
+		if cfg.toolsetpaths and cfg.toolsetpaths[cfg.toolset] and cfg.toolsetpaths[cfg.toolset][tool] then
+			return cfg.toolsetpaths[cfg.toolset][tool]
+		end
+
 		local toolset, version = p.tools.canonical(cfg.toolset or p.CLANG)
 		local value = clang.tools[tool]
 		if type(value) == "function" then
--- a/src/tools/emcc.lua
+++ b/src/tools/emcc.lua
@@ -22,7 +22,13 @@ emcc.tools = {
 -- flags correctly to emcc builds.
 emcc.shared.profile = nil
 emcc.ldflags.profile = nil
+emcc.getsharedlibarg = function(cfg) return "" end
 
 function emcc.gettoolname(cfg, tool)
+	-- Check toolsetpaths first
+	if cfg.toolsetpaths and cfg.toolsetpaths[cfg.toolset] and cfg.toolsetpaths[cfg.toolset][tool] then
+		return cfg.toolsetpaths[cfg.toolset][tool]
+	end
+
 	return emcc.tools[tool]
 end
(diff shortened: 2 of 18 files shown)


src/tools/clang.lua
```diff
@@ -350,6 +350,11 @@
 	}
 
 	function clang.gettoolname(cfg, tool)
+		-- Check toolsetpaths first
+		if cfg.toolsetpaths and cfg.toolsetpaths[cfg.toolset] and cfg.toolsetpaths[cfg.toolset][tool] then
+			return cfg.toolsetpaths[cfg.toolset][tool]
+		end
```
Jarod42, line 356: Would be fine to have something like:
`function toolset.gettoolname(toolset, config, tool)` to factorize that code.
but probably not possible as it is already used by exporters :-/

src/tools/emcc.lua
```diff
@@ -22,7 +22,13 @@ emcc.tools = {
 -- flags correctly to emcc builds.
 emcc.shared.profile = nil
 emcc.ldflags.profile = nil
+emcc.getsharedlibarg = function(cfg) return "" end
```
Jarod42, line 25: Seems unrelated to that PR.

tests/tools/test_clang.lua
```diff
@@ -48,6 +48,16 @@
 		test.isequal("windres-16", clang.gettoolname(cfg, "rc"))
 	end
 
+--
+-- Verify that toolsetpath overrides the default tool name.
+--
+function suite.toolsetpathOverridesDefault()
+	toolset "clang"
```
Jarod42, line 55: Is it needed?

tests/tools/test_clang.lua
```diff
@@ -48,6 +48,16 @@
...
+	toolsetpath("clang", "cc", "/path/to/my/custom/clang")
+	prepare()
+	test.isequal("/path/to/my/custom/clang", clang.gettoolname(cfg, "cc"))
+end
+
 --
```
Jarod42, line 61: `toolsetpaths` is not tested.
How it behaves if we mix `toolsetpath` and `toolsetpaths`?
At which level the table are merged?

theComputeKid:
Fixes #2754

Issue #2136, opened 2023-09-08
architecture ignored on macOS


bhundt:
See below or discussion here https://github.com/premake/premake-core/discussions/2123. 

### Discussed in https://github.com/premake/premake-core/discussions/2123

<div type='discussions-op-text'>

<sup>Originally posted by **bhundt** August 11, 2023</sup>
I am working on an M1 Mac running 13.3.1. I want to cross-compile to a target architecture of x86_64 and for that I tried the sample premake project and adapted the premake file as follows:

```
workspace "HelloWorld"
   configurations { "Debug", "Release" }
   architecture "x86_64"

project "HelloWorld"
   kind "ConsoleApp"
   language "C++"
   cppdialect "C++20"
   targetdir "bin/%{cfg.buildcfg}"

   files { "**.hpp", "**.h", "**.cpp" }

   filter "configurations:Debug"
      defines { "DEBUG" }
      symbols "On"

   filter "configurations:Release"
      defines { "NDEBUG" }
      optimize "Speed"
```

Unfortunately this creates the default architecture of ```arm64``` which I checked using the `lipo -info` command.
</div>

When I add
```
buildoptions "-arch x86_64"
linkoptions "-arch x86_64"
```
then the `lipo -info` command shows a `x86_64` build is generated.

perseoGI:
Same happens to me!
It seems that premake is only adding `-m64` on `ALL_LDDFLAGS` but it is also needed the `-arch x86_64` on `CFLAGS` and `LDDFLAGS`

mercury233:
Same for me.

https://github.com/mercury233/premake-macos-arch-test/blob/main/premake5.lua

![Image](https://github.com/user-attachments/assets/fadb4166-416b-4ef8-8e7e-2da9fa996289)

![Image](https://github.com/user-attachments/assets/e7b9f265-ef67-4e60-a034-5866608d71f2)

samsinsane:
This should be resolved with #2502.

PR #2468, opened 2025-05-17
[docs] Add warning for `os.hostarch`

mercury233:
**What does this PR do?**

Currently, there are issues with both the `os.hostarch` function and its documentation. The problem with the function is that it doesn't actually retrieve the runtime architecture, and the issue with the documentation is that it fails to point this out.

Nothing.

I think it is better to fix `os.hostarch` to reflect the actual arch.

### DIFF
--- a/website/docs/os/os.hostarch.md
+++ b/website/docs/os/os.hostarch.md
@@ -12,7 +12,9 @@ None.
 
 An architecture identifier; see [architecture()](architecture.md) for a complete list of identifiers.
 
-Note that this function returns the architecture for the OS that Premake is currently running on, which is not necessarily the same as the architecture that Premake is generating files for.
+:::warning
+Currently, this function actually returns the architecture of the system that Premake was compiled for. This means that if you run the Win32 version of Premake on a 64-bit Windows system, this function will return 'x86' instead of 'x86_64'.
+:::
 
 ### Availability ###
 


nickclark2016 (approved):

nickclark2016:
> Possible fix, untested:
> 
> ```c
> /**
>  * \file   os_hostarch.c
>  * \brief  Get the architecture for the current host OS we're executing on.
>  * \author Copyright (c) 2014-2024 Jess Perkins and the Premake project
>  */
> 
> #include "premake.h"
> 
> #if defined(_WIN32)
>   #include <windows.h>
> #elif defined(__unix__) || defined(__APPLE__)
>   #include <sys/utsname.h>
>   #include <string.h>
> #endif
> 
> /**
>  * Detect the actual CPU architecture at runtime.
>  * Falls back to compile-time PLATFORM_ARCHITECTURE on failure.
>  */
> static const char *detect_runtime_arch(void)
> {
> #if defined(_WIN32)
>     // Try the newer IsWow64Process2 (Windows 10+)
>     HMODULE hKernel = GetModuleHandleA("kernel32");
>     if (hKernel) {
>         typedef BOOL (WINAPI *LPFN_IsWow64Process2)(HANDLE, USHORT*, USHORT*);
>         LPFN_IsWow64Process2 fnWow2 =
>             (LPFN_IsWow64Process2)GetProcAddress(hKernel, "IsWow64Process2");
>         if (fnWow2) {
>             USHORT procMachine = 0, nativeMachine = 0;
>             if (fnWow2(GetCurrentProcess(), &procMachine, &nativeMachine)) {
>                 switch (nativeMachine) {
>                     case IMAGE_FILE_MACHINE_AMD64:  return "x86_64";
>                     case IMAGE_FILE_MACHINE_I386:   return "x86";
>                     case IMAGE_FILE_MACHINE_ARM64:  return "ARM64";
>                     case IMAGE_FILE_MACHINE_ARM:    return "ARM";
>                     default:                        return "unknown";
>                 }
>             }
>         }
>     }
>     // Fallback to GetNativeSystemInfo
>     {
>         SYSTEM_INFO si;
>         typedef void (WINAPI *LPFN_GetNativeSystemInfo)(LPSYSTEM_INFO);
>         LPFN_GetNativeSystemInfo fnNSI =
>             (LPFN_GetNativeSystemInfo)GetProcAddress(hKernel, "GetNativeSystemInfo");
>         if (fnNSI)
>             fnNSI(&si);
>         else
>             GetSystemInfo(&si);
> 
>         switch (si.wProcessorArchitecture) {
>             case PROCESSOR_ARCHITECTURE_AMD64:  return "x86_64";
>             case PROCESSOR_ARCHITECTURE_INTEL:  return "x86";
>             case PROCESSOR_ARCHITECTURE_ARM64:  return "ARM64";
>             case PROCESSOR_ARCHITECTURE_ARM:    return "ARM";
>             default:                            return "unknown";
>         }
>     }
> 
> #elif defined(__unix__) || defined(__APPLE__)
>     struct utsname u;
>     if (uname(&u) == 0 && u.machine[0] != '\0') {
>         const char *m = u.machine;
>         // Normalize uname values to our PLATFORM_ARCHITECTURE strings
>         if (strcmp(m, "x86_64") == 0 || strcmp(m, "amd64") == 0)
>             return "x86_64";
>         else if (strcmp(m, "i386") == 0 || strcmp(m, "i486") == 0 || strcmp(m, "i586") == 0 || strcmp(m, "i686") == 0)
>             return "x86";
>         else if (strcmp(m, "aarch64") == 0 || strcmp(m, "arm64") == 0)
>             return "ARM64";
>         else if (strncmp(m, "arm", 3) == 0)
>             return "ARM";
>         else if (strcmp(m, "riscv64") == 0)
>             return "RISCV64";
>         else if (strcmp(m, "loongarch64") == 0)
>             return "loongarch64";
>         else if (strcmp(m, "e2k") == 0)
>             return "e2k";
>         // Unknown or less common: return raw uname
>         return m;
>     } else {
>         /* Fallback to compile-time macro */
>         return PLATFORM_ARCHITECTURE;
>     }
> 
> #else
>     /* Other platforms: fallback */
>     return PLATFORM_ARCHITECTURE;
> #endif
> }
> 
> int os_hostarch(lua_State* L)
> {
>     const char *arch = detect_runtime_arch();
>     lua_pushstring(L, arch);
>     return 1;
> }
> ```

Unfortunately, this would break any existing code relying on the existing behavior (for some reason). We need to first deprecate the API before changing its functionality entirely.

nickclark2016:
Doing things based on premake host architecture also feels like an anti-pattern. I agree, this is definitely an edge case. I don't know what we would name and deprecate with. The correct solution may be to just fix it directly, since we are still just in beta and can have the breaking change.

mercury233:
I'm afraid filtering by architecture isn't particularly uncommon. Macs, for example, come with either Apple Silicon or Intel CPUs, corresponding to ARM64 and x86\_64 architectures respectively. Certain x86-specific optimizations, such as SSE, may not be includable in builds targeting the ARM64 architecture, and vice versa.

I understand that in such cases it might be more appropriate to use `os.targetarch`, but that function doesn't work either—though the documentation clearly states this.
https://github.com/mercury233/premake-macos-arch-test/actions/runs/15089710004/job/42416629418#step:5:1

Moreover, the `architecture` directive itself doesn't function fully as expected; see issue #2136.

nickclark2016:
You should likely be filtering using `filter "architecture:arm"`, for example. I'm saying it's probably an antipattern to filter based on the architecture of the computer running premake. It's perfectly reasonable (and expected) to filter based on the target system's architecture (via the filter tag)

mercury233:
> You should likely be filtering using `filter "architecture:arm"`, for example. I'm saying it's probably an antipattern to filter based on the architecture of the computer running premake. It's perfectly reasonable (and expected) to filter based on the target system's architecture (via the filter tag)

https://github.com/mercury233/premake-macos-arch-test/actions/runs/15089901144/job/42417030453#step:7:8

The `filter` only takes effect when `architecture` has been set. As for how to determine the `architecture`, if we want the compiled program to match the architecture of the machine doing the compiling, we need to use `os.hostarch`.

nickclark2016:
Okay, yeah, let's fix the actual issue then rather than updating docs (as @tritao suggested)

nickclark2016:
Just wanted to poke on this. An update to the docs is better than nothing, but if we can fix the root of the problem, that would be best.

mercury233:
> Can you try the suggested code @mercury233?

_I finally had the chance to test your patch. I set up an ARM-based TV box and installed the Armbian Linux system on it._

https://github.com/mercury233/premake-arch-test

This repository contains the program I compiled using your patch. Here are the test results:

https://github.com/mercury233/premake-arch-test/actions/runs/15649386571/job/44091879546

It successfully detected the current architecture on both Linux and Windows, but not on macOS. The original beta7 version, however, is a universal build, so it can correctly detect the architecture on macOS.

PR #2469, opened 2025-05-17
`os.findheader` support relative headerdirs

mercury233:
**What does this PR do?**

Let `os.findheader` can support searching relative paths.
Some libraries, by default, install header files within folders. For example, the header files for libfreetype-dev are located in:
/usr/include/freetype2/freetype/*.h
/usr/include/freetype2/ft2build.h

Previously, if a program included the header with `#include <ft2build.h>`, you had to use:
`os.findheader("freetype2/ft2build.h") .. "freetype2"`
to locate it.

Now you can simply use:
`os.findheader("ft2build.h", "freetype2")`
to achieve the same result.

Nothing.

Nothing.

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -159,7 +159,16 @@
 		elseif type(headerdirs) == "table" then
 			userpaths = headerdirs
 		end
-		paths = table.join(userpaths, paths)
+
+		for _, userpath in ipairs(userpaths) do
+			if path.isabsolute(userpath) then
+				paths = table.join({userpath}, paths)
+			else
+				for _, p in ipairs(paths) do
+					paths = table.join({path.join(p, userpath)}, paths)
+				end
+			end
+		end
 
 		local result = os.pathsearch (headerpath, table.unpack(paths))
 		return result
--- a/website/docs/os/os.findheader.md
+++ b/website/docs/os/os.findheader.md
@@ -6,13 +6,20 @@ p = os.findheader("headerfile" [, additionalpaths])
 
 ### Parameters ###
 
-`headerfile` is a file name or a the end of a file path to locate.
+`headerfile` is a file name of a file path to locate.
 
-`additionalpaths` is a string or a table of one or more additional search path.
+`additionalpaths` is a string or a table of one or more additional search path. Can be absolute or relative paths. If relative, they are relative to all the default search paths.
 
 ### Return Value ###
 
 The path containing the header file, if found. Otherwise, nil.
+
+### Example ###
+
+``` lua
+os.findheader("event.h") -- /usr/include
+os.findheader("ft2build.h", "freetype2") -- /usr/include/freetype2
+```
 
 ### Remarks ###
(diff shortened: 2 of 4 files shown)


website/docs/os/os.findheader.md
```diff
@@ -6,14 +6,21 @@ p = os.findheader("headerfile" [, additionalpaths])
 
 ### Parameters ###
 
-`headerfile` is a file name or a the end of a file path to locate.
+`headerfile` is a file name of a file path to locate.
 
-`additionalpaths` is a string or a table of one or more additional search path.
+`additionalpaths` is a string or a table of one or more additional search path. Can be absolute or relative paths. If relative, they are relative to all the default search paths.
```
nickclark2016, line 11: Suggestion: "The paths may be absolute or relative. If the path is a relative path, it is relative to each of the default search paths."

nickclark2016 (changes requested):

src/base/os.lua
```diff
@@ -159,7 +159,16 @@
 		elseif type(headerdirs) == "table" then
 			userpaths = headerdirs
 		end
-		paths = table.join(userpaths, paths)
+
+		for _, userpath in ipairs(userpaths) do
+			if path.isabsolute(userpath) then
+				paths = table.join({userpath}, paths)
```
Jarod42, line 165: I think "append" should be use instead of "prepend", to keep original order

src/base/os.lua
```diff
@@ -159,7 +159,16 @@
...
+			else
+				for _, p in ipairs(paths) do
+					paths = table.join({path.join(p, userpath)}, paths)
```
Jarod42, line 168: Same here

mercury233: I think the search should prioritize subdirectories. For example, if a user's computer has both `/usr/include/lua.h` and `/usr/include/lua5.1/lua.h`, and the user calls `os.findheader("lua.h", "lua5.1")`, they are more likely looking for the latter.

Jarod42: In addition you iterate and modify `paths`, so `os.findheader('lib.h', {'dir1', dir2})` will result in something like
`{'/usr/dir1/dir2', '/usr/include/dir2', '/usr/dir1', '/usr/include'}`

A `table.transform(userpaths` to transform relative paths into a table, and keep absolute path as-is would be better, then flatten userpaths, and finally `paths = table.join(userpaths, paths)`

Jarod42:
>> How does this PR change Premake's behavior?

> Nothing.

In fact behavior changes

```
os.findheader("header.h", "./submodules/library")
```
would no longer search from local directory.

Maybe creating another API would be cleaner, something like

```
function os.findheaderleaf(name, userpaths)
  local res = os.findheaderleaf(name, userpaths)
  if res then
    path.join(res, path.getdirectory(name))
  else
    return nil
  end
end
```

mercury233:
@nickclark2016 what do you think about creating another API like Jarod42 said?

PR #2471, opened 2025-05-17
Support powerpc

barracuda156:
**What does this PR do?**

This PR adds initial support for PowerPC.

P. S. Could someone explain how to force the build system actually use the patches .lua sources? Currently the build somehow sabotages applied changes, whether I build in a standard way or via boostrap script. I can build a working executable for powerpc with no issues, but it still does not acknowledge ppc among supported archs.
From earlier PRs adding support for `riscv` and `loongarch` it follows that nothing else needs to be changed, and a search across codebase does not point to any other instances of arch definitions. Yet the build ignores whatever is added.

It does not affect currently supported archs.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -172,6 +172,8 @@
 			{ "ARM64", "ARM64" },
 			{ "x86", "x86 (On macOS, same as x86_64.)" },
 			{ "x86_64", "x86_64" },
+			{ "ppc", "PowerPC 32-bit" },
+			{ "ppc64", "PowerPC 64-bit" },
 			{ "Universal", "Universal Binary (macOS only)" },
 			--
 			{ "Win32", "Same as x86" },
@@ -242,6 +244,14 @@
 		filter { "system:macosx", "options:arch=Universal" }
 			buildoptions { "-arch arm64", "-arch x86_64" }
 			linkoptions { "-arch arm64", "-arch x86_64" }
+
+		filter { "system:macosx", "options:arch=ppc" }
+			buildoptions { "-arch ppc" }
+			linkoptions { "-arch ppc" }
+
+		filter { "system:macosx", "options:arch=ppc64" }
+			buildoptions { "-arch ppc64" }
+			linkoptions { "-arch ppc64" }
 
 		filter { "system:windows", "options:arch=ARM" }
 			platforms { "ARM" }
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -30,6 +30,8 @@
 			p.ARM64,
 			p.RISCV64,
 			p.LOONGARCH64,
+			p.PPC,
+			p.PPC64,
 			p.WASM32,
 			p.WASM64,
 			p.E2K
@@ -968,6 +970,7 @@
 			"SSSE3",
 			"SSE4.1",
 			"SSE4.2",
+			"ALTIVEC",
 		}
 	}
 
--- a/src/base/_foundation.lua
+++ b/src/base/_foundation.lua
@@ -65,6 +65,8 @@
 	premake.ARM64       = "ARM64"
 	premake.RISCV64     = "RISCV64"
 	premake.LOONGARCH64 = "loongarch64"
+	premake.PPC         = "ppc"
+	premake.PPC64       = "ppc64"
 	premake.WASM32 = "wasm32"
 	premake.WASM64 = "wasm64"
 	premake.E2K = "e2k"
--- a/src/host/premake.h
+++ b/src/host/premake.h
@@ -75,6 +75,10 @@
 #define PLATFORM_ARCHITECTURE "loongarch64"
 #elif defined(__e2k__)
 #define PLATFORM_ARCHITECTURE "e2k"
+#elif defined(__ppc64__) || defined(__powerpc64__)
+#define PLATFORM_ARCHITECTURE "ppc64"
+#elif defined(__ppc__) || defined(__powerpc__)
+#define PLATFORM_ARCHITECTURE "ppc"
 #elif !defined(RC_INVOKED)
 #error Unknown architecture detected
 #endif
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -100,6 +100,7 @@
 			SSSE3 = "-mssse3",
 			["SSE4.1"] = "-msse4.1",
 			["SSE4.2"] = "-msse4.2",
+			ALTIVEC = "-maltivec",
 		},
 		isaextensions = {
 			MOVBE = "-mmovbe",
--- a/website/docs/architecture.md
+++ b/website/docs/architecture.md
@@ -15,6 +15,8 @@ architecture ("value")
 * `ARM64`
 * `RISCV64`
 * `loongarch64`
+* `ppc`
+* `ppc64`
 * `wasm32`,
 * `wasm64`,
 * `e2k`,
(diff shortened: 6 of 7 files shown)


nickclark2016:
Those changes just allow Premake to accept `riscv` and `loongarch` as architecture types. In order to push those changes throughout the system, the various exporters (i.e. `gmake`) and toolchains (`gcc`, `clang`) would need to have some of their internal maps updated to support `ppc`.

We can either add those changes onto this current PR, or you may want to follow them up in a future PR.

premake5.lua
```diff
@@ -172,6 +172,8 @@
 			{ "ARM64", "ARM64" },
 			{ "x86", "x86 (On macOS, same as x86_64.)" },
 			{ "x86_64", "x86_64" },
+			{ "ppc", "PowerPC 32-bit" },
```
nickclark2016, line 175: This may be better being "ppc32" instead of "ppc", since it makes it sound (in my mind) like 32 bit is the default, when 64 bit is the default on almost all desktop and server class hardware these days.

barracuda156: Like with `arm` and `mips`, convention is just `ppc`. Though I won’t really object if there is a strong preference to do otherwise.

nickclark2016: Good enough reason for me. Let's leave it.

barracuda156:
> Those changes just allow Premake to accept `riscv` and `loongarch` as architecture types. In order to push those changes throughout the system, the various exporters (i.e. `gmake`) and toolchains (`gcc`, `clang`) would need to have some of their internal maps updated to support `ppc`.
> 
> We can either add those changes onto this current PR, or you may want to follow them up in a future PR.

After I build `premake` on a powerpc machine, running `premake --help` does not show ppc among archs, and trying to use it to build something for ppc returns an error of unsupported arch. It looks like the build somehow overrides the changes applied to the sources.

gcc on its own supports ppc/ppc64, of course, though maybe the sources here needs some modification as well.

barracuda156:
> Got it, misunderstood the issue. Just to make sure, when you're running premake, you're using the most recently build binary, correct?

Initially I just added patches into the existing portfile from MacPorts https://github.com/macports/macports-ports/blob/ecf89c2ff86a0b3be80dff99ffc8d09edd83fac5/devel/premake5/Portfile and ran the build. It uses `gmake.macosx` config. After finding out that while I can make the build to work, arch support is not being added, I thought that I need to go via bootstrap procedure, and modified the portfile to invoke instead `make -f ./Bootstrap.mak osx` with setting `PLATFORM = ppc` in the script. However that did not lead to the intended result either.

Maybe I am not doing it right. Is there a canonical procedure to bootstrap premake on a new platform? I.e. I start from only having the toolchain, but no existing premake5. (There is an old premake4 for ppc, but I am not sure it is usable for this purpose.)

barracuda156:
@nickclark2016 Oh wait, I think I got it working. Give me a min to check.

nickclark2016:
> > Got it, misunderstood the issue. Just to make sure, when you're running premake, you're using the most recently build binary, correct?
> 
> Initially I just added patches into the existing portfile from MacPorts https://github.com/macports/macports-ports/blob/ecf89c2ff86a0b3be80dff99ffc8d09edd83fac5/devel/premake5/Portfile and ran the build. It uses `gmake.macosx` config. After finding out that while I can make the build to work, arch support is not being added, I thought that I need to go via bootstrap procedure, and modified the portfile to invoke instead `make -f ./Bootstrap.mak osx` with setting `PLATFORM = ppc` in the script. However that did not lead to the intended result either.
> 
> Maybe I am not doing it right. Is there a canonical procedure to bootstrap premake on a new platform? I.e. I start from only having the toolchain, but no existing premake5. (There is an old premake4 for ppc, but I am not sure it is usable for this purpose.)

Please verify your build outside of the macports configuration. I am not familiar enough with that ecosystem to help you out on that specific piece of your pipeline. I'm not sure who manages the MacPorts configuration.

barracuda156:
@nickclark2016 @tritao Ok, I can confirm that I have not only building, but actually working `premake5` on powerpc now. There is just one port in MacPorts which depends on premake5, and it was restricted to x86/arm, and now I can build it successfully.

premake

For the arch support what is in this PR is sufficient. (I did not include universal for powerpc, just not to overload the code; if desired, it can be added as `universal_ppc`, but it can work only with Xcode gcc, since clang is broken for ppc, while modern gcc needs driverdriver to be implemented, and until that cannot accept multiple arch flags.)

To have premake5 working on macOS ppc specifically, there are three more issues to address:
1. Securetransport does not build on < 10.7, and curl should use alternative tls. Mbedtls works; since Unix makefile uses that, I used it as well. MacPorts curl by default uses openssl, but that is much heavier to build, of course.
2. Darwin < 16 does not have clock_gettime and friends. MacPorts (or pkgsrc) have `legacy-support` library with implementation of that, otherwise some fallback will be needed.
3. `mbedtls` needs `-std=c99` or `-std=gnu99` – this is trivially fixable.

This results in a working premake: https://github.com/macos-powerpc/powerpc-ports/commit/13970564900173c020d1db9f03605a9397988eaa

I am not sure macOS-specific fixes are wanted, but PowerPC is worth supporting in any case, since both *BSD and Linux support it in their modern releases.

premake5.lua
```diff
@@ -243,6 +245,14 @@
 			buildoptions { "-arch arm64", "-arch x86_64" }
 			linkoptions { "-arch arm64", "-arch x86_64" }
 
+		filter { "system:macosx", "options:arch=ppc or arch=ppc" }
```
Jarod42, line 248:
```suggestion
		filter { "system:macosx", "options:arch=ppc" }
```

premake5.lua
```diff
@@ -243,6 +245,14 @@
...
+			buildoptions { "-arch ppc" }
+			linkoptions { "-arch ppc" }
+
+		filter { "system:macosx", "options:arch=ppc64 or arch=ppc64" }
```
Jarod42, line 252:
```suggestion
		filter { "system:macosx", "options:arch=ppc64" }
```

barracuda156: @Jarod42 Are these equivalent? I just copied the logic here from x86/arm cases, where both are used.

barracuda156: Ah, okay, I misunderstood what it does. Will fix this now.

barracuda156:
@Jarod42 Updated and rebated to master.

nickclark2016 (approved):

nickclark2016:
Given you appear to have access to more legacy Apple hardware than I have, would you be willing to tackle issues 1 and 2 that you brought up? (TLS and clock issues)

barracuda156:
> Given you appear to have access to more legacy Apple hardware than I have, would you be willing to tackle issues 1 and 2 that you brought up? (TLS and clock issues)

I have a patch to use `mbedtls` (like it is done for Unix otherwise) here: https://github.com/macos-powerpc/powerpc-ports/commit/13970564900173c020d1db9f03605a9397988eaa
But I do not know how to do it conditionally on macOS version (and what is a supposed threshold), and I guess there is no desire to switch _all_ macOS to `mbedtls`?

For `clock_gettime` the easiest solution is just using `legacy-support` (which does not _require_ MacPorts, since it has no dependencies, and it is used, I believe, by NetBSD folks with `pkgsrc` too).

Jarod42 (approved):

nickclark2016:
We have arguments in our build already for "system" vs vendored dependencies. We could probably just add an argument for explicit opt in to the `Bootstrap.sh`.

PR #2504, opened 2025-09-11
Added support for Visual Studio 2026

nickclark2016:
**What does this PR do?**

Adds support for VS 2026. This includes the new solution format, which is XML based. See schema here: https://github.com/microsoft/vs-solutionpersistence/blob/main/src/Microsoft.VisualStudio.SolutionPersistence/Serializer/Xml/Slnx.xsd

This also adds support for the new C++ toolchain, v145.

No changes to existing code.

No


Jarod42 (approved):

modules/vstudio/tests/sln2026/test_projects.lua
```diff
@@ -0,0 +1,155 @@
...
+		uuid "AE61726D-187C-E440-BD07-2556188A6565"
+		filter "configurations:SpecialRelease"
+			runtime "Release"
+		filter {}
+
+	sln2026.projects(wks)
+
+	test.capture [[
+<Project Path="MyProject1.vcxproj" Id="AE61726D-187C-E440-BD07-2556188A6565">
+	<BuildType Solution="SpecialRelease|Win32" Project="Release" />
```
samsinsane: Is this correct? The schema you linked doesn't say what this is for, but I would interpret it as being for mapping a configuration in the solution to a configuration in the project. I would have expected this to be controlled with `configmap` rather than `runtime`.

If this is correct, then we probably need to figure out how `configmap` comes into play because it doesn't look like we support it in this format?

samsinsane (approved):

modules/vstudio/tests/sln2026/test_projects.lua
samsinsane: I installed VS2026 preview and tested this, it does look like this should be controlled via `configmap` rather than `runtime`.

The remapping has some edge cases that might be worth supporting too:
- If you have two configurations and you remap one to the other (for all platforms), the output is simplified down to `<BuildType Project="Config" />`.
- If you have three (or more) configurations and you remap one to the other (for all platforms), the output is simplified down to `<BuildType Solution="Config|*" Project="Other" />`.
- The same thing happens with remapping platforms, but simplifies with `<Platform Project="x64" />` and `<Platform Solution="*|Win32" Project="x64" />`.

samsinsane (approved):

nickclark2016:
> I've opened an issue on microsoft's side cause adding normal custom platforms through the configuration manager doesn't work (while opening a regular old .sln with those same custom platforms works fine), so let's see what they reply https://developercommunity.visualstudio.com/t/Solutions-with-custom-Platform-names-no/10972388
> 
> EDIT: Okay, it's apparently "by design" now that custom platform names aren't supported

Just saw your edit. Really appreciate you opening that up on the VS side. I'm not really sure if there's something useful we can do from the Premake end to make it work, unfortunately.

codecat:
Would it make sense for Premake to do what the VS' sln conversion is doing? FWIW, [the premake configuration](https://premake.github.io/docs/Configurations-and-Platforms#platforms) mentions that platforms can be anything, but that currently doesn't hold true with the vs2026 generator.

nickclark2016:
We've had several changes go into the generator since our last release build. Are you testing against the release or a binary from master?

codecat:
Ah, I was using the beta8 build. This is indeed fixed in master. :+1:

PR #2505, opened 2025-09-15
Implements C++23 for Visual Studio non-makefile projects

nickclark2016:
**What does this PR do?**

Adds C++23 for non-makefile VS projects.

For non-makefile projects, uses the `stdcpp23` value in the vcxproj files. For makefile projects, this causes `/std:c++latest` to be used.

The value `/std:c++23preview` does exist, but it is a temporary flag that will go away with the full release of `/std:c++23`. I have opted to use `/std:c++latest`, as that flag will be around for the foreseeable future. Once `/std:c++23` is in full release, another PR will be needed to use it in `msc.lua` and in the makefile project configuration of `vs2010_vcxproj.lua`.


samsinsane (approved):

Jarod42:
How about https://github.com/premake/premake-core/blob/master/src/tools/msc.lua#L211 ?

PR #2509, opened 2025-09-17
Added ability for toolsets to specify tool output extensions

samsinsane:
**What does this PR do?**

Adds ability for toolsets to specify tool output extensions.

The gmake generator will now correctly use `.obj` when using the msc toolset.

N/A

### DIFF
--- a/modules/gmake/gmake_cpp.lua
+++ b/modules/gmake/gmake_cpp.lua
@@ -58,25 +58,33 @@
 	function cpp.initialize()
 		rule 'cpp'
 			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
-			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
+			buildoutputs  { "$(OBJDIR)/%{file.objname}.%{premake.modules.gmake.cpp.gettooloutputext('cxx', cfg)}" }
 			buildmessage  '$(notdir $<)'
 			buildcommands {'$(CXX) %{premake.modules.gmake.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
 
 		rule 'cc'
 			fileExtension {".c", ".s", ".m"}
-			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
+			buildoutputs  { "$(OBJDIR)/%{file.objname}.%{premake.modules.gmake.cpp.gettooloutputext('cc', cfg)}" }
 			buildmessage  '$(notdir $<)'
 			buildcommands {'$(CC) %{premake.modules.gmake.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
 
 		rule 'resource'
 			fileExtension ".rc"
-			buildoutputs  { "$(OBJDIR)/%{file.objname}.res" }
+			buildoutputs  { "$(OBJDIR)/%{file.objname}.%{premake.modules.gmake.cpp.gettooloutputext('rc', cfg)}" }
 			buildmessage  '$(notdir $<)'
 			buildcommands {'$(RESCOMP) $< -O coff -o "$@" $(ALL_RESFLAGS)'}
 
 		global(nil)
 	end
 
+	function cpp.gettooloutputext(tool, cfg)
+		local toolset = gmake.getToolSet(cfg)
+		if toolset.gettooloutputext ~= nil then
+			return toolset.gettooloutputext(tool)
+		end
+
+		return iif(tool == "rc", "res", "o")
+	end
 
 	function cpp.createRuleTable(prj)
 		local rules = {}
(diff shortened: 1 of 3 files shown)


Jarod42 (approved):

modules/gmake/gmake_cpp.lua
```diff
@@ -58,25 +58,33 @@
...
 			fileExtension ".rc"
-			buildoutputs  { "$(OBJDIR)/%{file.objname}.res" }
+			buildoutputs  { "$(OBJDIR)/%{file.objname}.%{premake.modules.gmake.cpp.gettooloutputext('rc', cfg)}" }
 			buildmessage  '$(notdir $<)'
 			buildcommands {'$(RESCOMP) $< -O coff -o "$@" $(ALL_RESFLAGS)'}
 
 		global(nil)
 	end
 
+	function cpp.gettooloutputext(tool, cfg)
```
Jarod42, line 80: Why gcc/clang don't provide it too?

Or move that function as helper method (usable by other toolset (as ninja))

src/tools/msc.lua
```diff
@@ -431,6 +431,14 @@
 		}
 	end
 
+--
+-- Return tool output extension
+--
+
+	function msc.gettooloutputext(tool)
+		return iif(tool == "rc", "res", "obj")
```
Jarod42: [`path.getextension`](https://premake.github.io/docs/path/path.getextension/) (as several programming languages (C++, python)) returns the dot too, seems more coherent to return `'.res'`/`'.obj'`.

modules/gmake/gmake_cpp.lua
samsinsane: > Why gcc/clang don't provide it too?

I couldn't decide if I should or not, but I'm leaning towards I should now.

This function needs to exist though with the fallback, otherwise all modules that introduce C and C++ toolsets won't use extensions on their files.

src/tools/msc.lua
samsinsane: Yeah, I considered this but I opted for this as the the usage would look weird otherwise - `%{}%{}` instead of `%{}.%{}`.

I don't really feel that strongly either way, and it does look like we have precedent to include the dot. @nickclark2016 thoughts?

nickclark2016: Technically speaking, the extension does not include the dot. I would honestly prefer we "break" existing code to fix that, but that's probably not viable. Let's just aim for consistency, so include the ".".

modules/gmake/gmake_cpp.lua
samsinsane: Fixed.

src/tools/msc.lua
samsinsane: Fixed.

modules/gmake/tests/test_gmake_file_rules.lua
```diff
@@ -81,6 +81,46 @@
...
+		toolset "msc"
+		files { "src/hello.cpp", "src/test.c" }
+		prepare()
+		test.capture [[
+# File Rules
+# #############################################
+
+$(OBJDIR)/hello.obj: src/hello.cpp
+	@echo "$(notdir $<)"
+	$(SILENT) $(CXX) $(ALL_CXXFLAGS) $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"
```
Jarod42, line 115: Note:
`%.o=%.d` should be `%.obj=%.d` but anyway `-MF` should not be part there.
Would be fixed in another PR normally, so OK.

modules/gmake/gmake_cpp.lua
```diff
@@ -58,25 +58,33 @@
 	function cpp.initialize()
 		rule 'cpp'
 			fileExtension { ".cc", ".cpp", ".cxx", ".mm" }
-			buildoutputs  { "$(OBJDIR)/%{file.objname}.o" }
+			buildoutputs  { "$(OBJDIR)/%{file.objname}%{premake.modules.gmake.cpp.gettooloutputext('cxx', cfg)}" }
 			buildmessage  '$(notdir $<)'
 			buildcommands {'$(CXX) %{premake.modules.gmake.cpp.fileFlags(cfg, file)} $(FORCE_INCLUDE) -o "$@" -MF "$(@:%.o=%.d)" -c "$<"'}
```
Jarod42, line 63: Shouldn't that `.o` also be changed?

modules/gmake/tests/test_gmake_file_rules.lua
samsinsane: Neither should be part of the MSC output, `-MF <path>` is the full flag. As you said, this is work for another PR, marking this as resolved.

modules/gmake/gmake_cpp.lua
samsinsane: ???

Jarod42: `"$(@:%.o=%.d)"` hardcode `.o`, should it be `"$(@:%%%{premake.modules.gmake.cpp.gettooloutputext('cxx', cfg)}=%.d)"`?

samsinsane: Not the line you've commented on here.

nickclark2016 (approved):

PR #2511, opened 2025-09-17
Updated gmake and gmakelegacy generators so that GCC dependency flags come from the toolset

samsinsane:
**What does this PR do?**

Updated gmake and gmakelegacy generators so that GCC dependency flags come from the toolset.

gmake and gmakelegacy will no longer emit the GCC dependency flag. GCC and Clang emit it. This may cause issues for modules that add new C and C++ toolsets and don't return the GCC/Clang cppflags.

N/A.

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -33,6 +33,10 @@
 
 	function gcc.getcppflags(cfg)
 		local flags = config.mapFlags(cfg, gcc.cppflags)
+
+		-- Request that dependencies be written out
+		table.insert(flags, '-MF "$(@:%.o=%.d)"')
+
 		return flags
 	end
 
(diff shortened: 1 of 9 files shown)


src/tools/gcc.lua
```diff
@@ -33,6 +33,10 @@
 
 	function gcc.getcppflags(cfg)
 		local flags = config.mapFlags(cfg, gcc.cppflags)
+
+		-- Request that dependencies be written out
+		table.insert(flags, '-MF "$(@:%.o=%.d)"')
```
Jarod42, line 38: Isn't `$(@:%.o=%.d)` Makefile specific? :-(

In premake-ninja, we hard code the rule with `-MF $out.d`

samsinsane: Oops, overlooked that. Thanks for pointing that out, going to close this, need to come up with a different approach.

PR #2518, opened 2025-10-01
Skip CI builds when making changes to docs

thomashope:
**What does this PR do?**

This change should make CI not trigger when PRs only make changes to files inside the website folder.

Triggering CI for website changes can cause unnecessary delay due to waiting for jobs to complete or jobs being canceled due to no runners being available.

This should not affect the behaviour of Premake itself. Only CI.

I'm not 100% confident in this change since I don't use Github Actions all that often.

It could also be the case that there is some reason to trigger the builds on website changes that I'm not aware of?

The motivation for this change came from submitting this [PR](https://github.com/premake/premake-core/pull/2517), however it looks like that PR failed some CI jobs for reasons that aren't clear to me?

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -1,6 +1,12 @@
 name: CI Workflow
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - 'website/**'
+  pull_request:
+    paths-ignore:
+      - 'website/**'
 
 jobs:
   linux:


nickclark2016:
I don't think this is a useful change. If someone deletes an existing doc for an API that still exists in the Lua code, this change would make it so that change isn't detected and our CI job never attempts to fail that change.

.github/workflows/ci-workflow.yml
```diff
@@ -1,6 +1,12 @@
 name: CI Workflow
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - 'website/**'
+  pull_request:
+    paths-ignore:
+      - 'website/**'
```
Jarod42, line 9: Can also include *.txt, *.md

thomashope: You mean txt and md files in the entire repo?

Jarod42: Yes, modifying text file doesn't impact build

nickclark2016:
> > I don't think this is a useful change. If someone deletes an existing doc for an API that still exists in the Lua code, this change would make it so that change isn't detected and our CI job never attempts to fail that change.
> 
> Maybe split the CI, one for code check, one for doc-check?

I don't really see the use for that. We still have to build premake for the doc check. By that point, we've built it, tests are relatively cheap. If we just say "hey, let's use XYZ runner to do doc-check on", we're just doing extra work. If we're trying to get around the fact that "but maybe X is degraded, so we want to use Y", that only works if "Y" is never degraded.

PR #2549, opened 2025-12-01
add "Fast" option to LinkTimeOptimization API which enables LTCG:incremental for vs2010, and behaves as "On" for other toolsets

withmorten:
add "Fast" option to LinkTimeOptimization API which enables LTCG:incremental for vs2010, and behaves as "On" for other toolsets

**What does this PR do?**

^ see commit description

This adds a "Fast" option to the LinkTimeOptimiation for vs2010 projects.

### DIFF
--- a/modules/gmakelegacy/tests/cpp/test_clang.lua
+++ b/modules/gmakelegacy/tests/cpp/test_clang.lua
@@ -61,3 +61,20 @@ ifeq ($(config),debug)
 		]]
 	end
 
+	function suite.usesCorrectCompilersAndFastLinkTimeOptimizationViaAPI()
+		linktimeoptimization "Fast"
+		make.cppConfigs(prj)
+		test.capture [[
+ifeq ($(config),debug)
+  ifeq ($(origin CC), default)
+    CC = clang
+  endif
+  ifeq ($(origin CXX), default)
+    CXX = clang++
+  endif
+  ifeq ($(origin AR), default)
+    AR = llvm-ar
+  endif
+		]]
+	end
+
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -71,6 +71,7 @@
 		},
 		linktimeoptimization = {
 			On = "-flto",
+			Fast = "-flto",
 		},
 		strictaliasing = {
 			Off = "-fno-strict-aliasing",
--- a/tests/tools/test_clang.lua
+++ b/tests/tools/test_clang.lua
@@ -39,12 +39,50 @@
 		test.isequal("windres", clang.gettoolname(cfg, "rc"))
 	end
 
+	function suite.tools_onLinkTimeOptimizationViaAPI()
+		linktimeoptimization "On"
+		prepare()
+		test.isequal("clang", clang.gettoolname(cfg, "cc"))
+		test.isequal("clang++", clang.gettoolname(cfg, "cxx"))
+		test.isequal("llvm-ar", clang.gettoolname(cfg, "ar"))
+		test.isequal("windres", clang.gettoolname(cfg, "rc"))
+	end
+
+	function suite.tools_onFastLinkTimeOptimizationViaAPI()
+		linktimeoptimization "Fast"
+		prepare()
+		test.isequal("clang", clang.gettoolname(cfg, "cc"))
+		test.isequal("clang++", clang.gettoolname(cfg, "cxx"))
+		test.isequal("llvm-ar", clang.gettoolname(cfg, "ar"))
+		test.isequal("windres", clang.gettoolname(cfg, "rc"))
+	end
+
 	function suite.tools_forVersion()
 		toolset "clang-16"
 		prepare()
 		test.isequal("clang-16", clang.gettoolname(cfg, "cc"))
 		test.isequal("clang++-16", clang.gettoolname(cfg, "cxx"))
 		test.isequal("ar-16", clang.gettoolname(cfg, "ar"))
+		test.isequal("windres-16", clang.gettoolname(cfg, "rc"))
+	end
+
+	function suite.tools_forVersion_onLinkTimeOptimizationViaAPI()
+		toolset "clang-16"
+		linktimeoptimization "On"
+		prepare()
+		test.isequal("clang-16", clang.gettoolname(cfg, "cc"))
+		test.isequal("clang++-16", clang.gettoolname(cfg, "cxx"))
+		test.isequal("llvm-ar-16", clang.gettoolname(cfg, "ar"))
+		test.isequal("windres-16", clang.gettoolname(cfg, "rc"))
+	end
+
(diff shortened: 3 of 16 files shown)


modules/gmakelegacy/tests/cpp/test_clang.lua
```diff
@@ -78,3 +78,20 @@ ifeq ($(config),debug)
...
+    CC = clang
+  endif
+  ifeq ($(origin CXX), default)
+    CXX = clang++
+  endif
+  ifeq ($(origin AR), default)
+    AR = llvm-ar
+  endif
+		]]
+	end
```
Jarod42, line 79: I would say that it is `clang.gettoolname` which should be tested instead

website/docs/linktimeoptimization.md
```diff
@@ -20,4 +21,6 @@ Project configurations
 
 ### Availability ###
 
-Premake 5.0-beta4 and later
\ No newline at end of file
+Premake 5.0-beta4 and later
+
+Fast only applies to Visual Studio 2010 and will behave as On for other toolsets.
```
Jarod42: and `'Fast'` is available from beta-8

modules/gmakelegacy/tests/cpp/test_clang.lua
withmorten: I just copied this from the `suite.usesCorrectCompilersAndLinkTimeOptimizationViaAPI()` test above, not sure how I would check for clang.gettoolname.

website/docs/linktimeoptimization.md
withmorten: done

modules/gmakelegacy/tests/cpp/test_clang.lua
Jarod42: I meant in test_clang.lua, near `function suite.tools_onDefaults()`, something like:

```
linktimeoptimization "Fast"
prepare()
test.isequal("ar", clang.gettoolname(cfg, "llvm-ar"))
```

withmorten: ~~Oh you don't mean the gmakelegacy one. Well I didn't see any tests there regarding linktimeoptimization, so I didn't add one there. So tests for clang lto seem to be generally missing?~~

I see, the gmakelegacy tests are used for clang.

Jarod42: You change behavior of `clang.gettoolname`, so it is that one which should be tested.
that function is already tested for regular case.

withmorten: It's tested by the test above, which I duplicated and changed the test from "On" to "Fast":

https://github.com/premake/premake-core/blob/master/modules/gmakelegacy/tests/cpp/test_clang.lua#L64

If I modify the behaviour of `clang.tools` to always return `ar` for ar here:

https://github.com/premake/premake-core/blob/master/src/tools/clang.lua#L370

Only the tests `suite.usesCorrectCompilersAndLinkTimeOptimizationViaFlag` and `suite.usesCorrectCompilersAndLinkTimeOptimizationViaAPI` fail.

I added the most similar test for "Fast".

nickclark2016: I agree with Jarod here. There should be a test in the clang tests. It's likely just an oversight that "On" changing the tool to "llvm-ar" wasn't tested.

src/tools/gcc.lua
```diff
@@ -71,6 +71,7 @@
 		},
 		linktimeoptimization = {
 			On = "-flto",
+			Fast = "-flto",
```
nickclark2016, line 74: GCC and Clang differ here. "Fast" should be `-flto -fwhole-program` for GCC, and Clang should be `-flto=thin`. GCC and Clang can no longer share flags for `linktimeoptimization`.

withmorten: I see. I saw the clang thin option, but I don't have the necessary overview/experience with how clang is used - and the shared gcc base - so I didn't want to add it as part of my PR. At least for now nothing is broken and LTO still works for gcc and clang ...

I don't have that much experience with gcc either to know whether `-fwhole-program` really is equivelant to `-flto=thin` or `/LTCG:incremental`.

I can try adding the missing tests though.

withmorten: Okay maybe this isn't as complicated as I thought. I would still not add `-fwhole-program` for gcc though.

nickclark2016: Okay yeah, that's fine. There's probably a better approach in GCC, looking at more recent releases for LTO. That can be done separately.

tests/tools/test_clang.lua
```diff
@@ -39,6 +39,33 @@
 		test.isequal("windres", clang.gettoolname(cfg, "rc"))
 	end
 
+	function suite.tools_onLinkTimeOptimizationViaFlag()
```
nickclark2016: The LTO flag was actually removed. The "via flag" test is no longer required. (Same comment on Clang, GCC, and MSC).

website/docs/linktimeoptimization.md
```diff
@@ -11,13 +11,16 @@ linktimeoptimization "value"
...
 
 Project configurations
 
 ### Availability ###
 
-Premake 5.0-beta4 and later
\ No newline at end of file
+Premake 5.0-beta4 and later
+
+Fast is available from Premake 5.0-beta8 or later. Applies to Visual Studio 2010 and clang, will behave as On for other toolsets.
```
nickclark2016: Given it's slightly more complex than just "Exists starting here", leave the "Premake 5.0-beta4 and later", then add a column to the table called Notes (see flags.md)

nickclark2016 (changes requested):
Minor changes, but overall happy with the code. Suggest merging in master, since there's a decent amount of changes that landed in the last 48 hours.

tests/tools/test_clang.lua
withmorten: Ah - I see. it would have been nice to keep that as deprecated option so I don't need to use different versions of premake for git bisecting old project files, but oh well :D

nickclark2016: All flags are being deprecated and removed ahead of the 5.0 full release. We'll be rolling out a new release soon with them all deprecated in favor of a dedicated API for each.

website/docs/linktimeoptimization.md
withmorten: Done - or get rid of the "available from Premake 5.0-beta8 or later" ?

withmorten:
> Minor changes, but overall happy with the code. Suggest merging in master, since there's a decent amount of changes that landed in the last 48 hours.

Done.

I think I just accidentally started a rewiew and deleted the pending comment, oh well. What a mess this interface is.

nickclark2016 (approved):

Issue #2551, opened 2025-12-02
fix: stream line the docs format for parsing and generating lua meta files


lolrobbe2:
_Before opening a new bug report, please read [Reporting Bugs](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#reporting-bugs) and consider if this is something [you can contribute](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#contributing-a-fix-or-feature) yourself. If this a new bug, help us help you by filling in the template below._

**What seems to be the problem?**
curently the documentation format is not fully standerdised, wich makes parsing the docs to generate lua meta files hard, and in some cases impossible.

**What have you tried so far?**
trying to work arround this by adding the different formats

the hard parts currently are, reliably aquiring the parameter type and parameter description in certain cases.
example: [vsprops](https://github.com/premake/premake-core/blob/master/website/docs/vsprops.md)

preferably the type of the parameter is part of the description:

## example:

* `path` **string** - is the relative path to be converted. It does not need to actually exist on the file system.
*  `file_list` **string[]** -  specifies one or more file patterns.

@KyrietS, @nickclark2016, any suggestion for a standardised format? 

*(You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to issues like these!)*

lolrobbe2:
To provide proper type hints i need to be able to properly parse the documentation files. 
For example: to provide code completion for the architecture types i need to be able to get the available options to generate type aliases etc.

Wich cannot be easely generated from the source files

lolrobbe2:
@samsinsane so i was able to succesfully generate meta files for the fields.

but how would i go about generating these for workspace, filter etc as these are not registered in fields etc...

(for now i have done these by hand)

PR #2552, opened 2025-12-02
fix: updated docs styling

lolrobbe2:
**What does this PR do?**
this pr aims to rectify #2551 

### params non determenistic:
for params wich are not deterministic IE have standard values like architecture the following style is used:

```MD
`<param_name>` **param_type** - <description>.
```

### params deterministic:
for params wich are deterministic a MD table is used:

for architecture:
arch is the enum/param name
```md

| Arch           | Description                                      |
|-------------|--------------------------------------------------|
| universal    | Universal binaries supported by iOS and macOS    |
| x86            | 32-bit x86 architecture                          |
| x86_64       | 64-bit x86 architecture                          |
| ARM          | 32-bit ARM architecture                          |
| ARM64       | 64-bit ARM architecture                          |
| RISCV64     | 64-bit RISC-V architecture                       |
```

### DIFF
--- a/website/docs/architecture.md
+++ b/website/docs/architecture.md
@@ -12,34 +12,36 @@ architecture ("arch")
 
 ### Parameters ###
 
-| Arch        | Description                                      |
-|-------------|--------------------------------------------------|
-| universal   | Universal binaries supported by iOS and macOS    |
-| x86         | 32-bit x86 architecture                          |
-| x86_64      | 64-bit x86 architecture                          |
-| ARM         | 32-bit ARM architecture                          |
-| ARM64       | 64-bit ARM architecture                          |
-| RISCV64     | 64-bit RISC-V architecture                       |
-| loongarch64 | 64-bit LoongArch architecture                    |
-| ppc         | 32-bit PowerPC architecture                      |
-| ppc64       | 64-bit PowerPC architecture                      |
-| wasm32      | 32-bit WebAssembly target                        |
-| wasm64      | 64-bit WebAssembly target                        |
-| e2k         | Elbrus 2000 architecture                        |
-| mips64el    | 64-bit MIPS little-endian architecture           |
-| armv5       | ARMv5 (only supported in VSAndroid projects)     |
-| armv7       | ARMv7 (only supported in VSAndroid projects)     |
-| aarch64     | AArch64 (only supported in VSAndroid projects)   |
-| mips        | MIPS (only supported in VSAndroid projects)      |
+| Arch        | Description                                        |
+| ----------- | -------------------------------------------------- |
+| universal   | Universal binaries supported by iOS and macOS      |
+| x86         | 32-bit x86 architecture                            |
+| x86_64      | 64-bit x86 architecture                            |
+| ARM         | 32-bit ARM architecture                            |
+| ARM64       | 64-bit ARM architecture                            |
+| RISCV64     | 64-bit RISC-V architecture                         |
+| loongarch64 | 64-bit LoongArch architecture                      |
+| ppc         | 32-bit PowerPC architecture                        |
+| ppc64       | 64-bit PowerPC architecture                        |
+| wasm32      | 32-bit WebAssembly target                          |
+| wasm64      | 64-bit WebAssembly target                          |
+| e2k         | Elbrus 2000 architecture                           |
+| mips64el    | 64-bit MIPS little-endian architecture             |
+| armv5       | ARMv5 (only supported in VSAndroid projects)       |
+| armv7       | ARMv7 (only supported in VSAndroid projects)       |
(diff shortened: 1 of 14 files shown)


website/docs/architecture.md
```diff
@@ -1,38 +1,47 @@
...
+title: architecture
+description: Specifies the system architecture to be targeted by the configuration.
+keywords: [premake, architecture, system, config, x86, x86_64, arm, arm64, riscv64, wasm]
+---
+
 Specifies the system architecture to be targeted by the configuration.
 
 ```lua
-architecture ("value")
+architecture ("arch")
```
Jarod42, line 10: Both `"value"`/`"arch"` are wrong there (not valid value). `value`/`arch` would be good (variable might have a correct value)

website/docs/basedir.md
```diff
@@ -1,14 +1,20 @@
+---
+title: basedir
+description: Sets the base directory for a configuration
+keywords: [premake, basedir, base directory, path, configuration]
```
Jarod42: Seems you use `config` in other places

website/docs/buildcustomizations.md
```diff
@@ -1,12 +1,18 @@
+---
+title: buildcustomizations
+description: Imports custom .props files for Visual Studio.
+keywords: [premake, buildcustomizations, visual studio, props, project]
```
Jarod42: 'project' versus 'project config'?

website/docs/architecture.md
lolrobbe2: the idea is to be able to extract the name and get the following: 
```lua
function architecture(arch) end
```

Jarod42: It is the quotes that bore me (which shouldn't avoid the name extraction).

Some APIs accept mostly any strings, some quote are OKish, some expect specific string, and providing unexpected ones is confusing IMO.
`architecture ("value")` was already wrong, and should have been `architecture (value)` (or `architecture ("%{value}")`).
And then you should only have renamed 'value' to 'arch'.

website/docs/callingconvention.md
```diff
@@ -1,17 +1,25 @@
+---
+title: callingconvention
+description: Sets whether or not the compiler should build STL modules.
```
KyrietS, line 3: I'm against adding `title` and `description` front-matters when they are not necessary. If not set, the title of the page is its filename and it's fine for 99% of our pages. Same goes for `description`. Docusaurus takes first paragraph as a description so there is no need to repeat it in the front-matter. It's bug-prone and this PR proves it. This description is copied from another page :)

website/docs/callingconvention.md
```diff
@@ -1,17 +1,25 @@
...
+keywords: [premake, callingconvention, cdecl, fastcall, stdcall, vectorcall, function calling convention, compiler, project config]
```
KyrietS, line 4: What is your algorithm for generating keywords for every page? Do they help you in any way? Because I am pretty much sure they will not affect searchability of our docs at all.

lolrobbe2: The doccheck script i am working has flagged this and enforces this.

Making the docs uniform will also make readabilty better

nickclark2016:
I'm not particularly a fan of having multiple things doing the same (or similar) things. Can we not do this in the existing Lua task?

> > if you have a script to check documentation, it should probably be called from https://github.com/premake/premake-core/blob/master/.github/workflows/website.yml to ensure consistency. It just has to return non-zero in case of failure from CI point of view, but messages help humans :-)
> 
> yes that is the goal, it is just a js script so can be run in the website task.
> 
> and logs precise errors when things are wrong

lolrobbe2:
> I'm not particularly a fan of having multiple things doing the same (or similar) things. Can we not do this in the existing Lua task?
> 
> > > if you have a script to check documentation, it should probably be called from https://github.com/premake/premake-core/blob/master/.github/workflows/website.yml to ensure consistency. It just has to return non-zero in case of failure from CI point of view, but messages help humans :-)
> > 
> > yes that is the goal, it is just a js script so can be run in the website task.
> > 
> > and logs precise errors when things are wrong
> 
> 

That script just checks that it exists at the moment

website/docs/callingconvention.md
KyrietS: By "uniform" you mean adding (a possibly, in my opinion) redundant information to each page and having keywords just for the sake of having them? Because what "compiler" keyword even is? This information is transparent to the reader and it does not affect readability. While I'm happy to uniform the way parameter lists are formatted, I cannot convince myself to having front-matter everywhere just because some script enforces it. Why would it be flagged and enforced in the first place? Why do you think we need this?

samsinsane:
I've commented on and closed the referenced issue, I'm going to close this PR too.

PR #2554, opened 2025-12-07
Ninja exporter for premake core

nickclark2016:
**What does this PR do?**

Implements a Ninja exporter in the core repository

No changes to exist behavior. Adds `ninja` action.

No additional context needed.


modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+
+function m.rules(prj)
+	local rulesDone = {}
+	
+	for cfg in project.eachconfig(prj) do
+		local toolset = ninja.gettoolset(cfg)
+		local toolsetKey = tostring(toolset)
+		
+		if not rulesDone[toolsetKey] then
+			m.ccrules(cfg, toolset)
```
Jarod42: You seems to allows several toolsets at the same time, but rule's name doesn't depend of toolset.
so you will have several times the `rule cc` (for clang*/gcc*/mvc*) which seems problematic

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+			local arname = toolset.gettoolname(cfg, "ar")
+			_p("rule ar")
+			_p("  command = %s -rcs $out $in", arname)
+			_p("  description = Archiving static library $out")
+			_p("")
+		else
+			local ldname = toolset.gettoolname(cfg, iif(cfg.language == "C", "cc", "cxx"))
+			local commands = string.format("command = %s -o $out $in $links $ldflags", ldname);
+			
+			commands = commands:gsub("^%s*(.-)%s*$", "%1")
```
Jarod42: `^%s*` seems also empty, as it begins with command, just need to trim last space.
Or providing a `string.trim` utility might be fine.

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+		_p("  command = %s -x %s $cflags -o $out -MD -c $in", pchname, headerType)
+		_p("  description = Generating precompiled header $in")
+		_p("  depfile = $out.d")
+	end
+	_p("")
+end
+
+function m.copyrules(cfg, toolset)
+	_p("rule copy")
+	_p("  command = cp $in $out")
```
Jarod42: should it be
```
_p(os.translateCommands("  command = {COPYFILE} $in $out"))
```
to handle 'cmd' shell?

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+	end
+	
+	for _, ruleName in ipairs(prj.rules) do
+		local rule = p.global.getRule(ruleName)
+		if rule then
+			local ruleNameEscaped = ruleName:gsub("[^%w_]", "_"):lower()
+			_p("rule %s", ruleNameEscaped)
+			_p("  command = $%s_command", ruleNameEscaped)
+			_p("  description = $%s_description", ruleNameEscaped)
+			_p("")
```
Jarod42, line 220: Not sure of the usefulness of several rules completely custom.

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+		for _, depname in ipairs(cfg.dependson) do
+			local depprj = p.workspace.findproject(cfg.workspace, depname)
+			if depprj then
+				local depcfg = project.getconfig(depprj, cfg.buildcfg, cfg.platform)
+				if depcfg then
+					local depTarget = path.getrelative(cfg.workspace.location, depcfg.buildtarget.directory) .. "/" .. depcfg.buildtarget.name
+					table.insert(depTargets, depTarget)
+				end
+			end
+		end
```
Jarod42: Can be in a sub-function as exact copy-paste of other places

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+	local ldflags = m.getLdFlags(cfg, toolset)
+	if #ldflags > 0 then
+		_p("ldflags_%s = %s", ninja.key(cfg), table.concat(ldflags, " "))
+	end
+	
+	local links = toolset.getlinks(cfg)
+	if #links > 0 then
+		local wksLinks = {}
+		for _, link in ipairs(links) do
+			if link:match("^%.%.") or (not link:match("^[/-]")) then
```
Jarod42: Not sure to understand that condition.
second condition seems to check for non-absolute (unix) path, is `path.isabsolute` wrong/not enough?
comment might help if not appropriate function exist.

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+end
+
+function m.copyrules(cfg, toolset)
+	_p("rule copy")
+	_p("  command = cp $in $out")
+	_p("  description = Copying file $in to $out")
+	_p("")
+end
+
+function m.prebuildcommandsrule(cfg, toolset)
```
Jarod42: Above rules use plural.
(I think i would prefer singular everywhere)

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+function m.getCFlags(cfg, toolset)
+	local flags = {}
+	toolset = toolset or ninja.gettoolset(cfg)
+	
+	local toolFlags = toolset.getcflags(cfg)
+	flags = table.join(flags, toolFlags)
+	
+	local escaper = p.escaper(p.quote)
+	local defines = toolset.getdefines(cfg.defines)
+	p.escaper(escaper)
```
Jarod42: Why just for `defines`?

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+	
+	local alreadyGenerated = false
+	if outputTracking then
+		for _, output in ipairs(outputs) do
+			if outputTracking[output] then
+				alreadyGenerated = true
+				break
+			end
+		end
+	end
```
Jarod42: It seems there a several duplication with above methods

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+local function touchCommand(file)
+	local shell = os.shell()
+	if shell == "posix" then
+		return "touch \"" .. file .. "\""
+	elseif shell == "cmd" then
+		return "type nul > \"" .. file .. "\""
+	else
+		return ""
+	end
+end
```
Jarod42, line 1189: `return '{TOUCH} "' .. file .. '"'`?

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+
+local function buildCommandString(cmds, message, touchFile)
+	local shell = os.shell()
+
+	local allcmds = table.deepcopy(cmds)
+	if message then
+		table.insert(allcmds, 1, 'echo \"' .. message .. '\"')
```
Jarod42: `{ECHO}` (even if translate into `echo` for all)

Jarod42: premake-ninja has
```
function ninja.quote(s)
  s = s:gsub('\\', `\\\\`)
  s = s:gsub("'", "\\'")
  s = s:gsub('"', `\\"`)
  return s
end
```
which would have sense for premake IMO
and would handle uniformly the extra quote at the different places.

modules/ninja/ninja_cpp.lua
```diff
@@ -0,0 +1,1389 @@
...
+		local commands = os.translateCommandsAndPaths(cfg.prebuildcommands, cfg.project.basedir, cfg.project.location)
+		local cmdstr = buildCommandString(commands, nil, prebuildTarget)
+
+		_p("build %s: prebuild%s", prebuildTarget, implicitDeps)
+		_p("  prebuildcommands = %s", cmdstr)
+	else
+		local commands = os.translateCommandsAndPaths(cfg.prebuildcommands, cfg.project.basedir, cfg.project.location)
+		local cmdstr = buildCommandString(commands, cfg.prebuildmessage, prebuildTarget)
+		_p("build %s: prebuild%s", prebuildTarget, implicitDeps)
+		_p("  prebuildcommands = %s", cmdstr)
```
Jarod42, line 1239: It seems that this case supersede above case; there are just `commands` to compute separately.

PR #2570, opened 2025-12-19
Add GCC support to VS on Linux

nickclark2016:
**What does this PR do?**

Adds official support for GCC on Linux Visual Studio (2017, 2019, 2022)
Adds support for functions defining the supported valid tools for actions.

No breaking changes.

None

### DIFF
--- a/modules/vstudio/vs2017.lua
+++ b/modules/vstudio/vs2017.lua
@@ -28,7 +28,13 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None", "Utility", "SharedItems", p.PACKAGING },
 		valid_languages = { "C", "C++", "C#", "F#" },
 		valid_tools     = {
-			cc     = { "msc"   },
+			cc     = function(prj)
+						if prj.system == p.LINUX then
+							return { "clang", "gcc" }
+						else
+							return { "msc" }
+						end
+					end,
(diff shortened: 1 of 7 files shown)


Jarod42 (approved):

modules/vstudio/vs2017.lua
```diff
@@ -28,7 +28,13 @@
 		valid_kinds     = { "ConsoleApp", "WindowedApp", "StaticLib", "SharedLib", "Makefile", "None", "Utility", "SharedItems", p.PACKAGING },
 		valid_languages = { "C", "C++", "C#", "F#" },
 		valid_tools     = {
-			cc     = { "msc"   },
+			cc     = function(prj)
+						local toolsets = { "msc" }
+						if prj.system == p.LINUX then
+							table.insert(toolsets, "gcc")
```
samsinsane: Why is `msc` a valid compiler for Linux? This should also be adding `clang`.

nickclark2016: In VS2017, we didn't have "Clang" there before, so I'll defer to @redorav on if that's correct. Also, does msc support cross compilation on VS for Linux, or is it only gcc and clang?

samsinsane: VS2017 shouldn't have Clang normally, but it should have Clang for Linux. MSC doesn't do cross compilation for non-Microsoft platforms, VS only supports GCC and Clang for Linux.

PR #2591, opened 2026-01-05
Unify Architectures, Fix Aliases

nickclark2016:
**What does this PR do?**

1. Uses AARCH64 for all 64 bit ARM architectures
2. Makes ARM64 an alias of AARCH64
3. Deprecates i386 and x32
  * `i386` - This should strictly be the original x86 architecture and mean `-m32 -march=i386` instead of an alias to `-m32`. This can be reintroduced as its own architecture in the future after deprecation period.
  * `x32` - This should be `-mx32` for the x32 ABI. Currently, this maps to `-m32` which is the x86 ABI, and thus is incorrect.

No breaking changes to core. Third party libraries relying on `ARM64` rather than `AARCH64` will break.

Preparation for 5.0

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -169,7 +169,7 @@
 		description = "Set the architecture of the binary to be built.",
 		allowed = {
 			{ "ARM", "ARM (On macOS, same as ARM64.)" },
-			{ "ARM64", "ARM64" },
+			{ "AARCH64", "AARCH64" },
 			{ "x86", "x86 (On macOS, same as x86_64.)" },
 			{ "x86_64", "x86_64" },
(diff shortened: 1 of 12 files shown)


samsinsane (approved):

premake5.lua
```diff
@@ -258,7 +259,7 @@
 		filter { "system:windows", "options:arch=ARM" }
 			platforms { "ARM" }
 
-		filter { "system:windows", "options:arch=ARM64" }
+		filter { "system:windows", "options:arch=AARCH64" }
```
Jarod42, line 262: Shouldn't those filters check both ARM64 or AARCH64?(as there are the "same").

PR #2117, opened 2023-08-09
Require unistd.h for macosx in libzip

nickclark2016:
**What does this PR do?**

Resolves implicit symbol in libzip and closes #2092 

**How does this PR change Premake's behavior?**

N/A

**Anything else we should know?**

N/A

**Did you check all the boxes?**

- [x] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [x] Add unit tests showing fix or feature works; all tests pass
- [x] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] Minimize the number of commits
- [x] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

*You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to requests like these!*

PR #2593, opened 2026-01-05
Add missing <unistd.h> header for macOS builds

ryanwisemanmusic:
This PR adds Mac-specific code to address a missing header that affects projects depending on premake-core (for example: Xenia). The header is added to two files: zipint.h and mkstemp.c. The fix specifically targets the __APPLE__ platform, addressing this Apple-exclusive issue.

This change only adds the necessary header; it does not alter premake’s functionality. Only these two files were updated, adding about six lines of code.

I verified this code compiles with:
./bootstrap.sh

### DIFF
--- a/contrib/libzip/zipint.h
+++ b/contrib/libzip/zipint.h
@@ -46,6 +46,10 @@
 #include <io.h>
 #endif
 
+#ifdef __APPLE__
+#include <unistd.h>
+#endif
+
 #ifndef _ZIP_COMPILING_DEPRECATED
 #define ZIP_DISABLE_DEPRECATED
 #endif


ryanwisemanmusic:
> Looking at the Xenia repo, it's not using the official Premake repository but a fork that hasn't been updated in 5 years. Given that we've already resolved this issue with #2117 back in August 2023, I'm going to close this PR.

I'll just write a .patch file for Xenia at this point using git diff since it seems that the codebase relies on very specific commits that. I apologize for not seeing the incorporation of unistd.h in the Lua file, thanks for your time and patience

Issue #2596, opened 2026-01-08
Deterministic order in embedded scripts


Dariqq:
**What problem will this solve?**

I regenerated the `src/scripts.c` from the release zip file for 5.0-beta7 on linux and was puzzled why the generated `src/scripts.c` is different.

This seems to be caused by a different order in the embedded scripts 

For me the order in the scripts corresponds to this order in the found manifests

```
src/_manifest.lua
modules/xcode/_manifest.lua
modules/vstudio/_manifest.lua
modules/self-test/_manifest.lua
modules/codelite/_manifest.lua
modules/raw/_manifest.lua
modules/gmakelegacy/_manifest.lua
modules/gmake/_manifest.lua
```

The release instead seems to have

```
src/_manifest.lua
modules/self-test/_manifest.lua
modules/gmake/_manifest.lua
modules/vstudio/_manifest.lua
modules/raw/_manifest.lua
modules/xcode/_manifest.lua
modules/codelite/_manifest.lua
modules/gmakelegacy/_manifest.lua
``` 

I tried to regenerate this on 2 other linux machines and got a different order each time so the order seems completely random to me.

**What might be a solution?**
Ensure the files are sorted deterministically, i.e. alphabetical order.

**What other alternatives have you already considered?**

Figure out why the order is different in the first place.
For dirent.h readdir the order seems to be filesystem dependant and no order is guaranteed.

**Anything else we should know?**

zip file order
```
"src/base/_foundation.lua"
"src/base/string.lua"
"src/base/table.lua"
"src/base/path.lua"
"src/base/os.lua"
"src/base/io.lua"
"src/base/tools.lua"
"src/base/tree.lua"
"src/base/globals.lua"
"src/base/semver.lua"
"src/base/http.lua"
"src/base/json.lua"
"src/base/jsonwrapper.lua"
"src/base/languages.lua"
"src/base/term.lua"
"src/base/field.lua"
"src/base/criteria.lua"
"src/base/detoken.lua"
"src/base/configset.lua"
"src/base/context.lua"
"src/base/container.lua"
"src/base/option.lua"
"src/base/action.lua"
"src/base/api.lua"
"src/base/global.lua"
"src/base/workspace.lua"
"src/base/group.lua"
"src/base/project.lua"
"src/base/config.lua"
"src/base/fileconfig.lua"
"src/base/rule.lua"
"src/base/usage.lua"
"src/base/oven.lua"
"src/base/validation.lua"
"src/base/premake.lua"
[...]

PR #2597, opened 2026-01-08
Ensure deterministic embed order.

Dariqq:
Sort table of embedded scripts by their name.

**What does this PR do?**

Ensures the order in the generated table is deterministic by sorting the files according to their name.
Fixes https://github.com/premake/premake-core/issues/2596

No

No.

### DIFF
--- a/scripts/embed.lua
+++ b/scripts/embed.lua
@@ -155,6 +155,10 @@
 
 	addScript(contentTable, "_modules.lua", "src/_modules.lua", "return {" .. table.implode(modules, '"', '"', ', ') .. "}")
 
+-- Sort scripts to ensure deterministic order
+
+	table.sort(contentTable, function(a, b) return a.name < b.name end)
+
 -- Embed the actual script contents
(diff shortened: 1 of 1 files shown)


scripts/embed.lua
```diff
@@ -156,6 +160,7 @@
 	addScript(contentTable, "_modules.lua", "src/_modules.lua", "return {" .. table.implode(modules, '"', '"', ', ') .. "}")
 
 -- Embed the actual script contents
+	table.sort(contentTable, compareScripts)
```
nickclark2016: Don't need a separate function, just use an anonymous function here. Also, leave a comment as to why we are sorting.

nickclark2016 (changes requested):

nickclark2016:
Can you confirm that this produces deterministic builds? I'm on Windows right now, so no matter what, the file hash is different on each build.

Jarod42 (approved):

PR #2602, opened 2026-01-12
[feature] Add `wholearchive`

Jarod42:
**What does this PR do?**

Add `wholearchive` API which correspond to link option
- [`/WHOLEARCHIVE`](https://learn.microsoft.com/en-us/cpp/build/reference/wholearchive-include-all-library-object-files)
- `--whole-archive` of [ld](https://linux.die.net/man/1/ld)
- [`-force_load`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang2-force_load)

Add an API

Tested with https://github.com/Jarod42/premake-sample-projects/tree/wholearchive/projects/wholearchive

### DIFF
--- a/modules/vstudio/tests/vc2010/test_link.lua
+++ b/modules/vstudio/tests/vc2010/test_link.lua
@@ -7,7 +7,6 @@
 	local p = premake
 	local suite = test.declare("vs2010_link")
 	local vc2010 = p.vstudio.vc2010
-	local project = p.project
 
 
 --
@@ -877,6 +876,26 @@
 <Link>
 	<SubSystem>Windows</SubSystem>
 	<AdditionalDependencies>kernel32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+	<ImportLibrary>bin\Debug\MyProject.lib</ImportLibrary>
+</Link>
+		]]
+	end
+
+
+--
+-- Whole archive should be listed in additional dependencies.
+--
+
+	function suite.wholearchive()
+		links { "MyProject2" }
+		wholearchive { "kernel32", "MyProject2" }
+		project "MyProject2"
+		kind "StaticLib"
+		prepare()
+		test.capture [[
+<Link>
+	<SubSystem>Windows</SubSystem>
+	<AdditionalDependencies>/WHOLEARCHIVE:kernel32;/WHOLEARCHIVE:bin\Debug\MyProject2.lib;%(AdditionalDependencies)</AdditionalDependencies>
 	<ImportLibrary>bin\Debug\MyProject.lib</ImportLibrary>
 </Link>
 		]]
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -1026,6 +1026,12 @@
 	}
 
 	api.register {
+		name = "wholearchive",
+		scope = "config",
+		kind = "list:string"
+	}
+
+	api.register {
 		name = "editorintegration",
 		scope = "workspace",
 		kind = "boolean",
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -394,6 +394,10 @@
 		}
 	}
 
+	function msc.wholearchive(cfg)
+		return table.translate(config.getwholearchive(cfg), function(libraryname) return "/WHOLEARCHIVE:" .. libraryname end)
+	end
+
 	function msc.getldflags(cfg)
 		local map = iif(cfg.kind ~= p.STATICLIB, msc.linkerFlags, msc.librarianFlags)
 		local flags = config.mapFlags(cfg, map)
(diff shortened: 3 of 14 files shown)


src/_premake_init.lua
```diff
@@ -1025,6 +1025,16 @@
 		}
 	}
 
+	api.register {
+		name = "wholearchive",
+		scope = "config",
+		kind = "string",
+		allowed = {
+			"Off",
```
nickclark2016: If it's just on/off, it should be a boolean and not a string.

src/tools/msc.lua
```diff
@@ -466,6 +466,11 @@
 			table.insert(links, link)
 		end
 
+		if cfg.wholearchive == p.ON then
+			-- Don't use "/WHOLEARCHIVE" as path.translate might be applied to result
+			links = table.translate(links, function (s) return "-WHOLEARCHIVE:" .. s end)
```
nickclark2016: Is the `path.translate` issue something that you did indeed find to be a problem? Or is this pre-emptive?

website/docs/wholearchive.md
```diff
@@ -0,0 +1,38 @@
+Apply whole archive flag to linked libraries.
```
nickclark2016: Reword suggestion: `Controls whether the linker should include all objects in the linked binary.`

src/_premake_init.lua
nickclark2016: We should probably have "Default" as well, because we should probably allow users to specify to use the default toolset/exporter behavior, rather than enforcing one behavior or another.

src/tools/msc.lua
nickclark2016: Also, this should probably go into the linker flags, rather than the getlinks functionality. Also, I think the proper way to do this would be: `link.exe foo.lib /WHOLEARCHIVE:foo.lib` -- Following up on this, if we don't support granular whole archive, we can just have a single `/WHOLEARCHIVE` without specifying the individual libs.

Jarod42: I encounter it in visual.
Using `-` as workaround is rejected in commandline though (as ninja).
I think we have to specify 'shell' and translate path here.

src/_premake_init.lua
Jarod42: I used `linkgroups` as model :-/

Ideally, I wanted libraries as well, but too lazy to do it, so currently an all or nothing.

Implementation would be more "On" or "Default" BTW :-)

src/tools/msc.lua
Jarod42: I might try that.
>  if we don't support granular

I hope to add that support.

src/_premake_init.lua
Jarod42: Not checked if we can have `boolean` or `string` (for individual library).

website/docs/wholearchive.md
Jarod42: Reworded to "Command linker to include all objects of given libraries."

src/_premake_init.lua
Jarod42: Changed to have list of strings instead of an ON/OFF flags.

src/tools/msc.lua
Jarod42: Changed to have granular support.
I follow the suggestion and use linker flags instead of `getlinks` functionality.

src/_premake_init.lua
Jarod42: Note:
Instead of
```
link { 'my_library' }
wholearchive { 'my_library' }
```
I also think about alternative API (similar to existing `:static`/`:shared` suffix syntax):
```
link { 'my_library:wholearchive' }
```
but it seems team's preferences go for the first syntax.

modules/vstudio/tests/vc2010/test_link.lua
```diff
@@ -883,6 +883,23 @@
 	end
 
 
+--
+-- Whole archive should be listed in additional dependencies.
+--
+
+	function suite.wholearchive()
+		wholearchive { "kernel32" }
```
Jarod42: Test with sibling projects is missing

website/docs/Project-API.md
```diff
@@ -160,6 +160,7 @@
 | [vectorextensions](vectorextensions.md)                   | Enable hardware vector extensions |
 | [vpaths](vpaths.md)                                       |  |
 | [warnings](warnings.md)                                   |  |
+| [wholearchive](wholearchive.md)                           |  |
```
nickclark2016, line 163: Out of scope of this PR, but I feel like I always forget to add things here. We should probably check to make sure all of the APIs are actually here.

Jarod42: Or remove it completely, as it doesn't bring more information that reference listing on the left panel.
I also note that for some pages, including Project-API, we lose the left panel.

PR #2615, opened 2026-02-01
Fixes filters using an aliased value

### LINKED ISSUE #2614: `filter` no longer matches `architecture:arm64`, starting in beta8
`targetsuffix` is no longer respected for ARM64, starting in beta8.
It was working fine for me in beta2 through beta7.
I've been using premake for ARM64 on Windows for years (cross-compiling and publishing generated ARM64 builds).

The `<TargetName>` should include the `targetsuffix`, but for ARM64 it no longer does.
I expected ARM64 to respect the `targetsuffix` as it did in beta2 through beta7.
In beta8 x86 and x64 include the suffix, but ARM64 does not.

1. I bisected which version of premake introduced this problem:  beta2 through beta7 respect `targetsuffix` for ARM64, but beta8 does not -- the regression was introduced in beta8.
2. I'm working on finding a way to work around this beta8 regression.
3. I may also track down what commit introduced the regression, after I find a workaround (or give up on finding a workaround).

Use the following premake5.lua file, run `premake5 vs2022`, and then observe the `<TargetName>` fields in the resulting `repro.vcxproj` file.
```lua
workspace("repro")
    configurations({"debug", "release", "final"})
    platforms({"x86", "x64", "arm64"})

project("repro")
    kind("consoleapp")

    filter {cfg, "architecture:x86"}
        targetsuffix("_x86")

    filter {cfg, "architecture:x64"}
        targetsuffix("_x64")

    filter {cfg, "architecture:arm64"}
        targetsuffix("_arm64")
```

- [x] Visual Studio 2022 (vs2022)
- [ ] I haven't tried any other versions of VS yet, but since VS is not even being invoked it seems unlikely to be related to the VS version.

```console
> premake5 --version
premake5 (Premake Build Script Generator) 5.0.0-beta8
```

nickclark2016:
**What does this PR do?**

Fixes #2614

No

N/A

### DIFF
--- a/modules/vstudio/tests/vc2010/test_target_name.lua
+++ b/modules/vstudio/tests/vc2010/test_target_name.lua
@@ -1,0 +1,128 @@
+--
+-- test_target_name.lua
+-- Test target names.
+-- Author: Nick Clark
+-- Copyright (c) 2026 Jess Perkins and the Premake project
+--
+
+    local p = premake
+    local suite = test.declare("vstudio_vs2010_target_name")
+    local vc2010 = p.vstudio.vc2010
+
+--
+-- Setup
+--
+
+    local wks, prj
+
+    function suite.setup()
+        p.action.set("vs2010")
+        wks, prj = test.createWorkspace()
+    end
+
+    local function prepare(platform)
+        local cfg = test.getconfig(prj, "Debug", platform)
+        vc2010.targetName(cfg)
+    end
+
+
+--
+-- Default target name is the project name
+--
+
+    function suite.defaultTargetName()
+        prepare()
+        test.capture [[
+<TargetName>MyProject</TargetName>
+        ]]
+    end
+
+
+    function suite.targetWithSuffix()
+        targetsuffix "_suffix"
+
+        prepare()
+        test.capture [[
+<TargetName>MyProject_suffix</TargetName>
+        ]]
+    end
+
+
+--
+-- Target name with targetsuffix with AARCH architecture
+--
+
+    function suite.targetWithAarch64FilterSuffix()
+        architecture "AARCH64"
+
+        filter { "architecture:AARCH64" } -- Fails if it is set to ARM64
+            targetsuffix "_ARM64"
+
+        prepare()
+        test.capture [[
+<TargetName>MyProject_ARM64</TargetName>
+        ]]
+    end
+
+
+--
+-- Target name with targetsuffix with ARM64 architecture
+--
+
+    function suite.targetWithArm64FilterSuffix()
+        architecture "ARM64"
+
+        filter { "architecture:ARM64" } -- Should match when architecture is AARCH64
+            targetsuffix "_ARM64"
+
+        prepare()
+        test.capture [[
+<TargetName>MyProject_ARM64</TargetName>
+        ]]
+    end
+
+
+--
+-- Target name with targetsuffix and x64 architecture
+--
+
+    function suite.targetWithX64FilterSuffix()
+        architecture "x64"
+
+        filter { "architecture:x64" }
+            targetsuffix "_x64"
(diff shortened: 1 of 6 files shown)


modules/vstudio/tests/vc2010/test_target_name.lua
```diff
@@ -0,0 +1,128 @@
...
+
+
+--
+-- Target name with targetsuffix with ARM64 architecture
+--
+
+    function suite.targetWithArm64FilterSuffix()
+        architecture "ARM64"
+
+        filter { "architecture:ARM64" } -- Should match when architecture is AARCH64
```
chrisant996, line 75: Would it be worth adding a lowercase test?

chrisant996: Or better yet, a mixed case test, e.g. `aRm64`.

modules/vstudio/vstudio.lua
```diff
@@ -34,6 +34,10 @@
 	vstudio.vs2010_architectures =
 	{
 		win32   = "x86",
+		x86       = "x86",
+		x86_64    = "x64",
+		ARM       = "ARM",
+		AARCH64   = "ARM64",
 	}
```
Jarod42, line 41: It is now identical to `vs200x_architectures`.
It seems they can be regrouped, and so below function might be simplified.

src/base/context.lua
```diff
@@ -100,9 +100,25 @@
...
 		elseif value ~= nil then
 			value = tostring(value):lower()
+
+			local field = p.field.get(key)
+			if field and field.aliases then
+				local aliased = field.aliases[value]
+				if aliased then
+					value = tostring(aliased):lower()
+				end
+			end
```
Jarod42: That pattern is repeated 3 times, maybe create a subfunction, something like:
```lua
function p.field.alias_resolved(key, value) do
  value = value:lower()
  local field = p.field.get(key)
  if field and field.aliases then
    local aliased = field.aliases[value]
	if aliased then
		value = tostring(aliased):lower()
	end
  end 
  return value
end
```

nickclark2016: These aren't quite the same. One is iterating over a table, one is a single value.

modules/vstudio/vstudio.lua
nickclark2016: Keeping is separate in case any plugins are relying on this.

modules/vstudio/tests/vc2010/test_target_name.lua
nickclark2016: We probably should, but that sort of test would be better fit elsewhere in the filter logic, not the VS logic.

src/base/context.lua
chrisant996: The pattern is identical in all three places, and can be factored to call the function Jarod42 suggested.

Here is what the 3 call sites would become:
_(Edit 1: fixed `.` to `:` for calling the function.)_
_(Edit 2: oops, Jarod42 already handled the missing `:lower()` issue in criteria.lua and factored further than I had initially noticed, and I've updated the call sites below accordingly.)_

**context.lua**
```lua
if type(value) == "table" then
	for i = 1, #value do
		value[i] = p.field:alias_resolved(key, value[i])
	end
elseif value ~= nil then
	value = p.field:alias_resolved(key, value)
end
```

**criteria.lua**
```lua
if prefix then
	word[1] = p.field:alias_resolved(prefix, word[1])
end
```

nickclark2016: Ah, misunderstood the comment

chrisant996: LGTM!  (and you're right, my `.` was right the first time, not `:`)

Jarod42 (approved):

PR #2617, opened 2026-02-06
add {COPYFILEIFNEWER} token

alchemyyy:
What does this PR do?**

Adds a `{COPYIFNEWER}` command token that copies a file only if the source is newer than the destination, or the destination doesn't exist.

No breaking changes. Existing behavior is untouched. This just adds a new `{COPYIFNEWER}` token alongside the existing `{COPY}`, `{COPYFILE}`, etc.

Nothing.

- [X] Mention any related issues
- [X] Follow our coding conventions
- [X] Align documentation to your changes

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -597,6 +597,9 @@
 			copyfile = function(v)
 				return "cp -f " .. path.normalize(v)
 			end,
+			copyfileifnewer = function(v)
+				return "cp -u " .. path.normalize(v)
+			end,
 			copydir = function(v)
 				return "cp -rf " .. path.normalize(v)
 			end,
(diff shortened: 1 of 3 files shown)


src/base/os.lua
```diff
@@ -675,6 +678,10 @@
 				-- Just use COPY instead, it actually works.
 				return "copy /B /Y " .. v
 			end,
+			copyfileifnewer = function(v)
+				v = path.translate(path.normalize(v))
+				return "echo F | xcopy /D /Y " .. v
+			end,
```
Jarod42, line 717: From https://stackoverflow.com/questions/14014763/vs-2012-post-build-xcopy-error-2/14022309#14022309
`echo F` hack doesn't work for some localized Windows.
The other hack proposed seems fragile too by adding `*` at the end (which might match other files) :-/

website/docs/Tokens.md
```diff
@@ -103,6 +103,7 @@ The available tokens, and their replacements:
 |------------|---------------------------------------------|-----------------------|
 | {CHDIR}    | chdir {args}                                | cd {args}             |
 | {COPYFILE} | copy /B /Y {args}                           | cp -f {args}          |
+| {COPYFILEIFNEWER} | echo F \| xcopy /D /Y {args}          | cp -u {args}          |
```
Jarod42: Should probably has extra column for "note" to mention version of inclusion of that token

alchemyyy: Done

src/base/os.lua
alchemyyy: Thanks for the link! The ```*``` hack shouldn't match other files since its just parsed as a hint in xcopy and isn't a wildcard matcher or anything. I've updated the code to use this approach.

EDIT: I'll mark this as resolved for now

website/docs/Tokens.md
```diff
@@ -99,19 +99,20 @@ Command tokens are replaced with an appropriate command for the target shell. Fo
...
-| {LINKFILE} | mklink {args}                               | ln -s {reversed args} |
-| {MKDIR}    | IF NOT EXIST {args} (mkdir {args})          | mkdir -p {args}       |
-| {MOVE}     | move /Y {args}                              | mv -f {args}          |
-| {RMDIR}    | rmdir /S /Q {args}                          | rm -rf {args}         |
-| {TOUCH}    | type nul >> {arg} && copy /b {arg}+,, {arg} | touch {args}          |
+| Token      | DOS/cmd                                     | Posix                 | Since |
+|------------|---------------------------------------------|-----------------------|-------|
+| {CHDIR}    | chdir {args}                                | cd {args}             |       |
+| {COPYFILE} | copy /B /Y {args}                           | cp -f {args}          |       |
+| {COPYFILEIFNEWER} | xcopy /D /Y {args}*                   | cp -u {args}          | 5.0-beta9 |
```
nickclark2016: Sorry for the late response. You don't need the "Since" field, since this will just be present in 5.0 stable (next release).

nickclark2016:
Tests fail. Please fix.

website/docs/Tokens.md
alchemyyy: fixed

nickclark2016 (approved):

tests/base/test_os.lua
```diff
@@ -360,6 +360,37 @@
...
+	function suite.translateCommand_windowsCopyFileIfNewer()
+		test.isequal('xcopy /D /Y a b*', os.translateCommands('{COPYFILEIFNEWER} a b', "windows"))
+	end
+
+	function suite.translateCommand_windowsCopyFileIfNewerWithSpaces()
+		test.isequal('xcopy /D /Y "a a" "b b*"', os.translateCommands('{COPYFILEIFNEWER} "a a" "b b"', "windows"))
+	end
+
+	function suite.translateCommand_windowsCopyFileIfNewerWithPaths()
+		test.isequal('xcopy /D /Y src\\dir\\file.txt dst\\dir\\file.txt*', os.translateCommands('{COPYFILEIFNEWER} src/dir/file.txt dst/dir/file.txt', "windows"))
```
Jarod42: paths should be enclosed by `%[..]` i.e `'xcopy /D /Y %[src\\dir\\file.txt dst\\dir\\file.txt*]'`

nickclark2016: Good catch here. I would agree this should be fixed.

PR #2626, opened 2026-02-24
Add dynamic debugging support for MSC (VS, Ninja)

nickclark2016:
**What does this PR do?**

Adds support for Dynamic Debugging (/dynamicdeopt) for Visual Studio and MSC via Ninja.

No breaking changes. Adds new API.

Context: https://devblogs.microsoft.com/cppblog/cpp-dynamic-debugging-full-debuggability-for-optimized-builds/

### DIFF
--- a/tests/tools/test_msc.lua
+++ b/tests/tools/test_msc.lua
@@ -879,3 +879,39 @@ end
 		prepare()
 		test.isequal({ "/MD", "/Z7" }, msc.getcflags(cfg))
 	end
+
+	function suite.cflags_onDynamicDebuggingOn()
+		dynamicdebugging "On"
+		prepare()
+		test.contains("/dynamicdeopt", msc.getcflags(cfg))
+	end
+
+	function suite.cxxflags_onDynamicDebuggingOn()
+		dynamicdebugging "On"
+		prepare()
+		test.contains("/dynamicdeopt", msc.getcxxflags(cfg))
+	end
+
+	function suite.cflags_onDynamicDebuggingOff()
+		dynamicdebugging "Off"
+		prepare()
+		test.excludes("/dynamicdeopt", msc.getcflags(cfg))
+	end
+
+	function suite.cxxflags_onDynamicDebuggingOff()
+		dynamicdebugging "Off"
+		prepare()
+		test.excludes("/dynamicdeopt", msc.getcxxflags(cfg))
+	end
+
+	function suite.linkerflags_onDynamicDebuggingOn()
+		dynamicdebugging "On"
+		prepare()
+		test.contains("/dynamicdeopt", msc.getldflags(cfg))
+	end
+
+	function suite.linkerflags_onDynamicDebuggingOff()
+		dynamicdebugging "Off"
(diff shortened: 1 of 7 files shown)


website/docs/dynamicdebugging.md
```diff
@@ -0,0 +1,33 @@
...
+| On | Allows dynamic debugging of source code. | For MSVC, debug [symbols](symbols.md) must be present. [Edit and Continue](editandcontinue.md) is not allowed. For a full list of flags that are required, see the [Microsoft Docs](https://learn.microsoft.com/en-us/cpp/build/reference/dynamic-deopt?view=msvc-170). |
+| Off | Disallows dynamic debugging of source code. | |
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0.0 or later for Visual Studio 2022 and later or MSC toolsets.
```
Jarod42, line 23: Not beta9?

website/docs/dynamicdebugging.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### Parameters ###
+
+`value` is one of:
+
+| Value | Description | Notes |
+|-------|-------------|-------|
+| Default | Uses the default dynamic debugging behavior. | |
+| On | Allows dynamic debugging of source code. | For MSVC, debug [symbols](symbols.md) must be present. [Edit and Continue](editandcontinue.md) is not allowed. For a full list of flags that are required, see the [Microsoft Docs](https://learn.microsoft.com/en-us/cpp/build/reference/dynamic-deopt?view=msvc-170). |
```
Jarod42, line 14: Does msvc warn for incompatible flags?
Else we should probably emit a warning.

tests/tools/test_msc.lua
```diff
@@ -879,3 +879,39 @@ end
...
+		dynamicdebugging "Off"
+		prepare()
+		test.excludes("/dynamicdeopt", msc.getcflags(cfg))
+	end
+
+	function suite.cxxflags_onDynamicDebuggingOff()
+		dynamicdebugging "Off"
+		prepare()
+		test.excludes("/dynamicdeopt", msc.getcxxflags(cfg))
+	end
```
Jarod42, line 905: No tests for `"Default"`?

website/docs/dynamicdebugging.md
nickclark2016: Nope. Next release is full release.

website/docs/dynamicdebugging.md
nickclark2016: It errors, I believe. That said, this would break paradigm. We don't do that anywhere else.

tests/tools/test_msc.lua
nickclark2016: I'm torn on this, personally. I typically treat default as "do whatever the default for the toolset/exporter is" -- So nothing should be emitted at all. We could test that we don't put `/dynamicdeopt` in there, but we don't typically test the "default" case (though the argument could be made that we should)

PR #2632, opened 2026-02-27
feat: added registries to community section

lolrobbe2:
**What does this PR do?**

This creates a new registries page (did no know where to put these otherwise?)

Any Modules are welcome to be registered!

There is a public rate limited search api available if other projects wan't to integrate the public registry

### DIFF
--- a/website/community/registries.md
+++ b/website/community/registries.md
@@ -2,11 +2,29 @@
 title: Registries
 ---
 
-These Registries are available from other developers; follow the links for more information. If you've created a registry you would like to share with others, feel free to [add a link](https://github.com/premake/premake-core/edit/master/website/community/registries.md) to the list!
+These registries are available from other developers for managing Premake modules and libraries. If you have created a registry you would like to share, feel free to [add a link](https://github.com/premake/premake-core/edit/master/website/community/registries.md) to the list!
 
 ## Premake Manager
 
-The following registries are integrated with **premake manager**
+The following registries are fully integrated with **premake-manager**.
 
-- [Public Registry](https://premake-registry-ywxg.onrender.com/)
-- [Common Registry](https://github.com/lolrobbe2/premake-common-registry)
+### [Common Registry](https://github.com/lolrobbe2/premake-common-registry)
+
+A collection of public and popular libraries designed to integrate seamlessly with **premake**.
+
+* **Dependency Management:** Libraries can declare dependencies, version ranges, and more.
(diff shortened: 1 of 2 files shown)


Jarod42 (commented):
It doesn't explain to users what it is.

Still unclear what it is for me ^_^

lolrobbe2:
@Jarod42
Explanation:
## common registry
[common registry](https://github.com/lolrobbe2/premake-common-registry)
A collection of public/popular libraries wich integrate with premake-manager.
These libraries can declare dependencies and ranges etc... Then a versions will be matched by premake-manager-cli and installed. these registries are created via github repo's.

>[!NOTE]
> People can create their own common registries and add them to their workflow via premake-manager

## public registry
[registry](https://premake-registry-ywxg.onrender.com/)
The public registry is for individual premake users who wan't to share their libraries and modules.
these can be registered via the UI and github login.
They also can integrate with premake-manager-cli but they also can be used standalone as git submodule etc..


might need some help to word this better :)

lolrobbe2:
if there are popular libraries you can think about, that i should add. Let me know.

website/community/registries.md
```diff
@@ -0,0 +1,30 @@
...
+The public registry is a hub for individual Premake users who want to share their libraries and modules with the community.
+
+* **Easy Registration:** Modules can be registered via a web UI using GitHub login.
+* **Flexible Integration:** While they integrate with `premake-manager-cli`, they are also designed to be used standalone (e.g., as Git submodules).
+* **Search API:** Provides a public API dedicated to searching for available modules and libraries, allowing for integration into external tools and scripts.
+
+---
+
+> [!NOTE]
+> Developers are encouraged to create their own common registries to manage internal dependencies within their team's specific Premake workflow.
```
samsinsane, line 30: This is way too much information and reads like this is an official part of Premake - especially with the note at the end. This should follow the same format as [Modules](https://premake.github.io/community/modules) and [Extensions](https://premake.github.io/community/extensions/), where each item is just a single line with just enough information to know what it is.

Something like this:
```md
## Registries

- [premake-common-registry](https://github.com/lolrobbe2/premake-common-registry) : The common registry for [Premake Manager](https://github.com/lolrobbe2/premake-manager).
```

samsinsane (changes requested):

lolrobbe2:
@samsinsane will reopen when all plugins are ready.

PR #2633, opened 2026-02-28
Add support for CYGWIN.

carlo-bramini:
**What does this PR do?**

This PR adds support for CYGWIN.
After these few fixes, premake can be compiled and used for this new platform.

These changes won't affect support for existing plaftorms and targets.

During my debug, I discovered that into `src/base/os.lua` there was a tiny mistake because it was adding `"/usr/lib64/"` instead of `"/usr/lib64"` (note the final `"/"` at the end of the string), so I fixed it. However, this is a minimal and safe fix, as you can see.

CYGWIN follows much more BSD than Linux.
This allows to avoid duplications of code into the makefile.
So, once you get the sources fixed by this PR, you can build premake for CYGWIN by using this command:
```
make -f Bootstrap.mak bsd
```
After that, you will find the executable of premake under the `bin/release` directory.
The new port also passed all tests successfully.

### DIFF
--- a/contrib/curl/lib/config-linux.h
+++ b/contrib/curl/lib/config-linux.h
@@ -185,7 +185,7 @@
 #define HAVE_GETHOSTBYNAME 1
 
 /* Define to 1 if you have the gethostbyname_r function. */
-#if !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__sun__)
+#if !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__sun__) && !defined(__CYGWIN__)
 #define HAVE_GETHOSTBYNAME_R 1
 #endif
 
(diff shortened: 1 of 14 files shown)


nickclark2016:
I am not a fan of changing our vendored libs outside of updating them. Are these libraries buildable with Cygwin itself? If so, let's figure out why we can't build them in Premake (or if we just need to update our versions). If not, they will break again when we inevitably update our libraries.

nickclark2016:
We definitely have made some minor changes, but we try to really keep them to an absolute minimum. We know stuff will break as soon as we try to update the libraries, and it's something I'm worried about. As for how to use system libraries, take a look at the build CI.

nickclark2016:
> Just for information, I contributed a patch to LUA for supporting CYGWIN:
> 
> https://groups.google.com/g/lua-l/c/8-ZBF5pW6cA
> 
> Now let's hope that it will be included soon into the next release.

Do we know if this is being included in a backport for 5.3.x, or will this be a 5.5.x only patch?

PR #2637, opened 2026-03-10
add os.findsubdirheader

mercury233:
**What does this PR do?**

Add an API `os.findsubdirheader` which can support searching relative paths.
Some libraries, by default, install header files within folders. For example, the header files for libfreetype-dev are located in:
> /usr/include/freetype2/freetype/*.h
/usr/include/freetype2/ft2build.h

Previously, if a program included the header with `#include <ft2build.h>`, you had to use:
`os.findheader("freetype2/ft2build.h") .. "freetype2"`
to locate it.

Now you can simply use:
`os.findsubdirheader("ft2build.h", "freetype2")`
to achieve the same result.

New API, no change to the existing APIs.

This PR replace and close https://github.com/premake/premake-core/pull/2469 .

### DIFF
--- a/src/base/os.lua
+++ b/src/base/os.lua
@@ -143,10 +143,22 @@
 		end
 	end
 
+---
+-- Attempt to locate and return the path to a header file.
+--
+-- Searches the same well-known system locations as os.findlib(), but replaces any /lib or /bin
+-- components with /include, and also searches in additional headerdirs provided by the caller.
+--
+-- @param headerpath
+--    A partial header file path.
+-- @param headerdirs
+--    Optional; a string or table of additional paths.
+--    If an input is an absolute path, it will be searched directly.
+--    If an input is a relative path, it will be treated as is, relative to the current working directory.
+-- @return
+--    The full path to the directory containing the headerpath if found; `nil` otherwise.
+---
 	function os.findheader(headerpath, headerdirs)
-		-- headerpath: a partial header file path
-		-- headerdirs: additional header search paths
-
 		local paths = get_library_search_path()
 
 		-- replace all /lib and /bin by /include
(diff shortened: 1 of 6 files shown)


src/base/os.lua
```diff
@@ -165,6 +165,47 @@
 		return result
 	end
 
+	function os.findsubdirheader(headerpath, additionalpaths)
+		-- headerpath: a partial header file path
+		-- additionalpaths: required; a string or table of subdirectory paths
```
Jarod42: Doc style is more often the one from `os.findlib` than `os.findheader`

website/docs/os/os.findsubdirheader.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+### Remarks ###
+Unlike [[os.findheader]], relative paths in `additionalpaths` are resolved against each default search path, allowing discovery of headers in named subdirectories without requiring an absolute path. When `additionalpaths` is non-empty, **only** the specified candidate directories are searched.
+
+`os.findsubdirheader` uses the same base paths as [[os.findheader]], which mostly uses the same paths as [[os.findlib]] but replace `/lib` by `/include`.
+
+### Availability ###
+
+Premake 5.0-beta9 or later.
```
Jarod42: No more beta9 now.

website/docs/os/os.findheader.md
Jarod42, line 30: Possibly add a `See Also` section for `os.findsubdirheader`.

nickclark2016 (changes requested):
Agree with @Jarod42's comments. Address those and I'll be happy to merge.

website/docs/os/os.findsubdirheader.md
mercury233: Which version should it use here?

nickclark2016: 5.0 -- The next release will be 5.0 stable

mercury233: Because Premake’s version numbers look rather conservative, I assumed that versions accepting new features would not become stable releases.

nickclark2016: We have decided to just cut it as a stable release. We won't be breaking any more APIs past 5.0 (hopefully)

Jarod42 (approved):

nickclark2016 (approved):

website/docs/os/os.findsubdirheader.md
mercury233: _So this won't be added in the coming 5.0 stable release?_

nickclark2016: I'll be honestly, I thought I merged these a while back. Merging.

PR #2653, opened 2026-04-15
Improve slnx format output

redorav:
**What does this PR do?**

Improves the configuration mapping for the slnx format

It accounts for the buildcfg and platform combinations so they map correctly to the project files

Some unit tests are failing due to some iterator not being set up correctly, don't think it's a big one but I need someone to review what I'm doing. Once that's fixed I can get to fixing any valid unit test failures and add new ones


modules/vstudio/vs2026_solution.lua
```diff
@@ -156,10 +180,86 @@ function m.projects(wks)
 			p.push('<Folder Name="%s">', groupName)
 		end
 		for _, prj in ipairs(projects) do
+
+			local typeID = vstudio.tool(prj)
+
+			local projectPathStrings = {}
```
Jarod42, line 186: Not sure why we use table for string concatenation (you even provide space)

modules/vstudio/vs2026_solution.lua
```diff
@@ -156,10 +180,86 @@ function m.projects(wks)
...
+				context.solCfg = solcfg
+				context.prjCfg = project.getconfig(prj, solcfg.buildcfg, solcfg.platform)
+				context.excluded = context.prjCfg == nil or context.prjCfg.excludefrombuild
+
+				if context.prjCfg == nil then
+					context.prjCfg = project.findClosestMatch(prj, solcfg.buildcfg, solcfg.platform)
+				end
+
+				context.arch = vstudio.archFromConfig(context.prjCfg, true)
+				context.buildType = context.solCfg.buildcfg.." "..context.prjCfg.platform
```
Jarod42, line 224:
```suggestion
				context.buildType = context.solCfg.buildcfg .. " " .. context.prjCfg.platform
```

modules/vstudio/vs2026_solution.lua
redorav: I guess I could just directly concatenate, I'll change it when I get a minute

modules/vstudio/vs2026_solution.lua
redorav: Yep can do when I get a minute

nickclark2016 (changes requested):
This PR causes failures in loading the solution file. Digging deeper, it's emitting multiple "Project" XML nodes for the same project, which is invalid.

More details after further investigation:

Each project is put into every group (folder node) in the solution. Something like this should suffice to show the issue:

```lua
workspace "Foo"
  project "Bar"
  kind "ConsoleApp"
  language "C++"
  
  group "Baz"
    project "Quux"
    kind "ConsoleApp"
    language "C++"
```

redorav:
Hmm could you give me an example of a script that does that? It's meant to just loop through the projects once. Ah I see, I haven't tested with group folders, I'll fix once I'm free this evening. I wonder if this is a bug that existed already in the original implementation and I just carried it over without much thought, I'll double check

nickclark2016:
> Hmm could you give me an example of a script that does that? It's meant to just loop through the projects once. Ah I see, I haven't tested with group folders, I'll fix once I'm free this evening. I think this is a bug that existed already in the original implementation, I just carried it over without much thought, I'll double check

This is not a bug present in the original version. It's a regression.

modules/vstudio/vs2026_solution.lua
```diff
@@ -148,161 +164,187 @@ function m.projects(wks)
...
-			p.pop('</Folder>')
-		end
-	end
-end
+		local tr = p.workspace.grouptree(wks)
+		tree.traverse(tr, {
+			onleaf = function(n)
+				local prj = n.project
 
+				local typeID = vstudio.tool(prj)
```
nickclark2016: Not sure if we actually need a tree traversal here, since we already have the projects in the sorted table.

nickclark2016 (approved):

PR #2654, opened 2026-04-16
Fix path encoding issues (#1153)

### LINKED ISSUE #1153: Path encoding issues on Windows
It seems the Windows path handling has some discrepancies when paths contain non-ANSI characters.

For instance, when running premake inside a directory with non-ANSI characters, it seems it notices there's a premake-system.lua file, but it can't load it:
```
C:\a>mkdir á
C:\a>cd á
C:\a\á>notepad premake-system.lua
C:\a\á>premake5.exe
Error: cannot open C:/a/├í/premake-system.lua: No such file or directory
```
but it works when using simple path names:
```
C:\a>mkdir a
C:\a>cd a
C:\a\a>notepad premake-system.lua
C:\a\a>premake5.exe
Type 'premake5 --help' for help
```

Another example, checking if a path exists shows some encoding issues:
```
C:\a\á>premake5.exe --interactive
> print(os.getcwd())
C:/a/├í
> print(os.isdir(os.getcwd()))
true
> print(path.getabsolute("../á"))
C:/a/á
> print(os.isdir(path.getabsolute("../á")))
false
```
and it works well with simple paths:
```
C:\a\a>premake5.exe --interactive
> print(os.getcwd())
C:/a/a
> print(os.isdir(os.getcwd()))
true
> print(path.getabsolute("../a"))
C:/a/a
> print(os.isdir(path.getabsolute("../a")))
true
```

Is there anything we can do to fix this issue (other than renaming all our paths? 😄)

jkriegshauser:
**What does this PR do?**

Solves #1153 by adding complete UTF-8 path support to both the embedded Lua 5.3 and premake5 functions

closes #1153 

Existing behavior should remain unchanged. There are a few other minor fixes that I will enumerate later

Don't think so, but feel free to ask away.

Can squash if necessary before merging. Also probably need to take a look at the documentation.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -224,6 +224,7 @@
 
 		multiprocessorcompile "On"
 		warnings "Extra"
+		symbols "On"
 
 		filter { "options:not zlib-src=none" }
 			defines { "PREMAKE_COMPRESSION" }
@@ -273,7 +274,6 @@
 
 		filter "configurations:Debug"
 			defines     "_DEBUG"
-			symbols	    "On"
 
 		filter "configurations:Release"
 			defines             "NDEBUG"
@@ -286,6 +286,14 @@
 
 		filter { "system:windows", "configurations:Release" }
 			incrementallink "Off"
+
+		filter { "system:windows", "action:vs*" }
+			characterset "Unicode"
+
+		filter { "system:windows", "action:not vs*" } -- TODO: mingw doesn't support `characterset "Unicode"`
+			defines { "UNICODE", "_UNICODE" }
+			buildoptions { "-municode" }
+			linkoptions { "-municode" }
 
 		-- MinGW AR does not handle LTO out of the box and need a plugin to be setup
 		filter { "system:windows", "configurations:Release", "toolset:not gcc" }
@@ -351,7 +359,7 @@
 			targetdir   "bin/release"
 
 		filter "system:windows"
-			links       { "ole32", "ws2_32", "advapi32", "version" }
+			links       { "ole32", "ws2_32", "advapi32", "version", "user32" }
 
 		filter { "system:windows", "toolset:msc*" }
 			files       { "src/**.rc" }
--- a/premake5.lua
+++ b/premake5.lua
@@ -351,7 +351,7 @@
 			targetdir   "bin/release"
 
 		filter "system:windows"
-			links       { "ole32", "ws2_32", "advapi32", "version" }
+			links       { "ole32", "ws2_32", "advapi32", "version", "user32", "shell32" }
 
 		filter { "system:windows", "toolset:msc*" }
 			files       { "src/**.rc" }
--- a/Bootstrap.mak
+++ b/Bootstrap.mak
@@ -106,7 +106,7 @@ mingw-clean: nix-clean
 
 mingw: mingw-clean
 	mkdir -p build/bootstrap
-	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_STATICLIB -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $(SRC) -lole32 -lversion
+	$(CC) -o build/bootstrap/premake_bootstrap -DPREMAKE_NO_BUILTIN_SCRIPTS -DLUA_STATICLIB -DUNICODE -D_UNICODE -municode -I"$(LUA_DIR)" -I"$(LUASHIM_DIR)" $(SRC) -lole32 -lversion
 	./build/bootstrap/premake_bootstrap embed
 	./build/bootstrap/premake_bootstrap --arch=$(PLATFORM) --os=windows --to=build/bootstrap --cc=gcc $(PREMAKE_OPTS) gmake
 	$(MAKE) -C build/bootstrap -j`getconf _NPROCESSORS_ONLN` config=$(CONFIG)_$(PLATFORM:x86=win32)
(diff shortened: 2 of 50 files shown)


nickclark2016 (commented):
I see a lot of what feels like duplicate code for handling MBCS conversions. Can we make a utility function somewhere that handles all of this in one spot? Definitely in favor of this change. It'll be nice having better UTF support within Premake.

jkriegshauser:
> I see a lot of what feels like duplicate code for handling MBCS conversions. Can we make a utility function somewhere that handles all of this in one spot? Definitely in favor of this change. It'll be nice having better UTF support within Premake.

I was doing a self-review and realized that as well. Going to make it an actual Lua auxiliary extension that hopefully I can work back into the Lua upstream. Working on that now.

nickclark2016:
> > I see a lot of what feels like duplicate code for handling MBCS conversions. Can we make a utility function somewhere that handles all of this in one spot? Definitely in favor of this change. It'll be nice having better UTF support within Premake.
> 
> I was doing a self-review and realized that as well. Going to make it an actual Lua auxiliary extension that hopefully I can work back into the Lua upstream. Working on that now.

Shall we defer this PR until it's in the Lua upstream, figure out what work needs to be done to bring in the latest Lua version, then readdress within Premake?

I'd like to get this change landed before 5.0 stabilizes, so we don't break any ABIs that may be relied upon by native Premake addons post 5.0.

jkriegshauser:
> Shall we defer this PR until it's in the Lua upstream, figure out what work needs to be done to bring in the latest Lua version, then readdress within Premake?

Personally I'd like to land it here first. Premake already is on an older version of Lua and has Lua modifications, plus I think it's a longer task to get any changes in Lua upstream.

> I'd like to get this change landed before 5.0 stabilizes, so we don't break any ABIs that may be relied upon by native Premake addons post 5.0.

100% agree, there shouldn't be any API/ABI affect with this other than making UTF-8 work for filenames

jkriegshauser:
@nickclark2016 alright, take a look now. All tests pass for me on linux and windows

nickclark2016:
Looks like there are still a fair amount of compilation failures. It's also worth noting that we attempt to support system installs of our vendored libraries, so we can't really change the interface without risking breaking that.

jkriegshauser:
> Looks like there are still a fair amount of compilation failures. It's also worth noting that we attempt to support system installs of our vendored libraries, so we can't really change the interface without risking breaking that.

Got it; I didn't previously realize that. I guess this support would be conditional based on whether one builds with the `contrib`. I'll work on getting that fixed up

jkriegshauser:
@nickclark2016 can you kick off another CI run? I've rewritten history to only 3 commits, but also fixed several issues with mingw/system libs/clang on windows/etc.

jkriegshauser:
Thanks, much closer. I'll look into the solaris compile issue and failing netbsd test tomorrow. Is there any way for me to be able to start CI jobs without adding a comment to request it manually?

nickclark2016:
I don't have permissions to allow that, but as soon as you land this, I believe any future PRs would not have this problem.

Also, those VM based actions are flakey. Looks like the issue was the action itself, not the build. Only thing left appears to be issues in the web actions.

Jarod42:
> Is there any way for me to be able to start CI jobs without adding a comment to request it manually?

You might enable actions in your Clone repo.
Check your repo settings.

jkriegshauser:
Alright, I think all the fixes are in now. Please kick off another build.

I believe that the website build issue is unrelated. Looks like the root cause is:
```
ValidationError: Invalid options object. Progress Plugin has been initialized
using an options object that does not match the API schema.
```

which doesn't seem related to my changes. Might need a fix like pinning the node version or a docusaurus update or a package-lock.json or something to lock the versions 🤷

nickclark2016:
Ah yeah, looks like we need to actually use a package-lock.json (pretty sure none of the core contributors are actually web devs)

jkriegshauser:
> Ah yeah, looks like we need to actually use a package-lock.json (pretty sure none of the core contributors are actually web devs)

Added another PR to fix the website [here](https://github.com/premake/premake-core/pull/2657). I enabled actions on my fork and it passes there now (at least the check part; i obviously cannot publish)

nickclark2016:
Oh good. The auto jobs are working now.

jkriegshauser:
> Oh good. The auto jobs are working now.

yeah i just needed to land a change 😅

jkriegshauser:
The dragonflybsd check that failed timed out while booting QEMU. Can you manually restart just that one job?

nickclark2016:
Giving this a deep dive now that everything is built and the tests are happy. Are the tests in a state where they should catch regressions? I'm looking at going in to update the dependencies in the very near future, so I don't want to break the improvements you've made here.

jkriegshauser:
> Are the tests in a state where they should catch regressions?

Yes, the tests should validate all of the changed functionality so regressions should be caught. Plus we define `[_]UNICODE` now so `wmain()` is used and win32 functions default to the `W` extension if not specified. It doesn't really affect the CRT functions though (`remove()`, `stat()`, etc.). It might be good to mention in the documentation somewhere that full UTF-8 is contingent on building with contrib Lua; otherwise we add the functions but we can't fix the actual internal Lua calls to ANSI functions. If you can point me at the desired place in the docs i can add a blurb.

If you have any issues updating dependencies just lmk. Next step is for me to try to get the Lua changes upstreamed there too.

nickclark2016:
https://premake.github.io/docs/Building-Premake

If we had better documentation on how to build with the contrib folder vs system, it would live here.  To not dump that additional work on you after this massive lift, would you mind just opening an issue, dumping a blurb there about UTF8 support, and setting me as the assignee (if you have permission to do so)? Then I can get it massaged into an fuller section on the various build options.

jkriegshauser:
@nickclark2016 can't assign, but mentioned you: #2658 

If I get some spare time I can try to help out too.

nickclark2016:
> @nickclark2016 can't assign, but mentioned you: #2658

On my list to address before 5.0 stable (hopefully soon), thanks!

premake5.lua
```diff
@@ -287,6 +287,13 @@
 		filter { "system:windows", "configurations:Release" }
 			incrementallink "Off"
 
+		filter { "system:windows" }
```
nickclark2016: Does `characterset "Unicode"` not work?

nickclark2016 (commented):
Overall, very happy with the state of this. General question to chew on:

Should we have the UTF8 changes better macro-ed off? This would let us turn on and off UTF8 support, which would be a nice to have for building using system dependencies instead of the contrib modules.

premake5.lua
jkriegshauser: > Overall, very happy with the state of this. 

☺️

> Should we have the UTF8 changes better macro-ed off? This would let us turn on and off UTF8 support, which would be a nice to have for building using system dependencies instead of the contrib modules.

I don't think it's really necessary. If you build with system Lua on Windows, the `_UTF8_ENABLED` doesn't get set (will evaluate as nil) which is how you know if you have UTF-8 support. You still get _some_ UTF-8 support but it cannot be relied upon (which is kind of how it was before this change). But ideally I manage to get the Lua changes upstream and we can update to that, so it might be a moot point.

> Does characterset "Unicode" not work?

It does for vs (i tried with vs2022), but not for `mingw`. I'll push an update and I can make an issue if you want.

nickclark2016: Let's go ahead an open an issue and have it reference this, I don't want to block this.

nickclark2016:
I'm going to leave this open for a few days for others to review, but LGTM at this point. Thanks!!

nickclark2016 (approved):

premake5.lua
jkriegshauser: Issue created: #2659

dfagnou:
approved

PR #2667, opened 2026-04-27
fix return value of os.writefile_ifnotequal()

mercury233:
**What does this PR do?**

- fix return value of os.writefile_ifnotequal()
- update the document for that function

Before #2654 , os.writefile_ifnotequal() returns `1` if successful, `0` if the string is identical to the current file contents, otherwise `-1` and an error message. #2654 seems to changed this behavior accidentally, return `0` on success too. This PR change the behavior back.

@jkriegshauser

### DIFF
--- a/src/host/os_writefile_ifnotequal.c
+++ b/src/host/os_writefile_ifnotequal.c
@@ -105,7 +105,7 @@ int os_writefile_ifnotequal(lua_State* L)
 
 		if (!error)
 		{
-			lua_pushinteger(L, error ? -1 : 0);
+			lua_pushinteger(L, 1);
 			return 1;
 		}
 	}
--- a/website/docs/os/os.writefile_ifnotequal.md
+++ b/website/docs/os/os.writefile_ifnotequal.md
@@ -13,7 +13,15 @@ ok, err = os.writefile_ifnotequal("text", "filename")
 
 ### Return Value ###
 
-True if successful, otherwise nil and an error message.
+The first return value:
+
+| Value | Explanation                                          |
+|-------|------------------------------------------------------|
+| 1     | The string was written to the file                   |
+| 0     | The string is identical to the current file contents |
+| -1    | An error occurred                                    |
+
+The second return value is an error message if the first return value is -1, otherwise nil.
 
 
 ### Availability ###


website/docs/os/os.writefile_ifnotequal.md
```diff
@@ -13,7 +13,7 @@ ok, err = os.writefile_ifnotequal("text", "filename")
 
 ### Return Value ###
 
-True if successful, otherwise nil and an error message.
+`1` if successful, `0` if the string is identical to the current file contents, otherwise `-1` and an error message.
```
nickclark2016: Suggest changing this to be a table

nickclark2016 (changes requested):

src/host/os_writefile_ifnotequal.c
```diff
@@ -105,7 +105,7 @@ int os_writefile_ifnotequal(lua_State* L)
 
 		if (!error)
 		{
-			lua_pushinteger(L, error ? -1 : 0);
+			lua_pushinteger(L, 1);
```
jkriegshauser, line 108: This should probably be

```suggestion
			lua_pushinteger(L, error ? -1 : 1);
```

If an error occurs during `write()`, we want `-1` pushed. Otherwise, a success should push `1`. You're correct though--my change inadvertently changed it to `0`.

jkriegshauser:
Please add unit tests to prevent future regressions

src/host/os_writefile_ifnotequal.c
mercury233: but this part is already covered by `if (!error)`

jkriegshauser: You're correct. This change is fine.

website/docs/os/os.writefile_ifnotequal.md
mercury233:
```suggestion
The first return value:

| Value | Explanation                                          |
|-------|------------------------------------------------------|
| 1     | The string was written to the file                   |
| 0     | The string is identical to the current file contents |
| -1    | An error occurred                                    |

The second return value is an error message if the first return value is -1, otherwise nil.
```

should it be like this?

mercury233:
Added tests, but I’m unsure whether they are appropriate, since the `test_runner.lua` intentionally uses functions like `stub_os_writefile_ifnotequal` to prevent actual operations.

jkriegshauser:
> Added tests, but I’m unsure whether they are appropriate, since the `test_runner.lua` intentionally uses functions like `stub_os_writefile_ifnotequal` to prevent actual operations.

If the test fails before your fix then it's appropriate 😄 

You may have to capture the original function value before it's stubbed out. Such as what I did in _test_lua_unicode.lua_:
```lua
    local real_io_open = io.open

-- ...

function suite.loadfile_UnicodeFilename()
	local p = tmpname()
	p = p .. "_café.lua"
	local f = assert(real_io_open(p, "w"))
	-- ...
end
```

jkriegshauser (approved):

mercury233:
I've reverted the test changes which I'm unsure, so this PR can be treated as ready. Will move the tests to another PR.

nickclark2016:
> I've reverted the test changes which I'm unsure, so this PR can be treated as ready. Will move the tests to another PR.

I would much rather have the changes tested in this PR than have a separate PR for tests. That defeats the purpose of having the code tested before going into the main branch.

mercury233:
> > I've reverted the test changes which I'm unsure, so this PR can be treated as ready. Will move the tests to another PR.
> 
> I would much rather have the changes tested in this PR than have a separate PR for tests. That defeats the purpose of having the code tested before going into the main branch.

I still don't know whether the approach in https://github.com/premake/premake-core/pull/2682 is acceptable.

jkriegshauser:
> > > I've reverted the test changes which I'm unsure, so this PR can be treated as ready. Will move the tests to another PR.
> > 
> > 
> > I would much rather have the changes tested in this PR than have a separate PR for tests. That defeats the purpose of having the code tested before going into the main branch.
> 
> I still don't know whether the approach in #2682 is acceptable.

I pulled down your change from #2682 and ran:
```
./Boostrap.sh
./bin/release/premake5 test
```
and noted the test errors. ❌

I then applied your changes from this MR over the top of #2682 and re-ran the above. Test errors are gone. ✅

I took a look at your tests too. They looks good, but i'd just add one small test that also tests the error case by attempting to write over a directory:
```
	function suite.writefile_ifnotequal_ReturnsError_OnInvalidPath()
		local dir = tmpdir()
		local ok, err = real_writefile_ifnotequal("Hello, World!", dir)
		test.isequal(ok, -1) -- cannot write to dir
		test.isequal("string", type(err))
		test.isequal(err, "unable to write file to '" .. dir .. "'")
		os.remove(dir)
	end
```

I only tested it locally on Linux. If it fails on other platforms we can ignore it for now (or fix it).

So please put your tests back in this PR with the additional test above.

jkriegshauser (changes requested):
See previous comment

Jarod42 (approved):

jkriegshauser:
Since this PR appears to be abandoned, I've combined this and the test PR here: #2694

nickclark2016:
Resolved by #2694

PR #2672, opened 2026-04-28
feat: print [New] or [Updated] in generation output

ak-slongchamps:
**What does this PR do?**

Prints [New] or [Updated] next to the generated files when they have been touched on the filesystem.

Adds extra info to pre-exiting output.
Uses global `_WORKING_DIR` instead of calling `os.getcwd()` as a minor optimization.
Adds a filesystem access to determine if the file existed prior.

Useful to know when files are new vs updated

### DIFF
--- a/src/base/premake.lua
+++ b/src/base/premake.lua
@@ -168,6 +168,7 @@
 			error(err, 0)
 		end
 
+		local fileexists = os.isfile(fn)
 		local f, err = os.writefile_ifnotequal(output, fn);
 
 		if (f == 0) then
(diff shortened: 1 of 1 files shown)


src/base/premake.lua
```diff
@@ -168,14 +168,19 @@
...
 		local f, err = os.writefile_ifnotequal(output, fn);
 
 		if (f == 0) then
 			return false -- file not modified
 		elseif (f < 0) then
 			error(err, 0)
 		elseif (f > 0) then
-			printf("Generated %s...", path.getrelative(os.getcwd(), fn))
+			if fileexists then
+				printf("Generated %s... [Updated]", path.getrelative(_WORKING_DIR, fn))
```
nickclark2016, line 180: This feels like we should do this only in the case of the "verbose" flag. If it's not specified, it should be using the original behavior.

nickclark2016 (changes requested):

Jarod42 (approved):

src/base/premake.lua
samsinsane: Would also be nice to avoid duplicating the `printf` so that updating the call doesn't require making the same change in multiple places.

PR #2675, opened 2026-04-28
fix: sort workspace configs and projects alphabetically for deterministic output

ak-slongchamps:
**What does this PR do?**

It patches iterators (eachconfig/eachproject) to sort by name, so that any users of these can produce deterministic output even when insertion order changes, which could differ depending on the client code (e.g., order of options, reflection from filesystem, etc).

Since this only changes the order without modifying the actual config or project elements, there is no behavior we're aware of that could result from this change. If there was an assumption by client code of the order of configs or projects, this could be changed, but Premake already makes no guarantee on the order.

The test in test_codelite_workspace.lua covers this change: project order and config orders are set in the captured output that's being validated and is already in the expected alphabetical ordering.
Note that I would still not make any promised in doc about the ordering returned by these functions, this is just a convenient way to drive baking output to be deterministic.

Been using this in production for years at this point, glad to contribute!

### DIFF
--- a/modules/codelite/tests/test_codelite_workspace.lua
+++ b/modules/codelite/tests/test_codelite_workspace.lua
@@ -206,17 +206,17 @@
 <CodeLite_Workspace Name="MyWorkspace" Database="" Version="10000">
   <Project Name="MyProject" Path="MyProject.project"/>
   <BuildMatrix>
-    <WorkspaceConfiguration Name="x86_64-Debug" Selected="yes">
-      <Project Name="MyProject" ConfigName="x86_64-Debug"/>
-    </WorkspaceConfiguration>
-    <WorkspaceConfiguration Name="x86-Debug" Selected="no">
+    <WorkspaceConfiguration Name="x86-Debug" Selected="yes">
       <Project Name="MyProject" ConfigName="x86-Debug"/>
     </WorkspaceConfiguration>
-    <WorkspaceConfiguration Name="x86_64-Release" Selected="no">
-      <Project Name="MyProject" ConfigName="x86_64-Release"/>
+    <WorkspaceConfiguration Name="x86_64-Debug" Selected="no">
+      <Project Name="MyProject" ConfigName="x86_64-Debug"/>
     </WorkspaceConfiguration>
     <WorkspaceConfiguration Name="x86-Release" Selected="no">
       <Project Name="MyProject" ConfigName="x86-Release"/>
+    </WorkspaceConfiguration>
+    <WorkspaceConfiguration Name="x86_64-Release" Selected="no">
+      <Project Name="MyProject" ConfigName="x86_64-Release"/>
     </WorkspaceConfiguration>
   </BuildMatrix>
 </CodeLite_Workspace>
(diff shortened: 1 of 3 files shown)


Jarod42 (approved):
Order might change the default project/config for some generator though.

nickclark2016:
I'm reluctant to merge this change because of [Hyrum's Law](https://www.hyrumslaw.com/). Can you describe the specific problem this is solving?

ak-slongchamps:
The specific problem this fixes is deterministic ordering in generator outputs when definitions files are loaded dynamically. Because configurations and projects are additive, the resulting order ends up dependent on the dynamic loading order.

Providing a sorted order only post-baking was a minimum impact way to solve the issue, which requires some sort of logic behind the ordering. Generators use those functions to output the corresponding lists and don't, as far as I can tell, depend on the ordering of elements to interpretation elements in that list. IMO, if a generator exists in the field that does, it's bound to break for some other client-provided ordering.

nickclark2016:
> The specific problem this fixes is deterministic ordering in generator outputs when definitions files are loaded dynamically. Because configurations and projects are additive, the resulting order ends up dependent on the dynamic loading order.
> 
> Providing a sorted order only post-baking was a minimum impact way to solve the issue, which requires some sort of logic behind the ordering. Generators use those functions to output the corresponding lists and don't, as far as I can tell, depend on the ordering of elements to interpretation elements in that list. IMO, if a generator exists in the field that does, it's bound to break for some other client-provided ordering.

To me, this is still speaking in generalities. Have you found a bug that this solves? Is there a specific feature that cannot be implemented without this change?

I'm a bit hesitant to bring in code without a tangible problem being solved.

ak-slongchamps:
In our setup, we load files dynamically for hiding platform info, and to support plugins. However, we cannot control the exact order these are defined, and so we end up with configurations and platforms that differ in order depending on the machine. While deduplication of configs and platforms allows to have a properly merged list, because it's not in a deterministic order we end up with solutions with different config set, leading to odd things like different defaults (since the first one is selected if not specified via `defaultconfig`). This hinders diffing, which customers do when they upgrade (since we ship the solutions, not just the lua definitions).

That lack of determinism is a bug, at least to us. I suppose if the risk is too high for others dependent of the current ordering, we could add an option to control the order, where we'd have Default and Sorted as choices. Would that mitigate the risk enough?

nickclark2016:
That specific use case is enough for me to call this a bug and bring this into core without any further concern for feature-gating it.

nickclark2016 (approved):

mercury233:
I found that this change is a serious breaking change. Actually, the project does not provide any `defaultconfig` API, so when using generic configuration names like `configurations { "Release", "Debug" }`, the only way to make Release the default is to place it first. If any CI scripts rely on this undefined behavior without explicitly specifying `make config=release`, they will break.

Therefore, I hope a `defaultconfig` or similar API can be added, and that this change is clearly documented in the release notes.

_From someone who spent 10 minutes wondering why local changes were not taking effect, only to realize the project had silently compiled the Debug configuration by default._

Jarod42:
@mercury233 : I think it is worst: depending of generator, default configuration is first or last.
Agree with a `defaultconfig` might help.

ak-slongchamps:
Fair point, I was mistaken with `defaultplatform` and the fact that we have `defaultconfig` in the gmake code itself, but missed that it's not exposed publicly.

Sounds like a simple addition, I'll look into it.

PR #2676, opened 2026-04-28
fix: add Precise to msc floatingpoint

ak-slongchamps:
**What does this PR do?**

Adds explicit /fp:precise entry as Precise in the msc floatingpoint list of supported values.

Simple one-liner to be able to explicitly set it.

Nope

### DIFF
--- a/src/_premake_init.lua
+++ b/src/_premake_init.lua
@@ -342,6 +342,7 @@
 			"Default",
 			"Fast",
 			"Strict",
+			"Precise"
 		}
 	}
 
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -50,6 +50,8 @@
 		flags = gcc.shared.flags,
 		floatingpoint = {
 			Fast = "-ffast-math",
+			Strict = "-ffp-model=strict",
+			Precise = "-ffp-model=precise"
 		},
 		strictaliasing = gcc.shared.strictaliasing,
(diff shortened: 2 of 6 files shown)


Jarod42 (commented):
Is there an equivalence for gcc/clang?

https://github.com/premake/premake-core/blob/master/src/_premake_init.lua#L337C1-L338C26 has not been changed to add that value, so I doubt your change is effective.

UT might have been added.

nickclark2016:
> Is there an equivalence for gcc/clang?
> 
> https://github.com/premake/premake-core/blob/master/src/_premake_init.lua#L337C1-L338C26 has not been changed to add that value, so I doubt your change is effective.
> 
> UT might have been added.

I don't think there's an equivalent for GCC without contracts, but Clang does have `-ffp-model=precise`

nickclark2016 (changes requested):
This should have UTs, both for MSC and vs2010_vcxproj

Additionally, this option is supported for Clang via `-ffp-model=precise`. I would request that this is added, as well as `Strict` support for clang via `-ffp-model=strict`.

ak-slongchamps:
@Jarod42 You're absolutely right, I missed `_premake_init.lua` in this backport (our patches over premake are squashed, this is a bit of an archeology project!).

Added the clang flags (love it!) and unit tests.

nickclark2016 (approved):

PR #2677, opened 2026-04-29
feat: add support for toolset-provided extensions in msc.getlinks

ak-slongchamps:
**What does this PR do?**

When using clang or gcc as a native toolset in Visual Studio as a cross-compiler, libraries may have .a or .so as extensions, but currently .lib is appended to the lib's name in the linker arguments list, which results in e.g., Foo.a.lib, Bar.so.lib for Foo.a and Bar.so, respectively.

We add the support for extensions that are typical of the toolsets on corresponding platforms and use them specifically in `msc.getlinks` to supplement the list of extensions when the active toolset defines some.
This only has an impact in the special case of a msc generated output with a non-msc toolset, typically for cross-compilation scenarios. Because we just add new extensions, it should not impact usages with clang-cl.

Nope.

### DIFF
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -483,10 +483,15 @@
 
 		-- Then the system libraries, which come undecorated
 		local system = config.getlinks(cfg, "system", "fullpath")
+
+		local toolset = config.toolset(cfg)
+		local toolsetExts = toolset and toolset.getLibraryExtensions and toolset.getLibraryExtensions() or {}
+		local libExts = table.merge(p.tools.msc.getLibraryExtensions(), toolsetExts)
+
 		for i = 1, #system do
 			-- Add extension if required
 			local link = system[i]
-			if not p.tools.msc.getLibraryExtensions()[link:match("[^.]+$")] then
+			if not libExts[link:match("[^.]+$")] then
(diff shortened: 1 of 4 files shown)


nickclark2016 (approved):

Jarod42 (approved):

src/tools/msc.lua
```diff
@@ -483,10 +483,15 @@
 
 		-- Then the system libraries, which come undecorated
 		local system = config.getlinks(cfg, "system", "fullpath")
+
+		local toolset = config.toolset(cfg)
```
nickclark2016, line 487: After sleeping on this and discussion with @samsinsane, we think it may make more sense to move the logic from `msc.lua` to the VS project generation file. This here is a bit smelly, since we are already in a toolset file.

Right now, VS is pretty hard coded to `msc`-isms. This may be a case where we want to be more agnostic in the VS exporter itself, loading in the specified toolset rather than just using MSC.

nickclark2016 (changes requested):

src/tools/msc.lua
ak-slongchamps: Interesting, worth investigating. I'll put this on pause until I get access to the specific platform that requires this change on our side to validate this approach.

PR #2703, opened 2026-06-08
Allow disabling implicit includes in compilecommands

### LINKED ISSUE #2693: premake5 compilecommands tweak
Please add a switch to tell it to not include `-isystem`.

Currently, it includes `-isystem`. I know you made it this way to try to ensure that the language server can find headers, but sometimes it causes issues itself.

For example: I got full of duplicated definitions errors by `clangd` today and these only went away if I manually removed all of the `-isystem` instances in the generated `compile_commands.json`. I know this doesn't normally happen. But some GCC/Clang builds are just problematic.

nickclark2016:
**What does this PR do?**

Allows disabling of implicit includes in `compilecommands` action.

No breaking changes. Default behavior remains the same.

`--implicit-includes=On` - Tries to fetch the implicit include directories from the toolset, adds to the compilation database if successful. This is the default behavior.
`--implicit-includes=Off` - Does not attempt to find the implicitly added include directories to the compilation database.

Closes #2693 

### DIFF
--- a/modules/compilecommands/_preload.lua
+++ b/modules/compilecommands/_preload.lua
@@ -28,6 +28,20 @@ newoption {
 	description = "Specify the output file path for compile_commands.json",
 }
 
+newoption {
+	category = "Compilation Database",
+
+	trigger = "implicit-includes",
+	description = "Include implicit include directories in compile_commands.json",
+	
+	allowed = {
+		{ "On" , "Include implicit include directories (default)" },
+		{ "Off", "Exclude implicit include directories" },
+	},
+
+	default = "On",
+}
+
 newaction {
 	trigger = "compilecommands",
 	shortname = "compilecommands",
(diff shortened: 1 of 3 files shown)


Jarod42 (approved):

modules/compilecommands/_preload.lua
```diff
@@ -28,6 +28,20 @@ newoption {
 	description = "Specify the output file path for compile_commands.json",
 }
 
+newoption {
+	category = "Compilation Database",
+
+	trigger = "implicit-includes",
```
samsinsane, line 34: Should this be prefixed like the others with `cc-`?

samsinsane (approved):
Any reason this is an option over an API?

nickclark2016:
> Any reason this is an option over an API?

It's more an option related to the environment you're working in, rather than a configuration on the project itself.

modules/compilecommands/_preload.lua
nickclark2016: I'm on the fence with this one. Do you have a preference? I don't mind changing it, though I am leaning towards renaming/aliasing the other ones to get rid of the `-cc` the more I think about it.

PR #2718, opened 2026-06-19
add defaultconfiguration API and config.getdefault

### LINKED ISSUE #2709: Default config changed from release to debug for gmake action

A regression, default config changed from release to debug for gmake action

config=release to stay the default.

none

```sh
premake gmake
make
```

- [ ] Visual Studio 2026 (vs2026)
- [ ] Visual Studio 2022 (vs2022)
- [ ] Visual Studio 2019 (vs2019)
- [ ] Visual Studio 2017 (vs2017)
- [ ] Visual Studio 2015 (vs2015)
- [ ] Visual Studio 2012 (vs2012)
- [ ] Visual Studio 2010 (vs2010)
- [ ] Visual Studio 2008 (vs2008)
- [ ] Visual Studio 2005 (vs2005)
- [x] GNU Makefile (gmake)
- [ ] GNU Makefile Legacy (gmakelegacy)
- [ ] XCode (xcode)
- [ ] Codelite
- [ ] Compile Commands
- [ ] Other (Please list below)

Saw the compilecommands action so had go with -dev commit 251f0a2a8b2e2db2e2a646ed1fcb8813ce9e9f16

Would be a bother to update the CI^^

*(You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to issues like these!)*

mercury233:
**What does this PR do?**

Close #2709

- Adds the `defaultconfiguration` API to specify the default build configuration. For example, in gmake this controls which configuration is used when running `make` without explicitly passing `config=release`.
- Adds `config.getdefault()` to centralize the logic for selecting the default configuration.
- Updates generators/modules to use the new shared default-selection logic.
- Add warning for `defaultplatform` when a non-existent value is specified, and update the documentation.
- Adds related tests.

When `defaultconfiguration` is specified, generators that support a default configuration will use it when selecting their default build target/configuration.

When `defaultconfiguration` is not specified, the expected behavior should remain consistent with the existing tests. The sorting performed by `config.getdefault()` is aligned with the workspace configuration ordering introduced by #2675, which sorted `workspace.eachconfig()` for deterministic output.

This can also be seen as a follow-up to #2675, extending the default-selection logic so it is shared and consistent across modules.

There were existing design comments using the name `defaultConfiguration`, so this PR preserves that API name conceptually while exposing it in Premake's usual lowercase style as `defaultconfiguration`.

#2675 did not update `project.eachconfig()` to sort project configurations. That may have been intentional, but whether project-level configuration iteration should also be sorted may be worth discussing separately.

### DIFF
--- a/modules/vstudio/tests/sln2005/test_platforms.lua
+++ b/modules/vstudio/tests/sln2005/test_platforms.lua
@@ -8,7 +8,6 @@
 	local suite = test.declare("vstudio_sln2005_platforms")
 	local sln2005 = p.vstudio.sln2005
 
-
 --
 -- Setup
 --
@@ -509,16 +508,16 @@ EndGlobalSection
 		prepare()
 		test.capture [[
 GlobalSection(SolutionConfigurationPlatforms) = preSolution
-	Debug|x64 = Debug|x64
 	Debug|x86 = Debug|x86
+	Debug|x64 = Debug|x64
 	Release|x64 = Release|x64
 	Release|x86 = Release|x86
 EndGlobalSection
 GlobalSection(ProjectConfigurationPlatforms) = postSolution
-	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x64.ActiveCfg = Debug|x64
-	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x64.Build.0 = Debug|x64
 	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x86.ActiveCfg = Debug|x86
 	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x86.Build.0 = Debug|x86
+	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x64.ActiveCfg = Debug|x64
+	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Debug|x64.Build.0 = Debug|x64
 	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Release|x64.ActiveCfg = Release|x64
 	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Release|x64.Build.0 = Release|x64
 	{C9135098-6047-8142-B10E-D27E7F73FCB3}.Release|x86.ActiveCfg = Release|x86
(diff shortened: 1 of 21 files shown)


mercury233:
_The CI failure was primarily caused by some NuGet API error (504) encountered during testing._

seragh:
Tested and works for me, thanks for the effort.

src/base/config.lua
```diff
@@ -191,6 +191,58 @@
...
+			if matchesConfiguration and matchesPlatform then
+				return cfg
+			elseif not bestConfiguration and matchesConfiguration then
+				bestConfiguration = cfg
+			elseif not bestPlatform and matchesPlatform then
+				bestPlatform = cfg
+			end
+		end
+
+		return bestConfiguration or bestPlatform or first
```
Jarod42, line 254: Should we warn if user set an non-existent configuration/platform?

modules/vstudio/tests/sln2005/test_platforms.lua
```diff
@@ -8,6 +8,17 @@
...
+	local function assertCapturedContainsInOrder(...)
+		local actual = p.captured()
+		local position = 1
+		for i = 1, select("#", ...) do
+			local fragment = select(i, ...)
+			local found = actual:find(fragment, position, true)
+			test.isnotnil(found)
+			position = found + #fragment
+		end
+	end
```
samsinsane: Why do we need this new assert function when we can just assert the full string result of `test.capture`?

modules/vstudio/vs2005_solution.lua
```diff
@@ -309,7 +310,13 @@
 		-- Now I can output the sorted list of solution configuration descriptors
 
 		-- Visual Studio assumes the first configurations as the defaults.
-		if wks.defaultplatform then
+		if wks.defaultconfiguration and defaultCfg then
+			p.push('GlobalSection(SolutionConfigurationPlatforms) = preSolution')
+			p.w('%s = %s', descriptors[defaultCfg], descriptors[defaultCfg])
+			p.pop("EndGlobalSection")
+		end
```
samsinsane: It may be simpler to just modify the sort above to put the `defaultCfg` value first - the checks with `defaultconfiguration` and `defaultplatform` are irrelevant as `defaultCfg` is always set to something and it's the only one that needs to be first. Then all the conditions here and below disappear and we just do a simple loop.

modules/xcode/xcode_common.lua
```diff
@@ -1726,8 +1726,8 @@
 
 
 	function xcode.XCBuildConfigurationList(tr)
-		local wks = tr.project.workspace
-		local defaultCfgName = xcode.stringifySetting(tr.configs[1].buildcfg)
+		local defaultCfg = p.config.getdefault(tr.project) or tr.configs[1]
```
samsinsane: `p.config.getdefault` doesn't return `nil`, so the `or ...` is unnecessary.

src/_premake_init.lua
```diff
@@ -1347,6 +1353,8 @@
 	api.alias("buildmessage", "buildMessage")
 	api.alias("buildoutputs", "buildOutputs")
 	api.alias("editandcontinue", "editAndContinue")
+	api.alias("defaultconfiguration", "defaultConfiguration")
+	api.alias("defaultplatform", "defaultPlatform")
```
samsinsane: The aliases here with mixed case exist because the API originally had mixed case and was changed to be all lowercase. We don't add new ones here.

tests/base/test_config_getdefault.lua
```diff
@@ -0,0 +1,201 @@
...
+--
+-- Setup/teardown
+--
+
+	local wks, prj
+
+	local function assertDefault(cfg, buildcfg, platform)
+		test.isequal(buildcfg, cfg.buildcfg)
+		test.isequal(platform, cfg.platform)
+	end
```
samsinsane: Not a huge fan of the helper function to wrap the `test.isequal` calls, I might be mistaken but I don't think we really do this in other tests.

website/docs/defaultconfiguration.md
```diff
@@ -0,0 +1,42 @@
...
+
+`configuration_name` - The name of the build configuration to use as default. If a value that is not present in `configurations` is specified, it has no effect.
+
+### Applies To ###
+
+Workspace configurations.
+
+### Availability ###
+
+Premake 5.0.0 or later.
```
samsinsane, line 19: @nickclark2016 is this the format we've used for the first non-beta release?

samsinsane (changes requested):

src/base/config.lua
nickclark2016: I second this. This feels like a warn situation.

tests/base/test_config_getdefault.lua
nickclark2016: I agree. This is one case where I am against DRY, since it's hiding the assertions being tested.

website/docs/defaultconfiguration.md
nickclark2016: This is (see `dynamicdebugging.md`)

nickclark2016 (changes requested):

modules/vstudio/tests/sln2005/test_platforms.lua
mercury233: I previously felt that the output could become very long and be easily affected by noise and other factors. It has now been changed to match the raw text.

samsinsane (approved):

nickclark2016 (approved):

src/_premake_init.lua
```diff
@@ -238,6 +238,12 @@
 		kind = "string",
 	}
 
+	api.register {
+		name = "defaultconfiguration",
+		scope = "project",
+		kind = "string",
```
nickclark2016, line 244: Actually, one thought before merging this. This should probably be a workspace setting, not a project. I can't imagine not building the entire solution in different configs.

mercury233: I agree that having multiple projects use different configurations is a very rare scenario, but Visual Studio does support it, and I've heard that Xcode supports it as well. Also, the current `defaultplatform` API is already defined at the project level.

image

PR #2727, opened 2026-07-04
Use TargetFrameworks tag if multiple frameworks are provided by dotnetframework

CapeGuyBen:
**What does this PR do?**

Allows multiple target .NET frameworks to be supported by a single dotnet project

Extends the `dotnetframework` property to switch from using `TargetFramework` to `TargetFrameworks` when multiple (semicolon-separated) target names are provided

### DIFF
--- a/modules/vstudio/tests/cs2005/test_netcore.lua
+++ b/modules/vstudio/tests/cs2005/test_netcore.lua
@@ -18,116 +18,116 @@ local project = p.project
 local wks, prj
 
 function suite.setup()
-    p.action.set("vs2005")
-    wks, prj = test.createWorkspace()
+	p.action.set("vs2005")
+	wks, prj = test.createWorkspace()
 	configurations { "Debug", "Release", "Distribution"}
-    language "C#"
+	language "C#"
 end
 
 local function targetFrameworkPrepare()
-    local cfg = test.getconfig(prj, "Debug")
-    dn2005.netcore.targetFramework(cfg)
+	local cfg = test.getconfig(prj, "Debug")
+	dn2005.netcore.targetFramework(cfg)
 end
 
 local function targetFrameworkVersionPrepare()
-    local cfg = test.getconfig(prj, "Debug")
-    dn2005.targetFrameworkVersion(cfg)
+	local cfg = test.getconfig(prj, "Debug")
+	dn2005.targetFrameworkVersion(cfg)
 end
 
 local function projectConfigurationsPrepare()
-    dn2005.projectConfigurations(prj)
+	dn2005.projectConfigurations(prj)
 end
 
 local function prepareNetcore()
-    dn2005.projectElement(prj)
+	dn2005.projectElement(prj)
 end
 
 local function prepareProjectProperties()
-    local localProj = test.getproject(wks, 1)
+	local localProj = test.getproject(wks, 1)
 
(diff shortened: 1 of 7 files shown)


modules/vstudio/tests/cs2005/test_netcore.lua
```diff
@@ -172,6 +172,21 @@ function suite.implicitUsings_off()
     ]]
 end
 
+function suite.multiple_target_dotnet_frameworks()
```
tritao, line 175: This has mixed spaces and tabs, should be normalized to tabs.

website/docs/dotnetframework.md
```diff
@@ -1,24 +1,20 @@
...
-dotnetframework ("version")
+dotnetframework ("versions")
 ```
 
 This value currently is only applied to Visual Studio 2005 or later, and GNU makefiles using Mono. If no .NET framework version is specified the toolset default is used.
 
 ### Parameters ###
 
-`version` is one of:
+`versions` is a semicolon separated list of version numbers where each element is one of:
```
Jarod42: Shouldn't we use table for list?

modules/vstudio/tests/cs2005/test_netcore.lua
CapeGuyBen: test_netcore.lua is now normalised to tabs

website/docs/dotnetframework.md
nickclark2016: I would agree. A table would be better.

To follow up, a table input *should* allow for single inputs as well.

nickclark2016 (changes requested):

website/docs/dotnetframework.md
CapeGuyBen: It could use a `list:string`? I think that would require a separate `dotnetframeworks` (plural version) property to avoid breaking backwards compatibility with existing uses of `dotnetframework`. This would mirror the `TargetFramework` and `TargetFrameworks` solution properties.

nickclark2016: I don't believe so. I would need to verify, but it should just let you "append" to a list/table if you have only a single input specified, rather than a table.

nickclark2016: See: https://github.com/premake/premake-core/blob/master/tests/api/test_list_kind.lua

CapeGuyBen: Updated to use a table (using `list = true`) for `dotnetframework`.

Jarod42 (approved):

nickclark2016:
Thinking a little more about this -- Does it make sense to group all of the dotnet stuff into the framework API? Or should we have a separate API that supports the old core vs framework dotnet versions and then the new merged version?

CapeGuyBen:
> Thinking a little more about this -- Does it make sense to group all of the dotnet stuff into the framework API? Or should we have a separate API that supports the old core vs framework dotnet versions and then the new merged version?

What Premake has now mirrors how dotnet project files represent them eg. they are all under the TargetFramework(s) tag and the framework naming convention is the one which dotnet uses so don't think separating them would add much IMHO.

nickclark2016:
Seems reasonable to me. Implementation seems sane. Happy to move this forward.

nickclark2016 (approved):

Issue #614, opened 2016-11-04
C# projects, add mechanism of being able to run custom action for invidividual file


tapika:
In C++ projects I'm including .lua configuration script as customly built file, for example like this:

    vpaths { ["*"] = { lua2compile } }
    files { lua2compile }

    filter ("files:**" .. lua2compile )
    buildrule {
         description = 'Generating project from $(SolutionName).lua file',
         commands = 
             "    premake5.exe --file=\"%Lua%\" " .. buildvsver .. "\r\n" ..
             ")\r\n" ..
             "echo 1>$(IntDir)" .. stateFile
         ,
         outputs = { "$(IntDir)" .. stateFile }
    }        
    filter {}

But I miss for C# to do the same.

There exists "buildaction", but currently you cannot specify custom command for it.

    files { includedFrom }
	filter ( "files:**" .. includedFrom )
    buildaction "Compile"
	filter { }

Using it as "Compile" - tries to compile .lua script, but not to use premake5.exe.

This is an important feature for integrating premake5 to be used for C# based projects.

PR #2729, opened 2026-07-07
Dotnet buildaction uses given build action verbatim if no matching mapping is found rather than using the None fallback

CapeGuyBen:
**What does this PR do?**

Allows user-specific build actions to be specified by using the given build action verbatim if no mapping is found rather than falling back to "None".

If a build action is provided that doesn't match one of the pre-defined actions then it will be applied to the associated files as specified instead of being ignored.

For example, MonoGame projects use the custom "MonoGameContentReference" buildaction. With this change it's possible to specify that when using a premake project.

### DIFF
--- a/src/tools/dotnet.lua
+++ b/src/tools/dotnet.lua
@@ -75,7 +75,11 @@
 				info.action = "Page"
 			end
 		else
-			info.action = "None"
+			if fcfg.buildaction == nil then
+				info.action = "None"
+			else
+				info.action = fcfg.buildaction
+			end
 		end
 
 		-- Try to work out any subtypes, based on the files in the project
--- a/modules/vstudio/tests/cs2005/test_files.lua
+++ b/modules/vstudio/tests/cs2005/test_files.lua
@@ -270,6 +270,17 @@
 		]]
 	end
 
+	function suite.fallbackUserAction()
+		files { "Hello.custom" }
+		filter "files:Hello.custom"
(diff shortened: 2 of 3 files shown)


CapeGuyBen:
Improves #614. Though there's a possibility of a name clash between premake's predefined actions and the user's custom name with this approach.

This allows custom build actions (defined via project dependencies) to be run but does not allow the user to specify new ones.

src/tools/dotnet.lua
```diff
@@ -75,7 +75,11 @@
 				info.action = "Page"
 			end
 		else
-			info.action = "None"
+			if (fcfg.buildaction == nul) then
```
Jarod42:
```suggestion
			if fcfg.buildaction == nil then
```

Jarod42 (commented):
Would i make sense to extend it to C/C++ actions too?

src/tools/dotnet.lua
CapeGuyBen: Oops, applied.

CapeGuyBen: I'm not sure if this makes sense for C++ projects in general. Is the concept applicable to all the generators? It's probably a bit beyond the scope of this PR.

nickclark2016: Agree that it's out of scope for this PR.

nickclark2016 (approved):

PR #2733, opened 2026-07-17
Add infoplist API + Info.plist embedding in Apple OS console apps

noresources:
**What does this PR do?**

The `infoplist` API adds a more explicit way to set the App bundle info in Xcode.
gmake, codelite and ninja generators now supports Info.plist by adding Apple toolchain specific flags provided by the new `toolset.getSections()` function. 

The `gcc.getSections()` allows other generator to embed the Info.plist in the binary.
This will allow command line tools to have bundle informations and security entitlements.

If `infoplist "file"` is set in premake script. The automatic detection of .plist in xcode will be ignored.

In the future, `toolset.getSections()` may be used to embed any file into the binary
by adding a new dedicated API or by handling
```lua
filter {"*.bin"}
	buildaction "Embed"
	``` 

### DIFF
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -535,7 +535,7 @@
 --
 -- get the link-time file embedding flags
 --
-	function gcc.getSections(cfg)
+	function gcc.getsections(cfg)
 		local result = {}
 		if table.contains(os.getSystemTags(cfg.system), "darwin") then
 			if cfg.infoplist then
--- a/src/tools/gcc.lua
+++ b/src/tools/gcc.lua
@@ -533,6 +533,19 @@
 	end
 
 --
+-- get the link-time file embedding flags
+--
+	function gcc.getSections(cfg)
+		local result = {}
+		if table.contains(os.getSystemTags(cfg.system), "darwin") then
+			if cfg.infoplist then
+				local relative = p.tools.getrelative(cfg.project, cfg.infoplist)
+				table.insert(result, "-Wl,-sectcreate,__TEXT,__info_plist," .. p.quoted(relative))
+			end
+		end
+		return result
+	end
+--
 -- get the right output flag.
 --
 	function gcc.getsharedlibarg(cfg)
--- a/tests/tools/test_gcc.lua
+++ b/tests/tools/test_gcc.lua
@@ -1559,7 +1559,7 @@ end
 
 		prepare()
 	
-		local actual = gcc.getSections(cfg)
+		local actual = gcc.getsections(cfg)
 		local expected = "-Wl,-sectcreate,__TEXT,__info_plist,MyInfo.plist"
 		test.contains({expected}, actual)
 	end
@@ -1570,7 +1570,7 @@ end
 
 		prepare()
 
-		local actual = gcc.getSections(cfg)
+		local actual = gcc.getsections(cfg)
 		local notExpected = "-Wl,-sectcreate,__TEXT,__info_plist,MyInfo.plist"
 		test.excludes({notExpected}, actual)
 		--test.contains ("", "")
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: tests/tools/test_gcc.lua [hunk 1]
```diff
@@ -1550,6 +1550,37 @@ end
...
+		system "MacOSX"
+		infoplist "./MyInfo.plist"
+
+		prepare()
+	
+		local actual = gcc.getSections(cfg)
+		local expected = "-Wl,-sectcreate,__TEXT,__info_plist,MyInfo.plist"
+		test.contains({expected}, actual)
+		--test.contains ("", "")
+		--test.contains ("", "")
```
Jarod42: Commented code to remove

tests/tools/test_gcc.lua
```diff
@@ -1550,6 +1550,37 @@ end
...
+		system "Linux"
+		infoplist "./MyInfo.plist"
+
+		prepare()
+
+		local actual = gcc.getSections(cfg)
+		local notExpected = "-Wl,-sectcreate,__TEXT,__info_plist,MyInfo.plist"
+		test.excludes({notExpected}, actual)
+		--test.contains ("", "")
+		--test.contains ("", "")
```
Jarod42, line 1577: Commented code to remove

website/docs/infoplist.md
```diff
@@ -0,0 +1,35 @@
...
+
+`path` is the file system path to the Apple property list file to use as Info.plist.
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0 beta 8 or later for Apple platform applications.
```
Jarod42: No longer "beta".

website/docs/infoplist.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Examples ###
+
+This project defines a default Info.plist and a different one for the Debug configuration.
+
+```lua
+project "MyProject"
+
+	infoplist "Default.plist"
```
Jarod42: Don't mix space and tab

website/docs/Project-API.md
```diff
@@ -96,6 +96,7 @@
 | [includedirs](includedirs.md)                             |  |
 | [includedirsafter](includedirsafter.md)                   |  |
 | [includeexternal](globals/includeexternal.md)             |  |
+| [infoplist](infoplist.md)                                   | Sets the Apple ap bundle infos.                  |
```
Jarod42, line 99:
```suggestion
| [infoplist](infoplist.md)                                   | Sets the Apple application bundle infos.                  |
```
?

website/docs/infoplist.md
noresources: OK, I was relying on git tags. There is no v5.0.0 tag.

Jarod42 (approved):

src/tools/gcc.lua
```diff
@@ -533,6 +533,19 @@
 	end
 
 --
+-- get the link-time file embedding flags
+--
+	function gcc.getSections(cfg)
```
nickclark2016: 99% of the code is all lower case. Should we make this match?

nickclark2016 (changes requested):

src/tools/gcc.lua
noresources: Since recent changes in source code introduced camelCase function. I supposed it was the new  policy. I can change this if needed.

nickclark2016: Could you point out the introduced change in question? I may have missed something. It's not a big issue, just would like to make us consistent if possible.

noresources: For example: `gcc.getLibraryDirectories()`,  `os.getSystemTags()`. Those functions wre introduced "recently".
I agree the need of consistency, and I understand the challenge of keeping relevant API names when everything is lowercased. If lowercase is still the rule I will be pleased to go for it.

nickclark2016: Yeah, those probably shouldn't have been let through as-is. I don't think I can rename those this close to 5.0, but I can probably do a lowercase alias for them for consistency. If you wouldn't mind changing this to lowercase, I'm happy to move this forward in its current state.

nickclark2016:
Hmm, I thought I fixed these in master. Going to bring it in, since these are unrelated build failures.

nickclark2016 (approved):

PR #2740, opened 2026-07-22
Fixed that build outputs are infinitely recursively processed

### LINKED ISSUE #1940: Build outputs are infinitely recursively processed
When using custom rules, if a buildoutput happens to use the same file extension as the rule itself, premake gets into an infinite recursion loop and later dies with stack smashing(?).

I had expected premake not to compile already compiled files again. Though, something like that can be by design, i.e. you do .bmp>.png>.jpg, but not .png>.png>... (unless specified so)

A dirty hack: create a file with a different extension and then copy it to the output directory.

```lua
rule "sampleRule"
    fileextension ".txt"

    -- buildmessage and buildcommands dont seem to matter
    buildmessage 'Copying %{file.relpath}'
    buildcommands 'cp %{file.relpath} new_%{file.relpath}'
    buildoutputs '%{file.basename}.txt'

workspace "sample"

project "sample"
    kind "Utility"
    rules { "sampleRule" }

    files { "**.txt" }
```
Keep in mind that for order for this to work you need to first have #1939 installed. 
My premake script uses a required script and overrides the gmake utility project generator, so the version is still relevant, nothing is touched except that.

- [ ] Visual Studio 2022 (vs2022)
- [ ] Visual Studio 2019 (vs2019)
- [ ] Visual Studio 2017 (vs2017)
- [ ] Visual Studio 2015 (vs2015)
- [ ] Visual Studio 2012 (vs2012)
- [ ] Visual Studio 2010 (vs2010)
- [ ] Visual Studio 2008 (vs2008)
- [ ] Visual Studio 2005 (vs2005)
- [ ] GNU Makefile (gmake)
- [x] GNU Makefile 2 (gmake2)
- [ ] XCode (xcode)
- [ ] Codelite
- [ ] Other (Please list below)

`5.0.0-beta1`

SanjeetBalamurugan:
**What does this PR do?**

Fixes infinite recursion when a custom rule generates a file with the same extension as its input. Previously, generated files could be treated as new inputs for the same rule, causing the rule to be applied repeatedly.

Generated files from custom rules are now tracked with the rule that created them. Before applying a rule, Premake checks whether the file was already generated by that same rule and skips it if so.

This prevents recursive rule application while keeping existing custom rule behavior unchanged. No breaking changes are introduced.

Closes #1940 

This addresses issue [1940](https://github.com/premake/premake-core/issues/1940).

The fix was verified with a regression case where a custom rule generates another file with the same extension as its input. The generated output is created once and is not recursively processed.

### DIFF
--- a/modules/gmake/gmake_cpp.lua
+++ b/modules/gmake/gmake_cpp.lua
@@ -199,7 +199,7 @@
 				table.insert(cfg._gmake.fileRules, file)
 
 				for _, output in ipairs(buildoutputs) do
-					cpp.addGeneratedFile(cfg, node, output)
+					cpp.addGeneratedFile(cfg, node, output, nil)
 				end
 			end
 		elseif filecfg.buildaction == "Copy" then
@@ -212,7 +212,7 @@
 				buildinputs = {'$(TARGETDIR)'}
 			}
 			table.insert(cfg._gmake.fileRules, file)
-			cpp.addGeneratedFile(cfg, node, output)
+			cpp.addGeneratedFile(cfg, node, output, nil)
 		else
 			cpp.addRuleFile(cfg, node)
 		end
@@ -233,7 +233,7 @@
 		return fileext;
 	end
 
-	function cpp.addGeneratedFile(cfg, source, filename)
+	function cpp.addGeneratedFile(cfg, source, filename, generatingRule)
 		-- mark that we have generated files.
 		cfg.project.hasGeneratedFiles = true
 
@@ -249,6 +249,11 @@
 		-- always overwrite the dependency information.
 		node.dependsOn = source
 		node.generated = true
+
+		-- if the generating node exists then use it
+		if generatingRule then
+            node.generatedByRule = generatingRule
+        end
 
 		-- add to config if not already added.
 		if not fileconfig.getconfig(node, cfg) then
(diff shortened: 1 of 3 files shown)


Jarod42 (approved):

SanjeetBalamurugan (commented):
.

SanjeetBalamurugan:
May I know if there is any problem with my pull request? @Jarod42

Jarod42:
I'm a simple contributor to Premake, like you. I am not a member of Premake.
So I don't have any right.
As contributors, I also do review to spot issue. and for your PR, I don't see problem (There is the fact there are no tests though which might be problematic.)

SanjeetBalamurugan:
@Jarod42, Sanjeet here.

Thanks for your review regarding my pull request. As for the tests, i still have doubts about how to add them. If its a really needed, I would try to add them. But i dont know how tests work in premake-core.

Thanks

modules/gmake/tests/test_gmake_customrules.lua
nickclark2016, line 1: There are no tests. We either need tests here or this file needs to be removed.

nickclark2016 (changes requested):

modules/gmake/tests/test_gmake_customrules.lua
SanjeetBalamurugan: Is the file removal enough or tests are required?

nickclark2016: Is there an effective way to unit test this? If so, tests are required.

SanjeetBalamurugan: I dont know how tests work here, its a bit confusing. I dont think so we could check for a specific error to occur(initial issue) using the current api. I might be wrong.

SanjeetBalamurugan:
@nickclark2016 Hmm, Is this better for merging?

nickclark2016 (approved):
Apologies, this dropped of my radar. I think this is good. Verifying CI before merging.

PR #2749, opened 2026-07-29
Make removal of symlinks more robust, fix FreeBSD build

nickclark2016:
**What does this PR do?**

Fixes an issue arising in FreeBSD 15.1 where remove on a symlink would fail to remove a symlink to a file.

**How does this PR change Premake's behavior?**

No changes to documented behavior. This make the FreeBSD behavior consistent before and after 15.1

**Anything else we should know?**

No

**Did you check all the boxes?**

- [x] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [ ] Add unit tests showing fix or feature works; all tests pass
- [ ] Mention any [related issues](https://github.com/premake/premake-core/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/premake/premake-core/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] Minimize the number of commits
- [ ] Align [documentation](https://github.com/premake/premake-core/tree/master/website) to your changes

*You can now [support Premake on our OpenCollective](https://opencollective.com/premake). Your contributions help us spend more time responding to requests like these!*

PR #2759, opened 2026-08-05
fix os.isdir() and os.mkdir()

mercury233:
**What does this PR do?**

* On MinGW, `_wstat` does not fully recognize reparse points (symbolic links); `GetFileAttributesW` needs to be used instead.
* Fixed an issue where `os.mkdir` incorrectly returned success when the target already existed and was a file, added the test.
* Simplified the implementation of `os.isdir`.

In the edge case where the target of `mkdir` already exists and is a file, the operation may now terminate earlier.

### DIFF
--- a/tests/base/test_os.lua
+++ b/tests/base/test_os.lua
@@ -179,6 +179,16 @@
 		test.isfalse(os.islink("folder/ok2.lua"))
 	end
 
+--
+-- os.mkdir() tests
+--
+
+	function suite.mkdir_ReturnsError_OnExistingFile()
+		test.istrue(os.isfile("folder/ok.lua"))
+		local ok, err = os.mkdir("folder/ok.lua")
+		test.isnil(ok)
+		test.isequal("string", type(err))
+	end
 
 --
 -- os.matchdirs() tests
--- a/tests/base/test_os.lua
+++ b/tests/base/test_os.lua
@@ -179,6 +179,24 @@
 		test.isfalse(os.islink("folder/ok2.lua"))
 	end
 
+--
+-- os.mkdir() tests
+--
+
+	function suite.mkdir_ReturnsTrue_OnExistingDirectoryLink()
+		test.istrue(os.linkdir("folder/subfolder", "folder/subfolder2"))
+		test.istrue(os.mkdir("folder/subfolder2"))
+		test.istrue(os.islink("folder/subfolder2"))
+		test.istrue(os.isdir("folder/subfolder2"))
+		test.istrue(os.rmdir("folder/subfolder2"))
+	end
+
+	function suite.mkdir_ReturnsError_OnExistingFile()
+		local ok, err = os.mkdir("folder/ok.lua")
+		test.isnil(ok)
+		test.isequal("string", type(err))
+		test.istrue(os.isfile("folder/ok.lua"))
+	end
 
 --
 -- os.matchdirs() tests
--- a/tests/base/test_os.lua
+++ b/tests/base/test_os.lua
@@ -7,6 +7,29 @@
 	local suite = test.declare("base_os")
 
 	local cwd
+	local real_io_open = io.open
+
+	local tmpname = function()
+		local p = os.tmpname()
+		if p:startswith("\\") then
+			p = "." .. p
+		end
+		os.remove(p) -- just needed on POSIX
+		return p
+	end
+
+	local tmpfile = function()
+		local p = tmpname()
+		local f = assert(real_io_open(p, "w"))
+		f:close()
+		return p
+	end
+
+	local tmpdir = function()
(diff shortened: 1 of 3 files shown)


mercury233:
The current test needs to be run outside the “sandbox” in order to pass, due to the reasons mentioned above.

* Run the test outside the sandbox.
* Skip this test on FreeBSD, or skip it only in the FreeBSD CI via an environment variable.
* Abandon this test completely.

I am not sure which option would be the most appropriate.

tests/base/test_os.lua
```diff
@@ -179,6 +179,33 @@
...
+		test.istrue(os.rmdir(link))
+		test.istrue(os.rmdir(target))
+		test.istrue(os.rmdir(root))
+	end
+
+	function suite.mkdir_ReturnsError_OnExistingFile()
+		local ok, err = os.mkdir("folder/ok.lua")
+		test.isnil(ok)
+		test.isequal("string", type(err))
+		test.istrue(os.isfile("folder/ok.lua"))
```
Jarod42: I would move `test.istrue(os.isfile("folder/ok.lua"))` at the beginning. It is the prerequire for the failure of `os.mkdir`

Jarod42 (approved):

tests/base/test_os.lua
```diff
@@ -179,6 +179,33 @@
 		test.isfalse(os.islink("folder/ok2.lua"))
 	end
 
+--
+-- os.mkdir() tests
+--
+
+	function suite.mkdir_ReturnsTrue_OnExistingDirectoryLink()
+		-- BSD CI mounts the workspace through SSHFS, so keep topology tests on local storage.
```
samsinsane: What was the actual problem?

samsinsane (changes requested):

tests/base/test_os.lua
mercury233: https://github.com/premake/premake-core/actions/runs/30967252835/job/92183785538
after `os.linkdir("folder/subfolder", "folder/subfolder2")`, `os.mkdir("folder/subfolder2")` fail.

samsinsane: Why did it fail?

mercury233: I’m not sure. I can only guess that it has something to do with SSHFS.

nickclark2016: Likely related to issues that I started addressing with #2749, if I had to guess. Our BSD VM seems to be breaking quite a bit of code that works elsewhere. Not sure if it's with the VM set up or if it's BSD itself.

nickclark2016:
I want to follow up on this. Currently, our CI jobs are passing. Is this a gap in our current testing?

mercury233:
> I want to follow up on this. Currently, our CI jobs are passing. Is this a gap in our current testing?

The main issue is that previously, `os.mkdir` would return success when the target already existed as a file. Fixing this requires reliably checking the attributes of the target, which introduced symlink detection and, in turn, caused an issue in the FreeBSD CI.

~~The current test works around this issue, but it introduces file operations outside the sandbox (in a temporary directory).~~ I've dropped the newly added test, I think this probably wouldn’t cause any issues in practice.

tests/base/test_os.lua
samsinsane: The comment reads as if it's obvious why you'd use local storage when using BSD and SSHFS, but then you're not sure why it doesn't work. We might just be calling the wrong functions on BSD? It would be nice to have more investigation into the issue so that we can clearly state why this test doesn't work with that configuration.

mercury233: > The comment reads as if it's obvious why you'd use local storage when using BSD and SSHFS, but then you're not sure why it doesn't work. We might just be calling the wrong functions on BSD? It would be nice to have more investigation into the issue so that we can clearly state why this test doesn't work with that configuration.

Sorry, I'm not interested about BSD. So I've just removed the test.

tests/base/test_os.lua
```diff
@@ -179,6 +179,16 @@
 		test.isfalse(os.islink("folder/ok2.lua"))
 	end
 
+--
+-- os.mkdir() tests
+--
+
+	function suite.mkdir_ReturnsError_OnExistingFile()
+		test.istrue(os.isfile("folder/ok.lua"))
+		local ok, err = os.mkdir("folder/ok.lua")
```
nickclark2016: If this "breaks" the `ok.lua` file, this will cause subsequent tests depending on the existence of this file to fail. Let's create a temp file or temp directory to do this.

nickclark2016 (changes requested):
One change requested in tests, but LGTM otherwise.

nickclark2016 (approved):

samsinsane (approved):

PR #2763, opened 2026-08-06
Fix clang linker on windows

### LINKED ISSUE #2762: Clang toolset emits GNU import-library linker flag when targeting Windows with `lld-link`
Premake’s Clang toolset emits the GNU linker option on Ninja for Windows:

```text
-Wl,--out-implib="<path>"
```

when generating a Windows shared library.

This works when Clang is targeting a MinGW/GNU linker, but it does not work when the compiler driver invokes `link.exe` or `lld-link` in MSVC mode.

In that case, `lld-link` reports:

```text
lld-link: warning: ignoring unknown argument '--out-implib=bin/Debug/shared-library.lib'
```

The MSVC-compatible equivalent is:

```text
/IMPLIB:<path>
```

When passed through a GCC-style compiler driver, it can be forwarded using:

```text
-Xlinker /IMPLIB:<path>
```

This would need to be fixed for all other linker options as well.

I wanted to put up a patch to fix this but it requires rethinking the architecture of toolsets and any change would be intrusive. Maybe clang-cl could be added as a separate toolset for makefile generators and it can maintain its list of compatible flags. Otherwise, I can't think of a clean way to direct clang to call GNU style linker on Windows. It is a shame because the rest of the toolset (compiler-side) works fine and it is the linker that is problematic.

Here is the issue location: https://github.com/premake/premake-core/blob/93ff1b42edb30f27960d04b8fd8c601cf8cba125/src/tools/clang.lua#L277

needs to be:

```
table.insert(r, '-Xlinker /IMPLIB:"' .. cfg.linktarget.relpath .. '"')
```

Happy to make this change if acceptable.

theComputeKid:
When clang is used on Windows for driving the linking, it will always call lld-link. Therefore, the call on Windows must be MS syntax, not GNU. 

Fixes #2762

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -273,8 +273,12 @@
 		kind = {
 			SharedLib = function(cfg)
 				local r = { clang.getsharedlibarg(cfg) }
-				if cfg.system == "windows" and cfg.useimportlib ~= p.OFF then
-					table.insert(r, '-Wl,--out-implib="' .. cfg.linktarget.relpath .. '"')
+				if cfg.system == "windows" then
+					if cfg.useimportlib ~= p.OFF then
+						table.insert(r, '-Xlinker /IMPLIB:"' .. cfg.linktarget.relpath .. '"')
+					else
+						table.insert(r, '-Xlinker /NOIMPLIB')
+					end
 				elseif cfg.system == p.LINUX then
 					table.insert(r, '-Wl,-soname=' .. p.quoted(cfg.linktarget.name))
 				elseif table.contains(os.getSystemTags(cfg.system), "darwin") then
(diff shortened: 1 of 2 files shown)


Jarod42 (approved):

src/tools/clang.lua
```diff
@@ -274,7 +274,7 @@
 			SharedLib = function(cfg)
 				local r = { clang.getsharedlibarg(cfg) }
 				if cfg.system == "windows" and cfg.useimportlib ~= p.OFF then
```
Jarod42: It seems `p.OFF` is not handled correctly neither.
I think `/NOIMPLIB` should be added in that case

theComputeKid: @Jarod42 I agree with your point. I have attempted to fix it now. Please review again.

Jarod42 (approved):

nickclark2016 (approved):

PR #2764, opened 2026-08-06
Enable runtime for clang

theComputeKid:
These are clang's equivalent to MSVC's /MT, /MTd, /MD and /MDd on Windows. Both compiler and linker flags are required.

### DIFF
--- a/src/tools/clang.lua
+++ b/src/tools/clang.lua
@@ -79,6 +79,32 @@
 		useshortenums = gcc.shared.useshortenums,
 	}
 
+	clang.shared.runtime = {
+		Debug = function(cfg)
+			if cfg.system ~= premake.WINDOWS then
+				return nil
+			end
+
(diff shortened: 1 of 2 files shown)


mercury233:
Premake supports omitting the `runtime` directive and automatically inferring it, but the newly added `staticruntime` change does not take effect when `runtime` is omitted.
https://github.com/mercury233/premake-core/actions/runs/31134127921/job/92729620718?pr=4

Jarod42 (approved):

theComputeKid:
@mercury233 : It would be easier if runtime and staticruntime were just one flag with 4 options. I agree with your comment and I have tried to incorporate it in my change, though it does make it messier to read. Could you check if the logic is correct?

Jarod42 (approved):

mercury233 (approved):

nickclark2016 (approved):

PR #2768, opened 2026-08-11
Require 'custom:' prefix for custom build actions

CapeGuyBen:
**What does this PR do?**

Update to changes from [2729](https://github.com/premake/premake-core/pull/2729) to require a 'custom:' prefix for verbatim user build actions.

This enables new premake builtin buildactions to be added in later versions without breaking existing custom actions which happen to use the same action name.

Fallback is returned to None as per the original implementation.

### DIFF
--- a/src/tools/dotnet.lua
+++ b/src/tools/dotnet.lua
@@ -56,6 +56,8 @@
 		-- Determine the build action for the file, falling back to the file
 		-- extension if no explicit action is available.
 
+		local customprefix = "custom:"
+
 		if fcfg.buildaction == "Compile" or ext == ".cs" or ext == ".fs" then
 			info.action = "Compile"
 		elseif fcfg.buildaction == "Embed" or ext == ".resx" then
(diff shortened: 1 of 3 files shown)


src/tools/dotnet.lua
```diff
@@ -75,11 +79,7 @@
 				info.action = "Page"
 			end
 		else
-			if fcfg.buildaction == nil then
-				info.action = "None"
-			else
-				info.action = fcfg.buildaction
-			end
+			info.action = "None"
```
Jarod42, line 85: Additional warning might be useful.

CapeGuyBen: Not sure it's exactly what you meant but I added a `premake.warn` when the `None` fallback is used when `fcfg.buildaction` is not `nil`.

Jarod42: It is that :)

nickclark2016 (approved):

nickclark2016:
This is not considered an API break, since this was never documented and I don't believe it had landed in an official release yet.

PR #2770, opened 2026-08-12
Enable assembly generation

theComputeKid:
- GCC and Clang use `-save-temps=obj` when `generateassembly` is `On`. `Verbose` additionally uses `-fverbose-asm`.
- MSVC uses `/FA` for `On` and `/FAs` for `Verbose`.
- Assembly is output in the object directory. Some logic has been added to avoid output name clashes, by deriving it from the object file name and assigning to `/Fa$(asm_name).asm` for msvc-syntax compilers. This requires modifying the vs/ninja/makefile generators for those compilers.

Future work can decouple assembly generation and output location. But having asm output to the object directory is a good starting point and a very common use case when you want to inspect the assembly for optimization purposes.

### DIFF
--- a/modules/ninja/ninja_cpp.lua
+++ b/modules/ninja/ninja_cpp.lua
@@ -776,7 +776,15 @@ function m.buildFile(cfg, node, filecfg, objFile, pchFile, prebuildTarget)
 	local flags = ""
 	local extraFlags = {}
 	local toolset = ninja.gettoolset(cfg)
+	local assemblyFile
 	local hasPerFileConfig = m.hasPerFileConfiguration(cfg, filecfg)
+	if toolset.getassemblyflags then
+		local assemblyCfg = filecfg.generateassembly and filecfg or cfg
+		if toolset.getassemblyoutput then
+			assemblyFile = toolset.getassemblyoutput(assemblyCfg, objFile)
+		end
+		table.insertflat(extraFlags, toolset.getassemblyflags(assemblyCfg, assemblyFile and p.quoted(assemblyFile)))
+	end
 	
 	if filecfg and filecfg.compileas and filecfg.compileas ~= "Default" then
 		if p.languages.isc(filecfg.compileas) then
@@ -845,7 +853,11 @@ function m.buildFile(cfg, node, filecfg, objFile, pchFile, prebuildTarget)
 			implicitDeps = implicitDeps .. " " .. table.concat(cfg._dependsOnTargets, " ")
 		end
 		
-		_p("build %s: %s_%s %s%s", objFile, rule, cfg.toolset, relPath, implicitDeps)
+		local implicitOutput = ""
+		if assemblyFile then
+			implicitOutput = " | " .. assemblyFile
+		end
+		_p("build %s%s: %s_%s %s%s", objFile, implicitOutput, rule, cfg.toolset, relPath, implicitDeps)
 		
 		if usePch then
 			if toolset == p.tools.msc then
--- a/modules/ninja/tests/test_ninja_build_rules.lua
+++ b/modules/ninja/tests/test_ninja_build_rules.lua
@@ -114,6 +114,25 @@ rule cxx_msc
 		]]
 	end
 
+	function suite.buildFile_generateAssembly_onMSVC()
+		toolset "msc"
+		language "C++"
+		files { "main.cpp" }
+		buildoptions { "/projectFlag" }
+		generateassembly "Verbose"
+		filter "files:main.cpp"
+		generateassembly "On"
+		local cfg = prepare()
+		local node = p.project.getsourcetree(cfg.project).children["main.cpp"]
+		local filecfg = p.fileconfig.getconfig(node, cfg)
+		local objFile = cpp.objectFile(cfg, node, filecfg)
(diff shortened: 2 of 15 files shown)


modules/ninja/tests/test_ninja_build_rules.lua
```diff
@@ -114,6 +114,30 @@ rule cxx_msc
...
+				if item.name == "main.cpp" then
+					node = item
+				end
+			end
+		}, false, 1)
+		local filecfg = p.fileconfig.getconfig(node, cfg)
+		local objFile = cpp.objectFile(cfg, node, filecfg)
+		cpp.buildFile(cfg, node, filecfg, objFile, nil, nil)
+		test.capture [[
+build obj/Debug/main.obj: cxx_msc main.cpp
```
Jarod42: obj/Debug/main.asm should be added as secondary output

src/tools/msc.lua
```diff
@@ -548,7 +552,17 @@
 			return string.format('/nologo /fo%s %s', output, input)
 		end
 		local lang = iif(tool == "cc", "/Tc", "/Tp")
-		return string.format('/nologo /Fo%s /c %s%s', output, lang, input)
+		local assembly = ""
+		if cfg.generateassembly == p.ON or cfg.generateassembly == "Verbose" then
+			local assemblyOutput = output
+			if output == '"$@"' then
+				assemblyOutput = '"$(@:%.obj=%.asm)"'
```
Jarod42: Look like Makefile's logic there.... :/

website/docs/generateassembly.md
```diff
@@ -0,0 +1,39 @@
...
+- MSVC uses `/FA` for `On` and `/FAs` for `Verbose`.
+- Assembly files are written to the object directory.
+
+### Applies To ###
+
+Project configurations.
+
+### Availability ###
+
+Premake 5.0.0-beta8 or later.
```
Jarod42:
```suggestion
Premake 5.0.0 or later.
```

theComputeKid: Fixed.

src/tools/msc.lua
theComputeKid: Sorry that was a silly mistake. Forgot about that. Fixed.

modules/ninja/tests/test_ninja_build_rules.lua
theComputeKid: Fixed.

Jarod42 (approved):

modules/ninja/ninja_cpp.lua
```diff
@@ -776,7 +776,12 @@ function m.buildFile(cfg, node, filecfg, objFile, pchFile, prebuildTarget)
 	local flags = ""
 	local extraFlags = {}
 	local toolset = ninja.gettoolset(cfg)
+	local assemblyFile
 	local hasPerFileConfig = m.hasPerFileConfiguration(cfg, filecfg)
+	if toolset == p.tools.msc and (cfg.generateassembly == p.ON or cfg.generateassembly == "Verbose") then
+		assemblyFile = path.replaceextension(objFile, ".asm")
+		table.insert(extraFlags, "/Fa" .. p.quoted(assemblyFile))
```
nickclark2016: I would prefer logic that is specific to flags be on the toolset if at all possible.

website/docs/generateassembly.md
```diff
@@ -0,0 +1,39 @@
...
+generateassembly ("value")
+```
+
+### Parameters ###
+
+`value` specifies the desired behavior:
+
+| Value   | Description                                                              |
+|---------|--------------------------------------------------------------------------|
+| Default | Do not request assembly output.                                          |
```
nickclark2016: Default should not be "off", but whatever the toolset or action's default is.

nickclark2016 (changes requested):

website/docs/generateassembly.md
theComputeKid: Fixed.

modules/ninja/ninja_cpp.lua
theComputeKid: This was a more invasive change than I expected. Can you please review my latest attempt?

theComputeKid:
@nickclark2016 : I have changed the flag's scope from config to per file, which makes more sense for a flag that enables assembly generation. Can you please review if I did it correctly?

src/tools/msc.lua
```diff
@@ -543,6 +558,12 @@
 		return iif(tool == "rc", ".res", ".obj")
 	end
 
+	function msc.getAssemblyOutput(cfg, objectOutput)
```
nickclark2016: Comment for all toolset functions -- names should be all lower for consistency

nickclark2016 (changes requested):
Overall code looks really good. Just one, hopefully trivial, comment to fix up.

src/tools/msc.lua
theComputeKid: Fixed. Actually, I started off lower case, and then I saw: https://github.com/premake/premake-core/blob/af5dfcb38db6711a71ec9f3197f5475dbbb510c5/src/tools/msc.lua#L44 

And then I thought maybe I should change. However, looking back now, most of the functions are lower case.

nickclark2016 (approved):

Issue #2767, opened 2026-08-11
C++20 module support


theComputeKid:
Is there a plan to support C++20 modules for non-vstudio generators? This is not an easy task, it would probably require a major structural change that adds a prebuild step to:

- Call toolset-specific tools (e.g. clang-scan-deps) to scan for module dependencies and emit this info
- Parse the output to determine dependencies
- Dynamically add these dependencies to the build graph for makefile/ninja

It seems that the major compiler vendors have now stabilised their module scanning tools and have adopted a single dependency generation format.

Is this within scope of the Premake project?

nickclark2016:
There are plans, but it will not make it into the 5.0 release. I would plan on this being added to Ninja first, and likely won't see any Makefile implementation soon.

I've been waiting for the toolsets to stabilize and for me to have enough time to implement them before starting. Hopefully the latter will be within the next few months, but I make no promises.

theComputeKid:
Here are the minimum commands to build the stdlib module across the major compilers:

```
rule std_msc
  command = cmd /c cl /nologo /c /std:c++latest /EHsc "%VCToolsInstallDir%modules\std.ixx" /ifcOutput std.ifc /Fostd.obj

rule std_gcc
  command = g++-16 -c -std=c++23 -fmodules --compile-std-module

rule std_clang
  command = clang++ -c -x c++-module -std=c++23 -stdlib=libc++ -Wno-reserved-module-identifier `clang -print-file-name=`../../../share/libc++/v1/std.cppm -fmodule-output=std.pcm
```

Notes:
- MSVC needs `cmd` rather than a direct call to `cl`. This is so it can parse the value of `%VCToolsInstallDir%` at compile time, which is set automatically with the appropriate vcvars. Luckily, MSFT has the fixed location of the std module to this environment variable.
- Alternative approach to clang is to read the module JSON file (`clang++ -print-file-name=libc++.modules.json`), but that is not as reliable across all clang distributions (e.g. homebrew).
- Note that on Windows, we can alternatively provide `"%VCToolsInstallDir%modules\std.ixx"` to clang and it will compile as it is meant to be compatible with msvc std lib by default. Therefore, if libc++ is not specified by the user, clang should compile msvc stdlib.

We would then need to add the appropriate module link flags for every file compiled e.g.:

```
rule cxx_msc
  cl /c /ifcSearchDir . /Fo$out.obj $in

or more explicit:

rule cxx_msc
[...]

PR #2779, opened 2026-08-21
Enable buildstlmodules for ninja-msc

theComputeKid:
Written in a modular way to allow clang and gcc to be added later, as well as makefile rules. Each toolset has its own definition of the variables in the cppmodules table, but I have only started off with `msc` as I have access to a Windows machine. To try:

```
import std;

int main() {
  std::println("Hello");
  return 0;
}

```
Needs:

```
language "C++"
cppdialect "c++latest"
buildstlmodules "On"
```

It will build `std.ifc` and `std.obj`:
- The IFC file is consumed by every file being compiled
- The OBJ file is consumed at program link time.

Note on shell command:
- Shell (`cmd`) is called because we need to expand the macro that defines the fixed location of the standard library module for MSVC: `"%VCToolsInstallDir%\modules\std.ixx"`. This macro is only available at project build time (when one would assume CL is present due to `vcvars` being called), not necessarily at project generation time.

### DIFF
--- a/src/tools/msc.lua
+++ b/src/tools/msc.lua
@@ -57,6 +57,15 @@
 		},
 	}
 
+	msc.cppmodules = {
+		bmiExtension = "ifc",
+		stl = "\"%VCToolsInstallDir%\\modules\\std.ixx\"",
+		bmiCompileFlags = "/nologo /c /interface",
+		bmiOutputFlag = "/ifcOutput ",
+		bmiReferenceFlag = "/reference ",
+		objOutputFlag = "/Fo",
+	}
+
 	function msc.getassemblyflags(cfg, assemblyOutput)
 		local flags = config.mapFlags(cfg, msc.generateassembly)
 		if assemblyOutput and #flags > 0 then
--- a/modules/ninja/ninja_cpp.lua
+++ b/modules/ninja/ninja_cpp.lua
@@ -45,6 +45,7 @@ function m.rules(prj)
 			m.cxxrule(cfg, toolset)
 			m.resourcerule(cfg, toolset)
 			m.linkrule(cfg, toolset)
+			m.cxxstdlibrule(cfg, toolset)
 			m.pchrule(cfg, toolset)
 			m.copyrule(cfg, toolset)
 			m.prebuildcommandrule(cfg, toolset)
@@ -56,6 +57,21 @@ function m.rules(prj)
 			rulesDone[toolsetKey] = true
 		end
 	end
+end
+
+function m.cxxstdlibrule(cfg, toolset)
+	toolset = toolset or ninja.gettoolset(cfg)
+	if not toolset.cppmodules then
+		return
+	end
+
+	local cppmodules = toolset.cppmodules
+	local cxxname = toolset.gettoolname(cfg, "cxx")
+	_p("rule cxxstdlib_%s", cfg.toolset)
+	_p("  command = $shell %s %s %s$bmi $cxxflags $stl %s$obj",
+		cxxname, cppmodules.bmiCompileFlags, cppmodules.bmiOutputFlag, cppmodules.objOutputFlag)
+	_p("  description = Building C++ standard library module $bmi")
+	_p("")
 end
 
 function m.ccrule(cfg, toolset)
@@ -634,6 +650,7 @@ function m.buildFiles(cfg)
 	local objList = {}
 	local copyList = {}
 	local pchFile = m.buildPch(cfg)
+	local stlModule, stlModuleObj = m.buildstlmodule(cfg)
 	
 	local prebuildTarget = m.buildPreBuildEvents(cfg)
 	cfg._hasPrebuild = (prebuildTarget ~= nil)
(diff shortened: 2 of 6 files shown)


theComputeKid:
@nickclark2016 : I don't want to interfere with your planned C++20 [module](https://github.com/premake/premake-core/issues/2767#issuecomment-5255469513) work, but I believe that because the std lib does not need dependency tracking, and since it is built before everything else, it can be added independently to proper c++20 module support. `import std` is a pretty big use case for modules, even if we don't use modules elsewhere in code.

nickclark2016:
How does this handle multiple projects specifying to build the STL module? Are there conflicts?  I'm also a bit concerned over invoking a shell/command prompt to find the STL module interface. I'd rather we detect this at configuration time rather than build time. What if I'm in a conflicting shell? What is the behavior?

theComputeKid:
Every project will have its own std module, so there shouldn't be clashes. The module is compiled into every projects objdir (based on config).

Ninja is shell agnostic because it calls commands directly using Win32 APIs (CreateProcessW). Therefore, regardless of the shell of the user, it will call `cmd` and there will be no conflicts. Note that the only purpose of the shell is to expand the macro at build time.

If we expand the macro (by using `os.getenv()`) at configure time, then we will need to specify that premake requires the build tooling present at configure time. Up until now, we can create an MSVC project even on MacOS by just specifying the --os parameter. This will not be possible if we expand the macro at generation time.

Do you prefer macro expansion on generation or build time? I will then make the change accordingly.

Equally, if you want to enable c++20 modules properly before enabling stdlib, we can just keep this as a proof of concept and not submit it. I just thought it might be good to demonstrate how we could achieve building the stdlib.

src/tools/msc.lua
```diff
@@ -57,6 +57,15 @@
 		},
 	}
 
+	msc.cppmodules = {
+		bmiExtension = "ifc",
+		stl = "\"%VCToolsInstallDir%\\modules\\std.ixx\"",
```
Jarod42, line 62: Trick to remove `\"` is to use single quote, i.e. `stl = '"%VCToolsInstallDir%\\modules\\std.ixx"',`
Not sure if there is a guideline in premake for quote to use though...

src/tools/msc.lua
```diff
@@ -57,6 +57,15 @@
...
+		bmiCompileFlags = "/nologo /c /interface",
+		bmiOutputFlag = "/ifcOutput ",
+		bmiReferenceFlag = "/reference ",
+		objOutputFlag = "/Fo",
```
Jarod42, line 66: Trailing space is inconsistent, which looks strange...

theComputeKid: That is on purpose. I put out msc as an example to get your initial thoughts but clang for example will have:

```
bmiOutputFlag = "-fmodule-output=",
```
MSVC needs a space whereas clang/gcc use `=`.

nickclark2016:
> Every project will have its own std module, so there shouldn't be clashes. The module is compiled into every projects objdir (based on config).

This honestly concerns me a bit. _Should_ every project have their own STL module? This feels excessive if so, and I would argue incorrect. I know CMake handles this by generating one STL module for each set of compiler flags that may cause issues, such as different toolsets or different C++ standards/versions.

> Ninja is shell agnostic because it calls commands directly using Win32 APIs (CreateProcessW). Therefore, regardless of the shell of the user, it will call `cmd` and there will be no conflicts. Note that the only purpose of the shell is to expand the macro at build time.
> 
> If we expand the macro (by using `os.getenv()`) at configure time, then we will need to specify that premake requires the build tooling present at configure time. Up until now, we can create an MSVC project even on MacOS by just specifying the --os parameter. This will not be possible if we expand the macro at generation time.
> 
> Do you prefer macro expansion on generation or build time? I will then make the change accordingly.

Due to the nature of modules being heavily entangled in the tooling, I think we should handle this at configuration time rather than build time. We have a few other cases in the code where this is done, namely finding implicit include paths.

> Equally, if you want to enable c++20 modules properly before enabling stdlib, we can just keep this as a proof of concept and not submit it. I just thought it might be good to demonstrate how we could achieve building the stdlib.

We still need to get C++20 modules ready. That architecture is still in the air, so I think this change is premature, as it could need a complete rework when we land C++20 modules.

For the future, I'd suggest opening an issue or a discussion before bringing in larger features, just so there is no wasted work on your end before going to implement things.

theComputeKid:
> CMake handles this by generating one STL module for each set of compiler flags

I think this is the correct approach but it requires a lot of maintenance to keep updated. CMake gets it wrong quite often, looking at the number of issues open, including some I have opened myself. But I agree with CMake's approach, which is that it is better to be excessive and have a binary that is correct, than accidentally do the wrong thing by optimizing the build system. It will eventually get there once it figures out BMI-impacting flags.

Also might be worth thinking how PCH interacts with modules when you decide to implement.

theComputeKid:
Feel free to close the PR. Modules is one of the most frustrating things about C++ these days, no matter how much we want it to work.

nickclark2016:
Marking closed for now

nickclark2016:
> Feel free to close the PR. Modules is one of the most frustrating things about C++ these days, no matter how much we want it to work.

Definitely understand that. And even when the build tools work, I still find the occasional ICE with them.

